PR #152, opened 2014-04-30
add pushover notification support

jcs:
A fairly mechanical diff, gleaned from existing Growl support.

I [saw](https://forums.sabnzbd.org/viewtopic.php?t=11694&f=9) that there was some hesitation to add Pushover support due to the limitations on notifications, but for public apps like this, each user would register their own app (by cloning the public one at https://pushover.net/apps/clone/sabnzbd) to get 7500 notifications per month, and then input that API token into SABnzbd.


shypike:
Now implemented in the "develop" branch.
The code needed quite some changes and extension, because Notifications are significantly different in 0.8.0
Thanks for the inspiration.

PR #168, opened 2014-07-07
Correct redirect if using incorrect path in URL: only rewrite path part

sanderjo:
Correct redirect, as decribed on https://forums.sabnzbd.org/viewtopic.php?f=2&t=10443&p=95175#p95175
Only the path is rewritten, not the host (as it happened before)

So: http://192.168.1.111:8080/isnotthere now redirects to the correct URL http://192.168.1.111:8080/sabnzbd/

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1538,7 +1538,6 @@ def main():
         browser_url = "https://%s:%s/sabnzbd" % (browserhost, cherryport)
     else:
         browser_url = "http://%s:%s/sabnzbd" % (browserhost, cherryport)
-    cherrypy.wsgiserver.REDIRECT_URL = browser_url
 
     sabnzbd.BROWSER_URL = browser_url
     if not autorestarted:
--- a/sabnzbd/panic.py
+++ b/sabnzbd/panic.py
@@ -279,10 +279,12 @@ def error_page_404(status, message, traceback, version):
     <head>
       <script type="text/javascript">
       <!--
-      location.href = "%s"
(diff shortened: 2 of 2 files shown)


Safihre:
I was wondering why this was not merged?
From javascript we cannot help the HTTP/HTTPS part, but at least when I am on a mobile device it will not redirect me to 'localhost' when making a typing error.

However, I see that some variable names changed and it is now:
sabnzbd.BROWSER_URL instead of cherrypy.wsgiserver.REDIRECT_URL

Maybe it needs a new updated PR?

shypike:
The only reason we need http-->https redirection is because Apple made an error in Bonjour: it doesn't support https. For this purpose, 0.8.0 works. Other than that, I don't think there's any real need to have a comprehensive redirection.

Safihre:
But this Javascript is still wrong and should be changed like Sander suggested. Now it will redirect wrongly when not on the localhost.

shypike:
@Safihre : Ok, you are right. The change to panic.py is OK, but not the one for SABnzbd.py
Also, this one is for 0.7.x.
@sanderjo : can you create a pull-request for the panic.py patch? If not, I can manually apply it to "develop".

sanderjo:
@shypike Should I still create the PR for panic.py, or is it solved by "This branch is up-to-date with the base branch"?

shypike:
@sanderjo Please do, because this one is based on 0.7.x

sanderjo:
OK, working on it as we speak ... ;-)

sanderjo:
OK, PR sent (develop -> develop). A 404's redirect now keeps the FQDN in place

PR #175, opened 2014-07-31
Update favicon, apple-touch-icon and mstiles

MasterRoot24:
As per #174:

On Apple iOS devices, the boring default icon is shown for SABnzbd in the favourites:
screenshot
I have regenerated the `favicon`s, `apple-touch-icon`s and created `mstile`s for Windows 8 Metro shortcuts.


Safihre:
I think the browserconfig.xml is wrong.
Should this not be:
 <square70x70logo src="./mstile-70x70.png"/>
(so with the . in front of the /)

Safihre:
Also the favicon of the config pages bad now, compared to those of Glitter and Plush.. very blurry.
(on Chrome & Firefox)

shypike:
What's you advice? Undo this one?
Can you fix it? It's a bit above my expertise.

Safihre:
Fixed it!
Turns out Firefox and Chrome will not use the regular 'faveicon.ico' when a icon of 16x16 size is defined. Since it is not very usefull and it is ugly, just deleting that reference (and the file) fixes it.
See here: cd2ba68f4f2b9e9799cd31772677fc61b2e6a63a 

In my commit #236 I fixed it in all templates and also added the touch icons to Plush.

PR #179, opened 2014-09-15
Add Upstart session job example

CameronNemo:


shypike:
We're not adding this. The Ubuntu/Debian package by JCFP already contains a more advanced solution.

PR #187, opened 2014-12-04
Fix for OSX version detection. 

digitalhoax:
This was keeping "dark mode" from working correctly.


shypike:
Fixed already, forgot to close the request.

PR #190, opened 2015-01-01
DEBUG info if retrieving URL leads to an exception

sanderjo:
Example:

2015-01-01 15:16:17,924::DEBUG::[urlgrabber:135] Exception <class 'urllib.ContentTooShortError'> trying to get the url https://www.blabla.com/nzedb/getnzb/fe1a1474a5026b74b5d63b78634f6b055add149a&i=3&r=de390776ff47336d50e5585876aa1704


shypike:
Thanks, I've put this in "develop".
The plan is to have no more 0.7.x releases.
(I'll keep the request open, just in case.)

PR #191, opened 2015-01-01
Better handling if import of pystone module does not work

sanderjo:


sanderjo:
Ouch, still a debug line "print pystonemodule" in the patch. Wait a second before accepting.

shypike:
Thanks.
To prevent a messy history, I squashed your contribution to just one commit.

PR #199, opened 2015-02-09
Determine and show public IPv4 address (if +Debug is on)

sanderjo:
Determine and show public IPv4 address (if +Debug is on)

Useful for determining actual Internet connectivity, and remote access.


shypike:
Thanks

PR #204, opened 2015-03-09
Find and log unrar version, and give a Warning if below 5.00

sanderjo:
As unrar version 5.xx is needed for current downloads, find and log the unrar version, and give Warning if unrar version is below 5.00


shypike:
Thanks, I used your code to improve the current check.

sanderjo:
But the logging.warning() about unrar < 5.00 in SABnzbd.py is not there? And that was the whole point of my patch ... ;-)

shypike:
The warning is displayed at the left top of the screen most of the time.
I don't see how people would be able to miss that.
But I'll add the extra logging line.

Issue #212, opened 2015-05-18
OK to move IP addresses determination into module


sanderjo:
I want to move the (optional) IP address determination into a module sabnzbd.utils.ipaddress. 

Goals: clean up SABnzbd.py, and make the determination accessible from the Dashboard I'm writing.

I'll call the module functions if sabnzbd.cfg.log_level() > 1 (like now), and if/when Dashboard is clicked. In that way, api.ipify.org will not be called with basic usage of sabnzbd.

OK? If so, I'll send a pull request.

```
#!/usr/bin/python -OO

import socket

def localipv4():
    try:
        s_ipv4 = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
        s_ipv4.connect(('1.2.3.4', 80))    # Option: use 100.64.1.1 (IANA-Reserved IPv4 Prefix for Shared Address Space)
        ipv4 = s_ipv4.getsockname()[0]
        s_ipv4.close()
    except:
        ipv4 = None
        pass
    return ipv4

def publicipv4():
    try:
        import urllib2
        #f = urllib.urlopen("http://api.ipify.org")     # urllib has no timeout parameter, so:
        f = urllib2.urlopen("http://api.ipify.org", timeout=2)    # timeout 2 seconds, in case website is not accessible
        public_ipv4 = f.read()
    except:
        public_ipv4 = None
        pass
    return public_ipv4

def ipv6():
    try:
        s_ipv6 = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
        s_ipv6.connect(('2001:db8::8080', 80))    # IPv6 prefix for documentation purpose
        ipv6 = s_ipv6.getsockname()[0]
        s_ipv6.close()
    except:
        ipv6 = None
    return ipv6

if __name__ == '__main__':
    print localipv4()
    print publicipv4()
    print ipv6()
```

shypike:
Go ahead.
Maybe we should add an opt-out option in "Special" to cater to the paranoid.

PR #213, opened 2015-05-19
Move IP address determination into a module sabnzbd.utils.getipaddress

sanderjo:
As discussed here: https://github.com/sabnzbd/sabnzbd/issues/212


thezoggy:
should add if dns is working (internet is reachable)..  and if the internet can reach it..

sanderjo:
Explicit DNS checking? Good point. I'll add that to the Dashboard.

thezoggy:
prob have a list of 3 urls... check each three and only report fail if all 3 are unreachable?
have this check tie in with the ip check.. so it can be disabled/enabled with it.. since they would check the same sites?

sanderjo:
More stuff into the dashboard:
- DNS Check
- Disk speed measuring (only when Refresh is clicked), based on separate module

On startup (no disk speed):
sabnzbd dashboard - on startup

After clicking Refresh:
sabnzbd - dashboard - after refresh

With no network at all:
sabnzbd - dashboard ---- no network

shypike:
Looks good.
Little suggestion. When no network address is found, why not show something like "not found", "cannot be determined", instead of nothing. That way people know that something is wrong.

sanderjo:
Something like this?

sabnzd - dashboard - message if not good

sanderjo:
Or maybe better: only in the template presentation (still None in the python code), and ... colors! Implemented for the Public IPv4 address in the picture below:

sabnzbd - dashboard with colors

The cheetah template code is:

```
        <!--#set $odd = not $odd#-->    <tr class="<!--#if $odd then "odd" else "even"#-->">
               <td>Public IPv4 address</td><td>  
                   <!--#if $publicipv4#-->$publicipv4
                   <!--#else#--><B><font color="red">No public IPv4. That is bad</font></B>
                   <!--#end if#-->
               </td></tr>

```

thezoggy:
prob need to create the relevant translation strings and use them there?

shypike:
Already merged through other pull request.

PR #218, opened 2015-05-26
Handle Rev-files in the correct way

sanderjo:


shypike:
Thank you for this contribution.

Issue #209, opened 2015-05-06
0.8.x unable to use ssl connections


lrusak:
From the log.
```
2015-05-06 14:35:19,937::INFO::[downloader:399] 1@ssl-eu.astraweb.com: Initiating connection
2015-05-06 14:35:19,938::INFO::[downloader:399] 2@ssl-eu.astraweb.com: Initiating connection
2015-05-06 14:35:19,938::INFO::[downloader:399] 3@ssl-eu.astraweb.com: Initiating connection
2015-05-06 14:35:19,939::INFO::[downloader:399] 4@ssl-eu.astraweb.com: Initiating connection
2015-05-06 14:35:19,940::INFO::[downloader:399] 5@ssl-eu.astraweb.com: Initiating connection
2015-05-06 14:35:19,941::INFO::[downloader:399] 6@ssl-eu.astraweb.com: Initiating connection
2015-05-06 14:35:20,943::ERROR::[downloader:588] Connecting 5@ssl-eu.astraweb.com failed, message=200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled).
2015-05-06 14:35:20,945::ERROR::[downloader:588] Connecting 6@ssl-eu.astraweb.com failed, message=200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled).
2015-05-06 14:35:20,945::ERROR::[downloader:588] Connecting 1@ssl-eu.astraweb.com failed, message=200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled).
2015-05-06 14:35:20,946::ERROR::[downloader:588] Connecting 2@ssl-eu.astraweb.com failed, message=200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled).
2015-05-06 14:35:20,947::ERROR::[downloader:588] Connecting 3@ssl-eu.astraweb.com failed, message=200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled).
2015-05-06 14:35:20,948::ERROR::[downloader:588] Connecting 4@ssl-eu.astraweb.com failed, message=200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled).
```

Other info
```
[...]

PR #221, opened 2015-05-28
convert unicode format to plain ascii because of pyopenssl 0.14 bug

sanderjo:
Solves SABnzbd issue with SSL newsserver connections  https://github.com/sabnzbd/sabnzbd/issues/209

Root cause: PyOpenSSL 0.14 has a bug not accepting unicode: https://github.com/pyca/pyopenssl/issues/15


shypike:
Only the first two conversions are needed (for username and password).
Committed to "develop".
Thanks for your help.

PR #228, opened 2015-06-15
Add support for restricting servers by category

cdheiser:
Add support to restrict which servers can be used to download items for a specific category by adding categories to the server config.  Selecting "Default" will allow a server to be used in all categories.  Any server created before this change will also default to being allowed by all categories.

screen shot 2015-06-14 at 5 24 18 pm

### DIFF
--- a/interfaces/Config/templates/config_server.tmpl
+++ b/interfaces/Config/templates/config_server.tmpl
@@ -81,6 +81,15 @@
                     <span class="desc">$T('srv-optional')</span>
                 </div>
                 <div class="field-pair">
+                    <label class="config" for="categories">$T('srv-categories')</label>
+                    <select name="categories" multiple>
+                    <!--#for $cat in $cats#-->
+                      <option value="$cat" <!--#if $cat == "Default"#-->selected<!--#end if#-->>$cat</option>
+                    <!--#end for#-->
+                    </select>
+                    <span class="desc">$T('srv-explain-categories')</span>
+                </div>
+                <div class="field-pair">
                     <input type="submit" value="$T('button-addServer')" />
                     <input type="button" value="$T('button-testServer')" class="testServer" />
                 </div>
@@ -176,6 +185,15 @@
                     <label class="config" for="send_group$cur">$T('srv-send_group')</label>
                     <input type="checkbox" name="send_group" id="send_group$cur" value="1" <!--#if int($servers[$server]['send_group']) != 0 then 'checked="checked"' else ""#--> />
                     <span class="desc">$T('srv-explain-send_group')</span>
+                </div>
+                <div class="field-pair">
+                    <label class="config" for="categories">$T('srv-categories')</label>
+                    <select name="categories" multiple>
+                    <!--#for $cat in $cats#-->
+                      <option value="$cat" <!--#if $cat in $servers[$server]['categories'] then 'selected' else ""#-->>$cat</option>
+                    <!--#end for#-->
+                    </select>
+                    <span class="desc">$T('srv-explain-categories')</span>
                 </div>
                 <div class="field-pair">
                     <input type="submit" value="$T('button-saveChanges')" class="saveButton" />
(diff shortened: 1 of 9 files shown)


shypike:
Nice contribution.
I have a few remarks though.
First, the files are incorrectly indented. You are using tabs in the files, while we have the strict rule that only spaces are allowed for indentation. Please change your files. I'm afraid this requires a new pull-request so that the history isn't polluted and future "blame" runs produce odd results.

Second, I see an unexpected behaviour.
Suppose you have category "movies' are you set server "A" to download only "movies". Then you delete the category "movies". The result will be that server "A" will start download from all categories, instead of from no categories, which I would expect. This may be an unavoidable side-effect, because selecting no categories means that all categories apply. Obviously the latter IS expected behaviour. I'm not saying that the scenario above is wrong, just that it's a bit counter-intuitive and may lead to a server being used for categories which it wasn't supposed to download. Note that when having two categories and then deleting one, does produce expected behaviour.

cdheiser:
I'm not sure why fixing the few places where there are tabs requires a new pull request since it's only lines changed by myself.  Maybe I'm missing something?

For the second unexpected behavior, I've been contemplating how to make upgrading compatible while supporting the use case you mention.  Maybe a change to the configuration loading to initialize a server to have Default in the list until the config overwrites it?  Then a config that doesn't have a categories entry will behave as if it was allowed to use all categories, and any server that had all the categories removed would work as you'd expect.

shypike:
If it's just a few lines, then ok. I was under the impression that the whole file was converted.
Setting "default" initially would be a good idea.

cdheiser:
I've updated my editor defaults to avoid tabs in the future (was editing on a new machine).  And I've added 2 more commits to the pull.  1 to fix the tabs, and 1 to set default initially for servers.

thezoggy:
squash and rebase then force push to cleanup history

cdheiser:
I believe I did that correctly.  Changes squashed and rebased.

shypike:
Great. I'll do some testing tonoight and accept the pull-request.

shypike:
I found an issue. Scenario:
Job A has category "testme".
Server A has prio 0, is allowed to download "testme", but doesn't have the post
Server B has prio 3, is not allowed to download "testme" and has the post.
Server C has prio 8, is allowed to download "testme" and has the post.

The result is that nothing is downloaded, the job just sits idly in the queue.
At first some articles are tried (and fail) on server A, but server B and C are never tried.
Disabling server B makes the problem go away.

cdheiser:
This looks like it's due to how priorities are implemented with the try_list.

It looks like one option is to change this implementation to short-circuit article fetching and just add the server to the try list so that the priority checking works.  There might be an alternative where the server check just marks all appropriate try lists up front.

cdheiser:
Hah... I think I found a fix.

Down in Article.get_article, it does a server priority check.  Adding a simple check to see if the server is allowed during the priority checks seems to fix this issue in my testing.

I'll add a commit to the pull request, and if it looks good, I can squash it with the others.

shypike:
Ok, it would be nice if this could go into Alpha3.
You should also rebase against the develop branch because I had to make a little fix to config.py

cdheiser:
Ok... rebased and squashed.  Merge conflict with config.py resolved.

cdheiser:
Something changed after rebasing.  Let me figure it out.

cdheiser:
Yeah... I managed to drop my own change when I rebased.  That's all fixed now and my change is still working.

Safihre:
Going to ask it here:
When adding a server only the category "Default" is selected and the text says "Only use this server for these categories."
So this means if I have all servers just "Default" selected, none of my jobs that have a category will be downloaded? 
Or: "Default" means that the server will be used for all categories (this is what seems to happen). In that case the explanation text should be changed, it's not very clear.

shypike:
Default = use server for all categories.
This was the only way to implement this without nasty surprises when a category is deleted.
No categories selected would then allow all categories. Now you'll need to select at least "default", which is also the maiden situation for each server.

Safihre:
I like the implementation, just think the explanation text should then make that clear. Now it sounds like it will only be used for the "Default" category and that every time you add a new category you also need to add that category to a server.

PR #233, opened 2015-07-03
if we get anything else than a plain IPv4 address, return None

sanderjo:
As discussed here: https://github.com/sanderjo/sabnzbd/commit/248bfbde996f2dc58c1eaad2d2eadc8aff15c8bb#commitcomment-11839843


shypike:
Thanks

Issue #234, opened 2015-07-03
no unrar => UnboundLocalError: local variable 'version' referenced before assignment


sanderjo:
On a system with no 'unrar'/'rar' installed, starting SABnzbd.py result in this error / exception / traceback
```
pi@raspberrypi ~/git/sabnzbd $ git pull
Already up-to-date.

pi@raspberrypi ~/git/sabnzbd $ git branch 
* develop

pi@raspberrypi ~/git/sabnzbd $ ./SABnzbd.py 
Traceback (most recent call last):
  File "./SABnzbd.py", line 1896, in <module>
    main()
  File "./SABnzbd.py", line 1116, in main
    sabnzbd.newsunpack.find_programs(sabnzbd.DIR_PROG)
  File "/home/pi/git/sabnzbd/sabnzbd/newsunpack.py", line 146, in find_programs
    version, original = unrar_check(sabnzbd.newsunpack.RAR_COMMAND)
  File "/home/pi/git/sabnzbd/sabnzbd/newsunpack.py", line 1705, in unrar_check
    return version, original
UnboundLocalError: local variable 'version' referenced before assignment

pi@raspberrypi ~/git/sabnzbd $ which unrar

pi@raspberrypi ~/git/sabnzbd $ unrar
-bash: unrar: command not found

pi@raspberrypi ~/git/sabnzbd $ rar
-bash: rar: command not found

```

sanderjo:
OK, problem solved in 

```
2015-07-04 00:31:46,951::WARNING::[SABnzbd:509] unrar binary... NOT found
```
and warning in the GUI:
```
	WARNING:	unrar binary... NOT found
```
I'll send a patch request

shypike:
True indeed.
If committed an alternative fix, tackling the problem where it is caused, in unrar_check() itself.
See: https://github.com/sabnzbd/sabnzbd/commit/a2c1b47a8fb538d6c7ce663b9ebe666791cca146

PR #235, opened 2015-07-03
handle correctly if no (un)rar is found

sanderjo:
as reported here: https://github.com/sabnzbd/sabnzbd/issues/234


shypike:
Thanks, but I chose a slightly different solution: https://github.com/sabnzbd/sabnzbd/commit/a2c1b47a8fb538d6c7ce663b9ebe666791cca146

PR #240, opened 2015-07-06
Fixing the favicons in all templates

Safihre:


thezoggy:
the legacy of the fav+touch icons is due to html5 boilerplate. if you want some of the info visit:
https://github.com/h5bp/html5-boilerplate/blob/5.2.0/dist/doc/extend.md#apple-touch-icons

but note that this practice is no longer used as with the age of modern browsers and how quick ios peeps have upgraded.. a lot of the legacy/hacky stuff can be dropped..

reasons:
https://mathiasbynens.be/notes/touch-icons

conversation:
https://github.com/h5bp/html5-boilerplate/issues/1367

so in the end, just drop all the variants and just use one..

PR #242, opened 2015-07-10
Added requirements.txt

MartijnBraam:
I've created a requirements.txt file so sabnzbd can be installed by using pip instead of the packagemanager of the distribution.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,8 @@
+Cheetah>=2.0.1.0
+feedparser
+gntp
+pyOpenSSL>=0.11
+configobj
+pydbus
+support
+-e hg+https://bitbucket.org/dual75/yenc#egg=yenc


thezoggy:
note that the yenc you link is only for python 2.6.

MartijnBraam:
It's the same source thats packaged in python-yenc in debian, it seems to work fine on my sabnzbd installation

thezoggy:
which is for python 2.6

MartijnBraam:
I've searched through some issues on the project and it seems to support python 2.7 or do you mean that the yenc dependency is only for running sabnzbd on python 2.6?

shypike:
feedparser, configobj and gntp are all included in SABnzbb's source distribution.
So adding them to requirements.txt isn't really needed.
On the other hand, SABnzbd is quite useless without par2cmdline and unrar...

Safihre:
Let's close this and also #366, it's just not going to happen because sabnzbd depends too much on tools like par2 and unrar that cannot be retrieved this way.

PR #246, opened 2015-07-19
Adding OZnzb to Glitter

Safihre:
Changed OZnzb so it will allow changing of up/down voting.
For out of retention it will only list the users servers, not all of them.
In Glitter you can only give a audio/video score once (just like on the website). Plush allows changing of audio/video rating and shows that change, but it is not actually changed on the website.

untitled


Safihre:
Minor bug: in IE8 and lower, the 'comment'/'other problem' text-field will not show up when it's selected, because it does not support the new CSS functions.
Will try to make a fix for that later.

Safihre:
Nevermind, just fixed the IE8 bug :)

shypike:
The items in the queue do not show rating info. Is that a concious choice?

Safihre:
Just noticed a small mistake I made for Firefox, fixed that a minute ago!
In Chrome you should see it in the queue? I am testing it on Windows and Ubuntu

shypike:
Great addition. Thank you.
Merged it "squashed", so I have to close the request manually.

PR #248, opened 2015-07-22
Updating setup wizard and Glitter upgrades

Safihre:
Updating Wizzard to be proper HTML and look like Glitter
Also includes small fixes to Glitter.


Safihre:
Next is the Config (minor changes expected), so that is why the bootstrap styles are placed in the config folder!

Safihre:
Glitter:
Added option to add name when adding NZB.
Added pause/resume tooltips to the main pause-buttons and the job pause-buttons.
Code for Delete-All-Orphaned cleaned so it could also allow Add-All or Selective-Add/Delete-Orphaned (but not impelemented).

PR #258, opened 2015-07-27
Fixing HTML issues in config

Safihre:


interfaces/Config/templates/config_notify.tmpl
```diff
@@ -142,7 +142,7 @@
         <div class="col2">
             <h3>$T('section-NC')</h3>
         </div><!-- /col2 -->
-        <div class="col1" <!--#if $have_ncenter then '' else 'style="display:none;"'#-->">
+        <div class="col1" <!--#if $have_ncenter then '' else 'style="display:none;'#-->">
```
shypike: This change is still not right;  in case of no ncenter, the end result will be:
<div class="col1" ">
Instead, the " at the end of the line should be removed.

interfaces/Config/templates/config_notify.tmpl
```diff
@@ -201,7 +201,7 @@
         <div class="col2">
             <h3>$T('section-OSD')</h3>
         </div><!-- /col2 -->
-        <div class="col1" <!--#if $have_ntfosd then '' else 'style="display:none;"'#-->">
+        <div class="col1" <!--#if $have_ntfosd then '' else 'style="display:none;'#-->">
```
shypike: Same as previous remark.

interfaces/Config/templates/config_notify.tmpl
Safihre: Fixed in 43088a6

Safihre:
Squashed changes.

shypike:
Is this one overruled by #260 or is it still valid?

Safihre:
It is included in #260, but even without redesigning config these changes should be made.

PR #260, opened 2015-07-27
Redesigning the Config

Safihre:
After updating the wizzard to look more like Glitter, I now have made a new version of the Config.

Initially I did not want to change too much, but the changes ended up more extensive than I anticipated. I did not completly ported the 'Glitter style' so it can still integrate somewhat with Plush.
However, I did go for the whole grey background with white foreground and bigger button's.
Since not all link's fit when the screen is made somewhat smaller (<1200px), I replace them with icons of which the text is displayed when you hover over them. There are still optimizations to be made for mobile devices, but I wanted to have input on this version first!
*I have only made changes to the CSS and the header, the config HTML and Javascript is still the same!*

Examples below:
**> 1200px**
big

**< 1200px**
smaller

**Mobile**
smallest


thezoggy:
before you submit pull requests, you should squash and rebase. and even if afterwards, you can squash and push+force. as rewriting history before it makes it into the repo is okay (before shy pulls it). this cleans up the history and eliminates uneeded changes. sab repo is already 90megs... (mainly from the binaries back from bzr days)... but anyways it also makes it easier to review changes.

Safihre:
Squashed them!

shypike:
Big change, please give some time to walk through it.

shypike:
Firefox:
When I access Config from the primary skin, I get a menu at the left side, reduced to "General" and "Help".
When I access it from the secondary skin (sabnzbd/m), the menu is at the top, with expanded text.

Safari:
Both primary and secondary skins show the Config menu at the top.
However, this time it's only small symbols which expand when the mouse hovers over them.

Chrome:
Both primary and secondary skins show the Config menu at the top.
Only text blocks are visible, no small icons.

I'm not too enthusiastic about a top menu. Most desktops are wide nowadays and height is scarce resource. Which is the look the new Config is supposed to have? It's different in three different browsers and in Firefox it fails in the primary skin.
Could Firefox be suffering from a cached old CSS?

shypike:
OK, the odd Firefox look was solved by clearing the cache. Something to keep in mind for people upgrading! Can that be avoided?

shypike:
Right, whether or not icons are shown depends on the width of the window.
Clever but it wasn't immediately obvious to me.

Safihre:
We can avoid the caching issue by giving the file a new name. Like config.css, then all browsers will reload. 
Basically on regular desktops (with minimum res of 1366) nothing really changes. Only when the browser is made smaller or on mobile devices the icons will show. 
But if the browser is made smaller than 1200px (that's the switch to the icons, see my screenshots) on the old configuration, it also lacks space in width and horizontal scrolling is nessecary there as well. 
I considered vertical menu on desktops, but that kind of feels against the idea of the redesign: making it look the same as wizard and Glitter.

thezoggy:
cache busting is pretty easy, just append `?####` at the end of the file.. in sb we used the pid of the app running. makes it unique between runs and static per session. as we could push out ui changes and never have any cache issues.

shypike:
I added the ?### construction and it works fine.

Safihre:
@shypike you didn't just add ### but replaced those with the pid or something like that, right? :P

shypike:
style.css?$pid
Of course in api.py build_header(), $pid is set to os.getpid()
That last change is needed as well of course.

Safihre:
(can squash the commits before pulling them, just wanted to show that I updated it to allow changing of the host-name after the recent update)

shypike:
Please squash, after that I'll pull it.

shypike:
I see now that the <textarea> for "Personal notes" box for the server needs a cols="60" parameter. It was wide in the previous config, now not anymore. The user can drag it wider using the bottom right corner, but it's a bit narrow now.

thezoggy:
Personally id just apply width of 400px to it.. actually already made that
change locally. 

 

.field-pair textarea {width: 400px;}

Safihre:
Squashed.

Added also:
* Server description/notes to Add Server (was not there yet)
* Width of Server Notes
* style.css?p=$pid

thezoggy:
looking at the sab config build script,
```
@import url('normalize.css');
@import url('formalize.css');
@import url('jquery-ui-1.8.18.custom.css');
@import url('jquery.qtip.css');
@import url('config.css');
```

looking at your css,
```
jquery-ui-1.8.18.custom.css
config.css
your edits
```

thus i'd be concern of browser differences a bit since normalize/formalize isnt used... theres a reason why frameworks use a reset/normalize css.

thezoggy:
looking at this in firefox:
sab ff

vs ie11
sab ff

is processed result broken suppose to show up? it does for the date sorting but not the other two...
notice the minor differences between firefox/ie.. stuff doesnt align.. different font sizes..

thezoggy:
if i type something in the test data field then erase it, the 'default' is used.. so just looks like something isnt initialized correctly (also happens on 0.7.20)

thezoggy:
also seeing tracebacks
```
2015-08-01 17:47:58,911::ERROR::[_cplogging:201] [01/Aug/2015:17:47:58] HTTP Traceback (most recent call last):
  File "C:\Program Files (x86)\sabnzbd_0.8.x\cherrypy\_cprequest.py", line 656, in respond
    response.body = self.handler()
  File "C:\Program Files (x86)\sabnzbd_0.8.x\cherrypy\lib\encoding.py", line 188, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "C:\Program Files (x86)\sabnzbd_0.8.x\cherrypy\_cpdispatch.py", line 34, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\interface.py", line 416, in tapi
    return api_handler(kwargs)
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\api.py", line 100, in api_handler
    response = _api_table.get(mode, (_api_undefined, 2))[0](name, output, kwargs)
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\api.py", line 644, in _api_eval_sort
    path = sabnzbd.tvsort.eval_sort(value, title, name, multipart)
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\tvsort.py", line 1224, in eval_sort
    path = sorter.get_final_path()
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\tvsort.py", line 772, in get_final_path
    if self.get_values():
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\tvsort.py", line 789, in get_values
    self.date_info['year'] = self.match_obj.group(3)
AttributeError: 'NoneType' object has no attribute 'group'
```

Safihre:
Regarding the CSS: don't forget that we also use the *bootstrap.min.css*, which has normalize.css included. I did not want to change any HTML of the config-pages so I could not use all the bootstrap-classes that make sure everything is aligned, will have to do some manual fixes. Also need to verify on more mobile devices.

shypike:
@Safihre There's no dogma against changing the html code of Config.

Safihre:
I know.. but if I know myself that if I start changing one input element, I would feel obliged to change them all :P

PR #266, opened 2015-08-01
Customized Add-NZB in Glitter, added extra tooltips

Safihre:


shypike:
In Firefox (on Mac) the Add dialog looks odd:
Snapshot

Safihre:
Force-refresh it, probably the cache at work again.

I think I will also add anti-cache to Glitter, but then with the version and build number. This way the cache is kept in between versions and start-up is just that tiny bit faster.

shypike:
The p=$pid trick solved the problem. Please add that fix too.

Safihre:
Yeah you're right! Will use the $pid.
Since both cache and cherrypy have to come from the disk, doesn't seem to have any benefit.

Safihre:
Updated.

PR #270, opened 2015-08-05
Saving Glitter refresh-rate and page-limits in general config. Adding a central speedhistory. Adding icons to Config.

Safihre:
refresh_rate and history_limit were already in the config but not used. Added queue_limit.
This way these settings are the same no matter what device or browser you use to connect to your SAB system.

Adding icon's to button's in the Config. 
TODO: Layout of RSS-details-page in the Config.


Safihre:
*Added 8ea182b (could not branch anymore since these changes build upon the previous ones)*

Adding a central speed history within SAB, a feature many download programs have and I think SAB could use! This way you can actually see what happend when the browser was not open. 
When there is no download-activity nothing really happens, it only calls and updates (in case 0's need to be added) when the main.tmpl file is called. In Glitter this is only once.
It does add consecutive numbers to a list, but this should not be a memory hog.

shypike:
@Safihre 
Alas, a "division by zero" crash at startup. Sent you the details by email.

Safihre:
@shypike Fixed and squashed the fix into the commit!

shypike:
BTW: is it a good idea to put refresh_rate and queue/history in the INI file?
I can imagine that you'd want different settings depending on the device (and screensize) you're using and whether you're local or remote.

thezoggy:
doesnt plush store it in cookie/localstorage?

shypike:
@thezoggy 
Plush uses cookies, not the INI file.

Safihre:
Glitter also uses cookies now. But I found these 2 (refresh_rate and history_limit) in the INI and thought it would be a good use.
Also this surved the second purpose that in order to create a speed-history I need to know the refresh-rate of the interface in the Python code.

shypike:
I think I prefer cookies because that leaves more flexibility and fits better should you use two different interfaces. The refresh-rate is a bit unfortunate: I understand the need to know it locally. For now there would be no issue, because the other skins use neither the INI values nor the speed-history.

Safihre:
Let's make a checkbox then in the Glitter *Status and interface options* window:
*Use global interface settings* (default: yes)

Since Glitter uses already both Cookies and INI, that checkbox would just turn of the update to the INI.

shypike:
Doesn't that leave you with a problem for the refresh rate?

Safihre:
Yes, it would use the default (or last applied) setting so the speed-history not lining up would be an unavoidable side-effect.
In the case of INI-refresh rate != cookie-refresh-ratethe speedhistory should also revert to only being locally (like it was before).

shypike:
Small issue.
The Status windows shows "Backup" for servers that do not have priority 0.
Plush shows just "Enabled".

thezoggy:
should use localstorage over cookies..  also sab has had a history of 'illegal cookies' with plush...

http://forums.sabnzbd.org/viewtopic.php?t=2399
https://www.google.com/search?q=illegal+cookie+sabnzbd&ie=utf-8&oe=utf-8#q=illegal+cookie+sabnzbd+site:forums.sabnzbd.org
http://sickbeard.com/forums/viewtopic.php?f=4&t=8466

Safihre:
Interesting! Never even thought of that.. Localstorage would  indeed be much more applicable for this situation. 
Will implement that for the next commit.

PR #275, opened 2015-08-09
Opt-out for global options in Glitter. RSS config page fixup. Config restart SAB fixup. Fixes for mobile/tablet.

Safihre:
1. Possible to use local settings instead of global settings from the INI in Glitter (default is to use global settings). Now using localStorage instead of cookies for saving settings. 
Also fixed Enabled/Backup for Servers in status-window for Glitter.
2. RSS page in config updated to new style.
3. Config restart now shows a splash screen while it's restarting, instead of redirecting to the restart-page. On the General-tab choosing 'Save settings' will now ask if you also want to Restart SAB as well.
4. Small fixes to Glitter and Config after more testing on iPad and mobile devices.


Safihre:
4. Small fixes to Glitter and Config after more testing on iPad and mobile devices.

shypike:
Nicely done, almost perfect.

- After changing the NZB or API key, the Save button is disabled, no matter what else I change after that.
- You suggest a restart for every change in Config->General although it is only required for host name/port, https and Skin settings.
  - The index will have the variable $restart_req set when a parameter requiring a restart has been touched
  - Should you be using a different API call for Save, please tell me so that I can add this variable
- You suggest a restart after saving Config->General, but the Restart dialog still appears (a bit overdone but acceptable)
  - An alternative for the extra prompt could be to enable the Restart button only when the variable $restart_req has been set

Safihre:
I implemented the code that will redirect to the new port in case it is changed (will test it a little more first).
Unfortunately due to 'Access-Control-Allow-Origin' I am not allowed to check from the old port if the new port is up and we can redirect back to the homepage as soon as it's ready.
So I will make it now that after 7 fails (35 seconds) it will just redirect to the new url, in the hopes that nothing else went wrong..

Safihre:
Updated and squashed:
* Will only ask for restart if the changed parameter requires it (helped by @shypike for the Python part).
* Removed all Restart SABnzbd buttons except on the General page, fixed Firefox issue.
* If port has changed the redirect will be to the new port, although only after 40 sec due to 'Access-Control-Allow-Origin'. Otherwise the redirect happens as soon as SAB is restarted.

shypike:
There are still issues.
Scenario:
- I setup https on port 8080 and http on port 8085
- I have a connection on https://localhost:8080/sabnzbd
- I restart (without changes).
- The UI tells me its waiting for a connection to https://localhost:8085/sabnzbd
- Auto-browser starts up a new UI at https://localhost:8080/sabnzbd
- The old UI keeps waiting for a connection, after a minute or so is says "SSL connection error"

Safihre:
So if I get it right to resolve this the UI should:
Check current browser connection, if https then read the https port and redirect to that one with https (also verify if https still checked). 
If current is http then redirect to http port.

shypike:
Something to that effect.
It maybe tricky to do because you need to predict what the complicated startup code of SABnzbd is going to do.

Safihre:
What else will it do? When would this fail?

Safihre:
Indeed cannot predict everything, but the standard restart page is even worse.. Doesn't take any ports or https in account. Indeed not a problem when browser on startup is turned on, but that's no help on remote installations.

shypike:
Implement what you can. You're right that your changes are better than the current situation.

Safihre:
Implemented:
Are we on General page, otherwise use protocol and port of URL we currently use.
If on General page and HTTPS is checked and we are currently visiting from the HTTPS URL, then redirect to HTTPS with the given HTTPS port
Otherwise: HTTP with the HTTP port.

Could you check if it works? I seem unable to get the SSL setup correctly on my Windows machine.

shypike:
Same scenario. Now the Restart page thinks it should go to http://localhost:8085 (note the missing /sabnzbd !)
While not incorrect, it's not the best destination. Also, it doesn't actually time-out and make the http connection.

Safihre:
Oke I got HTTPS working on my machine and think I have now covered everything.
* It will now redirect to the right port and right protocol at the right time.
* There is an exception build-in for when you connect to Config using HTTPS and then uncheck it, the redirect would then try through AJAX the HTTP but this is not allowed. So in that case after 40 seconds you get redirected automatically to the HTTP one.
* It will redirect to /sabnzbd/, and if you were using /sabnzbd/m/, it will redirect back to that one.

shypike:
It works quite well now.
When going from HTTP-only to HTTP+HTTPS it will remain on HTTP, while SABnzbd will auto-start the browser with HTTPS. This behavior is quite acceptable: the user keeps a working connection.
This is a welcome improvement for remote usage.
Thank you.

Safihre:
So if by default uniconfig=on, why keep the skin-specific config's? 
Time for a cleanup for 0.8?

shypike:
uni_configi is there for people who (did?) swear by the older smpl skin.
smpl is/was popular due to its raw speed.
Let's leave it like it is for 0.8.0 and see what feedback we get.

thezoggy:
uni_config was there so if people freaked out with the universal config and wanted to go back to the whole each skin did the config (plush/smpl/classic/etc).. which really sucked to maintain and support (as trying to direct a user to change a setting meant you had to know what skin they were using). 

anyways with 0.8.x uni_config does not need to be there anymore.. as it should be the only config 'skin'.

Safihre:
:thumbsup: clean-up!
They can continue to use smpl, just with regular Config when they have to change some settings.

PR #276, opened 2015-08-09
Adding nzo_id to addurl

Safihre:


thezoggy:
when I originally asked if addurl would have an nzo_id I was told even if it did return it, it would change when the nzb finished being fetched and the job got a new ID for the downloading of the articles

@shypike  does this still hold true?

thezoggy:
also, should add ability to override `new_nzb_on_failure` when adding by url.. so param to force it on/off.

use case scenario, sonarr can do the failed handling automatically and sab getting a different nzb (release) than it expected.. can have unexpected results.. thus for sonarr you'd want to be able to turn off that feature...  but you'd want it on globally so that other apps can still benefit from it.

shypike:
@thezoggy Last I checked it did, that's why I didn't implement this before.
The extra parameter is possible.
However both issues requires a lot of parameter passing over several layers. I'll have a look at it.

Safihre:
I'll just give up on interfering in the python code and leave it to the professionals

shypike:
I appreciate the effort anyway. This particular case happens to be a bit nasty.

PR #295, opened 2015-08-25
Updating KnockoutJS to latest version

Safihre:
Performance improvements for Firefox and Chrome.


shypike:
knockout-latest.js ??
Won't we be having browser cache problems when this one is updated later?

Safihre:
No, this is a Cheetah include and not a script tag. All Javascript gets included in the main file. It's (slightly) faster than having the browser send requests for each file to Cherrypy who then has to check if the file has been modified.

Safihre:
But I can change the filename to better reflect the version we use!

PR #296, opened 2015-08-25
GUI for Happy Eyeballs

sanderjo:
As discussed. 
And ignore previous pull request


Safihre:
Not adding the change of ipv6 check server to this?

PR #299, opened 2015-08-27
Adding animations to Glitter

Safihre:
A little more fun!


shypike:
Is this animation bearable on slow systems, like an older tablet or smartphone?

Safihre:
It's all css animations, so if the browser supports them (above android 3 and ios 7) it will show them and I can't really filter for device speed. 
But since they all animations only last 0.2 seconds,  they are short so bearable even if stuttered.

Safihre:
@sanderjo 
Is it not also a good idea to display the value of *_EXTERNAL_IPV6* in the dashboard?
External IPv6 connection: Yes/No

thezoggy:
why not show status icon for v4/v6 reachability. and then report back users public ip for each next to it

PR #300, opened 2015-08-27
Removing Mobile template

Safihre:
There is no need for it with Glitter being responsive-design.
 
@shypike I also removed the empty directory of Classic, although I did not really understand why it was still there anyway?


shypike:
There was a reason, I'll find out when building :)

PR #304, opened 2015-08-31
Fix for hover problems in Plush

### LINKED ISSUE #86: Auto refresh causes mouse menus to disappear
While in SAB's webpage, if I click on a menu such as to change the priority of an item, if the page refreshes while the menu is expanded, the menu will disappear which will require me to re-click on it to expand it again. Kind of like pressing escape while in a menu.

Safihre:
This fixes #86 (a problem which also occurred for me in Chrome, not just FF and is horribly annoying).
Why was this function added? 
I tried many scenarios in FF/IE/Chrome and deleted stuff with short and long refresh-times but it always seemed to continue refreshing when the hover was off the queue again.

### DIFF
--- a/interfaces/Plush/templates/static/javascripts/plush.js
+++ b/interfaces/Plush/templates/static/javascripts/plush.js
@@ -560,17 +560,6 @@ jQuery(function($){
     function(){ $.plush.skipRefresh=false; } // out
   );
 
-  // refresh on mouseout after deletion
-  $('#queue').hover(  // $.mouseout was triggering too often
-    function(){}, // over
-    function(){   // out
-      if ($.plush.pendingQueueRefresh) {
-        $.plush.pendingQueueRefresh = false;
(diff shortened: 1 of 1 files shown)


shypike:
The code was introduced in 2009 with this commit: 52f738770ba152735edc378cb4940377fca68477
I hope this makes it clear for you (it doesn't for me).

Safihre:
It seems that before when you deleted a job, instead of being removed it would fade to 25% transparency (but stay visible) until the next queue refresh.
But that functionality and it's code has now been (partially) removed and replaced by the popup that asks *Delete* or *Delete+Files*.
Line 598-615 is commented out in the current *plush.js*, so this change should be safe.
https://github.com/Safihre/sabnzbd/blob/develop-plushfix/interfaces/Plush/templates/static/javascripts/plush.js#L598-L616

PR #305, opened 2015-08-31
Plush fix for reporting

### LINKED ISSUE #197: small bug when reporting quality in history beyond 1 page
This isn't a very big deal, but it is consistent and reproducible.

In the "History" if you go back beyond page 1 and attempt to report the audio and/or visual quality of a download, the moment you do either it takes you back to page 1 yet the page still indicates you are on whatever page you were on before.  That is, it shows you the X  most recent listings in the history (ie. "page 1") but still highlights page Y in the navigation at the bottom.

Hope this helps.  Sorry if it was already reported.  Cheers.

Safihre:
Fixes #197.

### DIFF
--- a/interfaces/Plush/templates/static/javascripts/plush.js
+++ b/interfaces/Plush/templates/static/javascripts/plush.js
@@ -1104,7 +1104,7 @@ $.plush.histprevslots = $.plush.histnoofslots; // for the next refresh
       type: "POST",
       url: "tapi",
       data: {mode:'queue', name:'rating', value: nzo_id, type: videoAudio, setting: $(this).val(), apikey: $.plush.apikey},
-      success: $.plush.RefreshHistory
+      success: $.plush.Refresh
     });
   });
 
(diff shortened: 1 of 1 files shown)


Safihre:
Independent of this problem and this fix, reporting is horrible in FF when refresh rate is 1second. 
Since the history does not get paused on hover, the dropdown disappears almost immediately.

Pausing the history on hover does not seem like a good idea to me, since then it probably would be paused all the time..

shypike:
I don't see any difference.

Safihre:
Try what's described in the bug, it happens in all browsers. 
Go to any page that's not the first and there rate a download. Before this any rating action (thumbs up/down, video, audio) will jump the history back to the first page and leave the pagination at the wrong page. 
Now you will stay on the same page!

Safihre:
Could you reproduce it or you're trusting me on my baby blue eyes? :D

shypike:
Your eyes did the trick.

PR #306, opened 2015-09-03
Bugfixes for A4

Safihre:


thezoggy:
also could fix 'generic' sorting title to say Movies. since all the other sorting match their enable label

Safihre:
@shypike Now 0.8 needs new translations anyway, isn't this indeed the perfect time to change it?

shypike:
@Safihre Careful, I also don't want untranslated strings in highly visible places. Some of the translations haven't been changed for months, so this change would be an issue.
Corrections such as these I make in the "en.po" pseudo-translation file.

PR #307, opened 2015-09-03
Improve CPU usage on Firefox

Safihre:
*You were very fast with the previous merge :grin: Was still working on this. Maybe leave this PR open until you build Alpa4, so I can add stuff if I noticed any other bugs etc.*

Firefox uses very high CPU (up to 30%) for anything animated. All the others (IE, Chrome) use much less (max 2-7%).
Disables animations on progress-bar and make the History-'processing' a
block animation.
Can be removed if Firefox animation performance gets better in the future..

### DIFF
--- a/interfaces/Glitter/templates/static/stylesheets/glitter.css
+++ b/interfaces/Glitter/templates/static/stylesheets/glitter.css
@@ -1384,6 +1384,46 @@ td.delete label,
 }
 
 /***
+    SPECIAL FOR FIREFOX
+    It uses very high CPU for anything animated (Sep 2015)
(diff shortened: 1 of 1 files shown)


thezoggy:
test case: http://stackoverflow.com/questions/24627066/css3-keyframe-transition-animation-high-cpu-usage-on-mozilla-firefox-30-0
tried to find relevant mozilla bug to post, but there are just too many with css animations :(

Safihre:
Indeed, it is surprising how even IE9 outperforms FF.
Especially when 2 animations are active (progress-bar and the loading icon in the history) it seems to go wild.

PR #309, opened 2015-09-04
Cleaning skin texts

Safihre:
When changing skintext.py I noticed some texts are not really used anywhere, so I decided to write a script that actually checks which strings from the skintext.py actually are used.
Turns out there are more than 100 texts that are not used (anymore). Some obvious, like the 'X' button that is now replaced everywhere with an icon. But also 'Exit Wizard', which is not used because the first page of the wizard where this text is, there is no language system active yet.

I also change the label of genericSorting to movieSorting, but did not change the text itself to keep the translations active.

```python
from sabnzbd.skintext import SKIN_TEXT
import os

# Make one giant string of all the text for semi-efficient search
main_giant_file = '';

for dirname, dirnames, filenames in os.walk('interfaces') :
    for filename in filenames:
        # Only .tmpl and .htm(l) files
        if '.tmpl' in filename or '.htm' in filename:
            with open (os.path.join(dirname, filename), "r") as myfile:
                data = myfile.read().replace('\n', '')
            main_giant_file = main_giant_file + data;
            
# What do we have here?
not_found = [];
for key in SKIN_TEXT:
    if key not in main_giant_file:
        not_found.append(key)
    
# What line?    
not_found_numbers = [];
for lookup in not_found:
    with open('sabnzbd/skintext.py') as myFile:
        for num, line in enumerate(myFile, 1):
            if "'"+lookup+"'" in line:
                not_found_numbers.append(num)
 
# Make a normal human sorting, based on file-line              
re_sorted = sorted(zip(not_found_numbers,not_found), key=lambda pair: pair[0]);

# Print!
for a,b in re_sorted:
    print a,b

```

### DIFF
--- a/interfaces/Config/templates/_inc_header_uc.tmpl
+++ b/interfaces/Config/templates/_inc_header_uc.tmpl
@@ -8,7 +8,7 @@
         #if $pane == "Folders" then $T('cmenu-folders') else ""#
         #if $pane == "Switches" then $T('cmenu-switches') else ""#
         #if $pane == "Servers" then $T('cmenu-servers') else ""#
-        #if $pane == "Scheduling" then $T('Plush-cmenu-scheduling') else ""#
+        #if $pane == "Scheduling" then $T('cmenu-scheduling') else ""#
         #if $pane == "Email" then $T('cmenu-notif') else ""#
         #if $pane == "Categories" then $T('cmenu-cat') else ""#
         #if $pane == "Sorting" then $T('cmenu-sorting') else ""#
@@ -252,7 +252,7 @@
                 <li>
                     <a href="${root}config/scheduling/" #if $pane == "Scheduling" then 'class="active"' else ""#>
                         <span class="glyphicon glyphicon-time"></span>
-                        <strong>$T('Plush-cmenu-scheduling')</strong>
+                        <strong>$T('cmenu-scheduling')</strong>
                     </a>
                 </li>
                 <li>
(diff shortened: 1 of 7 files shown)


shypike:
Nice work.
However, with some testing I found that the first Wizard screen contains an error which makes some of texts not obsolete. After choosing a language, you can go back to the initial screen and now some texts are translated while others are not. (Example: the "Start Wizard" button.)
I'll need to fix that first before we remove the texts.

Safihre:
I as actually thinking to remove all translations from the first page of the wizard. It's sort of a nice universal landing page. 
Did you try using the back button from any step of the wizard? I remember also from one of the further steps backwards something is wrong when not on English.

shypike:
The Back button works fine. The best would be that the first page changes language as soon as you pick one, but that could be tricky.

Safihre:
Maybe that's a little too much for a page that users will see only once for a few seconds.. 
But I will add the translations back and also for the 2 other texts on the first page that are not in skintext yet!

Safihre:
I had time to test the wizard and like I remembered, the back button does not work as excepted:
If you select German and then click *Start Wizard* and then press *Zuruck* on the server-screen (one.html), the start screen is in English again. 
Same happens when I select German, fill the server-info correctly, go to *Schritt 2*. If you then press *Zuruck* to go back to the server-info, it's also in English again...

Safihre:
Aah, mystery solved by reading the actual back-end code:
For the first page the system-language takes precedence over the cfg.language on Windows, makes sense.
I applied a fix for Step 2->Step 1, where always the input was used instead of cfg.language if there was no input (see 78caa57).

Now all is fine :sunglasses:

PR #317, opened 2015-09-07
Interface fixes

Safihre:
Some updates after a chat with @thezoggy. More to come.


shypike:
@Safihre "More to come", so I'd better wait until you're ready?

Safihre:
All the changes are minor and could be in a next (beta :smirk:) release, so if you're on the verge of releasing A4 then please go ahead :grin:

PR #322, opened 2015-09-09
Interface fixes part 3

Safihre:
Just leave it open until you make Alpa4 ;-)


thezoggy:
some of these need to be in before alpha4

shypike:
A4 is delayed until next week, so take your time to improve.

Issue #46, opened 2012-06-13
Feature Request: Updates for Pause feature


thefotios:
Here are a couple of ideas I've had for the Pause button.
 * Instead of counting down, show "Paused until ..."
 * Provide a way to pause for "n hours" instead of just minutes
 * Provide a way to pause "until xx:xx"
   * This might be able to be done by Pausing and then creating a one-time scheduled event to unpause

shypike:
Fixes and adding option for custom pausing #342

PR #342, opened 2015-09-28
Fixes and adding option for custom pausing

Safihre:
As requested by #46 (granted, a little overdue :smile:)
The datetime-interpreter works very well, but is 30kb in size so will only be loaded when the button 'Custom' is clicked. 
The interpreter supposedly has multi-language interpretation, but I already noticed that this is far from complete, therefore only supporting English input for consistency. (For example 'tomorrow' in Dutch is not recognized)


thezoggy:
moment js ( http://momentjs.com/ ) would have also been good to use for this

Safihre:
As far as I could tell from testing and the docs it cannot parse 'human' text like DateJS.
Now you can just type things like "tomorrow 7pm" or "monday 20:00", or "300 hours" and it will work.

PR #343, opened 2015-09-28
One-time notice of new skin

Safihre:
Only for existing users

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -282,6 +282,12 @@ class MainPage(object):
             cfg.warned_old_queue.set(True)
             config.save_config()
             return panic_old_queue()
+        
+        if not cfg.notified_new_skin() and cfg.web_dir() != 'Glitter':
+            logging.warning(T('Try our new skin Glitter! Fresh new design that is optimized for desktop and mobile devices. Go to Config -> General to change your skin.'))
+        if not cfg.notified_new_skin()
+            cfg.notified_new_skin.set(True)
+            config.save_config()
 
         if kwargs.get('skip_wizard') or config.get_servers():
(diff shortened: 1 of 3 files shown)


sabnzbd/interface.py
```diff
@@ -282,6 +282,11 @@ def index(self, **kwargs):
             cfg.warned_old_queue.set(True)
             config.save_config()
             return panic_old_queue()
+        
+        if not cfg.notified_new_skin() and cfg.web_dir() != 'Glitter':
+            logging.warning(T('Try our new skin Glitter! Fresh new design that is optimized for desktop and mobile devices. Go to Config -> General to change your skin.'))
+            cfg.notified_new_skin.set(True)
+            config.save_config()
```
shypike: Move "cfg.notified_new_skin.set(True)" below the "if" statement so that it is always executed.
Otherwise you'll also get the message when going from Glitter to Plush.

shypike:
Almost there.
Suppose a new user starts up SABnzbd. He/she will be using Glitter. Then he/she selects Plush. SABnzbd restarts and a message pops up: "Try our new skin"...
For a solution, see my annotation of the code.

Safihre:
Fixed. Did not think of that :sleeping:

thezoggy:
this broke sab for me:
```
C:\Program Files (x86)\sabnzbd_0.8.x>python SABnzbd.py -f D:\test --new --no-api-log
  File "SABnzbd.py", line 87, in <module>
    import sabnzbd.interface
  File "C:\Program Files (x86)\sabnzbd_0.8.x\sabnzbd\interface.py", line 288
    if not cfg.notified_new_skin()
```

Safihre:
Oh no! Missing a : there.
@shypike probably faster if you fix this?

shypike:
Done

PR #347, opened 2015-09-29
sab 0.8.x -- dev pep8 pt4

thezoggy:
moved regex define before use (save memory when code branch isnt ran), calculate expire time once (save cpu cycles), reserved word (only some) & undefined & unused variables, depreciated `has_key`, more spelling and docstring cleanup


shypike:
Merged.

PR #351, opened 2015-10-01
Server-priority label

Safihre:
I made the label, @shypike will you actually sort them on server-priority in config-view?


sabnzbd/interface.py
```diff
@@ -1501,13 +1501,6 @@ def add_color(dir, color):
                 else:
                     wlist.append(rweb)
         conf['web_list'] = wlist
-
-        # Obsolete template variables, must exist and have a value
-        conf['web_colors'] = ['None']
-        conf['web_color'] = 'None'
-        conf['web_colors2'] = ['None']
-        conf['web_color2'] = 'None'
-
```
shypike: Are you sure these can go? No templates use them?

Safihre: Since we use the uniconfig now for all skins, this is not used anymore.

shypike:
I cannot sort in interface.py, because it passes a dictionary of servers, which is inherently unsorted.
So either I start sending a list, otherwise you need to sort the dictionary in the skin.

thezoggy:
just use ordered dict?

shypike:
No, I'll send the server as a list so that the Python code determines the order. Changing the skin code is fairly easy to do, but it needs to be done in sync.

Safihre:
Adding search functionality to the queue.

shypike:
I also added sorting the server on priority and name (in that order).

thezoggy:
@shypike @Safihre 

noticed that glitter status and config servers have different order:

config > servers:
sab3

glitter > status:
sab4

PR #354, opened 2015-10-04
Adding tabs to the Status and interface options

Safihre:
Even I could not deny it was just getting tooo long.
Now space to add extra option to choose extra column in queue (Category/Script/etc).


Safihre:
Added the option for extra column!

shypike:
Merged with rebasing, so without the dumb "merge" commit.

PR #356, opened 2015-10-05
Bootstrap update, less translations, more dateformats, user can set set history/queue deletion confirmation and bugfixes

Safihre:
You can leave this one open for a little more (at least a day) so I can push other interface bugfixes in case I find some :smirk:


shypike:
@Safihre 
Ready for merging?

Safihre:
Yes,  all good!

thezoggy:
so in plus, it said '100' in the box which is %, clicked on reset.. which empties out the box and I see it say bw limit set to 0. then a few moments later it comes back with 100. hit save, now I see it say bw limit set to 100.  @shypike  is bw limit 0 and 100 the same thing? (no limit vs 100% of bw set)

```
2015-10-08 17:46:14,006::INFO::[downloader:293] Bandwidth limit set to 0
2015-10-08 17:46:21,673::INFO::[downloader:293] Bandwidth limit set to 100
```

Safihre:
No 0 and 100 is not the same. 
If your connection is 5MB/s for example but you only put the bandwidth limit at 4MB/s then with 100% you will still be limited at 4MB/s while with 0 you will get your connection's max..

shypike:
I think we should translate 100% quietly into a  max speed (In the downloader code).

Safihre:
You mean max speed of the connection? Because in that case Glitter would possibly show the wrong max speed (in the slider) when you put it to 100% and that's not your connection's maximum.

shypike:
Yes, the max speed of the connection.

Safihre:
Not really sure what I should label that in Glitter under the slider though.. Something like:
*100% (5+ MB/s)*
Or maybe just only:
100%

shypike:
Just 100%.

PR #360, opened 2015-10-08
sab 0.8.x -- dev pep8 pt5

thezoggy:
More pep8 cleanup, (revert) unused variables by import / duplicate imports / misspelling / some minor renames to avoid using built-in names (list/set/id/etc). Will cleanup addID in another pull.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -297,7 +297,7 @@ GNU GENERAL PUBLIC LICENSE Version 2 or (at your option) any later version.
 
 def daemonize():
     try:
-        pid = os.fork()  # @UndefinedVariable - only available in UNIX
+        pid = os.fork()
         if pid > 0:
             sys.exit(0)
     except OSError:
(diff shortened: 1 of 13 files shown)


Safihre:
Do we really need all those ```@UndefinedVariable``` in the code, just for validation purposes?

thezoggy:
When you code with is specific commands it is. As the coder ide is only going to be able import/validate one is..  Then you have some funky imports that are done because python bugs or for py2exe. Now I can global ignore them but it can hide things that could be better addressed with code portability or wrappers/APIs. 

-----Original Message-----

thezoggy:
if @shypike wants I can set my ide to ignore `undefined variable from import` and cleanup those entries.

thezoggy:
alright, removed the suppression of undefined variables from import comments. forced pushed

shypike:
@thezoggy Thanks for removing the overload of overrides.

shypike:
Merged

PR #362, opened 2015-10-09
sab 0.8.x -- addid cleanup

thezoggy:
Removed addID from api, changed reference in Plush to use addurl instead. Tested and ensured plush still worked fine.


shypike:
I think we should keep "addid" in the table (pointing to addurl).
You never knoww whether some add-in or app is using it.

thezoggy:
added it back.. i understand the reason to want to keep legacy stuff around so you dont break apps.. but no app should be using addid.. as it was only for newzbin.. which hasnt been around for years. thus seems a bit pointless to still support it.

shypike:
Merged.
Thanks!

PR #375, opened 2015-10-24
Show available SSL protocols for a Usenet server.

shypike:
Determine available protocols.
Log information about SSL libraries.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1353,10 +1353,11 @@ def main():
                 logging.info("pygtk2 not found. No SysTray.")
 
     print_modules()
-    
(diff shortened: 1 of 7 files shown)


Safihre:
Can we also add this to the setup-wizard? Noticed there we only have SSL on/off but not which type..

shypike:
@Safihre I wish that I knew that before I wrote it :)
The Wizard should automatically pick the highest one. I'll look into that as well.

shypike:
@Safihre I simplified the Config->Server template code.
The Wizard will work fine now: when no ssl_type is set, then the highest available protocol will be picked.

shypike:
Merged

Issue #371, opened 2015-10-19
sab 0.8.x -- glitter - completed timestamp


thezoggy:
the current format I think is less than ideal.. and no matter what format we pick.. there is always going to be others that want something else. 

for example it shows currently:
`03-10-15 17:09`

the fact I look at it and can't tell if it is MM-DD-YY HH:MM or DD-MM-YY HH:MM at a glance is an issue. also is not any date standard.. personally if we have to go with just one, I'd use YYYY-MM-DD as it sorts nicely. however, we should just offer a set of options and let the user pick?

Taking a page from Sonarr they do a decent job offering a set of options:

Short Date Format:
```
<option value="MMM D YYYY">Mar 5 2014</option>
<option value="DD MMM YYYY">5 Mar 2014</option>
<option value="MM/D/YYYY">03/5/2014</option>
<option value="MM/DD/YYYY">03/05/2014</option>
<option value="DD/MM/YYYY">25/03/2014</option>
<option value="YYYY-MM-DD">2014-03-05</option>
```

Long Date Format:
```
<option value="dddd, MMMM D YYYY">Tuesday, March 25, 2014</option>
<option value="dddd, D MMMM YYYY">Tuesday, 25 March, 2014</option>
```

Time Format:
```
<option value="h(:mm)a">5pm/5:30pm</option>
<option value="HH:mm">17:00/17:30</option>
```

Also offer a switch for **Show Relative Dates**
Which shows Today/Yesterday/1 Week ago/etc instead of absolute dates.

Stanzilla:
http://timeago.yarp.com maybe?

Safihre:
The reason is simply that we don't have the translations for other than regular numeric dates. There are of course international time libraries available for Javascript, but didn't have the time to use those. 
The very long date format will just not fit in all the columns. The history, the error times, etc. 
I already included a commit in my PR that includes some extra numeric date formats.

thezoggy:
@Stanzilla  i like MomentJS better, http://momentjs.com/

Safihre:
@thezoggy I like MomentJS, but it is just so heavy (40kb extra).. Especially with the localisations (125kb..).
I tried just removing the localisations we don't use, but my system translates all the Chinese/Russian/etc characters to question marks when I save the file. I also tried cloning their Github and then removing all the localization files we don't use. But the building fails on my system (even without removing any files).
So, if anyone can build a minimized version with only the languages we use, then I will gladly implement it!

[...]

PR #376, opened 2015-10-24
Adding a hide button to the feedback window. Date-formatting. LocalStorage try/catch.

Safihre:


Safihre:
#371 @thezoggy There you go..

thezoggy:
you may want to add some logic to test if browser can even do localStorage or not...
https://gist.github.com/paulirish/5558557

while most browsers ie8+ shouldnt have a problem, i think there is some scenarios that you still need it.. like incognito/private mode, or certain mobile devices

Safihre:
Added try/catch with simple message to the user.

PR #394, opened 2015-11-20
Adding help-icons to config-sections

Safihre:
Since it is not clear that 'Help' on each Config page leads to a help page that is specific for the topic, this should make it more clear. Icon only appears on hover over the specific section.
With direct links to the sections of the help page, where applicable.


shypike:
Nice!

PR #395, opened 2015-11-22
Config fixes

Safihre:
Also found a bug that RSS feeds always got disabled when editing the URL.
Might find more things, so maybe keep it open until the next build?


shypike:
Please give me a sign when you think you're finished.
(I'll build on Wednesday probably).

thezoggy:
bootstrap 3.3.6 is out btw, not sure if you wanna just jump to that

Safihre:
@thezoggy Updated it, but don't think any changes were applicable to us.

thezoggy:
looks like this got pulled early?

thezoggy:
not sure if its by design, but you didnt enable autocomplete on the Categories folder/path field.
the 'Relative folders are based on: ' red is obnoxiously bright.. please revert it back to the dark red that we previously used.

Safihre:
Yes, will change the red.
Still want to change the auto-complete so that it works with relative-paths. Now the auto-complete will put the full path when selecting a value (it already suggest the right relative paths because the API understands it).

PR #399, opened 2015-11-28
Interfaces fixes 6

Safihre:


Safihre:
Changed the custom pause timer which includes changes to the text.
However, these were not in Launchpad yet so no translations were harmed!

PR #411, opened 2015-12-13
Translation things

Safihre:

PR #412, opened 2015-12-14
Enable Windows Notifications

Safihre:
I always wondered why this was not embedded into SAB!

untitled

Turns out to be really simple to implement and the PyWin32 code for it is already years old, so fully backwards compatible with older Windows versions.
Unfortunately only Metro-apps seem to be able to create persisting notifications in the Action Center of Windows 10. Could find no Python implementation anywhere that made this possible.
The **a** is because in Windows 10 it's called Action Center.
Next improvement would be to improve the icon for the Windows 10 notification, but documentation is very,very limited so no idea if that's even possible.

Also changed the sorting in notification config pane to be: Email - OS services - External services.

PS: Sorry for including the commits from my other PR, but created a new branch too late..

### DIFF
--- a/interfaces/Config/templates/config_notify.tmpl
+++ b/interfaces/Config/templates/config_notify.tmpl
@@ -73,52 +73,58 @@
     <div class="section">
         <div class="col2">
             <h3>$T('section-NC')</h3>
+            <table>
+                <tr>
+                    <td><input type="checkbox" name="ncenter_enable" id="ncenter_enable" value="1" <!--#if int($ncenter_enable) > 0 then 'checked="checked"' else ""#--> /></td>
+                    <td><label for="ncenter_enable"> $T('opt-ncenter_enable')</label></td>
+                </tr>
+            </table>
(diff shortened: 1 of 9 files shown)


Safihre:
Rebased this branch so it's only the notifications and the interface fixes are in #411

shypike:
Thanks, but I rebase any pull request regardless.
I hate those dumb "merge" entries that github puts in.
I prefer a linear history, despite that such a thing lies about the commit dates.

Safihre:
@shypike is it a bug or a feature, that when I select to be notified of 'Added NZB' it will also send notifications for pausing the queue using the pause button in Glitter or Plush?

shypike:
@Safihre "Pause" and "resume" are in the same group as "add nzb", which is "download".
It works as intended, only the visible group name "Add NZB" isn't very clear.

Safihre:
I don't get a notification for Resume though.. 

I will add this to the labels of all notification types then:
> Add NZB / Pause / Resume

PR #154, opened 2014-05-07
Detection and reporting of unrar binary version, and rar version of rar files

sanderjo:
Detection and reporting of unrar binary version, and rar version of rar files, as discussed on https://forums.sabnzbd.org/viewtopic.php?f=4&t=17490

Example output in sabnzbd.log:

2014-05-07 21:40:31,845::INFO::[SABnzbd:484] unrar binary version is: UNRAR 5.00 beta 8 freeware
...
2014-05-07 22:06:32,256::DEBUG::[newsunpack:599] Type of rar file is Rar4 or lower

PR #414, opened 2015-12-18
Log RAR-binary and RAR-file versions

Safihre:
Everything to make #154 work was already there, but the check of the rar-binary and it's ```log.debug()``` was called before the logger was active, so it would never show the version.
I implemented as @shypike suggested, based on @sanderjo placements!

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -146,10 +146,9 @@ def find_programs(curdir):
         version, original = unrar_check(sabnzbd.newsunpack.RAR_COMMAND)
         sabnzbd.newsunpack.RAR_PROBLEM = not original or version < 380
         sabnzbd.newsunpack.RAR_VERSION = version
-        logging.info('UNRAR version %s', version)
+        logging.info('UNRAR binary version %.2f', (float(version) / 100))
         if sabnzbd.newsunpack.RAR_PROBLEM:
             logging.info('Problematic UNRAR')
-
 
 def external_processing(extern_proc, complete_dir, filename, nicename, cat, group, status, failure_url):
     """ Run a user postproc script, return console output and exit value """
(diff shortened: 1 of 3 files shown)


shypike:
I disagree with this implementation.
All other tools are detected in the same module, don't make an exception for unrar by moving its code to SABnzbd.py.

Safihre:
Agreed, but I had to move ```find_programs()``` down after the logger started and where it would make sense in the startup log.
I tested on Windows and Ubuntu and it works fine, so changing it's location to before ```print_modules()``` doesn't affect anything else?

shypike:
I'll check later.

Safihre:
Fixed a forgotten ```return False```

sabnzbd/utils/rarfile.py
```diff
@@ -108,7 +108,11 @@ def is_rarfile(fn):
     '''Check quickly whether file is rar archive.'''
     try:
         buf = open(fn, "rb").read(50)
-        return buf.startswith(RAR_ID) or buf.startswith(RAR5_ID)
+        if(buf.startswith(RAR_ID)):
```
shypike: Incorrect: you should be checking here for RAR5_ID.

sabnzbd/utils/rarfile.py
```diff
@@ -108,7 +108,11 @@ def is_rarfile(fn):
...
+            return "rar5 signature"
+        if(buf.startswith(RAR_ID)):
+            return "rar4 (or lower) signature"
+        return False
```
shypike: For consistency: return None or an empty string.

sabnzbd/utils/rarfile.py
```diff
@@ -108,7 +108,11 @@ def is_rarfile(fn):
...
     except:
         return False
```
shypike: For consistency: return None or an empty string.

sabnzbd/newsunpack.py
```diff
@@ -143,14 +143,13 @@ def check(path, program):
         sabnzbd.newsunpack.PAR2C_COMMAND = sabnzbd.newsunpack.PAR2_COMMAND
 
     if not (sabnzbd.WIN32 or sabnzbd.DARWIN):
-        version, original = unrar_check(sabnzbd.newsunpack.RAR_COMMAND)
+        version, original = sabnzbd.newsunpack.unrar_check(sabnzbd.newsunpack.RAR_COMMAND)
```
shypike: The sabnzbd.newsunpack prefix probably isn't needed, since you moved the code back.

shypike:
Still some issues, most important: detecting the wrong RAR type.

sabnzbd/newsunpack.py
Safihre: Oh yes, oops! Changed that now.

shypike:
Merged.

PR #416, opened 2015-12-20
Interface fixes part 9

Safihre:


Safihre:
@shypike Is it a bug or a feature that when I use the API to change the priority **or** category of a paused job, it will always unpause it?
I could somehow see this could be useful for when I go from a lower to a higher priority, but it always does it.

Especially in Glitter when using multi-edit it feels weird that it's doing things you don't ask it..

shypike:
The priority field is also used to store the temporary pause state.
This has some side effects:
- Changing the category changes priority to the associated one
- Resuming the job will set the priority to Normal.

Obviously it would be better to keep a separate pause status,
but that's quite a bit of work and is hard to implement in a compatible way.

Safihre:
@shypike is there some way to order the ```stage_log``` for both Plush and Glitter?
Now it's often some weird order like:
*Download, Source, Unpack, Servers, Script, Repair*

Is there a way to make it logical?
*Source, Download, Servers, Unpack, Repair, Script*

I can make some JavaScript tricks, but would be better if done in the back-end so it's also for Plush fixed.

shypike:
Merged. Thank you!

PR #418, opened 2016-01-03
Preload power

Safihre:
This makes the first load instant(!) in Firefox and IE, but for some reason Chrome is less impressed.
Template rendering is only few msec slower and first-download size only increases little. 

@shypike hope I thought of everything for the Python code, added the try-block to make sure the JSON won't break it.


Safihre:
Now Chrome also loads instantly, the trick was to move the Javascript before any other elements.
Works like a charm!

shypike:
@Safihre 
I see no issues at all when using < and > on OSX with FF and Chrome, nor on Win10 with FF.

Safihre:
It works fine when entering, but goes wrong on any of the other pages like Sorting. 
This happens with your example. And did you also try to change or delete this entry? Because I can't anymore except with deleting it from the sabnzbd.ini!
image

Safihre:
@shypike The strip function doesn't seem to work completely, or am I using it wrongly?
Because with the code below I can still enter ```movies><test``` and it won't filter the < or > out of it..
But it does make ```<movies>``` into ```movies```..
```
newname = kwargs.get('newname', '').strip(' []"&<>')
```

shypike:
@Safihre You are using strip() wrong here. split() only removes the listed characters from the start and the end of the string. You need a number of replace() calls or translate(). The latter is like this:
```python
import string
string.translate('abc<ff>&d', None, '<>&')
```
Still scratching my head why I cannot reproduce your issues. I'll try again with your latest contributions.

Safihre:
I had to use ```newname = re.sub('[<>""\[\]]', '', kwargs.get('newname', ''))```, because ```str.translate``` doesn't work on the unicode strings. Rebased this change now.

Just tried and also on FF on Ubuntu as expected I get the same as the screenshot above (before my changes of course).

shypike:
Merged.
I skipped commit d7aa0e738bb158a80a2a94d66116fd3ca5f8242c (filtering \<\> etc.) because I'm not convinced it is a problem. At least it's inconsistent enough to warrant further investigation,

Safihre:
At least the " should be filtered out, they really are a consistent problem and cause un-fixable categories:
https://www.reddit.com/r/usenet/comments/3z9l3o/sabnzbd_categories_problem/

PR #419, opened 2016-01-07
Update the ad in the Wizard for Github Pages Future-Proofing

jeffgeorge:
We're eventually moving sabnzbd.org's hosting to Github Pages, so I needed to set up a new home for the few bits of utility code presently hosted under the main domain that wouldn't be portable.

The Wizard Ad is written in PHP, so I've relocated it here under the resources.sabnzbd.org subdomain which I'll keep hosting:

https://resources.sabnzbd.org/wizard/ad/en

This PR updates the iframe source in the wizard to point there instead. While I was at it I updated everything in the Wizard Ad to run over SSL.

At some future date when the site is in Github Pages we should evaluate porting this code to a simple JS include which could then be hosted in Pages, but for the moment just updating the URL should be enough.

### DIFF
--- a/interfaces/wizard/one.html
+++ b/interfaces/wizard/one.html
@@ -79,7 +79,7 @@
                     </div>
                     <div class="col-md-5">
                         <div class="clearfix"></div>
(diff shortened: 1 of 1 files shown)


shypike:
Merged as commit 1618e029a7cc655360055ca8ec37baa2dcfb4a5c
Thanks.

PR #424, opened 2016-01-12
Correct openssl package name in log message

jcfp:
Noticed someone using the incorrect package name in https://forums.sabnzbd.org/viewtopic.php?f=2&t=20018


Safihre:
Maybe also change this line to something better:
https://github.com/sabnzbd/sabnzbd/blob/develop/interfaces/wizard/one.html#L44
And should we maybe put some form of message as tooltip or next to the disabled SSL checkbox?

PR #425, opened 2016-01-13
Default skin not set correctly in Wizzard

Safihre:
On Linux 'Glitter - Night' is top of the list in Config, so on the first opening of the Config it would switch the layout to 'Glitter - Night' on save.

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -299,7 +299,7 @@ ntfosd_prio_queue_done = OptionBool('ntfosd', 'ntfosd_prio_queue_done', True)
 ntfosd_prio_other = OptionBool('ntfosd', 'ntfosd_prio_other', False)
 
 # [growl]
-growl_enable = OptionBool('growl', 'growl_enable', sabnzbd.DARWIN_VERSION < 8)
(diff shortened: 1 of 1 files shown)


Safihre:
I noticed that Growl is wrongly enabled by default on Linux because:
```
growl_enable = OptionBool('growl', 'growl_enable', sabnzbd.DARWIN and sabnzbd.DARWIN_VERSION < 8)
```
And ```sabnzbd.DARWIN_VERSION``` is 0 on Linux (and Windows I guess).

Safihre:
@jcfp While testing on Ubuntu I used to always get notifications from NotifyOSD, but not anymore in Beta4. 
Was ```python-notify``` purposely moved to suggested packages instead of recommended ones?

jcfp:
@Safihre python-notify was always suggested, for the rather large amount of deps is pulls in esp. on headless servers. Maybe fewer other packages need it these days, so it's no longer "just there" on desktops?

Safihre:
My Ubuntu 15.10 VM is a clean install without any other apps than the ones that came with the install and SABnzbd. 
Maybe indeed previously some other update/build-in app got it, I just restored a fresh-after-clean-install snapshot this morning.

PR #433, opened 2016-01-17
Interface fixes part 10

Safihre:


Safihre:
Re-ordered the Switches config page, now it is in logical order of:
(Pre)Queue, Detection
Verification, Unpack, Scripts, Deletes/cleanup after download

Issue #435, opened 2016-01-19
Notification script


domibarton:
Hi there!

First of all, awesome job! :) I love SABnzbd!

Have you ever thought about a "custom" notification script? I know there are custom post processing scripts per category and I already use them for SickBeard. But it would be great to get a custom notification script. If I'm not completely wrong I can use Email, Growl, Prowl, Pushover and Pushbullet. Why not adding a new option to execute a custom script? 

I want to use it to notify a Slack channel. Of course it would be even more awesome to get a specific Slack notification option.

What do you think? If you like the idea I'll have a look if I can integrate such an option and send you a PR.

Cheers
Domi

shypike:
A custom notification script would probably be better than a specific Slack option.
We don't want SABnzbd to become an exhibition of notification services.
Most of the real code is in sabnzbd/growler.py and there is the UI in sabnzbd/interface.py and the Config->Notifications template.

domibarton:
OK I understand. Makes sense :)
I'll have a look into it and try to build something.

Safihre:
Also take a look at Sonarr and Couchpotato for TV shows and movies, they have a really big amount of notification services build in. 
Of course only works for TV shows and movies then :)

domibarton:
There it is. Was quite a journey to be honest :) 

I think we could probably clean up a lot of the code because there's repetition of code everywhere. I tried to clean it up, but that requires more than just a day ^^ Would be great to get a proper abstract `Notification` class to inherit all concrete notification implementations from. 

The "management" of the priorities and additional custom parameters per notification type will become much easier. Something like that:

```python
class ProwlNotification(Prowl):
    ''' Notification class for Prowl. '''
    help_text = 'Requires a Prowl account'
    parameters = (
        NotificationParameter('apikey', 'API key for prowl', 'Personal API key for Prowl', required=True),
    )

    def notify(self, parameters):
        ''' Sends Prowl notification. '''

        # do some business logic here
```

The Notification class can handle then all the use cases, for example verifications and sending test messages (i.e. getting parameters from GET/POST instead from `sabnzbd.cfg`).

In this case someone could easily add new Notification types by just creating a new `Notification`-based class. Of course the HTML on the config front-end has to be rendered dynamically, but that's not very hard.

What do you think?

domibarton:
[...]

PR #436, opened 2016-01-20
Added custom notification script

domibarton:
Hi there!

As proposed in #435 I've added a new notification type which basically just executes a custom script. The script is executed with a __title__ and __msg__ CLI argument like this:

```bash
/path/to/the/script "TITLE" "MSG"
```

I tried to adopt the existing "code style" and tried to use your naming scheme. I saw that there is a _Notification Center_ notify which is shortened to `ncenter` in the code. So I decided to call it _Notification Script_ (verbose) and therefore `nscript` in the code. Hope that's OK!

Let me know if you're OK with it or if I've to change anything.

Cheers
Domi

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -37,7 +37,7 @@ from sabnzbd.constants import MAX_DECODE_QUEUE, MIN_DECODE_QUEUE
 from sabnzbd.articlecache import ArticleCache
 import sabnzbd.downloader
 import sabnzbd.cfg as cfg
-from sabnzbd.encoding import name_fixer
+from sabnzbd.encoding import yenc_name_fixer
 from sabnzbd.misc import match_str
 
 
@@ -245,7 +245,7 @@ def decode(article, data):
             try:
                 for i in xrange(min(40, len(data))):
                     if data[i].startswith('begin '):
-                        nzf.filename = name_fixer(data[i].split(None, 2)[2])
+                        nzf.filename = yenc_name_fixer(data[i].split(None, 2)[2])
                         nzf.type = 'uu'
                         found = True
                         break
@@ -267,7 +267,7 @@ def decode(article, data):
         # Deal with yenc encoded posts
         elif (ybegin and yend):
             if 'name' in ybegin:
-                nzf.filename = name_fixer(ybegin['name'])
+                nzf.filename = yenc_name_fixer(ybegin['name'])
             else:
                 logging.debug("Possible corrupt header detected => ybegin: %s", ybegin)
             nzf.type = 'yenc'
--- a/sabnzbd/encoding.py
+++ b/sabnzbd/encoding.py
@@ -71,17 +71,27 @@ def platform_encode(p):
 
 
 def name_fixer(p):
-    """ Return Unicode name of 8bit ASCII string, first try UTF-8, then cp1252 """
+    """ Return Unicode name of 8bit ASCII string, first try UTF-8, then codepage, then cp1252 """
     if isinstance(p, unicode):
         return p
     elif isinstance(p, str):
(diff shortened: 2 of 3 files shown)


Safihre:
I kind of like this more than I thought I would, since this would sort of excuse us from having to add any more notification services since those can be handled by scripts :smile:
On the other hand, the main notifications people care about (I think) will be job succes/failure, and those can be handled by regular end-of-job scripts. You would have to turn off 'Post-Process Only Verified Jobs' so it can also handle failed jobs.

But yes, some cleanup of the code would be welcome! growler.py kind of grew and grey over the years, even the filename reminds of that!

PS: Greetings from Zurich ;-)

shypike:
Thanks for your contribution.
Unfortunately, I'm too busy with 0.8.0 issues to review this now.
Also, we have a feature-freeze due to the pending 0.8.0 release.
It will of course be included in a release after 0.8.0

domibarton:
Awesome, thanks!

sabnzbd/growler.py
```diff
@@ -539,7 +545,57 @@ def send_pushbullet(title, msg, gtype, force=False, test=None):
...
+    if gtype == 'error':
+        prio = sabnzbd.cfg.nscript_prio_error()
+    if gtype == 'queue_done':
+        prio = sabnzbd.cfg.nscript_prio_queue_done()
+    if gtype == 'other':
+        prio = sabnzbd.cfg.nscript_prio_other()
+    if force:
+        prio = True
+
+    if prio:
```
Safihre: You don't need any of all this prio stuff, that's only for Prowl/Pushbullet etc. that support notification priorities. 
Look for example at the one for Windows notifications, all you need to do is call the script directly:
```python
def send_windows(title, msg, gtype):
    if sabnzbd.WINTRAY:
        sabnzbd.WINTRAY.sendnotification(title, msg)
    return None
```

domibarton:
@Safihre Are you sure? Isn't that required to decide which kind of messages (e.g. startup/shutdown) should be sent? 

I know on Prowl / Pushover the user can select the priority (integer) and I don't need that. But have a look at Pushbullet. The user can enable/disable notifications (boolean) based on the message type (e.g. startup, download started, post processing, download complete, disk full etc.). 

I thought it would be nice if we do the same thing for notification scripts, so that the user can select which notifications should invoke the notification script. However, if you want to send all notifications independent of the message type to the notification script, then I can remove it :)

Cheers
Domi

Safihre:
I forgot a part to post, the ```check_classes``` (see line 112) function will filter that only the ones that are checked will be send:
```python
    # Windows
    if sabnzbd.WIN32 and sabnzbd.cfg.acenter_enable():
        if check_classes(gtype, 'acenter'):
            send_windows(title, msg, gtype)
```
It should have also been this way for Pushbullet I think, maybe just a mistake.

Safihre:
@shypike besides the Pushbullet not needing the whole prio-if-things, I also saw 2 errors:
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/growler.py#L393
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/growler.py#L506
```disk-full``` should be ```disk_full```, that's the way it's variable is called in the ```cfg```

Probably should have a ```set_prio``` function that can work almost identical to the way ```check_classes``` does it instead of all those if's.

domibarton:
Sorry I was busy the last couple of days.
However, I fixed the errors and rebased my branch, which unfortunately closed this PR ;)

So I opened another one. 
Second one for the Pushbullet fix is coming soon...

shypike:
Closed with ccba56e073049c3c092f2be179b0a56d0ce4e798

Issue #369, opened 2015-10-15
Logout functionality is missing when http login is enabled


lpallard1:
While login functionality is enabled in the application's options, there are no ways to logout of the program.  To "logout" the browser needs to be closed and cookies deleted.

To me it is obvious that a way to logout is needed for privacy & safety considerations.  If you can login you should be capable of logging out.

shypike:
The "basic authentication" protocol we now use, doesn't provide a logout method.
Someday we'll implement better login that will also allow logout.

thezoggy:
note we use basic auth for sab login.. 
for cookies thats only used to store skin specific info.. (plush mainly) for glitter we use localstorage

Safihre:
Close! Added in #442 :)

Safihre:
Not closing? It's in there for a while now.. 
Already for so long maybe we can release 1.1.0 beta1 soon.. With the open pull requests and improvements in Glitter, Config and other details  we have plenty to add.

PR #442, opened 2016-01-24
Add HTML login page

Safihre:
Based on cookies, not CherryPy-sessions. We create an unique cookie based on user-IP, SAB-process-ID and a salt to facilitate log-out functionality (#369).
Yes, this cookie can be stolen when on a public wifi-network (same IP) when not on HTTPS, but when not on HTTPS basic-auth or API-requests (which have the API-key) can also easily be intercepted. Any restart of SAB would in-validate the cookie.

Could also add a 'Remember-me' option if you guys like my implementation.
If it would replace basic-auth we don't need any new translations (except 'Log out', but this can automatically be translated from other packages on Launchpad).

@shypike Since it's a new feature, I understand this is not anymore for the 0.8.0 release :)

image

### DIFF
--- a/interfaces/Config/templates/config_general.tmpl
+++ b/interfaces/Config/templates/config_general.tmpl
@@ -34,6 +34,11 @@
                     <span class="desc">$T('explain-web_password')</span>
                 </div>
                 <div class="field-pair">
+                    <label class="config" for="html_login">$T('opt-html_login')</label>
+                    <input type="checkbox" name="html_login" id="html_login" value="1" <!--#if int($html_login) > 0 then 'checked="checked"' else ""#--> />
+                    <span class="desc">$T('explain-html_login')</span>
+                </div>
+                <div class="field-pair">
                     <label class="config" for="web_dir">$T('opt-web_dir')</label>
                         <select name="web_dir" id="web_dir">
                         <!--#for $webline in $web_list#-->
--- a/interfaces/Config/templates/config_general.tmpl
+++ b/interfaces/Config/templates/config_general.tmpl
@@ -72,12 +72,7 @@
                     <label class="config" for="password">$T('opt-web_password')</label>
                     <input type="text" name="password" id="password" value="$password" />
                     <span class="desc">$T('explain-web_password')</span>
-                </div>
-                <div class="field-pair">
-                    <label class="config" for="html_login">$T('opt-html_login')</label>
-                    <input type="checkbox" name="html_login" id="html_login" value="1" <!--#if int($html_login) > 0 then 'checked="checked"' else ""#--> />
-                    <span class="desc">$T('explain-html_login')</span>
-                </div>   
+                </div>  
                 <div class="field-pair">
                     <label class="config" for="inet_exposure">$T('opt-inet_exposure')</label>
                         <select name="inet_exposure" id="inet_exposure" class="select">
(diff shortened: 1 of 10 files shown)


thezoggy:
while your at it, should add the whole 'whitelist' local ip features so only outsiders get the prompt... for those that want the pw prompt but only when they access remotely...

example code: https://github.com/midgetspy/Sick-Beard/pull/151/files

thezoggy:
also need to check to see if this works with web_root / reverse proxy.. and if changing sab from http<>https

Safihre:
I tested HTTPS, works fine! Maybe someone else can test reverse proxy (@sanderjo, if he has time).
For the white-listing, this would only work with the HTML login.
I decided to add an extra option to *External internet access* called: 
*Full Web interface - Only external access requires login* (still working on the python-code)

sanderjo:
I tested like below (correct?), and I get a working, non-popup login screen from SABnzbd behind my Apache2-as-HTTPS-proxy.

screenshot from 2016-01-24 19 50 45


```
mkdir safihre-sabnzbd
cd safihre-sabnzbd/
git clone https://github.com/Safihre/sabnzbd.git
cd sabnzbd/
git branch -a
git checkout newlogin

sudo service sabnzbdplus stop
./SABnzbd.py 
```

Safihre:
And usage was fine? Going to Config pages etc :)

Safihre:
So, I added the 'Remember me' checkbox, although after a SABnzbd restart or change of IP, nothing will be remembered.

I also added the **'Full Web interface - Only external access requires login'** option and the back-end code, since these things are intertwined: Without **Full Web interface** turned on, nothing will show anyway.

Safihre:
I made the HTML-login the default (with the option in Specials to switch it off). Because the 'Only external access requires login' only works with the HTML login.
Otherwise we have to explain how that option would only work with HTML login turned on etc.
We can always drop that commit from this PR in case it's a bad idea?

shypike:
@Safihre I think making HTML-login the default is fine. I think most people will expect a login page instead of the rare basic login prompt.

Safihre:
@thezoggy you maybe have time for that new or better old logo? :yum:
It looks a little odd now with the black shadow.

shypike:
@Safihre Isn't "Login" a better button name than "Send"?
What is the meaning of the checkmark before the text?

Safihre:
The checkmark is gone and I made it say "Log in", indeed better but required another new text so didn't do it before :) I just rebased this changes on top of the last commits.

This how it looks now:
image

thezoggy:
a very common thing is people forgetting their sab pw/not expecting it when they login (no clue how).. but anyways we should maybe offer a link to the wiki from this page on how to reset their password ( http://wiki.sabnzbd.org/faq#toc21 ) -- note we really should make a proper page since we dont want this page to change from this point forward to ensure that whatever we do link remains good.

thezoggy:
@Safihre ive completely dropped the ball on the logo :( let me go tinker with some stuff

Safihre:
I added an anchor to the wiki so now it's:
http://wiki.sabnzbd.org/faq#why-login

And made it this:
image

thezoggy:
lol that FAQ page needs some lovin, http://wiki.sabnzbd.org/faq#toc8

shypike:
@thezoggy I'm sure it would return the love you give it :)

PR #443, opened 2016-01-24
Interface fixes part 11

Safihre:
Fixes http://forums.sabnzbd.org/viewtopic.php?f=11&t=20114


thezoggy:
should also fix: https://github.com/sabnzbd/sabnzbd/issues/445

sabnzbd/misc.py
```diff
@@ -1307,6 +1307,9 @@ def get_base_url(url):
     url_host = urlparse(url).hostname
     if url_host:
         url_split = url_host.split(".")
+        # Exception for localhost and IPv6 addresses
```
shypike: Please explain why this is needed?
I have no idea what happens here...

Safihre: Because Google's Favicon service will only work on root domains (so nzbgeek.com, but nog api.nzbgeek.com) I updated the function a while ago to extra the root-domain instead of just the domain:
```return ".".join(len(url_split[-2]) < 4 and url_split[-3:] or url_split[-2:])```

But that is based on the assumption that there will be at least three ```.```, if there's less it breaks and gives a 500 page. So when using ```localhost``` or ```[::ffff]``` (or any IPv6).
This fixes that and just returns the whole domain instead of trying to find the root-domain.

shypike:
How much of this must go into 0.8.0? I want to keep changes to a minimum.

Safihre:
The ones with a issue # and the very first one about the RSS is essential.

shypike:
I cannot reproduce the issue.
I tried with two and three servers of the same priority, they balance nicely.
Even tried it with a random generator that fails messages with a 25% chance, to force usage of the other servers. Same result: works fine.

Issue #420, opened 2016-01-10
Replace par2-tbb with alternatives


thezoggy:
the version of par2 that comes with sab is 0.4 from 2003... but has modifications.. 

```
C:\Program Files (x86)\sabnzbd_0.8.x\win\par2\x64>par2
Not enough command line arguments.
par2cmdline version 0.4, Copyright (C) 2003 Peter Brian Clements.
Modifications for concurrent processing, Unicode support, and hierarchial
directory support are Copyright (c) 2007-2015 Vincent Tan.
Concurrent processing utilises Intel Thread Building Blocks 4.3 Update 1,
Copyright (c) 2007-2014 Intel Corp.
Executing using the x64 (64-bit) instruction set.
```

have you seen the forked and continue development of par2cmdline?
0.6.14 @ https://github.com/Parchive/par2cmdline

specifically I hear that this allows a lot more stuff to get repaired and runs a bit faster than the included one:
`-N       : No data skipping (find badly mispositioned data blocks)`

also to note there is:
https://github.com/jkansanen/par2cmdline-mt (multi-threaded for the brave)

performance comparison mention:
https://github.com/Parchive/par2cmdline/issues/27#issuecomment-102522465

havent taken the time to compile my own version yet, but they provided 0.6.13, so grabbed that real quick:

```
P:\>par2.exe -V
par2cmdline version 0.6.13

P:\>par2.exe --help
Usage:
  par2 -h  : show this help
  par2 -V  : show version
  par2 -VV : show version and copyright

  par2 c(reate) [options] <par2 file> [files] : Create PAR2 files
  par2 v(erify) [options] <par2 file> [files] : Verify files using PAR2 file
  par2 r(epair) [options] <par2 file> [files] : Repair files using PAR2 files

You may also leave out the "c", "v", and "r" commands by using "par2create",
"par2verify", or "par2repair" instead.

Options:

  -a<file> : Set the main par2 archive name
  -b<n>    : Set the Block-Count
  -s<n>    : Set the Block-Size (Don't use both -b and -s)
  -r<n>    : Level of Redundancy (%%)
  -r<c><n> : Redundancy target size, <c>=g(iga),m(ega),k(ilo) bytes
  -c<n>    : Recovery block count (Don't use both -r and -c)
  -f<n>    : First Recovery-Block-Number
  -u       : Uniform recovery file sizes
  -l       : Limit size of recovery files (Don't use both -u and -l)
[...]

PR #449, opened 2016-01-25
Multipar integration for Windows 

Safihre:
I added the Multipar interpreter for Windows (32bit and 64bit), for motivation (speedddd, at least 2x faster repair on my PC) see #420.

Multipar takes things a little different than the other par2-implementations. For example Multipar will first search for all complete files, relatively quickly ("Checking") and only then scan the broken ones ("Verify"). When there are obfuscated file names it will **always** rename the complete ones, but not the in-complete files if it can't repair. So if there are par-blocks missing, you're left with half-renamed set of files after the first repair run.
But that's alll been taken care of, just like joinables! 
The only case I couldn't find any NZB's for were these jobs where the rar-files themselves were then split again. But these are rare nowadays and can be added to the interpreter later. 
I have tested it with many downloads already and so far so good!

I enabled it by default now in the testing phase, with a message to the user. This way enough people will try it and we can get feedback from the users!


thezoggy:
while i dont think you should squash and rebase everything, see a few commits you could combine.. up to shy though if he cares about the commit noise

Safihre:
Yes, some can be combined indeed. Just to show the steps that were taken to get it working!

thezoggy:
1.2.9.1 came out yesterday btw.. looks like even more performance gains for some :)

https://multipar.eu/changelog
```
This version includes update for some non-popular PCs. It will detect proper cache size for AMD CPU. It will use more cores for Intel Xeon CPU. There would be no difference for most users. But, if someone find a problem, please report it.

Because par2j's "/lc" option was changed, those who use batch script or command-line should be careful. Setting by rate will be more useful than direct number. On MultiPar GUI's option, "CPU usage" slider means number of threads; 1/4, 2/4, 3/4, 4/4, 6/4 of physical cores.
```

Safihre:
@thezoggy interesting!

I have been using this now for a whole as my primary par2 option and I noticed that MultiPar is definitely slower in detecting Obfuscated filenames than par2-tbb is.
However as soon as it finds a file, it immiditaly renames it. So when a second run is done after it fetched more blocks, files are not Obfuscated anymore and it's just as fast.
It also seems to require more disk space during repair than par2-tbb. I use a small caching-SSD of 15GB as my temporary download folder. When a download is 14GB but it needs to repair 12 of 100MB rar's, it wants to reserve 1200MB for it. Because it can't, it will fail with disk-full error while par2-tbb somehow doesn't need all this space beforehand.
But of course the repair is super-fast.

shypike:
Closed?

Safihre:
Have been using this for a while now and works very well for 'normal' jobs, but on these special cases it doesn't behave the same as par2cmdline in those 0.1% of cases. Not the fault of Multipar, but of my interpreter probably. Will re-open when I fix it for all the weird NZB's.

thezoggy:
time to work on this again ?

Safihre:
@thezoggy Although repair performance is amazing with Multipar, verification performance is definitely worse. Not just initial verification, but also the verification after it finished the repair.
The good thing is that with the Prospective Par2 the number of verification-runs will be only 1 in 95% of downloads.
Probably pick this up after finishing the SSL stuff :)

mesa57:
I would like to ask you to pick this up also. 
I have problems with the current par2-tbb, because that only works on win-7 and not on my XP vm.
There are weird kernel.dll dependency's
Multipar however works perfect, so that would be a solution.

Par2 classic (par2cmdline) is not an option, because it cannot process renamed files in a rar set.
It will fail with a 'repair is not possible'. Multipar recognizes that the files should be renamed.

Safihre:
It's on the list, although it was much harder to implement than I originally thought.
Just to be sure: only in 1.1.1 we replaced par2classic by par2cmdline, did you try and see if that makes it better?

mesa57:
Running indeed 1.1.1 and yes, I tried par2cmdline, but it does not work on rar sets with renamed files in it.

Safihre:
Could you send me one of those NZB's to safihre@sabnzbd.org?

mesa57:
Of course :smile: 
Done.

Safihre:
Tested and it seems this is not Par2cmdline's fault, but SAB's!
Working on a fix.

Safihre:
Actually, it is par2cmdline's fault:
The job-folder ends in ```.par2```. this causes par2cmdline not to scan for additional files because it's confused. Even if you specify the paths to each file individually. 
Solution would be to let workdir never end on ```.par2```. 
What do you think @shypike ?

Safihre:
Tested and the job-folder ending on ```.par2``` is also a problem on Ubuntu, so clearly we should fix it.

mesa57:
That will be difficult. 
I could not get hold of par2 tbb to study the difference between the tbb and the normal version.

Safihre:
@mesa57 you should not add NZB files that have par2 in the filename, just remove that before adding. 
I can test both versions on my pc.

mesa57:
Just tested that, and indeed it works :+1:

mesa57:
Funny thing is that par2 tbb has no problem with it (tested with sabnzbd 1.1.0 / x64 / win).

Safihre:
Fixed: https://github.com/sabnzbd/sabnzbd/commit/e70fac98aa4182cccccaf4f1ba997cf18f322b4e

If you have any more NZB's that fail in par2cmdline for different reasons, just send them to me!

mesa57:
Thanks :+1: 
And of course we will if we encounter them :smile:

PR #453, opened 2016-01-27
sab 0.8.x -- config tweaks

thezoggy:
Add the module/binary information to the config root page since it can be a pain to see it if you dont access the log file or console directly..  

Some of the wiki pages for -0-8 did not exist, started to create them by using the 0-7 pages then updating...

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1365,6 +1365,7 @@ def main():
 
     import sabnzbd.utils.sslinfo
     logging.info("SSL version %s", sabnzbd.utils.sslinfo.ssl_version())
+    logging.info("pyOpenSSL version %s", sabnzbd.utils.sslinfo.pyopenssl_version())
     logging.info("SSL potentially supported protocols %s", str(sabnzbd.utils.sslinfo.ssl_potential()))
     logging.info("SSL actually supported protocols %s", str(sabnzbd.utils.sslinfo.ssl_protocols()))
 
--- a/interfaces/Config/templates/config.tmpl
+++ b/interfaces/Config/templates/config.tmpl
@@ -1,64 +1,111 @@
 <!--#set global $pane="Config"#-->
 <!--#set global $help_uri="configure-1-0"#-->
 <!--#include $webdir + "/_inc_header_uc.tmpl"#-->
-<div class="colmask">
-    <div class="section padTable">
-        <table id="infoTable">
-            <tbody>
-                <tr class="alt">
-                    <td class="infoTableHeader">$T('version'): </td>
-                    <td class="infoTableCell">$version [$build]</td>
-                </tr>
-                <tr>
-                    <td class="infoTableHeader">$T('uptime'): </td>
-                    <td class="infoTableCell">$uptime</td>
-                </tr>
-                <tr class="alt">
-                    <td class="infoTableHeader">$T('confgFile'): </td>
-                    <td class="infoTableCell">$configfn</td>
-                </tr>
-                <tr>
-                    <td class="infoTableHeader">$T('cache').capitalize(): </td>
-                    <td class="infoTableCell"><!--#set $msg=$T('ft-buffer@2')%($cache_art, $cache_size)#-->$msg</td>
-                </tr>
-                <tr class="alt">
-                    <td class="infoTableHeader">$T('parameters'): </td>
-                    <td class="infoTableCell">$cmdline</td>
-                </tr>
-                <tr>
-                    <td class="infoTableHeader">$T('pythonVersion'): </td>
-                    <td class="infoTableCell">$sys.version[:120]</td>
-                </tr>
-                <tr class="infoTableSeperator alt">
-                    <td class="infoTableHeader">$T('homePage') </td>
-                    <td class="infoTableCell"><a href="http://sabnzbd.org/" target="_blank">http://sabnzbd.org/</a></td>
-                </tr>
-                <tr>
-                    <td class="infoTableHeader">$T('menu-wiki') </td>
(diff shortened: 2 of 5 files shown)


Safihre:
Talking about the Wiki, @thezoggy made this very detailed page about the API: http://thezoggy.github.io/sabnzbd/api/
Why are we not using it @shypike ?

thezoggy:
that api page is really really old.. we were going to go with a readthedocs / docgen auto thing at one point but that never took off.. sab's api is horribly documented.. so its just a crazy task to start from scratch especially if you want to provide examples thats 'safe' :)

Safihre:
So what do you suggest? Expand your version maybe best?

thezoggy:
not sure, but prob should move this conversation (api doc) to its own github issue

Safihre:
Back to your PR: I tried it and liked it! Wondered if it really needed to be on tabs? This way it also hides the support part, and the page doesn't become that long (compared to the other Config-pages) when all is on 1 page. Also limits the need for extra translations..

thezoggy:
I had it all on the page but then it was getting a bit long.. so I went to tabs so would be shorter and easier to segment. Now segmenting the 'support' section to its own tab I figure may get some mixed reviews since some may want that as the default tab while others wouldnt... also the config root page is not in the navbar thus you can only get to this on the way to the config page from a skin.... so that support info probably isnt used a lot and may be best off being relocated else where / add config root page to the nav bar.

Safihre:
Adding the root page to the nav-bar isn't going to fit, it barely fits in English but in other languages I already had to use a smaller version of the logo to let the current pages to fit.
I would still vote for all on 1 page, but that's just me ;-)

thezoggy:
It is all on one page...

-----Original Message-----

Safihre:
Without tabs I meant :)

thezoggy:
well just copy the tables out of the tab content and you can see...  on linux it has a few extra rows (nice/ionice) and can get a little long.. but not a huge deal i just know you were trying to make everything mobile size.. why i did the tabs

Safihre:
Actually, why even show all this information on Windows and OSX? These come with all binaries included in the package. 
I would only show them for Linux or when they are not found (for development purposes, since yEnc and pyOpenSSL could be missing).

thezoggy:
not everyone installs using the binaries or checks out the source from git correctly as we found. like i run from git on windows, thus i do want to see the modules.. and if the binaries being used are the 64bit versions.. etc.

shypike:
Can you rebase to the latest "develop"?

thezoggy:
sure thing, also will look again about not doing tabs so we dont introduce more translations / make safihre happy :)

Safihre:
Appreciated ;-)
I still think when someone is running the binary version, we shouldn't show all this info, only when on Linux or Git. So when ```$build == 'unknown' or (not $nt and not $darwin)``` 
For the regular folk it's just too much information I feel..

shypike:
Shall I pick this up for 1.0.0 Final?
Seems low risk; dangerous words...

thezoggy:
let me rebase it and work on it some more..  hopefully my home internet is fixed today

Safihre:
And it might need new translations?

shypike:
@Safihre It's mostly tech info, missing translations isn't so bad.
@thezoggy I already fixed the Wiki links, so you can remove commit 4f3634bdda591fc7210609f639ab4d4ab7fbef98

thezoggy:
@shypike will do...

thezoggy:
alright redid it, no more tabs.. and ended up dropping most of it.. since the binary lines really dont say which one sab will use.. thus i dont want to cause more questions than it would answer.

sab12

Safihre:
Huh, why font awesome? We already have and use glyphicons from Bootstrap and never images. Why include another icon font?

thezoggy:
because images are quite inefficient at times, you have to make sacrifices when dealing with what bg color to use.. or make multiple images.. being able to change the 'icon' color on hover / action / scale depending on size.. just makes sense. I'd like for us to get away from glyphicons eventually.

Safihre:
Glyphicons is an icon font just like Fontawesome that I already use instead of images. The only image that is in the Config is the SABnzbd logo..
For just 2 icon's you would include all these extra icon font-files while you could have just used:
```html
<span class="glyphicon glyphicon-resize-full"></span>
<span class="glyphicon glyphicon-chevron-up"></span>
```

Bootstrap 4 will indeed not include Glyphicons any more, but then we can switch all (Glitter/Config/Wizard) at once to Fontawesome. Not include 2 icon fonts just for 2 icons on 1 page now for the final..

thezoggy:
i agree that font-awesome is overkill if we were to only use it for the two icons on the config root (which are in glyphicons as well). font-awesome does have far more icons than what bootstrap has.. which i think could benefit us greatly.. so anyways there are a few icons in glitter that dont show a hover state.. why i was thinking it was still images (last time i messed with bootstrap was 2.x)

thezoggy:
feel free to throw this in before 1.0.0 final

Safihre:
If this will go into 1.0.0, it needs some polishing because it doesn't match the rest of the new config. Sorry to scrutinize so much, but just spend quite a lot of time on that new config :relieved: 

image

thezoggy:
yes your being very nit picky but i'll humor you

the '+ Info' is what the translation shows.. I am just using whats available. thought you were anti creating more translations

the table does not have a border as its bootstrap and not custom css like the rest of the config, and the way you changed the custom css its a pain as its an alpha border and creates a funky effect if i try to apply it here

colors are just colors, green means proceed..  dunno what you have against green :)

two different tables with different content, of course they wont be aligned..  would have to apply css/bootstrap grid to specifically limit widths.. 

added collapse/fullscreen because of your love of wanting everything to be mobile friendly, and you gotta do it somewhere to show how its useful as I didn't want to go modify all the config pages with this 30 sec pull request

PR #460, opened 2016-02-05
Interface fixes part 12 (for 1.0.0 final)

Safihre:


Safihre:
The message about localStorage adds a new translation, but it's not essential that it gets translated before the final since just few people will see it.

PR #468, opened 2016-02-11
Test reliability of HTTPS certificate verification.

shypike:
If unreliable, disable it.


Safihre:
Implemented via #711.

PR #469, opened 2016-02-11
Improve handling of passwords embedded in the NZB.

shypike:
Show first password from NZB in the title of the user hasn't set one.
Try them all: user password, NZB-passwords, from password file.


thezoggy:
will check into this later, need to find sample nzb with passwords..  sadly i got my posting access taken away otherwise id just create my own and post

Safihre:
Looks good to me, just like I also was thinking of making it!

PR #474, opened 2016-02-15
Catch failing Windows Notification

Safihre:
I assumed Windows notifications could not fail, but clearly they can:
http://forums.sabnzbd.org/viewtopic.php?f=11&t=20211&p=104438
When no tray-icon is available, it will fail and in the case of completed NZB notification it will restart the whole of SABnzbd.

Should be in 1.0.0 I think :)

### DIFF
--- a/sabnzbd/growler.py
+++ b/sabnzbd/growler.py
@@ -541,5 +541,10 @@ def send_pushbullet(title, msg, gtype, force=False, test=None):
 
 def send_windows(title, msg, gtype):
     if sabnzbd.WINTRAY:
-        sabnzbd.WINTRAY.sendnotification(title, msg)
+        try:
+            sabnzbd.WINTRAY.sendnotification(title, msg)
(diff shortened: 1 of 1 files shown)


shypike:
Is it necessary to log a warning and a tracelog?
Cannot we just quietly ignore this?

Safihre:
We don't ignore it for any of the other notification methods? Like for OSX. The user could use the Test button in the config section to get some feedback why it's not working, that's true. I'll leave it up to you if warning is necessary :)

shypike:
@safihre A warning is only useful if the user can do something to fix it. Probably not so in this case.

Safihre:
So make it an info-level and the traceback debug-level?

shypike:
@Safihre Yes, please.

Safihre:
Done!

shypike:
Merged. Thanks.

PR #480, opened 2016-02-18
Minor config page fixes

thezoggy:
* Add Issues link back to config base page.
* Make sorting Key's look more like a button (not touching presets).
* Fix minor cosmetic issue with paths text being cut off.
* Remove bottom shadow from buttons when attached to inputs so it doesnt look so odd.


shypike:
Merged. Thanks.

PR #484, opened 2016-02-21
Fix missing trailing-slash in Glitter interface

jnikolich:
The "SABnzbd Config" link on the main screen is missing a trailing slash in (at least) the Glitter interface, causing failures when SABnzbd is accessed via a reverse proxy (tested with Pound and HAProxy)  This one-character change simply provides the missing slash.

PR #486, opened 2016-02-23
Interface fixes part 14

Safihre:
_Probably all of this should go into 1.0.0_
* Those ```/``` from #484, but then everywhere in Glitter 
* Improve message in Status window when a server is blocked and returning error, it did not display properly.
* Edited the first config page to use existing translation and add a link to the Wiki when not available. _@sanderjo @shypike maybe you can extend the information I put there? I don't know how yEnc should be installed and if my info about SSL install is correct.. http://wiki.sabnzbd.org/configure-1-0_
* Connection tab did not always refresh (not for 1.0.0, since other commit about this is also not in it)


Safihre:
@sanderjo @jeffgeorge I set the IPv6 and IPv4-address check to http://self-test.sabnzbd.org. 
Later the check for SSL can then more easily be added to this domain (if we want).
@jeffgeorge self-test is just as reliable as test-ipv6.sabnzbd.org was, right?

thezoggy:
on the config base page, for the ssl section: using `<em>` makes the protocols hard to read.. t and 1's are hard to distinguish when slanted

Safihre:
True, actually I was thinking we should replace those by how they are displayed in the Servers page, so TLSv1 instead of t1 for clarity.

thezoggy:
Sander would have to chime in on why he did this abbreviations 

-----Original Message-----

sanderjo:
Are you talking about the **last** line in this logging?
```
2016-02-14 15:07:35,834::INFO::[sabnzbdplus:1366] SSL version OpenSSL 1.0.1f 6 Jan 2014
2016-02-14 15:07:35,835::INFO::[sabnzbdplus:1367] SSL potentially supported protocols ['SSLv23', 'SSLv3', 'TLSv1']
2016-02-14 15:07:35,835::INFO::[sabnzbdplus:1368] SSL actually supported protocols ['t1', 'v23', 'v3']
```
So: the "actually supported protocols"? If so: AFAIK @shypike introduced that line (after I introduced the line before that.

Safihre:
I tested and it works fine by replacing ```t11``` by ```TLSv1.1``` etc as labels/dictionary-keys, the only problem is that here:
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/newswrapper.py#L209-L211
It uses the saved value for the server, so changing the keys would mean that if a beta/RC user set something else than default they would loose that setting.

thezoggy:
yeah not really seeing the point to make it pretty names when this is just debug info and (we) know what t1/v23/etc is.. plus the order is important

PR #487, opened 2016-02-23
Parse bandwidth limit as a float instead of an integer

chris-thorn:
To ensure speed limit value is calculated correctly when non-integer values are used for bandwidth limit.


shypike:
Merged. Thanks for your contribution!

PR #490, opened 2016-02-26
Update README.md

bobbysteel:
Update versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,4 +59,4 @@ Our many other command line options are explained in depth [here](http://wiki.sa
 
 ## About Our Repo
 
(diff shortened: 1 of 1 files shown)


jeffgeorge:
We really need to update the wording in the README.md more comprehensively. We're not really following Gitflow, we're sort of just doing . . whatever works. The README should actually describe how we plan on integrating code going forward.

bobbysteel:
Agreed. I'm trying to just remove confusion here. It's quite hard for non-technical folks to understand what's going on with v1.0+ at this point, as the only real way to follow is looking at the codebase itself. For example, some of the code in SickRage references 1.0 changes and there's no way at all I can see to easily find the 1.0 RC1 start point.

shypike:
After release of 1.0.0Final, I'll put a workflow plan in README.
See: https://github.com/sabnzbd/sabnzbd/issues/491

PR #493, opened 2016-02-27
Interface fixes part 15

Safihre:
**_All for 1.0.0 (so far)_**

* Glitter and Config small fixes.
* The feedback of _Test server_ and _Test notification_ was not the same. Now they are all displayed the same with a result-box under the test-button. 
 image
 image

* Implement message when non of the Enabled servers have the Default category selected, since we had a couple of reports of users that had no clue why nothing was downloading just because all their servers were linked to a category.
The message is hidden/shown immediately the user edits the settings (and on page-load), so the message appears before the user would press save.
 image


Safihre:
_The last 2 commits about the status-API and it's implementation in Glitter are not for 1.0.0_

PR #496, opened 2016-03-04
Handle 1/n-1 splitting

Tremax:
Some custom ssl library realization to prevent Rizzo/Duong-Beast attack encode whole data in separate chunks: in first chunk small amount of data(mostly 1 first byte) and append all other data in next chunk. In this case servertests don't work correctly, because **recv_chunk** function receive only first byte, and another chunk still in recv buffer. This situation could also happen in plain tcp when server will make some delay between sending chunks and when Nagle algorithm switched off -- but nobody will do it, so mostly ssl related.

Example:
Erlang ssl library split data into 2 chunk before send operation, for example: "200 welcome", would be sent like: "2" + "00 welcome", so in sabnzbd side(servertests.py) we receive first chunk: "2", and stop trying to receive more and servertests fails with **IndexError**(Server quit during login sequence)

Note:
It is related only to servertests part of functionality, while in downloading part reading from socket occur in infinity loop with select.

### DIFF
--- a/sabnzbd/utils/servertests.py
+++ b/sabnzbd/utils/servertests.py
@@ -21,6 +21,7 @@ sabnzbd.utils.servertests - Debugging server connections. Currently only NNTP se
 
 import socket
 import sys
+import select
 
 from sabnzbd.newswrapper import NewsWrapper
 from sabnzbd.downloader import Server, clues_login, clues_too_many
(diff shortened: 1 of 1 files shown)


thezoggy:
work on windows? what custom ssl library (do they even allow sab to work)?

Tremax:
Yes, works on windows just have a test.
Maybe i didn't clearly describe, but it is related to server side ssl. For example, ssl erlang library to prevent beast attack use 1/n-1 splitting, so server test functionality fails in sabnzbd, but all other features(downloading) works great, so it is related only to server test functionality.

Safihre:
Tested this and it works for all my servers with SSL enabled, since it's only called when clicking the _Test server_ button (so pretty harmless) I would say: merge! :+1:

shypike:
will do

Issue #498, opened 2016-03-05
1.0.0 RC2/RC4: Could not determine my public IPv4 address


sanderjo:
With 1.0.0 RC2 and RC4 I get "Could not determine my public IPv4 address"

RC1 is still OK.

I will investigate and report back. Related to the move of the ip address discovery to a different location?

I only discovered this now because @jcfp PPA still seems to be on RC1, so that is still on my Ubuntu/Debian boxes ...

```
2016-03-05 10:50:12,678::INFO::[SABnzbd:1244] SABnzbd.py-1.0.0RC4 (rev=1ba99769794134b18c8416631d51283ac4409b88)
2016-03-05 10:50:12,679::INFO::[SABnzbd:1245] Full executable path = /home/sander/sab-1.0.0-rc4/SABnzbd-1.0.0RC4/SABnzbd.py
2016-03-05 10:50:12,679::INFO::[SABnzbd:1257] Platform = posix
2016-03-05 10:50:12,680::INFO::[SABnzbd:1258] Python-version = 2.7.9 (default, Apr  2 2015, 15:33:21) 
[GCC 4.9.2]
2016-03-05 10:50:12,680::INFO::[SABnzbd:1259] Arguments = ./SABnzbd.py
2016-03-05 10:50:12,680::DEBUG::[SABnzbd:1266] My local IPv4 address = 192.168.1.245
2016-03-05 10:50:12,943::DEBUG::[SABnzbd:1274] Could not determine my public IPv4 address
2016-03-05 10:50:12,944::DEBUG::[SABnzbd:1278] My IPv6 address = 2001:470:blabla
```

Safihre:
Works fine here on Ubuntu 15.10 with RC4:
```
2016-03-05 10:11:44,502::INFO::[SABnzbd:1243] --------------------------------
2016-03-05 10:11:44,502::INFO::[SABnzbd:1244] SABnzbd.py-1.0.0RC4 (rev=1ba99769794134b18c8416631d51283ac4409b88)
2016-03-05 10:11:44,502::INFO::[SABnzbd:1245] Full executable path = /home/osboxes/Documents/SABnzbd-1.0.0RC4/SABnzbd.py
2016-03-05 10:11:44,503::INFO::[SABnzbd:1257] Platform = posix
2016-03-05 10:11:44,503::INFO::[SABnzbd:1258] Python-version = 2.7.10 (default, Oct 14 2015, 16:09:02) 
[...]

PR #501, opened 2016-03-06
Because of dual IPv4/IPv6 clients, finding the public ipv4 needs special attention

sanderjo:
as discussed here https://github.com/sabnzbd/sabnzbd/issues/498

### DIFF
--- a/sabnzbd/getipaddress.py
+++ b/sabnzbd/getipaddress.py
@@ -36,13 +36,39 @@ def localipv4():
     return ipv4
 
 def publicipv4():
+    # Because of dual IPv4/IPv6 clients, finding the public ipv4 needs special attention, 
+    # meaning forcing IPv4 connections, and not allowing IPv6 connections
     try:
         import urllib2
-        req = urllib2.Request("http://" + sabnzbd.cfg.selftest_host(), headers={'User-Agent': 'SABnzbd+/%s' % sabnzbd.version.__version__}) 
-        f = urllib2.urlopen(req, timeout=2)    # timeout 2 seconds, in case website is not accessible
-        public_ipv4 = f.read()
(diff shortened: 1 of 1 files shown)


thezoggy:
maybe a more elegant solution is just to have the page return a json result with both, then just parse the relevant one. 
```
{
    "ipv4": "8.8.8.8",
    "ipv6": ""
}
```

seems way better than some hack of trying to decipher which is which..

jeffgeorge:
I don't think that'll actually solve the problem @sanderjo is going after, because all I get from Cloudflare is the single environment variable `HTTP_CF_CONNECTING_IP`. It may be IPv4 *or* IPv6. If a user has access to both IPv4 *and* IPv6, I have no idea how it'd determine which protocol it actually connects over. I mean, I can parse it (so could you, by just checking the returned string for a `:`), but it's always going to be one or the other. Sander's problem is figuring out how to *force" a check over IPv4 if the client could connect over either.

Also: Is it asking too much to request that any connections to self-test.sabnzbd.org go over SSL? I *know* somebody is going to inspect these connections. When there's scrutiny I'd like to at least have it going over SSL.

thezoggy:
@inpheaux aww gotcha, thought it was just some php/code that you had that served up the info (aka, you had control)

Safihre:
Tried on my IPv6 VPN and works just as expected :) 
@jeffgeorge All we do is request the IP, something a sniffer/hacker would also know. Why now take the risk of more SSL relate problems when it's nothing that needs encrypting?

Safihre:
Merge this and get the 1.0.0 final release party started? :D

shypike:
@Safihre  Really? I'd rather do an extra RC. I get nervous when too many lines of code changes.
It's easy enough to forget a test case, as happened with RC3

Safihre:
This one only affects status information, not anything essential.

PR #502, opened 2016-03-07
Handling archives with multiple NZBs from API add_url call

paradix:
Please have a look, maybe it can be usable.

### DIFF
--- a/sabnzbd/dirscanner.py
+++ b/sabnzbd/dirscanner.py
@@ -70,23 +70,23 @@ def is_archive(path):
     if zipfile.is_zipfile(path):
         try:
             zf = zipfile.ZipFile(path)
-            return 0, zf
+            return 0, zf, '.zip'
         except:
-            return -1, None
+            return -1, None, ''
     elif is_rarfile(path):
         try:
             zf = RarFile(path)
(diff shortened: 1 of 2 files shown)


shypike:
Sorry for the late reaction.
Interesting change.
Do you happen to have a an example of an indexer serving multi-NZB files?

paradix:
I used to test it from a irc channel, serving such archives. Are you on irc?

shypike:
No, I'm not.
But I probably can setup some test situation where I can get a ZIP from a website.
I was mostly wondering abouy the use case, so it's practical usefulness.

paradix:
Well for me it works like that:
 - I get a notice with the URL from a bot on irc 
 - I catch it and send it to sabnzbd
 - if sabsnzbd is online it handles the archive download and the NZBs inside
 - if it's offline I save the archive locally, and upload it later once sabnzbd become available

That's pretty much my use case.

shypike:
Thank you.

thezoggy:
several indexers allow you to check off multiple items in the table, click download.. it serves up a zip of nzb's...  can easily see why someone would just want to load this up (especially if shared to another person for them to load).

PR #508, opened 2016-03-13
Interface fixes part 16

Safihre:
*For after the final :smirk:*

PR #512, opened 2016-03-17
update version numbers in README

neuhaus:


Safihre:
Already in #490 and for more files in #508, we are aware ;)

PR #513, opened 2016-03-19
Fix breaking Glitter bug with large script_log

Safihre:
If large chunks of HTML are put into the script_log by (for example) the old NZBDrone script, the ```preload_history``` and API call become so large that Glitter fails to load and chokes.

There is no need to have ```script_log``` in the API call, since all skins retrieve it using the ```/scriptlog?name=SABnzbd_nzo_lxyg7a``` call and it can make the API call way too large.

@shypike I verified the fix using the history.db of 2 of people that reported it!
Will you also put this in develop branch?

### DIFF
--- a/sabnzbd/database.py
+++ b/sabnzbd/database.py
@@ -495,7 +495,7 @@ def unpack_history_info(item):
         item['stage_log'] = [x for x in lst if x is not None]
 
     if item['script_log']:
-        item['script_log'] = zlib.decompress(item['script_log'][:])
+        del item['script_log']
     # The action line is only available for items in the postproc queue
     if not item.has_key('action_line'):
         item['action_line'] = ''


shypike:
A bit late, but doesn't removing the "script_log" element break existing third-party tools?

Safihre:
Maybe, but I opt to view this as a 'breaking API change', and that's allowed in a major release.
Since 1.0.0 isn't pushed officially to all clients through the latest-file, it's still oke I think.
Additionally we could add an official ```script_log`` API-call.

gargamol2000:
Thanks,

,,,, tk. 9

Amol Garg

Direct 1.425.77 0.8227

LinkedIn | Email

"If you play Godzilla backwards, it's about a giant lizard who rebuilds a city and then moonwalks into the ocean"8 8





On Sat, Mar 19, 2016 at 9:11 AM -0700, "shypike" <notifications@github.com> wrote:





A bit late, but doesn't removing the "script_log" element break existing third-party tools?


---

shypike:
On your advice I did change latest.txt on SF.net. Therefor I'll unbreak the API (by replacing script_log with a first line) and adding an extra API call.

Safihre:
Aah, I think my browser still had the latest-file cached when I checked today showing still 0.7.20. Now I see you changed it.. Oops! :sweat_smile:
That explains the increase in downloads I noticed!

```script_line``` in the API currently holds the _last _line of the script log.

PR #520, opened 2016-03-24
Improve interpreting of custom pause time for plain numeric input

savef:
Treat ambiguous numeric values as number of minutes for custom pause time.

Currently if you just type "100" into the custom pause field it'll think you want 143015 minutes, that's useless. A lot of people are probably used to the old Plush behaviour of entering the number of minutes you want to pause for, it's also a much saner default. So in the case that the user just enters some numbers and nothing else, this assumes they want to pause for that many minutes.

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.js
@@ -536,6 +536,12 @@ $(function() {
         
         // Update on changes
         self.pauseCustom.subscribe(function(newValue) {
+            // Is it plain numbers?
+            if(newValue.match(/^\s*\d+\s*$/)) {
+                // Treat it as a number of minutes
+                newValue += " minutes";
+            }
+
             // At least 3 charaters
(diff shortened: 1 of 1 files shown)


Safihre:
:like: simple but effective, great! 
Didn't think of that when adding this custom pause feature.

shypike:
Thank you!

savef:
That was quick, thanks! :smile: 

@Safihre, the feature is great in general, one of the highlights of Glitter for me (I've only just upgraded).

Safihre:
@savef I noticed now that for some weird reason it doesn't work for "40" or "30", no idea why.
But it does work for "40minutes", so I will change it to not do ```newValue += " minutes";``` but ```newValue += "minutes";``` (so without the space)!

savef:
@Safihre, I was able to replicate that, good find.

Safihre:
@shypike it's in 35d806b, in case you need it earlier for 1.0.1 :)
Don't really have enough other stuff to make a PR about it.

Issue #478, opened 2016-02-17
sab 1.0.x -- history


thezoggy:
requesting the history from sab can be a very costly thing.. i see many things we could do to help make this more efficient on both the backend and the frontend.

lets start with glitter requesting the current max amount of history:
api/history?start=0&limit=500

some notes:

* internally we limit history to `1000000`

* we search internally using python not sql.. thus we have to iterate over every result.. 

  > really could optimize this to just use sql... 

* we internally .reverse() the history results (which can get pretty large when your talking about the data set of 1000000), then we append post proc queue (.reverse there too)

  > not sure why we do all this ant not in sql.. as we have to hold a copy of the history in memory to do these actions..

one time hit is fine, but then to ask for the same thing when 99% of the time its going to be the same (until a new item is parsed) is just wasted cpu cycles
leverage the fact that history is showing the results in order by __completed__ time and stuff in post proc on top.. means we can easily sub out the limit from __500__ to just __1__ to check to see if something changed..
now to know if something changed, we just need to calculate the hash of the row data.. compare it to the last computed hash to see if its changed.. thus querying for one row to compare a string is much quicker/less memory than requesting the whole data
(since im not sure just querying/storing the latest `nzo_id` is reliable?)

now to put this all together all we need to do is add a hash param to the history (or just auto do it when limit=1) so that we dont rely on browsers/javascript to do it.. 
    api/history?start=0&limit=1&hash=1

really rough pseudo code
```
if no hash value:
  query history&limit=1, store it
  query history&limit=x
if has value:
  query history&limit=1
    if hash == last_hash
      return
    else
      query history&limit=x
```

shypike:
We **are** using SQL to extract the desired range.
database.py#260
```python
        t = (search, start, limit)
        cmd = 'SELECT * FROM history WHERE name LIKE ?'
        fetch_ok = self.execute(cmd + post + ' ORDER BY completed desc LIMIT ?, ?', t)
```
But I agree that history retrieval can be optimized.
Unlike the queue, it doesn't change all the time.

thezoggy:
not sure what your comment is regarding exactly...  yes sql is used to get the base data.. but then python is used to clean stuff up / search over / etc which can become very costly in memory/cpu.

shypike:
"We search internally using Python". That's not true. It's the conversion that is costly, so caching it is likely useful.

thezoggy:
[...]

PR #515, opened 2016-03-20
Feature/cherrypy5

shypike:
Update CherrryPy to official 5.1.0 release.

Re-apply almost all previous patches.
So far, CherryPy 5 seems to work fine.

PR #525, opened 2016-03-25
Do not calculate history every API call and more fixes

Safihre:
Investigating #478 I found the only time when the history is really updated is when the ```action_line``` is changed or when something is deleted.
And it works like a charm!
Is even faster when doing the 500ms delay in combination with this, but that requires some more work to fine tune.
image


shypike:
Nifty trick. A little bit dirty because a true optimization might be better.
Does this work for smpl, Plush and other API users (without changes)?
I cannot figure out what LAST_HISTORY_CALL actually carries. Last time someone used the API or last time the history was changed?

thezoggy:
hes setting the last time 'set_action_time' was called.. which happens on certain events.. when something is pp..  aka: `nzo.set_action_line()` is called.

Safihre:
Indeed I looked at the only time something happens with the queue and which functions are used then, and ```set_action_line``` is the common denominator! So ```LAST_HISTORY_CALL``` only houses a timestamp.
Also on Glitter preload the function is called without a ```last_history_call```, so always the freshest info and then only when something happens it will update. So also when multiple devices are viewing at same time, it will work since any pp-event or delete-event will trigger an update on any device.

@thezoggy only when a skin/call sets the ```&last_history_call``` parameter it is actually evaluated, otherwise it just behaves as normal and returns everything.

Safihre:
Found one more place that needed to update the ```LAST_HISTORY_CALL``` (see 1daa9e6) and accidentally pushed some unrelated commits in this branch :hamster: 
Maybe this can be merged (and #515) so others can test if any cases were missed?

I already noticed that idle CPU usage with "normal" sized queue/history lowered from 1% to 0.4% on my laptop's i5. So probably this will be even more beneficial for less powered devices.

shypike:
Which commits actually need to go into "develop"?

Safihre:
You mean if any should go into 1.0.1? I intended this feature  for develop, not for 1.0.1, since it might deserve some extra testing by others to make sure there are no unexpected freezes of the history. 
The other commits apply to Tabbed and Compact layouts, which are not in 1.0.x.

shypike:
You wrote "accidentally pushed some unrelated commits in this branch", but that remark wasn't valid for "develop".
So, only 597efc661ab8349c2a31049c932d5c21eb53b507 and 1daa9e66f686fb63dcf7717eea05b237dbee514e may go into 1.0.1 ?

Safihre:
Yes, but not sure if there might be some merge conflicts because of all the changes to glitter.js.
 And again can't promise the implementation is without faults yet..

shypike:
@Safihre In that case, this will not go into 1.0.1: safe bugfixes only.

Safihre:
Exactly what I was thinking  :)

PR #533, opened 2016-04-04
Prospective adding of par2 files when articles are damaged

Safihre:
Nowadays something is either (DCMA) removed or it is on the server with possible slight damage, like an article here and there malformed/damaged/etc.
Those last downloads fail QuickCheck, only to need 2 extra blocks or so.
So instead of waiting for the (possibly lengthy) par2-check, why not already download some extra blocks on the go when bad articles are detected?

My implementation uses a worst case scenario approach, where each bad article is expected to cause the need for 1 block. Could also be that 2 articles are damaged in the same block, in which case we download 1 block to many. I couldn't think of a better approach with the data available during the download.


sabnzbd/cfg.py
```diff
@@ -92,6 +92,7 @@ def validate_server(value):
 enable_filejoin = OptionBool('misc', 'enable_filejoin', True)
 enable_tsjoin = OptionBool('misc', 'enable_tsjoin', True)
 enable_par_cleanup = OptionBool('misc', 'enable_par_cleanup', True)
+prospective_par_download = OptionBool('misc', 'prospective_par_download', True)
```
shypike: Maybe we could combine this with the "enable_all_par" option.
But why should it be an option at all? Just always on?

Safihre: It was optional because I didn't know if everyone would like it by default, since there might be some overhead. But my thought was also to enable it by default. 
Don't know what would be a sensible combo with download all par, so will change it to enable by default :)

shypike: And add it to Config->Special ?

Safihre: Updated.

Safihre:
Some more real life testing today revealed that of course there should be a while-loop that adds par2 files until ```blocks_needed``` is met.
So far the implementation 1 bad article = 1 block seems to work very well, it has been accurate even for 10GB+ downloads. Really saving quite some time by eliminating those double verifications!

PR #536, opened 2016-04-05
Interface fixes part 18

Safihre:
Fixes and splitting ```glitter.js``` in sub-files that are combined on page load, it was just getting so big that it was hard to find what is doing what!


Safihre:
Only these are for 1.0.1:
41176bd 29cbf1d 0905606 3e53049

thezoggy:
fyi: http://blog.jquery.com/2016/04/05/jquery-1-12-3-and-2-2-3-released/

thezoggy:
why not use a build script on your side, so you can make it easy to dev/etc.. what makes it into sab upstream is the stripped slim down minified version.. what we used to do for the universal config and plush...

Safihre:
@thezoggy Still waiting for that final of jQueryUI 1.12, will give a nice reduction of size.
Will indeed consider a build script, but then it is only readable code for me and not easy for others to contribute like with the JS files of the first uniconfig.

thezoggy:
not like you cant create your own repo for it and then only provide the built files to sab when you want to update, etc. but still not the end of the world if its just priv since not like theres many others providing code changes to glitter...

PR #539, opened 2016-04-09
Prevent creating orphan items in "incomplete" when deleting job

shypike:
Prevent creating orphan items in "incomplete" when deleting downloading jobs.

Due to previous issues where articles could be lost,
the nzf.deleted and no.deleted flags were not obeyed.
This could lead to creation of orphans when lost articles would be flushed.

Better solution: drop articles only when job is in a final state.
Also prevent NZO files from being saved when job is in "deleted" state.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -933,8 +933,7 @@ def remove_data(_id, path):
             os.remove(path)
             logging.info("%s removed", path)
     except:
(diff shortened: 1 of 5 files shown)


Safihre:
Deleted jobs in multiple ways, high and low article caches, right when decoding of a file starts, etc
No orphaned jobs at all :+1:

PR #541, opened 2016-04-11
Safihre's fixes part 1

Safihre:
*For 1.1.0*

- When Queue is paused and downloading a job with priority Force that has missing articles, it would also download other jobs to just fill the connections. Not anymore.


shypike:
You forgot to add Glitter-propagating to skintexts.py
But what should be the actual text?
I wonder if the chosen column is the right one; it took me a while to see where the indication was. An alternative would the way delayed url-grabs are shown.

Maybe make a separate PR for propagation delay, because the other commits look OK to me.

Safihre:
Oke rebased the Propagation delay out of this PR and added a commit that includes the new logo in Config and Login page.
All good to go now!
image
image
image

thezoggy:
may want to add svg to gzip...
https://github.com/sabnzbd/sabnzbd/blob/develop/SABnzbd.py#L1451

thezoggy:
also you can save a request/logic and put the fallback image in the svg tag: https://css-tricks.com/svg-fallbacks/

or use the logic to detect if svg can be used, and act upon that:
https://css-tricks.com/test-support-svg-img/

Safihre:
Thought we decided to not to do fallback since we don't support IE8 anyway? 

Will add svg to gzip!

thezoggy:
@Safihre gotcha, yeah just IE8 would suffer.. 
http://caniuse.com/#search=svg

thezoggy:
@Safihre actually svg may already be handled..
https://github.com/sabnzbd/sabnzbd/blob/develop/SABnzbd.py#L1458

Safihre:
I choose actually to include the SVG's directly into the template using Cheetah, saves a request and no need to do gzip :)
https://github.com/sabnzbd/sabnzbd/pull/541/files#diff-0c33e6ce627e74ba80a2b09ae78c7485L9

thezoggy:
gotcha, well then i've got nothing else :) carry on. almost time for 1.0.1 ?

Safihre:
It's already time for 1.0.1 :wink: but from this PR only the forced items problem should be included in it.

Issue #545, opened 2016-04-14
Feature request getapikey without the existing apikey or with disabled api.


Hellowlol:
I would love to have some sort of functionality like to grab the apikey i like couchpotato has.

My idea is to:
If api is disabled enable it and make a new apikey, if webauth is enabled, username and password is required

This would nice to use with the zerfoconfig to automatically find the url and use the api to get the apikey :)

I can send a pr for this if your interested in this type of feature

Safihre:
Sure go ahead and make a PR, seems no harm in this feature.
There are already some supporting functions to check username/password and if API access is allowed externally (```check_access()```) etc in interface.py.

shypike:
Can you give us some more details? Because how I read it, it sounds like a security gap.

Hellowlol:
I dont think anybody uses sabnzbd without using the api. If your using sabnzbd without a username and password you already have a huge security gap. It doesn't really change the status que regarding this. 

If no auth: enable Apii and generate apikey
If auth: enable api generate apikey requires the username and password
If apikey exist, raise a error and point to the regenerate api that already exist

I'm open to other suggestion, but that was what I had planned.

shypike:
There's no security issue if you don't have a password. Most people use SABnzbd within their own PC or within their local network (which may or may not be safe). The API key (as a session key) was originally introduced to protect against cross-site scripting vulnerabilities where an external website could gain access to your purely locally running SABnzbd instance.

How does a third-party tool gain access to SABnzbd?
I don't know how CP does this, can you explain?

Hellowlol:
I guess that depends on the app. In this case I'd pla to use https://github.com/home-assistant/netdisco to find sabnzbd (it simply construct the sab URL) and query sabnzbd to get the apikey if the apikey wasn't in ha config file.

Cp:

../getKey?username=xxx&password=yyy 

If a username and password is set you need to md5 them if not just remove xxx and yyy
I don't think it will enable the api, i think the ui requires it and handles that that seperatly.

I'm sorry for the lack off formatting I only have access to my phone.

PR #548, opened 2016-04-16
Get apikey

Hellowlol:
This is a inital pr for https://github.com/sabnzbd/sabnzbd/issues/545

Must be true: Only requests from your localnetwork is allowed

Any of:
List of iprange,
disabled apikey,
blank username and password

if username and password set you you need to supply the ma_username and password or some of the other contions ablemust be true.
If apikey is given it needs to be correct.

I havnt really tought about the security risks so i have no idea if this is the worst idea ever.. :P

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -794,7 +794,7 @@ def _api_undefined(name, output, kwargs):
 def _api_browse(name, output, kwargs):
     """ Return tree of local path """
     compact = kwargs.get('compact')
-    
+
     if compact and compact == '1':
         paths = []
         name = platform_encode(kwargs.get('term', ''))
@@ -814,9 +814,7 @@ def _api_config(name, output, kwargs):
 
 def _api_config_speedlimit(output, kwargs):
     """ API: accepts output, value(=speed) """
-    value = kwargs.get('value')
-    if not value:
-        value = '0'
+    value = kwargs.get('value', '0')
     Downloader.do.limit_speed(value)
     return report(output)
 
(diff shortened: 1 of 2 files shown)


thezoggy:
which apikey? full or nzb? we already provide QR code for apikey's to make it easy to add to mobile..

Hellowlol:
Full key. Yeah and that awesome for mobile, in this case it would be sabnzbd would be automatically discovered on the local network and the thirdparty app would automatically fetch the apikey and be ready for use.

Safihre:
Evaluating your code I realized this exact function is already build-in to SABnzbd for a while and that it is also in the wiki:
http://wiki.sabnzbd.org/api#toc48
http://localhost:8080/sabnzbd/api?mode=config&name=set_apikey

This function will ask for username/password if those are set or for api-key when that one is set.
Also note that SAB will always auto-generate an API-key when it is first started. 
So I wouldn't know for sure how auto-discovery would work, only if that install by default also had ```disable_apikey``` turned on.

thezoggy:
he doesnt want to set an apikey.. which is also bad to do.. especially if the existing apikey is already set and used for other apps. he just wants to mimic apps like emby which listen to a multicast call to return info for discovery.. basically an app sends out a broadcast signature for sab.. and see if something comes back.. if so.. get said data (apikey) and use it. instead of expecting the user to know the sab info/apikey.... personally i think this request is dumb as your asking the user to give the app the username+pw of the sab instance.. which in my mind is much more sensitive than the apikey.. as a lot of time people use the username/pw in other places.. thus if this info is sniffed.. thats far worse than someone learning of the apikey.

Hellowlol:
@thezoggy It only returns the current apikey or generate a new one if it dont exist. The multicast thing already exist in sabnzbd.

Safihre:
Indeed I was mistaken!
@Hellowlol So this would just be in case you know the username and password but not the API key?
Because if you already know the API-key, you know it. And if the user disabled the API-key and has no username/password, you don't need it anyway.

Safihre:
@shypike I don't really see the general use-case for this, so for now we can close it (in my opinion).

PR #551, opened 2016-04-19
New ICO files - and other fixes

Safihre:
image


Safihre:
In my first attempts to speed up more of the downloading process of SABnzbd, I found that my download speed was mostly limited by the decoder pausing the downloader. 
Working my way through the steps the first thing I found was that [this for-loop](https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/decoder.py#L404-L406) is particularly slow (around 20-30msec during decoding of each article). 
This causes the decoder to pause the downloader en lowering speeds. I first increased the number of jobs allowed to be in the decoder-queue. But I realized that already doing this replacing of ```..``` to ```.``` in the downloader is less blocking.

I verified this on my work's gigabit connection by downloading the same NZB 3 times with and without these changes, showing a substantial increase in speed:
```
Before: 21.2 MB/s
With changes: 29.0MB/s

NZBGet: 70.1MB/s
```

I hope to find more improvements by actually seeing what the article looks

Stanzilla:
Oh wow, why is NZBGet so much faster in that case?

Safihre:
Optimized C++ vs Python, just much closer to the metal so to say. Also much of SABnzbd's central code is from back in the 2004 or earlier days and hasn't really been optimized since.

Although SABnzbd's performance is not linear on my system, downloading at limit of 12MB/s gives a steady 7-8% CPU usage (25% is max for 1 core, since it single threaded). 
However, with the limit on 24MB/s the usage is suddenly 22-25%. Suggesting the downloading itself is not the bottleneck now, but it's the decoding or other downstream processes that can't keep up with the downloading.

Safihre:
@shypike 
Wow, I found that setting the decoding queue limit to something super-high like:
```
MAX_DECODE_QUEUE = 1000000000
```
I can get a stable **50.2MB/s** and that moving the for-loop actually 'slows' things down to ~43MB/s. But: memory usage also goes crazy (saw up to 2GB because the decoder just couldn't keep up). 
However: when I do move the for-loop (see: a198290) memory usage is normal, about 150MB max at **~43MB/s**. While this NZB* downloads on 1.0.x at **27MB/s**.
That's a huge difference!

* I switched to a very recent NZB because previous results were provider limited because the post was 100d old.

Safihre:
Another NZB, on 1.0.x I can get max 8.7 MB/s while with the for-loop move and the super high limit: 25MB/s.
Wow!
image

sabnzbd/interface.py
```diff
@@ -273,6 +273,8 @@ def log_warning(txt):
     if not (cfg.disable_key() or nokey):
         key = kwargs.get('apikey')
         if not key:
+            key = kwargs.get('session')
```
shypike: Why the need for the session key? Skins should use TAPI instead of API.

Safihre: There is a security 'gap' with the new HTML login form because CherryPy doesn't force basic auth for the TAPI. So in https://github.com/sabnzbd/sabnzbd/pull/551/commits/69b6d5fead8491953b50692b3e43af214681bf64 I switched the TAPI to also use ```check_apikey```. Otherwise if the user disables his API key but has username/password turned on, there is no check for username/password because ```check_session``` assumes CherryPy's basic auth handles it.

But because I switched it, all the Config pages broke because they send a ```session=xxx``` instead of an ```apikey=xxx```.

Issue #550, opened 2016-04-17
SABnzbd is unusable with a very large queue


untotren:
I go through Sickbeard and backlog all my shows, a few thousand episodes.  And all these shows add to sab, then after they add SABNZBD starts to fall apart, lag itself into oblivion so I can not interface with the UI at all.

jeffgeorge:
1. **How many jobs are you trying to cram into SABnzbd simultaneously?** We can't try to diagnose this until we know the scale we're talking about.

1. **How are these jobs being added?** Is SABnzbd having to request the NZB over HTTP, or is the full NZB payload being provided as part of the original call? This will help us determine if more of the issue is before or after ingesting the NZB.

1. **What hardware / OS are you running SABnzbd on?** Disk and CPU info will be especially helpful here, as the bottleneck will likely be one or the other. If you can tell where the bottleneck is, that would be _even more helpful_.

1. **Is there any difference in performance if you try and do all this with SABnzbd paused?** It could be that part of what's freaking out is constantly shifting what jobs are loaded into memory for processing.

1. **Does *any* Usenet client perform to your standards under this kind of load?** It could be that what you're attempting to do is just completely out of bounds of reasonable operation, and is thereby not our fault.

shypike:
SABnzbd has issues with extremely large queues.
This is not an issue for most users, but in extreme cases SABnzbd should at least degrade gracefully. Currently it does not. We must try to fix this somehow.

thezoggy:
why the hell do you have thousands of episodes in your sickbeard backlog...

labrys:
@jeffgeorge As I have this issue (although it has improved significantly) I'd like to answer your questions for my case.

* **1. Jobs:** regularly > 1000-2000 jobs
* **2. Add Method:** HTTP, not full payload, from [Medusa](https://github.com/pymedusa/sickrage) (a sickrage/sickbeard fork)
* **3. Hardware:**
 * 6-Core AMD Phenom II X6 1100T Black Edition at 3.7GHz
 * Samsung 840 pro SSD
 * 32 GB RAM
* **4. Paused performance**:
 * Some increase in performance but still unusable with really large queues
 * Worst offender for performance has been mostly resolved (UI used to crash with lots of jobs and no display limit)
 * Lots of NZBs in queue that have not been resolved into a payload can cause the queue to stop resolving
[...]

PR #553, opened 2016-04-21
Optimize queue API

Safihre:
While our user in #550 originally used a somewhat unfriendly tone, he pointed out a valid point: SAB's API is slow for large queues.
This triggered @shypike to take a good look at the code and realized there was great room for improvement and already [made changes that drastically improved performance](https://github.com/sabnzbd/sabnzbd/commits/feature/optimize-api).
I also did some more tweaking of those and the results are really great!

Using a queue of 600 items totalling to 1.5TB with the global queue paused, these are the timings of the ```_api_queue_default``` function averaged over 2 minutes:

```Develop``` | @shypike  fixes | + @safihre fixes
------------ | -------------  | -------------
85.58 ms | 12.31 ms | 9.24 ms
- | **6.95x** faster | **9.26x** faster

- Plush also greatly benefits, SMPL not so much because it does 3 simultaneous API requests that slow each other down.
- When pausing the queue it would normally write the whole queue to the disk, this caused a complete freeze of up to 60 seconds with such queues. Now only the article cache is saved to disk.
- Removed the option to view files of a job on the main page in SMPL, this was ugly code that shouldn't be in the back-end functions.

Surely there are more API functions that could be improved, but this is the first one to be tackled!

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.main.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.main.js
@@ -302,7 +302,7 @@ function ViewModel() {
         // Only update the title when page not visible
         if(!pageIsVisible) {
             // Request new title 
-            callSpecialAPI('./queue/', {}).done(function(data) {
+            callSpecialAPI('./queue/', { limit: 1, start: 0 }).done(function(data) {
                 // Split title & speed
                 var dataSplit = data.split('|||');
 
--- a/interfaces/Plush/templates/_inc_header.tmpl
+++ b/interfaces/Plush/templates/_inc_header.tmpl
@@ -10,7 +10,7 @@
 <html>
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-  <title>SABnzbd $version - $T('queued'): $mbleft $T('MB')</title>
+  <title>SABnzbd $version</title>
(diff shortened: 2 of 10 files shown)


Safihre:
CherryPy's overhead is huge.. Even though API handlers take 2 or 3 msec to return data (including JSON encoding), the whole request still takes 25msec.
Even a simple HelloWorld example of CherryPy takes 15msec to respond.

Tried Tornado (because CouchPotato uses it and it also supports IPv6 and SSL, at least in CouchPotato), it only takes 2 or 3msec for it's HelloWorld.

Maybe we should consider switching at some point?

shypike:
A great unknown, Tornado, at least for us.
Any idea how much changes due to this?
We use a lot of secondary functions of CP.

shypike:
Merged manually and pushed to my branch.

thezoggy:
the sickrage guys switched from cherrypy to tornado... they regretted it. tornado is far more buggy than cherrpy and they didnt really see any speed differences that was worth anything.

Safihre:
Didn't know that, only saw CP using it and the benchmarks. But indeed Google for 'sickrage tornado' gives many hits, also from this year. 

Would also be a lot of work, so maybe not worth it. Probably other things that could be optimized.

PR #556, opened 2016-04-28
systemd service: add wiki link, remove group handling, housekeeping

jcfp:
Removing Group=%I because it (wrongly) assumes every username always has a matching groupname. Not specifying a group will simply make systemd use the default group for the given user.


shypike:
Thanks.

PR #565, opened 2016-05-02
Fixes part n+1

Safihre:
Without changing too much of the actual downloader code yet, there is performance to gain!

- Remove ```asserts``` throughout the code. Even though the ```-OO``` option is specified in the file header, it is not actually applied in many cases. For example JCFP's Ubuntu package will not use it. Gives performance boost and reduction of CPU usage of 10-20% in these cases.
- Decoder had a small queue of only max. 10 jobs, but on a fast connection with 50 connections this will fill up instantly and the decoder will pause the downloader because the Queue is full. Now the decoder queue is part of the article-cache so it is basically not limited. Speed before was limited to 25MB/s, now I get 45MB/s (although NZBGet still gets 75MB/s on the same NZB).
- ~~The downloader-while-loop was without a "speedlimit", causing it too loop too often without doing any action but still performing all the checks. Especially on slow connections it makes no sense to check so often. By limiting the while loop on my 60Mbit connection the CPU usage dropped from 31% to 19%.~~
- ~~Increasing the receive buffer size allows for making the downloader-while-loop even slower. But for some reason this is not possible for SSL connections, due to some kind of lock.~~

~~The current speedlimit on the downloader-while-loop does make SSL connections slower on high speed (above 25MB/s it seems to be limited to 28MB/s on the same NZB as above, while non-speed-limited it gets to 32MB/s). 
For the next iteration going to try other approaches to make the downloader-while-loop less heavy or maybe a speedlimit of the loop that's dependent on the current speed.~~


Safihre:
Also implemented:

- Only 1 error message will be shown when there's a connection problem, not 1 per connection.
- Removed the support for UUEncode since it would only be interesting if we actually supported multipart-UUencode, but we don't.. Only the (never seen) singlepart one. Since we only got 1 complaint about multioar-UUencode, clearly nobody is using that either.
- Added a 'Restore Defaults' button to the Switches page

sabnzbd/config.py
```diff
@@ -786,7 +786,6 @@ def _read_config(path, try_backup=False):
 def save_config(force=False):
     """ Update Setup file with current option values """
     global CFG, database, modified
-    assert isinstance(CFG, configobj.ConfigObj)
```
shypike: The isinstance() calls are there to help static analysers understand the code.
Removing them is a loss of information.
The alternative is:
if 0: assert isinstance(CFG, configobj.ConfigObj)
This should incur no performance penalty.

thezoggy: https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement

assert is similar to throwing an exception if a given condition isn't true... so it's good to have and should be minimal hit for anything like checking if CFG is an instance before we save. An important difference is that assert statements get ignored if you compile your code with the optimization option. Since the `__debug__` variable gets set to false...  anyways the fact that we arent acting upon any of the asserts makes them pretty pointless however..  as shypike said, he uses them with WING ide.. thus you can leave them in but just add the if 0 statement before (which means it'd never run).. even if `__debug__` is set to true.

https://wingware.com/doc/edit/helping-wing-analyze-code
```
if 0:
    assert <whatever>
```

sanderjo:
@Safihre 

I got your code (https://github.com/Safihre/sabnzbd.git) and ran it on an old Atom. A few different newsservers, no SSL, Fast Ethernet on a 100/100 Mbps Fiber connection. 

Resulting speed with your code: 
`Downloaded in 2 mins 59 seconds at an average of 5.5 MB/s`

I switched back to SABnzbd 1.0.1RC1, and did the same download: 
`Downloaded in 2 mins 46 seconds at an average of 6.0 MB/s`

So your code seems to be slower on my setup?

Safihre:
@sanderjo Damm, did not expect that. Seems I have to go back to the drawing board. 
I removed that last commit about the downloader. Could you maybe try again and compare 1.0.1 vs. mine? Might have to remove it and clone it again since probably it doesn't like that I removed a commit.
Just to make sure: on both 1.0.1 and when you run my version yEnc is working (shows a checkmark on first Config page).
This NZB is also undamaged so no fetching of extra blocks, right?

I don't have a slower system myself to test on, so indeed will need input from you :+1:

thezoggy:
wonder if the atom box isnt cpu bound but io bound. thus the optimizations arent really going to do much for it, if thats the case. move d/l directory to flash/ssd/etc? check stats.. cpu maxxed? memory maxxed? io maxxed?

@Safihre just force push to update your repo. to slow down your box you can do all sorts of stuff.. but an easy one is to just limit process affinity to one cpu.. so that way you have more of a chance to be cpu starved.

shypike:
@sanderjo Did you set a cache and how big is it?
Without a cache there might not be a sufficient decoder queue.

sanderjo:
git pull (correct?):
```
sander@netbook:~/git/safihre-sab$ git pull
From https://github.com/Safihre/sabnzbd
 + ec50390...16d9516 develop    -> origin/develop  (forced update)
Already up-to-date.

```
Result:
`Downloaded in 2 mins 51 seconds at an average of 5.8 MB/s
`

CPU load during download doesn't look bad; see screenshot:

screenshot from 2016-05-03 08-07-10

sanderjo:
@shypike 

Yes, cache was set during the tests: 777M
The Atom system has 1GB RAM. I will also try a lower cache. I don't expect anything from that, because there is still 792 MB RAM free.


```
sander@netbook:~$ free -m
             total       used       free     shared    buffers     cached
Mem:           991        709        282         36        140        369
-/+ buffers/cache:        198        792
Swap:         2212         45       2166

```

Safihre:
@shypike I changed the code for the decoder queue since yesterday morning, now when no cache is set it will use the old maximum decoder length.

@sanderjo Maybe you can try again tonight with both 1.0.1 and mine right after the other? To make sure that there's no other things at play. Interesting that load is not maximum, indeed it might be more I/O bound than CPU bound? I would expect it to consume all CPU power for decoding/assembling, but it doesn't?

sanderjo:
Hmmm: I downloaded with a certain other NZB-downloader, and I got 7.5 MB/s. So: better speed, but no line speed.

Safihre:
Added the ```asserts``` back using the ```if 0:```. 
Also found that using ```sys.getsizeof``` did not return the actual size of the data-list that was being put in the decoder/articlecache, most probably because it was just a reference instead of the real object.
Found that multiplying it by 32 is a small overestimation (about 2-3%), but better to be on safe side.
Tested this with NZB's with all kind of article sizes.

Will continue to work on the real performance improvements, so far it's mostly fixes. Updated the title accordingly.

Safihre:
@sanderjo I made my VM to only use 20% of CPU to reproduce a slow system and the changes that are still in the request do improve the speed on my system.
```
#1 (small)
Downloaded in 53 seconds at an average of 7.0 MB/s
vs
Downloaded in 1 min 1 sec at an average of 5.7 MB/s

#2 (bigger)
Downloaded in 2 mins 20 seconds at an average of 7.1 MB/s
vs
Downloaded in 3 mins 2 seconds at an average of 5.6 MB/s
```
That's probably just from increasing the decoder size. Cache was set to 150M.

thezoggy:
i should check out the fixes on this laptop, it is constantly pegged at 100% trying to download which ends up limiting me to 12-14mb/s... you can easily tell when your not getting max speed as the download bar is super spikey..  when downloading in sab I see windows defender takes up like 24% cpu... wtf..

Safihre:
Yep same here, another reason not to use defender. In other anti virus you can at least turn of the checking if nntp connections combined with excluding the temp. Folder..

thezoggy:
i disabled defender via regedit (since i couldnt do global policy on my version of win10).
but we should check if this provides a 'fix' and recommend people adding an exclusion for the incomplete folder / sab exe etc... maybe offer a button to auto set it?
http://windows.microsoft.com/en-us/windows-10/add-an-exclusion-to-windows-defender

shypike:
Thank you!!

Issue #606, opened 2016-06-22
Server warnings never get cleared


Safihre:
So there are server warnings (```server.warning```) and server errors (```server.error```).
But it seems those warnings never really get cleared?

I was working on showing them a bit better in Glitter, since now they get cruedly placed in the ```serverconnections``` output field (for some compatibility reason probably).
For example my primary server is not being used because the current downloads are out of his retention-limit, but it has been showing this warning for hours now, even though the server is perfectly fine:
![image](https://cloud.githubusercontent.com/assets/5703454/16266670/5fc7724c-3886-11e6-9cb2-2b62388a5c35.png)

Safihre:
Fixed? I didn't fix this, just referenced from the PR that I found this issue when working on something else.

PR #607, opened 2016-06-23
Fixes and adding sabnzbd.ini to logging

Safihre:
Sometimes a bug is caused or only present in a very specific combination of settings, but without information about those settings bug-hunting take much longer.
Therefore it seemed like a good idea to attach the ```sabnzbd.ini``` to the end of the log when a user download's it on the status page.
The usernames/passwords/api-keys get cleared using the same routines as used to clear the log.
Added a message to the top of the log to inform users that their settings-file is attached.

Also fixed some bugs and made the server warnings displayed better (#606).

### DIFF
--- a/ABOUT.txt
+++ b/ABOUT.txt
@@ -10,10 +10,10 @@ SABnzbd also has a fully customizable user interface,
 and offers a complete API for third-party applications to hook into.
 
 There is an extensive Wiki on the use of SABnzbd.
-http://wiki.sabnzbd.org/
+https://sabnzbd.org/wiki/
 
 IMPORTANT INFORMATION about release 1.0.0:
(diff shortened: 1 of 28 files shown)


shypike:
There's not quite enough censoring of the INI file.
These fields remain untouched:

email_pwd
pushover_token
pushover_userkey
pushbullet_apikey
prowl_apikey
growl_password
growl_server

Safihre:
:+1: you are right. I added an extra regular expression that will filter these out.
If ever we add a new variable that should be hidden, we need to manually add it to the regular expression  though..

Safihre:
@shypike @jeffgeorge Updated all the references to ```http://wiki.sabnzbd.org``` to the new addresses.

PR #609, opened 2016-07-01
Feature/cherrypy602

shypike:
Time we upgraded again.
Some bugs are resolved, now needing less patches.


thezoggy:
just internal note from [pyOpenSSL](https://pyopenssl.readthedocs.io/en/stable/changelog.html): 
`Python 2.6 support has been deprecated. Our main dependency cryptography deprecated 2.6 in version 0.9 (2015-05-14) with no time table for actually dropping it. pyOpenSSL will drop Python 2.6 support once cryptography does.`

PR #617, opened 2016-07-08
Add .pyd to .gitignore

labrys:
This adds .pyd (essentially python .dll files on windows) to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 #Compiled python
-*.py[co]
+*.py[cod]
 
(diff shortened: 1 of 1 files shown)


thezoggy:
would think its user error to have .pyd files on a source install...

labrys:
@thezoggy True it's not really necessary, but it doesn't really hurt either.  Just noticed it was not included in the gitignore.  Up to you if you do/don't want to include.

shypike:
Done.

Issue #618, opened 2016-07-09
Restarting SABnzbd behind reverse proxy redirects to incorrect URL


phobozad:
When restarting the server after e.g. a config change, the browser is redirected to an incorrect URL.

This is seen with the following setup:

Reverse proxy (nginx) listening on port 80
SABnzbd listening on port 8079

URL used to access SABnzbd: http://sabnzbd.example.com
URL that SABnzbd redirects to after restarting: http://sabnzbd.example.com:8079

thezoggy:
can you post your redirect/proxy info?

phobozad:
Here is the nginx config:
```
server {
        listen          [::]:80;
        server_name     sabnzbd sabnzbd.example.com;
        location / {
                proxy_pass              http://127.0.0.1:8079;
                proxy_set_header        X-Real-IP $remote_addr;
                proxy_buffering off;
        }
}
```

sanderjo:
@phobozad 

No problems with my SAB-behind-a-proxy (Apache, BTW).

So:
which SABnzbd version do you use?
which skin do you use?

phobozad:
This is on 1.0.3 with the Glitter skin. 

@sanderjo Are you running your SAB server on a different port than your reverse proxy?

sanderjo:
> This is on 1.0.3 with the Glitter skin.

OK, good; so the skin is not the problem.

> Are you running your SAB server on a different port than your reverse proxy?

Of course (two servers on one port is not possible):
SAB on port 127.0.0.1:11111
Reverse proxy on [::]:80

Safihre:
@sanderjo He is right, this does not work and also can't work in your case:
When you change a setting on the _General_ page of the Config that requires a restart and you click yes when it asks you to, you will be redirected to the port that is specified on the General page.
This is by design.. How can SAB know what your proxy settings are? All it knows is that on that page you set port 8079, so this is what it will redirect you to. Maybe you changed it, that's why we redirect to that port, and not the port you came from.

Technically we could only redirect to the port set in the Config when it's changed, of course.

sanderjo:
> When you change a setting on the General page of the Config that requires a restart and you click yes when it asks you to, you will be redirected to the port that is specified on the General page.

Ah ... we were talking about different Restart use cases:

[...]

PR #633, opened 2016-07-23
Only redirect browser on restart if port configuration changed

phobozad:
Possible fix for #618 

This will only redirect the user's browser to a new URL if the http/https port configuration changed.

In the case where it changes, the same guessing logic currently in place will be used to smartly redirect the user's browser to the new destination port.

In every other case, it will reload using the same URL/port that the user is currently accessing the server with (i.e. the restart will function as if the user selected "Restart SABnzbd" from the hamburger menu.

### DIFF
--- a/interfaces/Config/templates/config_general.tmpl
+++ b/interfaces/Config/templates/config_general.tmpl
@@ -21,7 +21,7 @@
                 </div>
                 <div class="field-pair">
                     <label class="config" for="port">$T('opt-port')</label>
-                    <input type="number" name="port" id="port" value="$port" size="8" />
+                    <input type="number" name="port" id="port" value="$port" size="8" data-original="$port" />
                     <span class="desc">$T('explain-port')</span>
                 </div>
                 <div class="field-pair">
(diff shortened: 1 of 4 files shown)


Safihre:
But.. How does this solve your problem? It will still redirect incorrectly.

I think we can solve your problem on the client side without introducing another global. By changing both
````html
<input type="number" name="port" id="port" value="$port" size="8" />
<input type="number" name="https_port" id="https_port" value="$https_port" size="8" />
```
to
```html
<input type="number" name="port" id="port" value="$port" size="8" data-original="$port" />
<input type="number" name="https_port" id="https_port" value="$https_port" size="8" data-original="$https_port />
```
and then doing a check in the redirect javascript:
```javascript
var urlPort = ($('#port').val() == $('#port').data('original')) ? location.port : $('#port').val();
var urlPort = ($('#https_port ').val() == $('#https_port ').data('original')) ? location.port : $('#https_port ').val();
```

phobozad:
Good call, I'm too much of a jQuery noob to have realized it could be done that simply.

I used the same base methodology to redirect, but implemented the logic slightly differently than what you have above.  If we just throw in those ternary ops it does technically work.  However, when using port 80/443, location.port is empty so when we build the URL, it would look like: 

`https://sabnzbd.example.com:/sabnzbd`  _(note the colon thrown in there)_

So instead of that, after checking which config page is being saved, it now checks to see if the new/old port configs are different.  If not, we redirect the same way as on a normal restart.  If it is different, then we redirect to a new URL based on the new http/https port value.

BTW the previous commit did work to solve the issue, not sure why you said it would incorrectly redirect.  But I think your method doing it all client-side is much better than adding in another global (yuck). 😄

Safihre:
Looks good to me :+1:
Yes I understand now how the previous version worked :)

PR #638, opened 2016-07-26
Allow 7zip to detect archive type

tcare:
This change fixes an issue with 7zip extraction failing for a supported archive type. The '-t7z' parameter will only allow 7z file types, and excludes others, e.g. Split type (.7z.001). By removing -t7z we allow 7zip to detect the archive type, and prevent 7z from errorring out with a valid archive.

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -1020,8 +1020,7 @@ def seven_extract_core(sevenset, extensions, extraction_path, one_folder, delete
     if not os.path.exists(name):
         return 1, T('7ZIP set "%s" is incomplete, cannot unpack') % unicoder(sevenset)
 
-    command = [SEVEN_COMMAND, method, '-y', '-aou', '-t7z', case, password,
-               '-o%s' % extraction_path, name]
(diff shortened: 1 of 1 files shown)


shypike:
I'm not sure why we added the -t7z parameter. It may have been a security issue, but I'm not sure.
At the time this was implemented, 7zip posts were very rare and they're still not common.
Do you happen to have a few example NZB files that demonstrate the difference that you patch makes? bugs@sabnzbd.org

Safihre:
It was because of those vulnerabilities found when unpacking non-7z.
See: https://github.com/sabnzbd/sabnzbd/commit/462a47c9270b7e8aa99e6a6f2ddd76cfc20878ae

shypike:
@Safihre So the sad by-effect is that split archives are no longer recognized?
That sucks.

tcare:
@shypike I'll email an example.

@Safihre That's a shame - however an alternative approach may be to fall back to -tSplit?

shypike:
@tcare That is probably possible. But the parameter can only be used on actual split files. So SABnzbd should first determine whether it is actually a set. This should be easy enough: #files > 1 means split file.

shypike:
Well, that was easy: 3a426e9367c722340949967f512f3f53e0961399
Should have thought of that before.

Issue #634, opened 2016-07-23
1.1.0RC2 post process script-output "more" link broken


OldskoolOrion:
As the topic states. It's of course not an urgent issue : minor glitch that's all.
Included a clear partial screenshot of the issue - I don't think much more additional info is needed from the logfile (of sab and/or script), but if need be I can post it.

![sabnzbd_1 1 0rc2_small_issue](https://cloud.githubusercontent.com/assets/13029885/17080447/7b2e53d0-5130-11e6-9df5-fe11b6a6aec8.jpg)

Safihre:
Oops, my fault. Will fix!

OldskoolOrion:
Damn.. i'll have to find the location in the source first, so I won't race you :-)  Thanks for all the effort put in !

PR #639, opened 2016-07-26
Add HTTPS checkbox to Switches

Safihre:
@sanderjo / @shypike What do you think about the wording? Did not want to add info about minimal requirements etc.

It doesn't work when SSL is not available, right? I added that check.

### DIFF
--- a/interfaces/Config/templates/config_switches.tmpl
+++ b/interfaces/Config/templates/config_switches.tmpl
@@ -31,6 +31,11 @@
                         <span class="desc">$T('explain-ampm')</span>
                     </div>
                     <div class="field-pair">
+                        <label class="config" for="enable_https_verification">$T('opt-enable_https_verification')</label>
+                        <input type="checkbox" name="enable_https_verification" id="enable_https_verification" value="1" <!--#if int($enable_https_verification) > 0 then 'checked="checked"' else ""#--> />
+                        <span class="desc">$T('explain-enable_https_verification')</span>
+                    </div>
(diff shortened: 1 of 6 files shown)


shypike:
@Safihre It's more subtle.
We test for presence of PyOpenSSL. Standard Python should contain SSL support, but not at the level that NNTP needs. I think that "out-of-the-box" Python always supports https. Maybe some deranged embedded versions do not, but those are not likely to be 2.7.9+ anyway.
So the SSL test isn't really needed. Besides, it won't hurt if the checkbox is available while https isn't working.

Safihre:
@shypike Removed the check for SSL. Do we really need a check if it's the right Python version?
Can't we just make 2.7.9 the minimum requirement for SABnzbd and we done with it?
Because let's face it: from the [download statistics](https://sabnzbd.org/download-stats), 95+% of our users are on Windows/OSX, and the Linux PPA also has the right version. 
Plus we know that QNAP and SynnoCommunity also offer the latest versions.

shypike:
@Safihre No, we cannot make 2.7.9 a requirement. That would exclude too many embedded systems. We don't see the statistics, because they download third-party packages.

shypike:
BTW: There's no need to suppress the actual option on pre 2.7.9 versions.
The enable/disable code already checks for presence of verification capabilities.

Safihre:
Added the #634 fix, if you're fine with the HTTPS text then this is good to go!

Issue #636, opened 2016-07-24
sab 1.1.0RC2 -- glitter erronous error


thezoggy:
error for saving admin folder reported in glitter, however everything worked just fine.
![sab](http://zoggy.net/firefox_2016-07-23_03-28-11.png)

havent dug into it, but I assume it error saved on the folder.. it retried/cleanuped folder and worked on the next time.. thus the error being reported is a little premature/pointless.

thezoggy:
relevant log entries:
```
2016-07-23 03:26:22,174::DEBUG::[assembler:337] file in rar file: JWXJN8cPzMowm3Gs3u2MH2v1iNOPqx5J3D85BsAtBu.mkv
2016-07-23 03:26:22,174::INFO::[assembler:88] Decoding \\?\C:\Users\zoggy\downloads\incomplete\Mr.Robot.S02E03.eps2.1 k3rnel-\JWXJN8cPzMowm3Gs3u2MH2v1iNOPqx5J3D85BsAtBu.vol003+004.par2 yenc
2016-07-23 03:26:22,204::INFO::[assembler:88] Decoding \\?\C:\Users\zoggy\downloads\incomplete\Mr.Robot.S02E03.eps2.1 k3rnel-\JWXJN8cPzMowm3Gs3u2MH2v1iNOPqx5J3D85BsAtBu.vol001+002.par2 yenc
2016-07-23 03:26:22,210::ERROR::[__init__:892] Saving \\?\C:\Users\zoggy\downloads\incomplete\Mr.Robot.S02E03.eps2.1 k3rnel-\__ADMIN__\SABnzbd_nzo_wi7zd0 failed
2016-07-23 03:26:22,211::INFO::[__init__:893] Traceback: 
Traceback (most recent call last):
  File "sabnzbd\__init__.pyo", line 882, in save_data
RuntimeError: dictionary changed size during iteration
2016-07-23 03:26:22,213::INFO::[__init__:945] Saving data for totals10.sab in C:\Users\zoggy\AppData\Local\sabnzbd\admin\totals10.sab
2016-07-23 03:26:22,246::INFO::[assembler:88] Decoding \\?\C:\Users\zoggy\downloads\incomplete\Mr.Robot.S02E03.eps2.1 k3rnel-\JWXJN8cPzMowm3Gs3u2MH2v1iNOPqx5J3D85BsAtBu.vol007+008.par2 yenc
[...]

PR #641, opened 2016-07-29
Bugfixes/shypike

shypike:
This should solve a vulnerability, work-around database locks and prevent failed saving of NZO files.

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -187,36 +187,7 @@ class Decoder(Thread):
             if logme or not found:
                 # Add extra parfiles when there was a damaged article
                 if cfg.prospective_par_download() and nzo.extrapars:
-                    # How many do we need?
-                    bad = len(nzo.nzo_info.get('bad_art_log', []))
-                    miss = len(nzo.nzo_info.get('missing_art_log', []))
-                    killed = len(nzo.nzo_info.get('killed_art_log', []))
-                    dups = len(nzo.nzo_info.get('dup_art_log', []))
-                    total_need = bad + miss + killed + dups
-
-                    # How many do we already have?
-                    blocks_already = 0
-                    for nzf in nzo.files:
-                        # Only par2 files have a blocks attribute
-                        if nzf.blocks:
-                            blocks_already = blocks_already + int_conv(nzf.blocks)
-
-                    # Need more?
-                    if blocks_already < total_need:
-                        # We have to find the right par-set
-                        for parset in nzo.extrapars.keys():
-                            if parset in nzf.filename and nzo.extrapars[parset]:
-                                extrapars_sorted = sorted(nzo.extrapars[parset], key=lambda x: x.blocks, reverse=True)
-                                # Loop untill we have enough
-                                while blocks_already < total_need and extrapars_sorted:
-                                    # Add the first one
-                                    new_nzf = extrapars_sorted.pop()
-                                    nzo.add_parfile(new_nzf)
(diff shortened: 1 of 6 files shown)


shypike:
Should solve:
https://forums.sabnzbd.org/viewtopic.php?f=2&t=21173
https://github.com/sabnzbd/sabnzbd/issues/634
https://github.com/sabnzbd/sabnzbd/issues/636

Safihre:
:+1: running smooth so far..

Safihre:
Something is wrong with the changes.
When I download a post that is clearly gone from the servers, [like this one](https://github.com/sabnzbd/sabnzbd/files/406434/The.Big.Bang.Theory.S09E18.1080p.HDTV.X264-DIMENSION.nzb.gz), and ```prospective_par2``` is turned on, I can see in the console that it never discards articles because they are missing from all servers. So it will just go on forever at 50KB/s.. 
Turning off ```prospective_par2``` will make it discard articles again.

**But:**
When I switch back to my ```develop``` from #643  (also with ```prospective_par2``` turned on), articles are nicely discarded when both of my servers are tried and then it stops the download when its hopeless.

shypike:
OK, back to the drawing board.

sabnzbd/decoder.py
```diff
@@ -187,36 +187,9 @@ def run(self):
...
-                                    new_nzf = extrapars_sorted.pop()
-                                    nzo.add_parfile(new_nzf)
-                                    nzo.extrapars[parset] = extrapars_sorted
-                                    blocks_already = blocks_already + int_conv(new_nzf.blocks)
-                                    logging.info('Prospectively added %s repair blocks to %s', new_nzf.blocks, nzo.final_name)
-                                    # Reset all try lists
-                                    NzbQueue.do.reset_all_try_lists()
+                    nzo.prospecive_add()
+                    # Reset all try lists
+                    NzbQueue.do.reset_all_try_lists()
```
Safihre: This resets the TryList after *every* missing article. 
So it needs to go...

sabnzbd/nzbstuff.py
```diff
@@ -1185,6 +1193,41 @@ def remove_extrapar(self, parfile):
...
+                if parset in nzf.filename and self.extrapars[parset]:
+                    extrapars_sorted = sorted(self.extrapars[parset], key=lambda x: x.blocks, reverse=True)
+                    # Loop untill we have enough
+                    while blocks_already < total_need and extrapars_sorted:
+                        # Add the first one
+                        new_nzf = extrapars_sorted.pop()
+                        self.add_parfile(new_nzf)
+                        self.extrapars[parset] = extrapars_sorted
+                        blocks_already = blocks_already + int_conv(new_nzf.blocks)
+                        logging.info('Prospectively added %s repair blocks to %s', new_nzf.blocks, self.final_name)
```
Safihre: Here!
Only when a prospective block is added, we reset all.

I think we could limit the resetting to only when it's a ```.par2``` file, but not sure enough to change it now before the final.

sabnzbd/decoder.py
shypike: Case of bad refactoring.
I'll fix that.

shypike:
I pushed a fix.
Should the reset be inside or outside the while-loop.
Inside looks like overkill to me, but could there be situations where the while-loop would run zero times?

Safihre:
This will return False when there are no more par2-files left, right?
```
if parset in nzf.filename and self.extrapars[parset]
```

In that case, the location outside the loop would be best.

PR #643, opened 2016-08-02
Airplane-mode fixes

Safihre:
Long wait in an airport without internet and in the airplane made me find number of small improvements, mostly in the Config

* Display message to help translate when user selects non-English in the Config
* Update text to Scripts Folder (removed the 'post-processing' from the label)
* fail_hopeless=True by default (it has proven itself and is very usefull in the current age of DCMA-every-day)
* Move AM/PM switch to Specials (only applies to SMPL really)
* Don't show Sanatize for Windows, on Windows. Format AllPar2 explain text
* Fix Wiki link on first page of Config (linked to FAQ, instead of Wiki-main)
* Encoding was logged twice during startup, due to @sanderjo and @shypike both adding it
* Show Encoding on first Config page (so people don't have to check the log when we need that info)
* Human readable SSL labels on first Config page
* Remove code for v0.5.x (it's 2016)
* Change text to 'HTTPS certificate verification'

### DIFF
--- a/ABOUT.txt
+++ b/ABOUT.txt
@@ -15,28 +15,17 @@ https://sabnzbd.org/wiki/
 IMPORTANT INFORMATION about release 1.0.0:
 https://sabnzbd.org/wiki/introducing-1-0
 
-
 Please also read the file "ISSUES.txt"
-
-
-*******************************************
-*** Upgrading from 0.7.x or 0.6.x       ***
-*******************************************
-Empty your current queue
-Stop SABnzbd.
-Install new version
-Start SABnzbd.
-
-
-*******************************************
-*** Upgrading from 0.5.x                ***
-*******************************************
-Stop SABnzbd.
-Uninstall current version, keeping the data.
-Install new version
-Start SABnzbd.
 
 The organization of the download queue is different from 0.7.x (and older).
 1.0.0 will not finish downloading an existing queue.
 Also, your sabnzbd.ini file will be upgraded, making it
 incompatible with older releases.
+
+*******************************************
+*** Upgrading from 0.7.x and below      ***
+*******************************************
+Empty your current queue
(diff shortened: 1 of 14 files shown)


Safihre:
From some discussions on the forum and on Reddit, I noticed some of Glitter's features might be a little *too* hidden. Things like multi-edit, or extra queue/history columns, or the option to delete history items for a specific search term, or custom pause, etc.
So added a notification with a link to the page https://sabnzbd.org/wiki/extra/glitter-tips-and-tricks, which I will fill with these kind of features. Notification will show once and if closed won't appear again.

shypike:
"Move AM/PM switch to Specials"
Why? And why does the commit text say it only applies to Smpl?

"Human readable SSL labels on first Config page"
Sensible. However, when a symbol begins with an underscore, it's not supposed to be used externally. So either change its name or create an access function for it.

"Change text to 'HTTPS certificate verification'"
There go the translations...

Safihre:
> Why? And why does the commit text say it only applies to Smpl?

Only when you use SMPL, this switch will actually change anything. For Glitter there are many more date-formats available, so you need to pick that in the Glitter settings and changing this switch here doesn't actually change anything for the user.

> Sensible. However, when a symbol begins with an underscore, it's not supposed to be used externally. So either change its name or create an access function for it.

Oke will do.

> There go the translations...

There are no translations of this text yet, because you didn't push it to Launchpad since we added this switch to the Switches page ;)

shypike:
> Only when you use SMPL, this switch will actually change anything. For Glitter there are many more date-formats available, so you need to pick that in the Glitter settings and changing this switch here doesn't actually change anything for the user.

Makes sense.

> There are no translations of this text yet, because you didn't push it to Launchpad since we added this switch to the Switches page ;)

Ahum

PR #648, opened 2016-08-09
Glitter history fix

Safihre:
Added ```index```-field to history-API because in Glitter items that were post-processing used to be in a weird order, because it only had the ```completed```-timing field to use, which was updated while a job was waiting to be post-processed. Now the ordering is normal.
I should have fixed this months and months ago :-1:


Safihre:
@shypike You already fixed the TryList, so removed that one from this pull-request :)

PR #653, opened 2016-08-10
Diving into NZO details

Safihre:
~~Finally, I think I found the main reason for the ```Saving SABnzbd_nzo_XXX failed```. There are many, many, places where nzo properties get directly edited, too many to all put in IO_LOCK.
But in ```register_article``` there was a call to ```save_data``` immediately followed by setting ```nzo.next_save```. By just putting the ```save_data``` after this modification: problem solved :+1:
I know 99.99% sure it's this one, because in the log the ERROR is _always_ followed by ```Saving data for totals10.sab``` (also in @thezoggy's log in #636), which is the next command (```BPSMeter.do.save()```) in  ```register_article```.~~
So, I reverted the making copies in ```BPSMeter.do.save()```, because it's not the cause of the problem.

* I also noticed in my logs that the NZO-info is saved **a lot**, there is no need to save it every 6 seconds, every 30 seconds is just fine.
* Also found that upon saving the queue, the ```save_data``` for each NZO is called twice. Fixed that.
* Disabled checking of the rar-contents when there's no need, saves disk operations and CPU cycles. This still could use some work.. why do we need to check *every* rar file for encryption and then seperatly for unwanted extensions? ```RarFile``` is called twice for _every_ rar, on high-speed connections this happens **a lot**.
* Added a possibility to upload multiple NZB's using the upload-form
* Both for the form and drag-and-drop there is now a proper uploads-counter, that counts correctly and until everything is done.

Downloaded 500GB in the last 24hours for testing.. Poor SSD :P


sabnzbd/assembler.py
```diff
@@ -326,7 +326,7 @@ def rar_contains_unwanted_file(filepath):
     # returns False if no unwanted extensions are found in the rar file
     # returns name of file if unwanted extension is found in the rar file
     unwanted = None
-    if is_rarfile(filepath):
+    if is_rarfile(filepath) and cfg.unwanted_extensions():
```
shypike: You may want to change the order of this test.
Testing the option is cheaper than checking the content.

sabnzbd/nzbstuff.py
```diff
@@ -893,8 +893,8 @@ def __init__(self, filename, pp, script, nzb=None,
         if not self.password and self.meta.get('password'):
             self.password = self.meta.get('password', [None])[0]
 
-        # Set nzo save-delay to 6 sec per GB with a max of 5 min
-        self.save_timeout = min(6.0 * float(self.bytes) / GIGI, 300.0)
+        # Set nzo save-delay to 30 seconds
+        self.save_timeout = 30
```
shypike: This will make it save more often for larger jobs. The larger the job, the more expensive the save is.
I'm OK with increasing the lower limit.

Safihre:
Implemented now that only the first **valid** rar will be checked if it's encrypted. 
I tested and if the first rar is damaged, RarFile will report that it's not encrypted and so we will check until the first usable rar file. 
Only if there's multiple sets of rar's inside the NZB and the first set happens to be un-ecrypted but the second one *is* encrypted, we would miss that. But this is rare, in my opinion.
Thoughts? :)

sabnzbd/nzbqueue.py
```diff
@@ -813,12 +813,13 @@ def register_article(self, article, found=True):
...
             if file_done:
                 if nzo.next_save is None or time.time() > nzo.next_save:
-                    sabnzbd.save_data(nzo, nzo.nzo_id, nzo.workpath)
                     BPSMeter.do.save()
                     if nzo.save_timeout is None:
                         nzo.next_save = None
                     else:
                         nzo.next_save = time.time() + nzo.save_timeout
+                    # After setting next_save, to avoid threading issues!
+                    sabnzbd.save_data(nzo, nzo.nzo_id, nzo.workpath)
```
shypike: How can this solve the problem?
Both calls still come from the same thread and cannot interfere with each other.
It may be because the timing has changed, but this doesn't solve the root cause.

sabnzbd/nzbstuff.py
shypike: Before, a 5G job would save every 30 seconds and a 20G job every 2 minutes.
Now it's 30 sec always. This will make 20G jobs slower due to all the saving.

sabnzbd/cfg.py
```diff
@@ -148,7 +148,7 @@ def validate_server(value):
 propagation_delay = OptionNumber('misc', 'propagation_delay', 0)
 folder_rename = OptionBool('misc', 'folder_rename', True)
 folder_max_length = OptionNumber('misc', 'folder_max_length', DEF_FOLDER_MAX, 20, 65000)
-pause_on_pwrar = OptionBool('misc', 'pause_on_pwrar', True)
+pause_on_pwrar = OptionNumber('misc', 'pause_on_pwrar', 1)
```
shypike: Why this change?

shypike:
@Safihre 
I agree with just checking the first file.
However, ultimately we may need to check the content of each file anyway. Due to the increasing obfuscation of names.

sabnzbd/cfg.py
Safihre: Because it's not a bool, it's a number. It can be 0=Nothing, 1=Pause or 2=Abort

sabnzbd/nzbqueue.py
Safihre: Hmmm, yes you are right. I was mixed things up. I read python does the file-wryting in a seperate process, but it still locks the current thread for that. I will revert this.

sabnzbd/nzbstuff.py
Safihre: Agreed, will just set a minimum of 30.

Safihre:
@shypike I checked all the places where ```nzo.``` was modified directly and identified 2 spots in articlecache and downloader where it happens a lot, I moved those to thread-safe functions: 334ac16

What do you think?

Safihre:
The only places I can't 'catch' are the TryLists, since those are updated all the time, on Article/NZF/NZO level. What can we do about that?

shypike:
@Safihre 
> The only places I can't 'catch' are the TryLists, since those are updated all the time, on Article/NZF/NZO level. What can we do about that?
Nothing. There's no need, because the try-lists are saved to disk but never used on restore. More importantly (for this aspect) is that TryList only uses lists and not dictionaries. Lists don't seem to suffer from the "save" problem.

shypike:
@Safihre Please job my memory about 58176f4b78b04df99c6f40c2fe71dfd2844eaa2e
Is this a generic solution to getting the job size from the RSS entry (based on enclosures)?

Safihre:
> Is this a generic solution to getting the job size from the RSS entry (based on enclosures)?

I was discussing with the guy that does some deving for nZEB/Newzab (or whatever that indexer software is called) and he gave me a test account for his testing-server. 
I noticed that no 'size' was detected when using his RSS feed because he reported it as ```Size: 4GB``` and the regular expression expects a value with a dot like ```Size: 4.3GB``` . But upon inspection of all my indexers I saw that *all* of the paid ones use the ```<enclosures>``` to supply the URL and the size of the download in bytes.
In case an indexer doesn't use it, it will now still fall-back to trying to get the size using regular expressions from the description. Verified that using all the free searchers like Binsearch.

shypike:
@Safihre 
One remark still about
> So, I reverted the making copies in BPSMeter.do.save(), because it's not the cause of the problem.

It may not be the cause, but it may go wrong occasionally too. Given that it uses dictionaries.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -709,10 +705,14 @@ function ViewModel() {
...
             } else {
-                // Hide notification
-                hideNotification('.main-notification-box-uploading')
                 // Refresh
                 self.refresh();
+                // Hide notification
+                hideNotification('.main-notification-box-uploading')
+                // Reset the form
+                $('#modal-add-nzb form').trigger('reset');
+                $('#nzbname').val('')
```
thezoggy: you have trailing `;` above but dont do it on other lines..  cosmetic but could lead to issues if not explicitly defined when making code changes later on.

Issue #646, opened 2016-08-08
Notes on os.path.join()


sanderjo:
<Work in progress>

Here are some notes on os.path.join() used by SABnzbd in 142 lines of code, which can cause new problems as of SAB 1.0.0, which introduced Unicode as the default encoding.

Goal of this post: document, possible solutions, constructive and concrete feedback.

**The problem with os.path.join()**
 
Examples of reported problems caused by os.path.join() in SAB 1.x:
- https://github.com/sabnzbd/sabnzbd/issues/644#issuecomment-238053518
- https://github.com/sabnzbd/sabnzbd/issues/570#issuecomment-219504191

**Examples of bad and good**

In a nutshell: a join of a unicode string with a string with non-ASCII chars in a non-unicode encoding gives a traceback. Example:

```
>>> print os.path.join( u"/blabla", u"müller".encode('cp1252') )
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xfc in position 2: ordinal not in range(128)
```

Please note the following joins go OK:

Join of unicode and cp1252 but no special characters in the cp1252:
```
>>> print os.path.join( u"/blabla", u"muller".encode('cp1252') )
/blabla/muller
```

Join of plain ASCII with a cp1252-with-non-ASCII-characters:
```
>>> print os.path.join( "/blabla", u"müller".encode('cp1252') )
/blabla/m�ller
>>> print os.path.join( u"/blabla".encode('ascii'), u"müller".encode('cp1252') )
/blabla/m�ller

```

**So ... what could we do**

- We could do nothing; we consider the above problem a niche aka edge case
- We could build a 'fence' around SABnzbd to only accept ASCII or unicode as input, and actively reject any other encodings of files and strings
- We could re-actively wait for problems, and then solve/workaround it
- We could pro-actively replace all os.path.join() with a safe function

**Solve/workarounds and Safe function**

1) In https://github.com/sabnzbd/sabnzbd/issues/644#issuecomment-238054629 I just remove characters >= 128. That looks safe in that situation.
2) in `path` variables, only use Unicode if necessary. If not necessary, use ASCII encoding. Then, if a strange `file` name appears, much less change that the os.path.join() gives a traceback
[...]

Issue #570, opened 2016-05-09
Post Processing Failed/UnicodeDecodeError


vinzz73:
I get the following error on certain spots:

:ERROR::[postproc:523] Post Processing Failed for ...
UnicodeDecodeError: 'utf8' codec can't decode byte 0xfc in position 10: invalid start byte

Sabnzbd version:
git status
On branch develop
Your branch is up-to-date with 'origin/develop'.

shypike:
Not enough info, I'm afraid.
Operating system?
Example NZB?

vinzz73:
Debian Wheezy
PHP 5.6.19
NZB example, see file..

[Miscellaneous Magazines 11.zip](https://github.com/sabnzbd/sabnzbd/files/255016/Miscellaneous.Magazines.11.zip)

shypike:
Please check this issue: https://github.com/sabnzbd/sabnzbd/issues/543

sanderjo:
As of SABnzbd 1.0.2, there is a special log line for the issue Shypike is reffering too:

```
$ cat ~/.sabnzbd/logs/sabnzbd.log | grep -i encoding

2016-05-07 13:17:01,709::INFO::[sabnzbdplus:1262] Preferred encoding = UTF-8
2016-05-08 12:10:51,045::INFO::[sabnzbdplus:1262] Preferred encoding = UTF-8
2016-05-08 12:27:11,153::INFO::[sabnzbdplus:1262] Preferred encoding = UTF-8
```

So ... what does your sabnzbd.log say about "Preferred encoding"? On Linux it should be something with "UTF-8", @vinzz73 ?

vinzz73:
Strange:

INFO::[SABnzbd:1345] Preferred encoding = UTF-8
locale:
LANG=en_US.UTF-8

Also the test download nzb with german and korean langauge in #543 I can download with no issues..

I don't get then..

sanderjo:
> INFO::[SABnzbd:1345] Preferred encoding = UTF-8
> Also the test nzb with german and korean langauge in #543 I can download with no issues..

Wow ... interesting. 

One difference I can see is that 'your' post contains a ZIP-file. Shall I create a test-post with a ZIP-file in it containing the German and Korean files to verify that hypothesis?

vinzz73:
I zipped that file myself only for Git to be able to upload it here.

sanderjo:
No, inside the rar set is yet another zip file (size 670 MB):

[...]

PR #654, opened 2016-08-12
Bugfixes/pathfix

shypike:
@sanderjo This should solve the path issues on Linux as described in #570 and #646


Safihre:
Simple but effective fix for the joining.
Was this supposed to include the commit from the other PR #502?

shypike:
@Safihre 
No, just found it out myself.
I will revert it.

shypike:
Not needed, we found the root cause.

PR #655, opened 2016-08-12
Another try at the Downloader

Safihre:
I tried (and failed) in most of my endeavours to make SABnzbd download faster, I kind of gave up on it.
But it annoyed me too much that when I am using a 100Mbit connection, SABnzbd will average at about 25% CPU usage (this is 1 core 100%) and download at 11.3MB/s. 
But if I set the speedlimit to 10.5MB/s, the CPU usage will only be 12% or so. That's just too much of a difference and it's all caused by the previously diagnosed hyperactive Downloader-loop on systems where the CPU is not the limiting factor.

Instead of just making the loop always slower (what I tried before and killed performance on high-speed connections), I opted to look at how many of the connections are 'ready'.
If less than 25% of open connections are ready, then we can easily wait a bit before going on.
This works very well! :smiley:  My CPU usage without speedlimit is now 15% on the 100Mbit and it does not effect the speed when I am on the 1Gbit connection. There is a slight increase in max-speed from about 37MB/s to 39MB/s, but this could also be other things.

**Also changed:**  
Bug in RC3 where if you pause a download that already has some files complete, it will show ENCRYPTED.
Moved the logging of 'Decoding article blabla' to only when ```sabnzbd.LOG_ALL``` since it adds no diagnostic information and just pollutes the logs.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -482,6 +482,10 @@ class Downloader(Thread):
 
             if readkeys or writekeys:
                 read, write, error = select.select(readkeys, writekeys, (), 1.0)
+                
+                # Why check so often when so few things happend?
+                if len(read) < len(readkeys)/4:
+                    time.sleep(0.05)
 
             else:
                 read, write, error = ([], [], [])
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -484,7 +484,7 @@ class Downloader(Thread):
                 read, write, error = select.select(readkeys, writekeys, (), 1.0)
                 
                 # Why check so often when so few things happend?
-                if len(read) < len(readkeys)/4:
+                if len(readkeys) >= 8 and len(read) < len(readkeys)/4:
                     time.sleep(0.05)
 
             else:
--- a/interfaces/Glitter/templates/static/javascripts/glitter.main.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.main.js
@@ -168,6 +168,8 @@ function ViewModel() {
         // Same sizes? Then it's all 1 disk!
         if(response.queue.diskspace1 != response.queue.diskspace2) {
             self.diskSpaceLeft2(response.queue.diskspace2_norm)
+        } else {
+            self.diskSpaceLeft2('')
         }
         
         // Did we exceed the space?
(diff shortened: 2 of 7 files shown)


sabnzbd/decoder.py
```diff
@@ -102,7 +102,8 @@ def run(self):
                     if nzo.precheck:
                         raise BadYenc
                     register = True
-                    logging.debug("Decoding %s", art_id)
```
shypike: Why this one?
I always use the decoding messages as a sign that things go well.
Otherwise nothing is logged.

sabnzbd/downloader.py
```diff
@@ -482,6 +482,10 @@ def run(self):
 
             if readkeys or writekeys:
                 read, write, error = select.select(readkeys, writekeys, (), 1.0)
+                
+                # Why check so often when so few things happend?
+                if len(read) < len(readkeys)/4:
+                    time.sleep(0.01)
```
shypike: I'm surprised that this helps.
I would have thought that more sleep was needed here.

Safihre: Could probably do more, but this is on safe side. Let me try with 0.05 to be sure.

sabnzbd/decoder.py
Safihre: Hmm right, that makes sense.

Safihre:
@shypike Tried and tested, seems 0.05 also works well!  Tested with some fast and slow servers and with low and high connection counts.
Rebased the changes.

Safihre:
Interestingly, with this change, having a higher number of connections becomes beneficial since it can "wait more".
With 8 connections I get 12 MB/s and ~13% CPU
With 35 connections also 12 MB/s and ~8% CPU

shypike:
Congratulations!
Still amazed that is is such a good solution.

Safihre:
Also seems to work well for SSL servers, just to make sure.

But only helps when CPU isn't the limit.. for CPU-limited it just stays the same :weary:
We need to do something about that at some point..

thezoggy:
@Safihre now what `Article Cache Limit` settings are you using in sab for these tests?

Safihre:
1G, and on my slower Intel NUC the SAB default of 450M.

PR #658, opened 2016-08-17
Fixes and Rating upgrade

Safihre:
Fixes for number of issues.

The real change is that to make the rating-functionality more attractive, it has been made more flexible, [discussed ](https://github.com/nZEDb/nZEDb/issues/2235) with the people from the indexer software nZEBb.
Required fields can now also be supplied using ```x-rating-*``` headers, instead of locked into the NZB meta-tags. The headers take precedence over the meta-tag information.
There is a new field ```x-rating-url``` that allows to specify a full URL where the rating should be send to for each NZB, allowing multiple indexers to work with SABnzbd.
If not supplied, the ```rating_host``` and default-URL from the Config will be used.
For now only updated the English text, but if this catches on and gets implemented in more indexers, we can later make it correct for all localizations.


sabnzbd/config.py
```diff
@@ -380,7 +380,7 @@ def __init__(self, name, values):
         self.enable = OptionBool(name, 'enable', True, add=False)
         self.optional = OptionBool(name, 'optional', False, add=False)
         self.retention = OptionNumber(name, 'retention', add=False)
-        self.ssl_type = OptionStr(name, 'ssl_type', 't1', add=False)
+        self.ssl_type = OptionStr(name, 'ssl_type', 'v23', add=False)
```
shypike: I think this is wrong (both original and new code).
It should be empty so that the highest level will automatically be used.
The same goes for the other two places.

Safihre: You're right, fixed!

sabnzbd/assembler.py
```diff
@@ -169,6 +169,11 @@ def _assemble(nzf, path, dupe):
     decodetable = nzf.decodetable
 
     for articlenum in decodetable:
+        # Break if deleted during writing
+        if nzf.nzo.deleted:
```
shypike: Good find, but is this enough?
Should you test for nzf.nzo.status in (Status....) ?

Safihre: It's the same that we use a few lines higher (after this function)
https://github.com/sabnzbd/sabnzbd/pull/658/files#diff-3ffdc5044910fc8e1d1913955f7ee55fR91

Safihre: Maybe we just need a nzo.is_gone() function for this check?

shypike: OK, that's true, should be good enough.
The is_gone() function would test for a series of Statuses, I assume?

Safihre:
I changed the solution in the assembler. In that case we only care about ```nzo.deleted```, because for ```FAILED``` or ```COMPLETED``` files can stay. 
But sometimes we were too late, so this new leftover file that was created still needs to be removed.
Seemed safest to me to just the ```nzo.purge_datat()``` command.

Safihre:
Actually nevermind, I was wrong. It removed too aggressively because ```nzo.deleted``` is also used when it succesfull went to history.
Choose a more safe method now that seems also to work and only when it's really _really_ deleted.

shypike:
@Safihre Your last commit fixes what you said here?
> Actually nevermind, I was wrong. It removed too aggressively because nzo.deleted is also used when it succesfull went to history.
> Choose a more safe method now that seems also to work and only when it's really really deleted.

Safihre:
@shypike Yes all good now :)

PR #660, opened 2016-08-19
Improved network stuff: unresponsive DNS and HappyEyeballs-with-memory

sanderjo:
Solve delays in case of non-responding DNS

Delays were in:
- startup of SABnzbd: 2 times 20 seconds
- Status screen (with the wrench symbol): no status-screen at all.

Thank you @Safihre for the decorator approach.


sanderjo:
"Happy Eyeballs: remember results to avoid redundant lookups" solves the "Very, very slow initial connexion to Usenet serve" bug as described on https://forums.sabnzbd.org/viewtopic.php?f=11&t=21198

PR #663, opened 2016-08-21
Prospective par2 was broken

Safihre:
Always wondered why my prospective par2 was **not** doing very well when there were multiple sets in a job. Turns out the for-loop above it changed the ```nzf``` reference to the last one in the list.
This was highlighted by moving it to nzbstuff.py after which PyLint added that ```nzf=None```.

Now it works the way it was intended, but this also means I have to re-test (tomorrow) with a lot of jobs to make sure there's still no stalling issues.

### DIFF
--- a/interfaces/Config/templates/config_folders.tmpl
+++ b/interfaces/Config/templates/config_folders.tmpl
@@ -59,9 +59,8 @@
                     <span class="desc">$T('explain-email_dir')</span>
                 </div>
                 <div class="field-pair">
-                    <!-- Here we add "search" to the ID as an exception to stop Password Managers (like LastPass)! --> 
-                    <label class="config" for="password_file-search">$T('opt-password_file')</label>
-                    <input type="text" name="password_file" id="password_file-search" value="$password_file" />
+                    <label class="config" for="password_file">$T('opt-password_file')</label>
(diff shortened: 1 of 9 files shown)


Safihre:
After experiencing how aggressive Lastpass and Chrome autofill has become, I now really made sure that things are not auto-filled anymore.
This is especially important since we now have a HTML-login form. If you choose to save that username/password in Lastpass or Chrome, it will be filled *everywhere* throughout your Config. Fixed that now.

JackDandy:
Hello @Safihre
I'm puzzled, or I may be missing something, but why prevent the use of password managers? Seem to so me that if everyone did this, then a great many of us would be doomed or forced to use one memorable password which kinda defeats one of the great uses of this type of tool.
tia

Safihre:
I should maybe have written it more clear. It's indeed only for the Config, the login of course is not changed. The problem was that since we now have a HTML form, Lastpass/Chrome/etc will ask to save the password after first login. If you choose to do so, it would then input it *everywhere* in the Config. I lost all my Server logins.

Also reported a couple of times on the forums, here and Reddit. People will blame SABnzbd for not saving password changes to servers, while it's actually the browser auto-filling.

Issue #661, opened 2016-08-20
'Checking' does not show progress in 1.1.0


Safihre:
When 'Check before downloading' is turned on, no progress of the checking is shown in 1.1.0.
In 1.0.3 you nicely see the progress by the blue-progress bar in Glitter, but in 1.1.0 the API doesn't report the progress anymore, all you see is the general 'Queue size left' counter go down untill all is checked, then it goes back up again when the actual download starts.

shypike:
Must be a side effect of all those optimizations, but which one?

Issue #664, opened 2016-08-22
Lots of missing articles, correct download-report?


Safihre:
Not sure if this intended behavior, or due to our optimizations in 1.1.0.
If a download is 1000MB (**including** par2) and 100MB is missing, it will currently show in the Queue that it downloads until 900MB and then goes to post-processing. So it really indicates how much it has actually _downloaded_, not how much it has _progressed_ through the amount of MB declared in the NZB.

For example this NZB, it says in the Queue 765MB/2542MB, after having finished 9 files of 103MB. So shouldn't it show 900MB / 2542MB? 
![image](https://cloud.githubusercontent.com/assets/5703454/17847800/49a4f7d8-6850-11e6-9610-8a2c3496137f.png)

shypike:
Same cause as pre-download-check no longer showing progress.
So far, I don't know why either happens.

Safihre:
I see the reason is simple: before we used to loop over ```nzo.files``` to determine the number of bytes left, so would only count the bytes of files that still needed to be done.
How to fix it.. I dunno..

thezoggy:
due to: https://github.com/sabnzbd/sabnzbd/commit/f17d3bd6b31f185b036157944ed36e0fd72089d5 ?

Safihre:
No, a change long ago to improve the speed of the API to not loop over *all* files in *every* download.
This will fix it even better than 1.0.x-behavior: 134ebc2 
Keeps proper check of bytes that were tried.

But, when there's an old queue loaded on startup the loaded ```nzo``` objects will not have the ```nzo.bytes_tried``` and give errors. 
Maybe @shypike can take and improve my idea or maybe has a better one ;)

shypike:
I was thinking of adding an NZO level byte counter and update that together with the update of the counter at NZF-level. Or go back to looping over the NZFs, or is that already too expensive? I mean, just to get the totals.

Safihre:
That's exactly what is in my commit 134ebc2, a counter of all bytes tried at NZO level.

shypike:
In `nzbstuff.NzbObject.__setstate__()` add these lines just before `TryList.__init__(self)`:
``` python
        if self.nzo.bytes_tried is None:
            self.nzo.bytes_tried = 0
```

Safihre:
:+1: 
And how to copy this attribute when a job is Retried?
Now it will show 0% completed when Retried because ```nzo.bytes_tried``` is empty.

[...]

PR #665, opened 2016-08-23
Real bytes counter

Safihre:
Fixing #664 and improving API speed even a bit more, this seems to me the way the bytes counter should have always been: a true account of the bytes tried and the bare minimum of loops required.

Timings also improve, ```queue_info()``` averaged over 1 minute with 120 jobs in the queue:

Branch | Time
------------ | -------------
This PR + PR #663   |  2.45ms
Old | 9.43ms
1.0.x | 24.33ms

**Important:** it does need fix d597fbc from my other PR #663, to make sure par2's are removed properly from all the lists that they are on.

### DIFF
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -890,8 +890,8 @@ class NzbQueue(TryList):
 
         for nzo in self.__nzo_list:
             if nzo.status != 'Paused':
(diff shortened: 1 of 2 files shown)


Safihre:
Just tested and this also fixes #661 :smiley:

Safihre:
Adding this to #663 and closing :)

Issue #628, opened 2016-07-17
NNTPS servers: the good and the bad


sanderjo:
FWIW & FYI:

- I wrote a script that checks the SSL/TLS connection of a newsserver on port 563, requiring a correct connection. It needs ssl.SSLContext(), which was introduced in Python 2.7.9. NB: I used `context.check_hostname = True` which is quite strict. To do: run without that requirement.
- I created (with brute-force) a list of potential newsservers based on https://www.reddit.com/r/usenet/wiki/providers

... and I combined that. Result below.

Goal: see the state of secure SSL/TLS NNTP connections, and what would happen if SABnzbd would require a correct SSL/TLS connection for the newsserver (that is: _if_ you have set SSL)

Script run from Python 2.7.12 #on Ubuntu 16.04

**The good (with hostname checking on)**

```
Check: news.4ux.nl 563 : Version: TLSv1.2 Welcome is 200 Welcome to 4 Usenet Xs 
Check: news.anarqy.com 563 : Version: TLSv1 Welcome is 200 Welcome NNTPSwitch-0.12, 112170 groups available, posting allowed, slot 7, connections 1 
Check: ssl.astraweb.com 563 : Version: TLSv1 Welcome is 200 news.astraweb.com NNRP Service Ready (posting ok) (yEnc enabled). 
Check: news.budgetnews.net 563 : Version: TLSv1.2 Welcome is 200 Welcome to budgetnews.net (fx46.am4) 
Check: news.cheapnews.eu 563 : Version: TLSv1.2 Welcome is 200 Welcome to Cheapnews 
Check: news.eweka.nl 563 : Version: TLSv1.2 Welcome is 200 Welcome to Eweka (fx21.am4) 
Check: news.extremeusenet.nl 563 : Version: TLSv1.2 Welcome is 200 Welcome to ExtremeUsenet 
Check: news.forteinc.com 563 : Version: TLSv1.2 Welcome is 200 Welcome to Forte Premium.  Please Authenticate. (fx36.am4) 
Check: news.giganews.com 563 : Version: TLSv1.2 Welcome is 200 News.GigaNews.Com 
Check: news.hitnews.com 563 : Version: TLSv1.2 Welcome is 200 Welcome to hitnews.com (fx10.fr7) 
Check: ssl.hitnews.com 563 : Version: TLSv1.2 Welcome is 200 Welcome to hitnews.com (fx11.fr7) 
[...]

PR #677, opened 2016-09-08
New SSL approach

Safihre:
## **Of course this is not for 1.1.0, but 1.2.0**

After discussion in #628, this is my attempt at removing our dependency on pyOpenSSL and using native ```ssl``` module.

- The user can no longer select what protocol to use, it will always use the best possible. The used protocol and cipher will now be logged after connection is completed.
- There is a new Switch, where you can select the level of verification: Off, Normal (only require a good certificate) and Strict (also force hostname check). 
- To test:
```news.surfino.com``` which only works when it's set to Off. 
```eunews.frugalusenet.com``` won't work in Strict-mode.
- Updated CherryPy to version 8.1.0 to use ```ssl``` instead of PyOpenSSL as SSL provider. They have patched almost everything we did, but I still had to make 1 custom patch, see: https://github.com/cherrypy/cherrypy/issues/1497
- 301 redirect HTTP->HTTPS on the HTTPS port doesn't work yet, but that's for later.

So, what do you think? 🐹

### CODE REVIEW

**Comment by Safihre** [comment]
On my home 100Mbit connection the ```newssl``` branch uses 9-12% CPU when downloading using NNTPS and Glitter via HTTPS.
The ```develop``` branch uses for the same thing 14-15% CPU and the ```1.0.x``` branch 18-20% CPU.
Not bad 😃 👍

File: sabnzbd/interface.py [hunk 1]
```diff
@@ -1264,6 +1265,14 @@ def index(self, **kwargs):
 
         conf['have_unzip'] = bool(sabnzbd.newsunpack.ZIP_COMMAND)
         conf['have_7zip'] = bool(sabnzbd.newsunpack.SEVEN_COMMAND)
+        conf['have_yenc'] = HAVE_YENC
```
thezoggy: wrap in bool like the others?

sabnzbd/newswrapper.py
```diff
@@ -139,6 +139,10 @@ def con(sock, host, port, sslenabled, write_fds, nntp):
                 except WantReadError:
                     select.select([sock], [], [], 1.0)
 
+            # Log SSL/TLS info
+            logging.info("%s@%s: Connected using %s (%s)",
```
thezoggy: how costly is this to do over and over? maybe only do it once per server/per nzb? as logging == disk io.. which hurts sab performance all around

thezoggy:
wonder if we could use this:
https://github.com/pinae/certificate-chain-creator

thezoggy:
also need to enforce python 2.7.9+ for this pull...

sabnzbd/interface.py
Safihre: ```HAVE_YENC``` is an actual bool, while the others are strings with the path to the command.

sabnzbd/newswrapper.py
Safihre: It's only done when the connection is opened, which is only once per NZB, so very little.
I also noticed that for example with my 50 connections to Frugal, not all used the exact same cipher.

Safihre:
> wonder if we could use this:
https://github.com/pinae/certificate-chain-creator

Unfortunately not, he uses ```os.system('openssl')``` calls and for example on Windows this won't work from the commandline, unless you specifically install OpenSSL. I tried that!

> also need to enforce python 2.7.9+ for this pull... right now we only enforce 2.7.x

I'm all for enforcing it! But because of the check in ```newswrapper``` SABnzbd will work fine on <2.7.9, just without SSL support.

thezoggy:
if someone installs sab.. if its one of the builds we do.. its going to have ssl. if someone installs via package manager.. prob going to have ssl. etc etc etc. the only scenario these days where it prob wont have ssl is someone installing from source (git clone) and just doesnt have it installed..  and those users prob will complain about stuff when it doesnt work because they dont understand.. so perhaps we should just force sab to require ssl... would simplify a lot of stuff and reduce a bit of checks..

Safihre:
Yes ```ssl``` will be available, but not the ```openssl``` command. But I have found a way to generate self-signed certificates using only ```cryptography```, working on that now.

Just added the check to make sure we have Python 2.7.9+.

thezoggy:
then we should just include `cryptography` with sab then (which does all the heavy lifting of pyopenssl), so that way we dont have to worry about what versions people have and if its installed..

Safihre:
Completed the PyOpenSSL-ectomy! 
```cryptography``` will now take care of the self-signed certificate generation.
@shypike maybe you can test if this eases the binary creating?

It's still a work-in-progress:
- CherryPy still needs a patch to do 301 redirect HTTP->HTTPS on the HTTPS port. That became a bit more complicated due to the way CherryPy changed where it throws an error in that case.
- It sometimes throws ```SSLError: ('_ssl.c:574: The handshake operation timed out',)```, have to see what causes that and how to correctly catch it. Seems to happen when a connection is attempted before CherryPy is ready, for example when a previous Glitter is still open in the browser while SABnzbd is (re)starting.

**Comment by shypike** [comment]
@Safihre I'll check the redirection.
Actually, it's only needed for Bonjour support (because you cannot supply https in a Bonjour record, duh!)
So, if it doesn't work sometimes in a browser, that's no big deal.

**Comment by sanderjo** [comment]
@Safihre 

Impressive work!

Checking: news.newsgroup.ninja & HappyEyeballs on & Strict ... does that work?

**Comment by sanderjo** [comment]
With `enable_nntps_verification = 1` it looks good:

```
2016-09-09 22:58:02,882::INFO::[downloader:436] 1@news.newsgroup.ninja: Initiating connection
2016-09-09 22:58:03,081::INFO::[happyeyeballs:138] Quickest IP address for news.newsgroup.ninja (port 563, ssl True, preferipv6 True) is 81.171.92.220
2016-09-09 22:58:03,082::DEBUG::[happyeyeballs:140] Happy Eyeballs lookup and port connect took 35 ms
2016-09-09 22:58:03,083::DEBUG::[downloader:114] For server news.newsgroup.ninja, using IP 81.171.92.220 as server
2016-09-09 22:58:03,209::INFO::[newswrapper:144] 1@news.newsgroup.ninja: Connected using TLSv1.2 (DHE-RSA-AES128-SHA)
2016-09-09 22:58:04,121::DEBUG::[downloader:564] Server login problem: 502, 502 Authentication Failed
2016-09-09 22:58:04,121::ERROR::[downloader:591] Failed login for server news.newsgroup.ninja
2016-09-09 22:58:04,122::INFO::[downloader:626] Server news.newsgroup.ninja ignored for 10 minutes
```
That's impressive.

With `enable_nntps_verification = 2` it is NOT going well:

```
2016-09-09 23:01:41,121::INFO::[downloader:436] 1@news.newsgroup.ninja: Initiating connection
2016-09-09 23:01:41,372::INFO::[happyeyeballs:138] Quickest IP address for news.newsgroup.ninja (port 563, ssl True, preferipv6 True) is 81.171.92.220
2016-09-09 23:01:41,372::DEBUG::[happyeyeballs:140] Happy Eyeballs lookup and port connect took 76 ms
2016-09-09 23:01:41,372::DEBUG::[downloader:114] For server news.newsgroup.ninja, using IP 81.171.92.220 as server
2016-09-09 23:01:41,502::ERROR::[newswrapper:282] Server 81.171.92.220 uses an untrusted HTTPS certificate
2016-09-09 23:01:41,502::INFO::[newswrapper:290] Failed to connect: Server 81.171.92.220 uses an untrusted HTTPS certificate 1@81.171.92.220:563
```
Two things:
- "untrusted HTTPS certificate" ... it's not HTTPS. I would say "untrusted SSL/TLS certificate"
- Happy Eyeballs might be causing problems as it gives back an IP address as `host`, but an IP address can never have a valid certificate.

**Comment by sanderjo** [comment]
Testing with news.cheapnews.eu, which has IPv4 and IPv6 (so HappyEyeballs will kick in):

```
host news.cheapnews.eu
news.cheapnews.eu has address 78.152.55.164
news.cheapnews.eu has IPv6 address 2a02:d28:10:1b:0:431:ff02:164
```

news.cheapnews.eu has a 100% good score for SSL/TLS:
```
"news.cheapnews.eu","IPv4-IPv6","TLSv1.2","200 Welcome to Cheapnews","OK","OK","OK","-","-"
```

That goes well for the current SAB 1.1.0 RC4: happyeyeballs finds the quickest IP address, and gives that back. Then SAB does a connect (SSL or not) on that IP address. So far so good. 

But with the NNTPS SSL check in our SAB code, with Strict 2, the hostname-check does a check against an IP address, which will always fail :-(

Solution: SAB should connect to the IP address, but still feed the original name to the connect. BRB

```
2016-09-10 09:04:20,535::INFO::[downloader:436] 1@news.cheapnews.eu: Initiating connection
2016-09-10 09:04:20,697::INFO::[happyeyeballs:138] Quickest IP address for news.cheapnews.eu (port 563, ssl True, preferipv6 True) is 78.152.55.164
2016-09-10 09:04:20,697::DEBUG::[happyeyeballs:140] Happy Eyeballs lookup and port connect took 8 ms
2016-09-10 09:04:20,698::DEBUG::[downloader:114] For server news.cheapnews.eu, using IP 78.152.55.164 as server
2016-09-10 09:04:20,747::ERROR::[newswrapper:282] Server 78.152.55.164 uses an untrusted HTTPS certificate
2016-09-10 09:04:20,748::INFO::[newswrapper:290] Failed to connect: Server 78.152.55.164 uses an untrusted HTTPS certificate 1@78.152.55.164:563
```

**Comment by Safihre** [comment]
@sanderjo, oops! I should able to fix that, I think the hostname is available through one of the fields at that point.

**Comment by sanderjo** [comment]
> There is a new Switch, where you can select the level of verification: Off, Normal (only require a good certificate) and Strict (also force hostname check).

And Normal will be the default?

Of the 121 NNTPS-servers in my overview, 9 will fail the Normal test, as they result in "SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)"

For example:
```
news.aeton.nl
 Issuer                       self-signed (NOT ok)
 Trust (hostname)             certificate does not match supplied URI
 Chain of trust               NOT ok (self signed)
 EV cert (experimental)       no 
 Certificate Expiration       expired! (2007-08-20 12:16 --> 2009-08-19 12:16 +0200)
```
Very bad setup.


The list of 9 that will fail Normal NNTPS check:

```
news.aeton.nl
news.flashnewsgroups.com
news.newsguy.com
news.nntpjunkie.com
news.speedplaza.net
news.surfino.com
news.united-newsserver.de
news.usenet-to-go.de
news.z51.biz
```

**Comment by Safihre** [comment]
@sanderjo I made an update, now it should use the original hostname.
I also fixed the generation of self-signed certificates, I made a typo before.
Is it working better now? I don't have IPv6 available to test.

File: sabnzbd/newswrapper.py [hunk 3]
```diff
@@ -219,7 +219,7 @@ def __init__(self, host, port, info, sslenabled, send_group, nw, user=None, pass
             if(sabnzbd.cfg.enable_nntps_verification() == 0):
                 ctx.verify_mode = ssl.CERT_NONE
 
-            self.sock = ctx.wrap_socket(socket.socket(af, socktype, proto), server_hostname=host)
+            self.sock = ctx.wrap_socket(socket.socket(af, socktype, proto), server_hostname=nw.server.host)
         elif sslenabled and not HAVE_SSL:
             logging.error(T('Error importing OpenSSL module. Connecting with NON-SSL'))
```
thezoggy: bad grammar.. I would just say without SSL.. not non-ssl. as someone may think that is some type of ssl..

Safihre: I know, but the text (and thus it's translations) have been there for a long time. 
Although it might be worth a change..

Safihre:
Oops, I commented in the wrong PR. Wanted to ask @jcfp about Python in Ubuntu about this commit, not the other one.

Browsing around there seems nothing we can do about Python on Ubuntu 14.04, can't just forcefully upgrade the system wide Python to 2.7.9+ while other tools might expect 2.7.6.

I don't want to mess up the code just to build pyOpenSSL back in. There's a measurable advantage in speed and CPU usage when not using pyOpenSSL, and according to @shypike building the binaries is also easier.

shypike:
@Safihre Do I understand correctly that Python before 2.7.9 cannot serve us without pyopenssl?

thezoggy:
ssl was native to python in 2.6.
2.7.9 is when they added the ability to validate server identity , 2.7.9's changelog:

https://www.python.org/downloads/release/python-279/
```
-    The entirety of Python 3.4's ssl module has been backported for Python 2.7.9. See PEP 466 for justification.
-    HTTPS certificate validation using the system's certificate store is now enabled by default. See PEP 476 for details.
-    SSLv3 has been disabled by default in httplib and its reverse dependencies due to the POODLE attack.
-    The ensurepip module module has been backported, which provides the pip package manager in every Python 2.7 installation. See PEP 477.
```

so 2.7.9 is when they added Server Name Indication (SNI) support.. so thats why were wanting 2.7.9 or newer. we previously relied on pyOpenSSL to add the support.

Safihre:
> we previously relied on pyOpenSSL to add the support.

No we didn't, we just never did any verification of HTTPS or NNTPS.
Have now fixed it so when on Python < 2.7.9 it will fall back to the 'old' situation in which you can connect via SSL, it just won't verify the certificates.

thezoggy:
we used it to verify certs

Safihre:
No we didn't, certificate verification was always turned off for NNTPS.

thezoggy:
yes for nntp, not https

Safihre:
In the current code, we use ```ssl.create_default_context``` for **HTTPS also** which isn't available until Python 2.7.9.

```python
def set_https_verification(value):
    prev = False
    try:
        import ssl
        if hasattr(ssl, '_create_default_https_context'):
            prev = ssl._create_default_https_context == ssl.create_default_context
            if value:
                ssl._create_default_https_context = ssl.create_default_context
            else:
                ssl._create_default_https_context = ssl._create_unverified_context
    except ImportError:
        pass
    return prev
```

shypike:
@Safihre @thezoggy 
So, any Python below 2.7.9 will just not verify any certificate.
I can live with that. We can always write a recommendation to upgrade Python or else just recommend the fine Docker solution.
The `set_https_verification` part is already supporting pre and post 2.7.9.

Safihre:
After some tweaking, this approach now also works with pre-2.7.9. 
It will show a warning in the Config in that case and the checkbox/dropdown to set HTTPS/NNTPS verification are disabled on the Switches page.
image

Some other things like ```SSLWantReadError``` are also not available before 2.7.9, so had to create work-arounds. But it's still quite an elegant solution.

sanderjo:
Works great on my Ubuntu 14.04 LTS with Python 2.7.6
- correct screen (like you showed)
- I can connect via SSL / NNTPS.

Good!

screenshot from 2016-09-16 14 25 42

sanderjo:
Loggings looks:

```
2016-09-16 14:24:05,465::INFO::[newswrapper:158] 2@news.caiway.nl: Connected using TLSv1/SSLv3 (AES256-SHA)
2016-09-16 14:24:05,468::INFO::[newswrapper:158] 1@news.caiway.nl: Connected using TLSv1/SSLv3 (AES256-SHA)
2016-09-16 14:24:05,469::INFO::[newswrapper:158] 3@news.caiway.nl: Connected using TLSv1/SSLv3 (AES256-SHA)
2016-09-16 14:24:05,472::INFO::[newswrapper:158] 4@news.caiway.nl: Connected using TLSv1/SSLv3 (AES256-SHA)
2016-09-16 14:24:05,828::INFO::[newswrapper:158] 1@newsreader3.eweka.nl: Connected using TLSv1/SSLv3 (DHE-RSA-AES128-SHA)
2016-09-16 14:24:05,853::INFO::[newswrapper:158] 4@newsreader3.eweka.nl: Connected using TLSv1/SSLv3 (DHE-RSA-AES128-SHA)
2016-09-16 14:24:05,866::INFO::[newswrapper:158] 2@newsreader3.eweka.nl: Connected using TLSv1/SSLv3 (DHE-RSA-AES128-SHA)
2016-09-16 14:24:05,868::INFO::[newswrapper:158] 3@newsreader3.eweka.nl: Connected using TLSv1/SSLv3 (DHE-RSA-AES128-SHA)
```
TLSv1 is right: my Ubuntu / Python does not support higher.
However: "SSLv3"? Can't the logging be more specific and leave that out?

A minor thing: clicking on Test Server gives this stdout logging:

```
2016-09-16 14:29:49,170::INFO::[newswrapper:275] -1@-1: Connected using TLSv1/SSLv3 (AES256-SHA)

2016-09-16 14:29:58,935::INFO::[newswrapper:275] -1@-1: Connected using TLSv1/SSLv3 (DHE-RSA-AES128-SHA)
```
A bit strange. I would expect the newsserver's name or FQDN in there.

Safihre:
@sanderjo, you are right. In Python pre-2.7.9 I cannot get the actual protocol version that was used. I can only get the cipher and which protocol enforced that cipher-suite.
https://docs.python.org/2/library/ssl.html#ssl.SSLSocket.cipher
And the Test Server is just the way it's constructed, it doesn't pass the server name etc in the same way as during regular connection. Since it's only Debug logging, I didn't make special code to make it look prettier..

Safihre:
Certificate validation is now a per-server setting, so 1 bad server doesn't lower the bar for the rest!

jcfp:
@Safihre as I understand it your later commits have removed any potential issues with older python releases, i.e. sab will run with any python 2.7 but just won't do validation < 2.7.9?

Safihre:
@jcfp Exactly :)
And they will get a warning in the first page of config (see post from @sanderjo a bit up). Will probably clarify the text of that a bit.
Was thinking:
"Secure (SSL) connections from SABnzbd to newsservers and HTTPS websites will be encrypted, however, validating a server's identity using its certificates is not possible (requires Python 2.7.9 or above)." (followed by a "?" button that links to the Wiki where we can put more info.

thezoggy:
fyi: https://www.openssl.org/news/secadv/20160922.txt

Safihre:
Glitter status popup will now show SSL protocol and cipher (after a connection is establish, otherwise it's just the checkmark that SSL is enabled):
image
Thought you might like this feature @sanderjo :wink:

Safihre:
Make a new special ```ssl_ciphers```, where a user can set a OpenSSL cipher format string for SABnzbd to use. This way you can force for example RC4-MD5, for a few extra % of performance.
Not a per-server setting, since normal users have no need to change this setting.
If left empty (the default) it will use the best available, as usual.
If you put an invalid cipher or one that's not supported by the client/server it will thrown an error when starting the connection: ```No cipher can be selected```.

sanderjo:
I'm very impressed by all the SSL/TLS work you've done on SABnzbd, @Safihre . Thank you. 

And indeed I like that you can see the resulting SSL/TLS encryption in Status popup.

screenshot from 2016-10-03 16-48-44

PR #683, opened 2016-09-12
Generate new HTTPS-certificates

Safihre:
@sanderjo Could you test it to verify its workings? Thanks!
It does add 44ms to start-up time on my laptop, in case SSL is enabled. Seems manageable.


sanderjo:
TL;DR: works!

Long:

I did
```
rm ~/.sabnzbd/admin/server.*
```
Then restarted SABnzbd. Afterwards the files are there:
```
sander@flappie:~/git/sab-safihre-12sept2016$ ll ~/.sabnzbd/admin/server.*
-rw-rw-r-- 1 sander sander  985 sep 12 23:07 /home/sander/.sabnzbd/admin/server.cert
-rw-rw-r-- 1 sander sander 1708 sep 12 23:07 /home/sander/.sabnzbd/admin/server.key
```
And openssl says it's good:
```
sander@flappie:~/git/sab-safihre-12sept2016$ openssl x509 -noout -text -in /home/sander/.sabnzbd/admin/server.cert
Certificate:
    Data:
        Version: 1 (0x0)
        Serial Number: 1473714473 (0x57d71929)
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=Certificate Authority
        Validity
            Not Before: Sep 12 21:07:53 2016 GMT
            Not After : Sep 10 21:07:53 2026 GMT
        Subject: CN=SABnzbd
```

Safihre:
@sanderjo Discussed with @shypike that forcing update might not be best idea, since it will prompt the browser to show the untrusted certificate warning again for the domain, which might come unexpected.

Therefore opted to add a button to the Config, which will let people re-generate their certificates by simply deleting them and then restarting, forcing the generation of new certificates when HTTPS is enabled.

Modified this PR, removed the previous commit and added the new one to add the button.

Safihre:
Yeah that implementation wasn't the best solution after all.
I just updated it so that now it will immediately generate new keys (only if the user hasn't selected different certificates) and then restart SABnzbd.

Safihre:
@jcfp With this change, users would be required to have Python 2.7.9+ to have SSL downloading support, the rest (including HTTPS access to the interface) would keep working.
@sandero already found that Ubuntu 14.0.4 LTS has Python 2.7.6. Is there a way for users to easily upgrade?
What do you think?

thezoggy:
@Safihre yes its easy for ubuntu users to upgrade using ppa as noted in https://github.com/sabnzbd/sabnzbd/issues/628



were already forcing people to use python 2.7.x, forcing them to use 2.7.9 prob isnt a big deal.. for the binaries/builds we do.. no issue there. just going to affect those that do nas/manual install.. ubuntu 14.04 LTS for example still has 2.7.6 :(

http://packages.ubuntu.com/search?keywords=python2.7
thankfully there is a ppa to deal with it,
https://launchpad.net/~fkrull/+archive/ubuntu/deadsnakes-python2.7

looking at synocommunity (synology) were good there.
https://synocommunity.com/packages

qnap, were good there too:
https://apps.qnap.community/

drobo is good:
http://www.droboports.com/app-repository/python2/

PR #687, opened 2016-09-19
Feature/sxx eyy as match filter

shypike:
- Add  "From Show SxxEyy" accept filter
- Use "Apply filters" button instead of automatic re-evaluation after each filter change
- Small optimization of season/episode analysis

@Safihre How do you add a "busy" indication?

### DIFF
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -98,11 +98,6 @@ class URLGrabber(Thread):
                         logging.debug('Dropping URL %s, job entry missing', url)
                         continue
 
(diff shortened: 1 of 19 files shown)


Safihre:
The button always shows the first time you look at a feed, is that supposed to happen? Or is it just because my feeds are disabled so they never ran after I restarted SAB?
I expected the button to only show after you add/modify a filter?

sabnzbd/urlgrabber.py
```diff
@@ -25,6 +25,7 @@
 import re
 import logging
 import Queue
+import urllib
```
Safihre: You're right!

PR #701, opened 2016-10-06
Develop bugfixes

Safihre:
Fixes for 1.2.0:
- Cipher gave a problem on Python <2.7.8

Fixes for 1.1.1:
- User reported an error when the Retry button was clicked twice (so when a job is already gone)
- CSS fix for Config
- [After enough problem with Max line speed](https://forums.sabnzbd.org/viewtopic.php?f=2&t=21551), I made it more clear now using a select at the end of the input. The default (so when nothing is set yet for example when upgrading from 0.7, is ```MB/s```)

image

### DIFF
--- a/interfaces/Config/templates/config_general.tmpl
+++ b/interfaces/Config/templates/config_general.tmpl
@@ -170,10 +170,15 @@
         </div><!-- /col2 -->
         <div class="col1">
             <fieldset>
-                <div class="field-pair">
-                    <label class="config" for="bandwidth_max">$T('opt-bandwidth_max')</label>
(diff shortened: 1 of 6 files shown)


Safihre:
```IndexError``` is reported, because of the ```[0]``` that is in there when a retry get's fired for a job that's already gone.
https://i.reddituploads.com/a00776cbc7b9450fb738adbf1670155e?fit=max&h=1536&w=1536&s=2bee5e5ca36c11f38699a31d62481738

Safihre:
@shypike Update, now only catching the extra ```IndexErrror```.

shypike:
Branch can go too?

PR #704, opened 2016-10-07
Feature/saf fixes

Safihre:
More fixes..


sabnzbd/newsunpack.py
```diff
@@ -1240,7 +1240,7 @@ def PAR_Verify(parfile, parfile_nzf, nzo, setname, joinables, classic=False, sin
         # Allow options if not classic or when classic and non-classic are the same
         if options and (not classic or (PAR2_COMMAND == PAR2C_COMMAND)):
             command.insert(2, options)
-    logging.debug('Par2-classic = %s', classic)
+    logging.debug('Par2-classic/cmdline = %s', classic)
```
shypike: Not just "par2cmdline"?

sabnzbd/newsunpack.py
```diff
@@ -1572,7 +1572,7 @@ def PAR_Verify(parfile, parfile_nzf, nzo, setname, joinables, classic=False, sin
         used_joinables.extend(reconstructed)
 
     if retry_classic:
-        logging.debug('Retry PAR2-joining with par2-classic')
+        logging.debug('Retry PAR2-joining with par2-classic/cmdline')
```
shypike: Same here

win/par2/README_par2cmdline.txt
```diff
@@ -0,0 +1,336 @@
+par2cmdline is a PAR 2.0 compatible file verification and repair tool.
```
shypike: This file should be called "Licence_par2cmdline.txt" and the other one (for multi-core) "Licence_par2.txt". "COPYING" should be renamed to "GPL2.txt".
The x64 folder has the same issue.

README.md
```diff
@@ -24,14 +24,17 @@ SABnzbd has a good deal of dependencies you'll need before you can get running.
 
 - `python` (only 2.7.x and higher, but not 3.x.x)
 - `python-cheetah`
-- `python-dbus`
```
shypike: Are these no longer a requirement??

INSTALL.txt
```diff
@@ -68,19 +68,21 @@ Windows
 Essential modules
     cheetah-2.0.1+          use "pip install cheetah"
     par2cmdline >= 0.4      http://parchive.sourceforge.net/
+                            Beware effect of -N parameter introduced in v0.6.13
```
shypike: "Beware". How is the user to interpret this? Do we have a link to more info?

sabnzbd/newsunpack.py
Safihre: No because on Mac it is still par2-classic..

README.md
Safihre: python-dbus, cryptography, unzip, 7zip and yenc are not required, SAB will work without them.
Just missing some features.
Now it's the same as in INSTALL.txt:
https://github.com/sabnzbd/sabnzbd/blob/master/INSTALL.txt

INSTALL.txt
Safihre: In the current release (v0.6.13) -N will include missing blocks, but in upcoming release the par2cmdline guy has inverted the behavior and -N will exclude missing blocks. Very confusing..

README.md
shypike: OK, makes sense

sabnzbd/newsunpack.py
thezoggy: would think it should be `par2-classic/par2cmdline`

PR #711, opened 2016-10-14
Bugfixes and new features

Safihre:
## Fixes
- ```retry_all``` API call was broken for a veryyy long time
- Some CherryPy upgrade broke the correct HTTPS binding of the ```attach_server()``` function, causing extra HTTP bindings on ports that should actually be HTTPS bindings.
-  ```ipv6_hosting``` disabled by default, was ignored and IPv6 was bound anyway when user has set ```localhost``` as host
- ```build_filelists()``` refactored so files are put in the correct lists, before it would put 7zip split files also in the rar-list

## New features
- Auto-detect if par2cmdline has the faulty ```-N``` parameter and if so, we add it
- Verify certificate validation by connecting to ```self-test.sabnzbd.org```, certificate is backed by CloudFlare. @jcfp I specifically do a socket-connect and don't receive any data, I hope this way you can leave the check in? Since all we test is the handshake, no other transmission.
- If validation fails, we show message in the first Config page.
- Now enabled HTTPS certificate validation by default, if there's something wrong with the user's setup we will disable it on startup.
- Check for ```retry-after``` header when getting an NZB, even if the server sends a HTTP-error code.
- Post-processing (the par2 and unrar stages) can now be aborted, which is shown in glitter using a stop-icon instead of delete icon.
- Jobs waiting to be post-processed can now also be deleted, for some reason this was always possible but blocked in Glitter.
- Update UnRar and 7zip for Windows/Mac.

## Other
- Removed Dockerfile, there are better alternatives available that are more regularly updated. Now also mentioned in our wiki: https://sabnzbd.org/wiki/installation/install-unix, so user can make his/her own choice.
- Cleanup of some excess whitespace/tabs.


jcfp:
@Safihre is the cert validation actually ever a problem on debian or ubuntu? afaik python on these platforms always integrates with the system ca store, and verification always works (assuming the python version can handle it). My impression is that these issues occur only with nas-based installs and their custom firmware/addons setup.

Safihre:
@jcfp If you're sure of that, this check is easy to cut out I guess. Unfortunately the interplay of Python version, OpenSSL version and CA certificates is so delicate that I gave up on trying to assume it works fine generally :P

osx/unrar/license.txt
```diff
@@ -1,34 +1,127 @@
- ******    *****   ******   UnRAR - free utility for RAR archives
```
shypike: I think we should keep the original license file, the new one feels inappropriate.

interfaces/Config/templates/config_switches.tmpl
```diff
@@ -52,6 +52,11 @@
                         </select>
                         <span class="desc">$T('explain-load_balancing')</span>
                     </div>
+                    <div class="field-pair <!--#if int($have_ssl) == 0 then "disabled" else ""#-->">
```
shypike: When the user fills in an invalid cipher, SSL connections will no longer work. Is it possible to ignore invalid cipher names?

interfaces/Glitter/templates/include_history.tmpl
```diff
@@ -184,7 +184,9 @@
                         </div>
                         <!-- /ko -->
                     </div>
-                    <a href="#" class="hover-button" data-bind="click: deleteSlot, css: { disabled : processingDownload() || processingWaiting() }"><span class="glyphicon glyphicon-trash"></span></a>
+                    <a href="#" data-bind="click: deleteSlot">
```
shypike: Is it possible to terminate the running job if you abort (shutdown or ^C) SABnzbd?

PR #717, opened 2016-10-17
Update rarfile to support RAR5

Safihre:
Update ```rarfile```, the latest version now supports RAR5 and other improvements.


Safihre:
**Contains also the commits from #711 due to merge-conflicts otherwise.**

thezoggy:
this pull should only have the related feature ones..

Safihre:
Unfortunately the RarFile changes are  dependent on ```Avoid duplicates in build_filelists() that break 7zip support```, otherwise it will create merge conflicts. Don't see any other way..
As soon as the other one is merged into develop, all those commits will dissappear from this pull.

PR #731, opened 2016-11-11
Update to 1.1.1 and set version

Safihre:
Direct update of files to avoid problems with merge-conflicts .
Also fix ```version.py``` to ```1.1.1```, for people running from ```master``` (like all QNAP's).


shypike:
You squashed this?
Maybe that is a better way for this particular usage.

PR #746, opened 2016-12-09
SABYenc

Safihre:
Work in progress on implementing the newly created ```sabyenc``` [module](https://github.com/sabnzbd/sabnzbd-yenc) that I created over the past months.

### Why create a new module? 

The `yenc` module is a great improvement over doing yEnc decoding in Python. However, it only accepts pure yEnc and not in the format it actually is downloaded, which includes newline characters and the header information like the filename and CRC. Python code is still needed to remove unnecessary characters extract all this info. By doing offline tests I found this to be about ~45% of actual calculation time. By doing all the steps in a C-module we can achieve great increases in performance, especially on low power devices or very fast connections.
Here are the results when running the decoding 250 times on data cut in 16K chunks, this is the maximum size of data when SSL is enabled):
```
---
     yEnc C New  took   666 ms
    Base Python  took     0 ms = 0%
---
     yEnc C Old  took  1460 ms   Diff  794 ms >>>>> (54% slower) <<<<<<
    Base Python  took   608 ms   Diff  607 ms
---
    yEnc Python  took  2469 ms
---
```
Actual increases in maximum speed are a bit lower, but I've seen around 35% increase in speed.

For reference: On my system SAB will then do on a single CPU-core 55MB/s, while NZBget will go to 80MB/s but using 2 cores fully...  We're pushing everything out of Python's single core! :bowtie:

### Installing

Installing is simple: ```pip install sabyenc```. 
For now you'll need a compiler installed, but later we can add pre-compiled versions for Windows/macOS.

### The module
https://github.com/sabnzbd/sabnzbd-yenc

The speed comparison above can be found [here](https://github.com/sabnzbd/sabnzbd-yenc/blob/master/tests/speedtest.py).

### What still needs to be done
Testing, testing, testing to make sure that it is stable and decodes everything correctly.
And then cleanup the code of both the module and the integration of SAB.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -428,6 +428,11 @@ def GetProfileInfo(vista_plus):
 
 def print_modules():
     """ Log all detected optional or external modules """
+    if sabnzbd.decoder.HAVE_SABYENC:
+        logging.info("SABYenc module... found!")
+    else:
+        logging.info("SABYenc module... NOT found!")
+
     if sabnzbd.decoder.HAVE_YENC:
         logging.info("_yenc module... found!")
     else:
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -431,7 +431,7 @@ def print_modules():
     if sabnzbd.decoder.HAVE_SABYENC:
         logging.info("SABYenc module... found!")
     else:
-        logging.error("SABYenc module... NOT found! Expecting v%s", sabnzbd.constants.SABYENC_VERSION)
+        logging.error("SABYenc module... NOT found! Expecting v%s - https://sabnzbd.org/sabyenc", sabnzbd.constants.SABYENC_VERSION)
 
     if sabnzbd.decoder.HAVE_YENC:
         logging.info("_yenc module... found!")
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -431,7 +431,7 @@ def print_modules():
     if sabnzbd.decoder.HAVE_SABYENC:
         logging.info("SABYenc module... found!")
     else:
-        logging.info("SABYenc module... NOT found!")
+        logging.error("SABYenc module... NOT found! Expecting v%s", sabnzbd.constants.SABYENC_VERSION)
 
     if sabnzbd.decoder.HAVE_YENC:
         logging.info("_yenc module... found!")
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -429,7 +429,7 @@ def GetProfileInfo(vista_plus):
 def print_modules():
     """ Log all detected optional or external modules """
     if sabnzbd.decoder.HAVE_SABYENC:
-        logging.info("SABYenc module... found!")
+        logging.info("SABYenc module (v%s)... found!", sabnzbd.constants.SABYENC_VERSION)
     else:
         logging.error("SABYenc module... NOT found! Expecting v%s - https://sabnzbd.org/sabyenc", sabnzbd.constants.SABYENC_VERSION)
 
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -432,11 +432,11 @@ def print_modules():
         logging.info("SABYenc module (v%s)... found!", sabnzbd.constants.SABYENC_VERSION)
     else:
         logging.error("SABYenc module... NOT found! Expecting v%s - https://sabnzbd.org/sabyenc", sabnzbd.constants.SABYENC_VERSION)
-
-    if sabnzbd.decoder.HAVE_YENC:
-        logging.info("_yenc module... found!")
-    else:
-        logging.warning(T('_yenc module... NOT found!'))
+        # Only now we care about old-yEnc
+        if sabnzbd.decoder.HAVE_YENC:
+            logging.info("_yenc module... found!")
+        else:
+            logging.warning(T('_yenc module... NOT found!'))
 
     if sabnzbd.HAVE_CRYPTOGRAPHY:
         logging.info('Cryptography module (v%s)... found!', sabnzbd.HAVE_CRYPTOGRAPHY)
--- a/interfaces/Config/templates/config.tmpl
+++ b/interfaces/Config/templates/config.tmpl
@@ -68,6 +68,15 @@
                     </td>
                 </tr>
                 <!--#end if#-->
+                <!--#if not $have_yenc#-->
+                <tr>
+                    <th scope="row">SABYenc:</th>
+                    <td>
+                        <span class="label label-danger">$T('notAvailable')</span>
+                        <a href="$helpuri$help_uri#no_yenc" target="_blank"><span class="glyphicon glyphicon-question-sign"></span></a>
+                    </td>
+                </tr>
+                <!--#end if#-->
                 <!--#if not $have_unzip #-->
                 <tr>
                     <th scope="row">$T('opt-enable_unzip'):</th>
--- a/interfaces/Config/templates/config.tmpl
+++ b/interfaces/Config/templates/config.tmpl
@@ -59,7 +59,7 @@
                     </td>
                 </tr>
                 <!--#end if#-->
-                <!--#if not $have_yenc#-->
+                <!--#if not $have_yenc and not $have_sabyenc#-->
                 <tr>
                     <th scope="row">yEnc:</th>
                     <td>
@@ -68,12 +68,12 @@
                     </td>
                 </tr>
                 <!--#end if#-->
-                <!--#if not $have_yenc#-->
+                <!--#if not $have_sabyenc#-->
                 <tr>
                     <th scope="row">SABYenc:</th>
                     <td>
                         <span class="label label-danger">$T('notAvailable')</span>
-                        <a href="$helpuri$help_uri#no_yenc" target="_blank"><span class="glyphicon glyphicon-question-sign"></span></a>
+                        <a href="$helpuri$help_uri#no_sabyenc" target="_blank"><span class="glyphicon glyphicon-question-sign"></span></a>
                     </td>
                 </tr>
                 <!--#end if#-->
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -172,7 +172,7 @@ def _assemble(nzf, path, dupe):
             break
 
         # Sleep to allow decoder/assembler switching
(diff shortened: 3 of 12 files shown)


Safihre:
Anyone interested in testing this? I think I've almost finished it.
Make sure to update/install `sabyenc` to latest version by running `pip install sabyenc --upgrade` (SAB shouldn't be running).
Just check out the `feature/sabpower` branch and you're good to go!

Getting steady 25-35% increases in speed.

Hellowlol:
Ill test it if i can install it :)

Safihre:
There's guides how to run from git  on our website! Even regular testing is helpful, just need to know it doesn't crash for some exotic reason!

OverFlow636:
running this now, i think my connection was maxed before and it still is with this, so thats good!

Safihre:
Any testing is helpful, just to verify that everything works as usual :)

Safihre:
I rebase this PR on `develop`, so to update you will have to do:
```
git fetch
git reset --hard origin/feature/sabpower
```
(or replace `origin` with how you called the sabnzbd repo remote)

thezoggy:
HEAD is an alias for whatever the repo is used.. 

thus, you can just do:
`git reset --hard HEAD`

Safihre:
@thezoggy it seems that's only the case if you made local changes. When wanting to reset to the repo, you need to specifically reference it.

thezoggy:
aww yeah forgot your force pushing which is the need of the discard and re-pull down again

shypike:
This should wait until after 1.2.1 or do you want to create a 1.2.x branch?

Safihre:
@shypike Until after 1.2.1. Then we can create that 1.2.x branch.

Safihre:
Created a `1.2.x` branch, so this can be merged into `develop` for testing.

thezoggy:
looks like you merged this into develop...

Safihre:
Yes, as intended, for testing :)

thezoggy:
so is sabyenc optional or required on develop?

Safihre:
For now it will shown an ERROR, in the final release it will be only a warning in the Config.
You can also read that in the URL provided in the current ERROR message:  https://sabnzbd.org/sabyenc

We have only been listing things that are missing on the first Config page, in an effort to simplify the Config for less experienced users.
In the final I also intend to not offer the `enable_sabyenc` or `nr_decoders` settings, they will be fixed. Seems to be no reason to disable it, besides testing.

thezoggy:
hmm so far sabyenc is resulting in the same speed if not slightly slower than stock yenc.. normally sab maxes out at 18MB/s, so far sabyenc is only hitting 17~. newsbin gets ~20  :(

Safihre:
That's very strange. I go from 33 to 67MB/s on my i5 and from 25MB/s to 49MB/s on a Pentium.

thezoggy:
moved test instance from typical hd to ssd so it matches my main sab install, (-f d:\test -> -f c:\test) and still seeing the same results.. also to note every time I asked sab to go from ui page to the config root.. the speed takes a little hit per the graph... never saw cpu go over 11% in python during download.. I run newsbin on the 'traditional' hd because thats where the space is.. and it gets the full 20MB/s.

**System performance (Pystone)** 234774 Intel(R) Core(TM) i7-4790K…
**Download folder speed** 693.1 MB/s (c:\test\downloads\incomp…)
**Complete folder speed** 683.6 MB/s (c:\test\downloads\comple…)

Safihre:
11% is a weird number and should be much more.
Is all this correct?
- Started with `python -OO SABnzbd.py`
- `enabled_sabyenc` turned on
- `nr_decoders` set to 2
- Article Cache set to at least `450M`?

Different NZB? NZB's with small article sizes are annoying.

Safihre:
Actually 17MB/s is really really slow for a i7-4790K that can go up to 4.4Ghz boost, if my poor laptop i5 with max of 2,2Ghz can get the speeds mentioned above even without `sabyenc`.

thezoggy:
I took my sab main install instance, cloned it.. told the sab instance I run from source I tell it to use this admin folder (`python SABnzbd.py -f c:\test\sabnzbd.ini --new`) run it once so it does the upgrade.. then go in and disable rss feeds/scheduler entries/etc so its leaner and wont startup or have anything running that may skew my tests). Then I shutdown.. let it save then start it back up so I have a 'clean' test bed. 

Not sure why you think 11% cpu used is too low.. the normal sab (win binary 1.2.1RC1) instance for example is yielding 10% cpu used during downloads.. I've never seen sab use much cpu (never have been cpu limited because of the hardware I use).

I pay for 150/150 Mbps connection but I realistically get 175/175 when doing speed tests... so roughly 21MB/s max.. but with overhead and the fact its mainly for burst.. 20MB/s stable is realistically the max... I see for any length of time.

Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] [cp1252]

enabled_sabyenc is set to on, as its the default.
nr_decoders is set to 2, as its the default.
article cache is set to 512M

About the article cache as I've mentioned on other github issues..  for me it just doesnt end up being very much help because of the type of nzb's I send to sab..  for example on today's tests I'm only seeing about 25MB of cache used (because these are small 1-2gb downloads and small files.. so small articles and not a lot to cache before going to the next part)..  which is what sab normally sees from my download usage..

so article cache generally never helps me out too much for my normal use.. now if I wanted to load up a 1:1 bluray (50gb) or something.. that is where it would come into play.. not in this case.

thezoggy:
Tried with `python -OO ...` still see it not passing the 18MB barrier :(
UI says 18MB during the middle of the download, click to go to config root to check cache usage..
`Cached 34 articles (25 MB)`
Go back to main ui.. see that it dropped to 17.6 then slowly starts to make its way back towards 18MB before it finishes.. result:
`Downloaded in 1 min 6 seconds at an average of 17.4 MB/s`

Safihre:
The dip is cheetah doing its caching thing on the first load of any template after a restart. 
Maybe 1 last thing to try: set nr_decoders to 1 and restart. 

I truly don't understand how this is possible and why your hardware in general doesn't easily max out the connection, but by slower hardware does. Strange..

Safihre:
@thezoggy On side-note I realized that the cache info is also in the `queue`-API call, so now the cache-info in the Status Window will be updated with every Queue update. 
Just pushed this to `develop`.

Safihre:
@thezoggy let's try something else:
You have the source for `_yenc`, right? So try this:
First make sure that `_yenc` is completly removed (have to manually delete it from the `site-packages` directory!) and then modify the source by replacing this:
```C
output_len = decode_buffer(input_buffer, output_buffer, input_len, &crc, &escape);
```
by:
```C
Py_BEGIN_ALLOW_THREADS;
output_len = decode_buffer(input_buffer, output_buffer, input_len, &crc, &escape);
Py_END_ALLOW_THREADS;
```
And then install and run the `tags/1.2.x` branch or `develop` with `sabyenc` turned off.

Safihre:
@thezoggy, did you ever see this presentation? https://www.youtube.com/watch?v=Obt-vMVdM8s
It shows how the GIL actually causes python programs to run slower when there are more cores, maybe this is what's happening on your PC.

thezoggy:
removed yenc from site-packages.. 
then went to config > special > enable_sabyenc and set that to off.. save. restart sab. 
it still finds it and does NOT show a warning that sabyenc isnt used... 

importing yenc and _yenc, I see that these are also in the lib folder...
>>> print yenc
<module 'yenc' from 'C:\Python27\lib\yenc.py'>
>>> print _yenc
<module '_yenc' from 'C:\Python27\lib\_yenc.pyd'>

removed those as well..

```
C:\Program Files (x86)\sabnzbd_git>python
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] on win32
>>> import yenc
ImportError: No module named yenc
>>> import _yenc
ImportError: No module named _yenc
>>> import sabyenc
>>> print sabyenc
<module 'sabyenc' from 'C:\Python27\lib\site-packages\sabyenc.pyd'>
```

thus, we prob want to put something in logs saying after sabyenc is found that its being disabled per special setting....

thezoggy:
found something interesting...
using this test `ubuntu-16.04.2-desktop-amd64.iso` nzb from a.b.test, 
shared here for reference: http://zoggy.net/ubuntu-16.04.2-desktop-amd64.iso.nzb

we see that with no yenc/sabyenc.. it fails in sab.
with sabyenc enabled it d/ls just fine....

--

with no yenc and sabyenc disabled:

See it hit 17.8 MB/s @ 12% cpu hitting the 4 physical cores
2017-02-21 14:08:53,448::INFO::[__init__:673] Adding ubuntu-16.04.2-desktop-amd64.iso.nzb
2017-02-21 14:09:37,088::DEBUG::[nzbstuff:1043] Abort job "ubuntu-16.04.2-desktop-amd64.iso", due to impossibility to complete it
sab-test
Size: 758 MB
Source: ubuntu-16.04.2-desktop-amd64.iso.nzb
Download: Downloaded in 43 seconds at an average of 17.3 MB/s
Age: 1d -- 354 articles were malformed
Servers: SuperNews=758 MB

--

with sabyenc enabled:

See it hit 17.8 MB/s @ 9% cpu, later goes to hit 18 MB/s but peaks at 15% cpu.. 
2017-02-21 14:15:32,210::INFO::[misc:890] Creating directories: \\?\c:\test\downloads\incomplete\ubuntu-16.04.2-desktop-amd64.iso
2017-02-21 14:17:06,279::INFO::[postproc:577] Cleaning up ubuntu-16.04.2-desktop-amd64.iso (keep_basic=False)
sab-test2
sab-test3
Size: 1.5 GB
Source: ubuntu-16.04.2-desktop-amd64.iso.nzb
Download: Downloaded in 1 min 26 seconds at an average of 17.6 MB/s
Age: 1d
Servers: SuperNews=1.5 GB

Safihre:
I should test that without sabyenc/_yenc again. 
It's a bit weird since none of the code changed, if it works for _yenc it should work when neither sabyenc or _yenc are present.

Should also be noted that the CPU Downloader-optimisation only kicks in after ~15 seconds.

PR #749, opened 2016-12-12
Bugfixes for 1.2.0

Safihre:
Mostly bug fixes and update of the year.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -47,7 +47,7 @@ Release Notes  -  SABnzbd 1.2.0
   built-in post-processing options that automatically verify, repair,
   extract and clean up posts downloaded from Usenet.
 
-  (c) Copyright 2007-2016 by "The SABnzbd-team" \<team@sabnzbd.org\>
+  (c) Copyright 2007-2017 by "The SABnzbd-team" \<team@sabnzbd.org\>
 
 ### IMPORTANT INFORMATION about release 1.x.x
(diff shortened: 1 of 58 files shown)


README.mkd
```diff
@@ -28,6 +28,7 @@ Release Notes  -  SABnzbd 1.2.0
 - Dropped dependency on PyOpenSSL, now only requires cryptography package
```
shypike: I agree with Zoggy. Don't remove yet.

Safihre:
@shypike 

- Skimming through SABnzbd.py I removed the ```RotatingFileHandler``` for Windows that was added long ago but never used and removed CherryPy config settings that don't exist anymore
- Updated all the Copright's already to 2017, just so we don't forget
- On shutdown/restart the notifier would send a 'Shutdown completed' notification to the Windows Icon, even though it was already gone.
- Updated the check if there's already a version of SABnzbd active to also include the the current `__version__` so it also works for people on `develop`.
- When there was a recursive unpacking in the destination folder, the path was not shortened on Windows causing UnRar to fail when extracting.

war59312:
Typo on line 564.

> Explain erver optional tickbox

To:

> Explain server optional tickbox

Missed the s, whoops. ;)

Safihre:
@war59312 Luckily just in the comment :)

war59312:
Indeed ;)

PR #757, opened 2016-12-23
Config refactoring

Safihre:
In collaboration partly with @sanderjo some re-working of the Config:

- Changed the General page, making a separate _Security_ section and move things like _Update check_ setting there since these are General settings. HTTPS in the _SABnzbd Web Server_ settings (collapsed when not enabled), since it's about the Web Server.
- The Servers page had become way too long with many confusing settings. Now only the things that actually matter er show, and rest is hidden behind a new 'Advanced' button. Just like in the Wizard, enabling SSL will auto-change the port setting from 119 to 563 and vica versa.
- Add 4 default Categories for new users: `tv`, `movies`, `software`  and `audio`. So users can get an idea of what categories can be used for. These are also Newznab categories that SAB now auto-matches.
- Moved disabled_api_key to Specials, but also removed the texts and some Javascript that was used for it (a bit more than @shypike put in #756 😉 )

### DIFF
--- a/ABOUT.txt
+++ b/ABOUT.txt
@@ -1,5 +1,5 @@
 *******************************************
-*** This is SABnzbd 1.1.x               ***
+*** This is SABnzbd 1.2.x               ***
 *******************************************
 SABnzbd is an open-source cross-platform binary newsreader.
 It simplifies the process of downloading from Usenet dramatically,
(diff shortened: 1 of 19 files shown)


Safihre:
@jcfp In order to support more attributes from newznab indexers I had to extend feedparser. But I build-in `excepts` to make sure that if the adapted version is not used it still works.

:christmas_tree:

jcfp:
@Safihre thanks for the heads-up :snowman:

PR #767, opened 2017-01-08
Fix  %s mismatch in NL translation

jcfp:
introduced in commit f8d6286


Safihre:
@shypike it's already updated on Launchpad, can you do the upload of the POT and the update of the translations? :)

shypike:
Already solved by pulling new translations from Launchpad.
Pushed latest POT to Launchpad, in case some new text weren't there yet.

PR #769, opened 2017-01-11
Last stuff for 1.2.0 Final

Safihre:
Only small fixes.

I think that now in 2017 with all these high-speed connections, not having `_yenc` is bad enough to warrant a warning on startup. In preparation for `sabyenc`!


interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -691,6 +691,9 @@ function ViewModel() {
         var file = files[fileindex]
         fileindex++
 
+        // Check if it's maybe a folder, we can't handle those
+        if(!file.type && file.size % 4096 == 0) return;
```
shypike: I understand "! file.type", but I don't understand "file.size % 4096 == 0"

sabnzbd/utils/feedparser.py
```diff
@@ -529,6 +529,7 @@ class _FeedParserMixin:
         'http://www.w3.org/XML/1998/namespace':                  'xml',
         'http://www.newznab.com/DTD/2010/feeds/attributes/':     'nZEDb',
         'http://www.newznab.com/DTD/2010/feeds/attributes/':     'newznab',
+        'http://www.newznab.com/DTD/2010/feeds/attributes/':     'nntmux',
```
shypike: Patching feedparser.py is OK, but I think a number of package creators (like JCFP) are using the already installed version. You would need to change the name to cover this (and warn package makers).

SABnzbd.py
```diff
@@ -431,10 +431,7 @@ def print_modules():
     if sabnzbd.decoder.HAVE_YENC:
         logging.info("_yenc module... found!")
     else:
-        if hasattr(sys, "frozen"):
-            logging.error(T('_yenc module... NOT found!'))
-        else:
-            logging.info("_yenc module... NOT found!")
+        logging.warning(T('_yenc module... NOT found!'))
```
shypike: It's still an error to miss _yenc in a "frozen" app.
Are you sure about the warning? First time in 10 years.
If you have no choice, it's no fun to get a warning every time.
But then again, it is way better to have it.

interfaces/Glitter/templates/static/javascripts/knockout-extensions.js
```diff
@@ -51,6 +51,10 @@ ko.bindingHandlers.longText = {
             // Nothing special
             outputText += value.join('<br />');
         }
+
+        // Replace link to our website
+        outputText = outputText.replace('https://sabnzbd.org/not-complete', '<a href="https://sabnzbd.org/not-complete" class="history-status-dmca" target="_blank">https://sabnzbd.org/not-complete</a>')
+
```
shypike: You have the same URL in three different places.
It would be better to put it in "constants.py" and send it as a parameter to the Web-UI.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
Safihre: According to the docs, files that are actually folders are reported without a type and a size that's exact multiples of `4096`. No idea why, but it works :P

Issue #494, opened 2016-02-28
Cleanup leftover files after crash on importing/grabbing NZB 


zeon:
Does anyone encounter this before? Jobs I added to the queue just stay there forever and no error logs or anything. See the screen capture below:

![image](https://cloud.githubusercontent.com/assets/450073/13380285/f6979470-de79-11e5-8552-2e0f36fa0b7f.png)

and if I copy that download url sabnzbd tries to fetch the nzb file in the browser it downloads just fine. Any clues?

Safihre:
What version of Sabnzbd are you on?

zeon:
0.7.20

Safihre:
Are you running it locally or on your NAS/server?

zeon:
NAS

Safihre:
There are problems with SSL certificates (for the https connection) on Synology NAS. Could you try adding the same URL but then with http instead of https. Browsing through the forum I found updates to Python on the NAS might solve this problem. There is also a way to disable certificate validation, but can't find the guide on the forum for that now @sanderjo is the expert there :)

thezoggy:
config > special > enable_https_verification -- set that to off

Safihre:
@thezoggy he's on 0.7.2, that special and function were only introduced in 1.0.0.

thezoggy:
oh yeah, just downgrade python to pre 2.7.9 or upgrade sab version

labrys:
This is appears to be a duplicate of what I was experiencing in #404.  After reading this I tried @thezoggy's fix and continued to have the same issues.  I looked into it some more and found that it appears the URL grabber appears to be dying.  I watched the console logs to see what might be triggering it and found that sometimes when the URL grabber runs, it prints out the content of the NZB file to the log and after that the URL grabber never runs again.   If you remove the offending job and restart SAB then the URL grabber will continue to function as normal and resolving jobs, until it chokes on another job.  Rinse lather and repeat until all jobs are resolved and all offending jobs have been removed.  If you copy the offending jobs URL, download it via your browser, and add the NZB that way, it will proceed with the job as normal, however it appears the majority of those NZBs fail to download.

Tested on Win 10 running Python 2.7x64 and SAB, various 1.0.0 versions from dev and master.

Safihre:
Could you maybe post the part of the log where it goes wrong? Or attach it to the message, think Github allows that.

labrys:
@Safihre Redacting it right now so I can post it.

[...]

Issue #373, opened 2015-10-23
sab 0.8.x -- windows + files named Con.*


thezoggy:
`Con.Man.s01e01.Stalled.720p.WEB-DL.AAC2.0.H.264` - https://nzbcache.info/nzb/64ce652425293504ab056d4f52eb81bb6d99517a.nzb.gz

sab 0.7.x wont even load the nzb without puking on it.
sab 0.8.x loads it, d/les it and extracts.. but then you hear a series of beeps.. and windows internally is freaking out? sab logs look okay oddly. but thought it was interesting. ill send you logs via email.. not really anything for you to do.. but thought it was different :p

shypike:
Known problem in WIndows. You cannot have files called con.* or *.con.
In the (very) old days MSDOS reserved the name "con", so you could not create a file called "con", but you could create "con.txt". Then some complete moron in the Windows NT team decided that "*.con" or "*.con" were also forbidden. This problem is in all versions of Windows since NT.
The same is true for "prn", "aux" and few more of those. BTW: "con" means console.

thezoggy:
yep, i knew about the reserved words.. just thought its odd how sab loading that nzb causes system beeps..

Stanzilla:
*.con works, just con.* works here, I just created a con.txt on Win8.1 just fine.

shypike:
@Stanzilla. What, Microsoft finally solved that problem?

Stanzilla:
Yeah well `con.txt`still fails so... :D

thezoggy:
Stanzilla, 'con' isnt the issue. but 'con.' so try 'con.man' and see.. but then try `mkdir \\?\E:\Test\Con.man` will work just fine.

shypike:
@thezoggy That is odd, because 0.8.0 uses long path notation. I'll look into it further.

Stanzilla:
I said exactly that :p

thezoggy:
@shypike yes, which is why sab 0.8.x can handle it but 0.7.x cant..

Stanzilla:
I reported it as a problem via the Windows 10 feedback app for good measure..

shypike:
Should be fixed now: https://github.com/sabnzbd/sabnzbd/commit/d3d5f37d49e563bfb2a60b72ad9f58a20580608a

thezoggy:
does that only sanitize the file or both the file and folder?

shypike:
The folder name only.
[...]

PR #770, opened 2017-01-11
Fix URL grabber crashing on invalid windows file names

labrys:
When attempting to write an invalid file (such as device files on Windows) the file doesn't get created and the URL grabber stops responding.  As discussed in #494 modifying the filename can lead to other consequences such as issues with unrar.  This raises an IOError instead when the file doesn't get written so the error can propagate but the URL grabber can still continue.

http://superuser.com/questions/613313/why-cant-we-make-con-prn-null-folder-in-windows

### DIFF
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -228,6 +228,10 @@ class URLGrabber(Thread):
                 f = open(path, 'wb')
                 f.write(data)
                 f.close()
+                # make sure file exists after writing to avoid crashing
+                # the URL grabber
+                if not os.path.exists(path):
+                    raise IOError('Unable to create temp file: ' + path)
                 del data
 
                 # Check if nzb file


labrys:
This may be a regression of #373 and may just need `replace_win_devices` applied, however even with that, this could add value by making sure the file exists before continuing.  This could also be an assertion instead of an IF but as I didn't notice any assertions in a quick review of the production code, I opted for an IF instead.

shypike:
This is not the right solution.
It will crash the urlgrabber (and restart) which is a rather crude method.
You suggestion of calling "replace_win_devices" makes more sense. There's no risk involved here, because it's just the name of the NZB.

Safihre:
The PR was based not based on the `develop` branch, so I accidentally closed it when I deleted the `feature/more_for_1.2.0` branch!
But we (well @shypike :P) will work on a similar solution!

labrys:
No problem.  I personally preferred the replace option (it's what I did locally) but I figured there might have been a reason you chose not to.  I had planned on closing the PR anyway after your earlier response so no problems :).  I assume you want future changes to be made against develop, so I'll do so in the future.

Safihre:
@labrys Turns out the solution in #771 also causes problems when the NZB still downloads files starting with `Con` (it will hang the whole rest of the Queue). 
So we will try to either: detect the download and abort it with a proper error message, or handle these downloads correctly. 
But the completly correct handling requires little tweaks throughout the process. So something for 1.2.1!

thezoggy:
Safihre, if you saw my email about this a few weeks/month ago.. you can see that the nzb downloads just fine (the folder gets sanitized with `_` in front) and as long as the actual filename isnt invalid.. all is well.. which was the case. The only problem was the saving of the NZB using the illegal filename.

labrys:
@Safihre Personally I would recommend at least a hacky fix before then because as it sits right now the entire Queue completely hangs as soon as a single file starting with Con is found.  Essentially SAB queue is broke until I manually identify and delete offending items from the queue.  Even if its just raising a warning "Can't download this file" and skipping it for the next one in the list.

PR #771, opened 2017-01-12
Windows: prevent NZB names with forbidden elements

shypike:
Replaces #770

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -33,7 +33,8 @@ from sabnzbd.encoding import TRANS, UNTRANS, unicode2local, \
     reliable_unpack_names, unicoder, platform_encode, deunicode
 import sabnzbd.utils.rarfile as rarfile
 from sabnzbd.misc import format_time_string, find_on_path, make_script_path, int_conv, \
-    flag_file, real_path, globber, globber_full, short_path, get_all_passwords, renamer
+    flag_file, real_path, globber, globber_full, get_all_passwords, renamer, clip_path, \
+    has_win_device
 from sabnzbd.tvsort import SeriesSorter
 import sabnzbd.cfg as cfg
 from sabnzbd.constants import Status, QCHECK_FILE, RENAMES_FILE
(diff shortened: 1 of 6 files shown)


thezoggy:
example reported via email a few weeks back on 1.1.1, can see we sanitize the folder but not the actual nzb name:
`Saving \\?\C:\Users\test\downloads\incomplete\_Con.Man.S02E01.What.Goes.Up.7\__ADMIN__\Con.Man.S02E01.What.Goes.Up.720p.WEB-DL.DD5.1.H.264-CM.nzb.gz failed `

Safihre:
@shypike Maybe we should move the filesystem functions out of `misc.py` to a seperate `filesystem.py`? `misc.py` is getting big..

shypike:
@Safihre Yes, but not in 1.2.1

Safihre:
# For future reference:
Windows does not support paths >256 and has special reserved words.
Programs handle this differently:

- Python uses `\\.\` and this works fine
- `Par2tbb`/`par2cmdline`: Paths should start with the alternative long-path notation: `\\.\`
- `Unrar`: Does not understand `\\?\` or `\\.\` for path to RAR, but does understand and it for the extraction path.
- `Par2` and `unrar` can/will create 'bad' names, so after each step we need to clean the output

shypike:
@Safihre Python uses \\?\ and this works fine.

thezoggy:
windows: reserve 
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396#namespaces

sabnzbd/newsunpack.py
```diff
@@ -1192,6 +1171,7 @@ def PAR_Verify(parfile, parfile_nzf, nzo, setname, joinables, classic=False, sin
     stup, need_shell, command, creationflags = build_command(command)
 
     # par2 wants to see \\.\ paths on Windows
+    # See: https://github.com/sabnzbd/sabnzbd/pull/771
```
shypike: Dit soort links werken misschien later niet meer

thezoggy: `This kind of links might later work no more` 
yes, we should just make a hidden wiki page for dev notes? and then just link to the dev note  ID there instead?

shypike: yes

shypike: Or just more comments or commit text

Safihre:
Turns out we were partly wrong, par2cmdline doesn't actually support the `\\.\` or `\\?\` notation. It allows to specify the par2-file in this way, but it ignores any of the other files in the folder:
````
Ignoring non-existent source file: \\.
```
So this means that it will fail on obfuscated posts. 
We have to implement a work-a-round so that it doesn't freeze in case of a special Windows word.

PR #772, opened 2017-01-13
Set hard limit on Decoder-queue

Safihre:
The calculation of memory usage for articles in the Decoder-queue is not correct, which can cause slow growth of that Queue and its corresponding memory usage.

Tried multiple ways to calculate a correct size, but python's `sys.getsizeof(data)` doesn't return the right amount. So for now, as a safety patch, set a hard limit of 100 articles.
This is oke because the problem is actually the smaller articles, which take relatively longer to go through the whole decode process than bigger ones.


thezoggy:
https://code.tutsplus.com/tutorials/understand-how-much-memory-your-python-objects-use--cms-25609

Safihre:
@thezoggy Agreed, we need a better solution. But unfortunately that function takes too long, for each article about ~5ms. 
For reference, yEnc decoding of 1 article using `_yenc` takes ~3ms.

PR #780, opened 2017-01-17
Bugfixes after 1.2.0

Safihre:


sabnzbd/interface.py
```diff
@@ -2833,17 +2833,23 @@ def make_item(job):
...
-    # Sort in reverse order of time stamp for 'Done'
-    names.sort(lambda x, y: int(jobs[y].get('time', 0) - jobs[x].get('time', 0)))
-    done = [make_item(jobs[job]) for job in names if jobs[job]['status'] == 'D']
+    try:
+        # Sort based on actual age, in try-catch just to be sure
+        good.sort(key=lambda job: job[8], reverse=True)
+        bad.sort(key=lambda job: job[8], reverse=True)
+        done.sort(key=lambda job: job[10], reverse=True)
+    except:
+        pass
```
shypike: Ugly, hides problems.
Which kind of trouble do you expect here?
At least create a log statement with a trace.

interfaces/Config/templates/config_rss.tmpl
```diff
@@ -401,16 +401,16 @@
                             <form action="download" method="get">
```
shypike: You commit message says "super unclear what was happening".
It still is. Please summarize in the commit comment.

po/main/en.po
```diff
@@ -60,10 +60,6 @@ msgstr "Disable server:"
 msgid "enable server"
 msgstr "Enable server:"
 
-#: sabnzbd/skintext.py:661
-msgid "Generic Sorting"
-msgstr "Movie Sorting"
-
 #: sabnzbd/emailer.py:117
 msgid "The server didn't reply properly to the helo greeting"
 msgstr "The server didn't reply properly to the hello greeting"
```
shypike: Not part of your commit, but this correction is funny.
It really is called the "helo" greeting, so the correction is not really justified.
The actual text should be something like "Email server didn't respond according to protocol".

sabnzbd/utils/feedparser.py
```diff
@@ -1775,11 +1775,15 @@ def _start_newznab_attr(self, attrsD):
...
-            context['newznab'][attrsD['name'] + '_parsed'] = _parse_date(attrsD['value'])
+        # Don't crash when it fails
+        try:
+            # Add keys
+            context['newznab'][attrsD['name']] = attrsD['value']
+            # Try to get date-object
+            if attrsD['name'] == 'usenetdate':
+                context['newznab'][attrsD['name'] + '_parsed'] = _parse_date(attrsD['value'])
+        except:
+            pass
```
shypike: Please use explicit except and at least log something.

sabnzbd/newsunpack.py
```diff
@@ -534,7 +534,7 @@ def rar_extract_core(rarfile_path, numrars, one_folder, nzo, setname, extraction
```
shypike: Please explain what this change actually does.

README.mkd
```diff
@@ -1,53 +1,13 @@
-Release Notes  -  SABnzbd 1.2.0
```
shypike: Please separate "readme" changes from functional changes.
Combined commits make future cherry-picking more difficult.

sabnzbd/newswrapper.py
```diff
@@ -241,10 +241,9 @@ def __init__(self, host, port, info, sslenabled, send_group, nw, user=None, pass
             if not block:
                 Thread(target=con, args=(self.sock, self.host, self.port, sslenabled, write_fds, self)).start()
             else:
-                # if blocking (server test) only wait for 4 seconds during connect until timeout
-                self.sock.settimeout(4)
```
shypike: 15 sec is almost forever.
Any specific reason?

sabnzbd/interface.py
Safihre: Jobs of older feeds that don't have the right fields. Not sure what error would be thrown in that case.

interfaces/Config/templates/config_rss.tmpl
Safihre: Ok

sabnzbd/utils/feedparser.py
Safihre: Ok

sabnzbd/newsunpack.py
Safihre: This change is later removed because we remove the check for Expected files, so not important anymore.

sabnzbd/newswrapper.py
Safihre: As mentioned in the commit message:

> Some servers (Altopia) use 10 second timeouts with wrong credentials to prevent flooding.

shypike: Right!

sabnzbd/newsunpack.py
shypike: ok

PR #792, opened 2017-01-24
CPU reporting now also on Windows and MacOS (fka OSX)

sanderjo:


sabnzbd/utils/getperformance.py
```diff
@@ -1,14 +1,39 @@
...
+                if myline.startswith(('model name')):
+                    # Typical line:
+                    # model name      : Intel(R) Xeon(R) CPU           E5335  @ 2.00GHz
+                    cputype = myline.split(":", 1)[1]	# get everything after the first ":"
+                    break # we're done
+
     except:
-        # probably not on Linux
+        # An exception, maybe due to a subprocess call gone wrong
+        cputype = "Exception happened"
```
Safihre: This means that if will show this in the Status window, isn't that maybe a bit too dramatic?

sanderjo: IMHO: No. It should never happen, and so it should never appear. And if it happens I want to know it so that we know where to look for the problem. So a king of assert.
It's a different case than on 'unknown' OSes; then the output is None, and nothing appears.

thezoggy:
could also just use: https://github.com/workhorsy/py-cpuinfo

PR #800, opened 2017-01-29
new userxbit() as workaround for certain mounted filesystems

sanderjo:
Workaround for problem with mounted filesystems ("mergerfs") where x-bit was not reported by os.access(), as reported on https://forums.sabnzbd.org/viewtopic.php?f=2&t=22359

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -1521,6 +1521,18 @@ def fix_env():
     else:
         return None
 
+def userxbit(filename):
+    # Returns boolean if the x-bit for user is set on the given file
+    # This is a workaround: os.access(filename, os.X_OK) does not work on certain mounted file systems
+    # Does not work on Windows, but it is not called on Windows
+
+    # rwx rwx rwx
+    # 876 543 210      # we want bit 6 from the right, counting from 0
+    userxbit = 1<<6 # bit 6 
+    rwxbits = os.stat(filename)[0] # the first element of os.stat() is "mode"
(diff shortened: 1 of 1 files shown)


sanderjo:
Some testing:

With these files in /tmp:
```
$ ll /tmp
drwxr-xr-x  2 sander sander  4096 jan 29 17:34 dir-with.py/
-rwxr-xr-x  1 sander sander     0 jan 29 17:34 hallo-exec.py*
-rw-r--r--  1 sander sander     0 jan 29 17:34 hallo-no-exec.py
-rw-r--r-x  1 sander sander     0 jan 29 17:40 others-x-bit.py*
```

the new function userxbit() results in:
```
sander@Stream-13:~/git/sanderjo-sab-28jan2017$ python
Python 2.7.12+ (default, Sep 17 2016, 12:08:02) 
[GCC 6.2.0 20160914] on linux2
>>> import sabnzbd.newsunpack
>>> sabnzbd.newsunpack.userxbit('/bin/bash')
True
>>> sabnzbd.newsunpack.userxbit('/tmp/dir-with.py')
True
>>> sabnzbd.newsunpack.userxbit('/tmp/hallo-exec.py')
True
>>> sabnzbd.newsunpack.userxbit('/tmp/hallo-no-exec.py')
False
>>> sabnzbd.newsunpack.userxbit('/tmp/others-x-bit.py')
False

```
So:
* Correct behaviour for the files
* The x-bit on the directory is technically correct. Do we agree on that? Is it OK for SABnzbd?

PR #819, opened 2017-02-13
Try os.rename 3 times, before using shutil.move()

Safihre:
To try and prevent a copy (instead of an instant rename), we should first try 3 times with `os.rename()` in the hopes that it works.
At max we lose 9 seconds before the `shutil.move()` is started.
Hoping this will fix the problems of users that see a copy happen when renaming `__UNPACK__` to final name on a NAS from Windows.

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -918,6 +918,7 @@ def move_to_path(path, new_path):
             renamer(path, new_path)
         except:
             # Cannot rename, try copying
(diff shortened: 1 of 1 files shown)


thezoggy:
I personally have seen this due to apps trying to read/generate metadata off the media file.. and os thumbnail generation / mediainfo gets a lock on the file... so hopefully this does resolve that :)

Issue #798, opened 2017-01-29
sab 1.2.0 -- unable to validate cert / RemoteCertificateNameMismatch / RemoteCertificateChainErrors


thezoggy:
running sab 1.2.0 binary on windows, using https with sab generating its own self signed cert

per browser its unable to verify because the issuer is unknown.
Owner: `This website does not supply ownership information.`

apps like radarr/sonarr show:
`17-1-28 20:55:23.3|Debug|X509CertificateValidationPolicy|Certificate validation for https://192.168.0.2:8000/api?mode=queue&start=0&limit=1&apikey=(removed)&output=json failed. RemoteCertificateNameMismatch, RemoteCertificateChainErrors`

but everything 'works'.. just a bit of logging noise because of it all.

thezoggy:
looking through the code, i see mention of `key.pem` and `cert.cert` but cant find where they actually end up..  in the admin folder I see `server.key` and `server.cert`

thezoggy:
side note, X.509 certs per mmc should have the file name ext of: `*.cer;*.crt`

thezoggy:
was about to resolve the `RemoteCertificateChainErrors` part.. cant get rid of the other one though. even tried importing it into trusted certs and other places..  digging more into this.. looks like its because the CN is not my domain we use `SABnzbd` which will never be correct..  thus i need to just hack in sab to generate the key using my ip instead

thezoggy:
ok nuked the certs from the browser/store/etc.

ok ran sab from source on the box and changed out the CN from `SABnzbd` to `192.168.0.2` and had it generate a new key/cert. shut down that instance and copied over the relevant .key/cert to sab. restarted sab+sonarr.. now I see:

`17-1-28 21:52:37.7|Debug|X509CertificateValidationPolicy|Certificate validation for https://192.168.0.2:8000/api?mode=queue&start=0&limit=0&apikey=(removed)&output=json failed. RemoteCertificateChainErrors`

So looks like that fixed the CN mismatch.. now about the ChainError..
> You are not providing an issuer for your certificate and in this way it can't be trusted - this is the cause of the RemoteCertificateChainErrors Exception.

So I've imported my sab cert into trusted chain.. now no more message and all good.

`17-1-28 21:57:07.9|Debug|SabnzbdProxy|Url: https://192.168.0.2:8000/api?mode=queue&start=0&limit=0&apikey=(removed)&output=json`

thezoggy:
[...]

PR #822, opened 2017-02-15
Update self-signed cert code to use SubjectAltName

thezoggy:
Addresses https://github.com/sabnzbd/sabnzbd/issues/798

### DIFF
--- a/sabnzbd/utils/certgen.py
+++ b/sabnzbd/utils/certgen.py
@@ -12,7 +12,7 @@ from cryptography import x509
 from cryptography.x509.oid import NameOID
 import datetime
 import os
-import struct
+from sabnzbd.getipaddress import localipv4, ipv6
 
 
 # Ported from cryptography/utils.py
@@ -20,25 +20,16 @@ def int_from_bytes(data, byteorder, signed=False):
     assert byteorder == 'big'
     assert not signed
 
-    if len(data) % 4 != 0:
-        data = (b'\x00' * (4 - (len(data) % 4))) + data
-
-    result = 0
-
-    while len(data) > 0:
-        digit, = struct.unpack('>I', data[:4])
-        result = (result << 32) + digit
-        # TODO: this is quadratic in the length of data
-        data = data[4:]
-
-    return result
+    # call bytes() on data to allow the use of bytearrays
+    return int(bytes(data).encode('hex'), 16)
 
 
-# Ported from cryptography/utils.py
+# Ported from cryptography/x509/base.py
 def random_serial_number():
     return int_from_bytes(os.urandom(20), "big") >> 1
 
 
+# Ported from cryptography docs/x509/tutorial.rst (set with no encryption)
 def generate_key(key_size=2048, output_file='key.pem'):
     # Generate our key
     private_key = rsa.generate_private_key(
@@ -53,19 +44,32 @@ def generate_key(key_size=2048, output_file='key.pem'):
             encoding=serialization.Encoding.PEM,
             format=serialization.PrivateFormat.TraditionalOpenSSL,
             encryption_algorithm=serialization.NoEncryption()
+            # encryption_algorithm=serialization.BestAvailableEncryption(b"passphrase")
         ))
 
     return private_key
 
 
-def generate_local_cert(private_key, days_valid=356, output_file='cert.cert', LN='', ON='', CN=''):
+# Ported from cryptography docs/x509/tutorial.rst
+def generate_local_cert(private_key, days_valid=3560, output_file='cert.cert', LN=u'SABnzbd', ON=u'SABnzbd', CN=u'localhost'):
     # Various details about who we are. For a self-signed certificate the
     # subject and issuer are always the same.
     subject = issuer = x509.Name([
         x509.NameAttribute(NameOID.LOCALITY_NAME, LN),
         x509.NameAttribute(NameOID.ORGANIZATION_NAME, ON),
-        x509.NameAttribute(NameOID.COMMON_NAME, CN),
+        # x509.NameAttribute(NameOID.COMMON_NAME, CN),
     ])
+
+    # build SubjectAltName list since we are not using a common name
+    san_list = [x509.DNSName(u"localhost")]
+    # add local ipv4 address
+    mylocalipv4 = localipv4()
+    if mylocalipv4:
+        san_list.append(x509.DNSName(u"" + mylocalipv4))
+    # add local ipv6 address
+    myipv6 = ipv6()
+    if myipv6:
+        san_list.append(x509.DNSName(u"" + myipv6))
 
     cert = x509.CertificateBuilder().subject_name(
         subject
@@ -76,11 +80,12 @@ def generate_local_cert(private_key, days_valid=356, output_file='cert.cert', LN
     ).not_valid_before(
         datetime.datetime.utcnow()
     ).not_valid_after(
-        # Our certificate will be valid for 10 days
         datetime.datetime.utcnow() + datetime.timedelta(days=days_valid)
     ).serial_number(
         random_serial_number()
-    # Sign our certificate with our private key
+    ).add_extension(
+        x509.SubjectAlternativeName(san_list),
+        critical=True,
     ).sign(private_key, hashes.SHA256(), default_backend())
 
     # Write our certificate out to disk.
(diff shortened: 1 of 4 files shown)


sabnzbd/utils/certgen.py
```diff
@@ -12,33 +12,24 @@
 from cryptography.x509.oid import NameOID
 import datetime
 import os
-import struct
+from sabnzbd.getipaddress import localipv4, ipv6
```
Safihre: So clearly it's not a standalone module anymore, maybe we can just move it out of the `utils` folder and into the `sabnzbd` folder?
And then also move the `create_https_certificates` function into the file.

sabnzbd/utils/certgen.py
```diff
@@ -53,20 +44,33 @@ def generate_key(key_size=2048, output_file='key.pem'):
...
             encryption_algorithm=serialization.NoEncryption()
+            # encryption_algorithm=serialization.BestAvailableEncryption(b"passphrase")
         ))
 
     return private_key
 
 
-def generate_local_cert(private_key, days_valid=356, output_file='cert.cert', LN='', ON='', CN=''):
+# Ported from cryptography docs/x509/tutorial.rst
+def generate_local_cert(private_key, days_valid=3560, output_file='cert.cert', LN=u'SABnzbd', ON=u'SABnzbd', CN=u'localhost'):
```
Safihre: Maybe not pre-fill SAB defaults to keep it a tiny bit general? :)

Safihre: Since the `CN` is not used, it can be removed as a parameter probably.

Safihre:
Could you also test the default situations,  so with host set to `127.0.0.1` or `0.0.0.0`: if the apps are happier with the new certificates generated like that  than the old style ones?

sabnzbd/utils/certgen.py
thezoggy: certgen is ran for one of 3 scenarios..  we do on refresh.. we do on startup.. user can manually from the dir.  putting the defaults here ensures that all 3 cases result in the same cert.

sabnzbd/utils/certgen.py
thezoggy: the main scenario is sab kicking off the cert gen process..  tried to keep the code as close to the tutorial as possible (which is what we had before).. and the changes we made to the file locally I just cleaned up so it covers the scenario if the user wanted to run it manually and less likely to get different results

thezoggy:
i can test other variations of the host being set, but the main thing is how does the user access sab.. do they go to `https://localhost` or `https://127.0.0.1` or what..  localhost is automatically added.. so that scenario is covered.. then whatever sab has as their local ip.. we add that.. so thats covered as well.

thezoggy:
if i tell sab host is `0.0.0.0`, it then relaunch web browser to `https://127.0.0.1:8092` then telling it to generate new cert.. the SAN list does not change (not that I expected it to), since the computers local ip didnt change.. I end up with the same entries, localhost and my local ip.

So now that sab is bounded to `0.0.0.0` instead of ` ` does that change how the cert is validated against? no. because apps would still only validate off the SAN list...

Also to note, radarr does not like sab source reporting back `develop`, reported this: https://github.com/Radarr/Radarr/issues/769

I can see that if I put in `127.0.0.1` for the access ip it fails to validate but if i use the local ip or localhost.. its fine... so in the end the certs stuff is the same as before just handles the case when people access sab via their local ip or as localhost.. which covers the majority of use cases. For those that are trying to do self-signed certs for remote.. they need to generate their own/bind to a proper dns name/use reverse proxy/etc.. all outside the scope of our self signed certs.

Safihre:
Thanks for testing. Can you not just also add `127.0.0.1` to the SAN list?

Safihre:
@thezoggy IPv6 is causing problems:
```
  File "C:\Users\<USERNAME>\Documents\GitHub\sabnzbd\sabnzbd\misc.py", line 1254, in create_https_certificates
    cert = generate_local_cert(private_key, days_valid=3560, output_file=ssl_cert, LN=u'SABnzbd', ON=u'SABnzbd', CN=u'localhost')
  File "C:\Users\<USERNAME>\Documents\GitHub\sabnzbd\sabnzbd\utils\certgen.py", line 89, in generate_local_cert
    ).sign(private_key, hashes.SHA256(), default_backend())
  File "C:\Python27\Lib\site-packages\cryptography\x509\base.py", line 564, in sign
    return backend.create_x509_certificate(self, private_key, algorithm)
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\multibackend.py", line 404, in create_x509_certificate
    return b.create_x509_certificate(builder, private_key, algorithm)
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\backend.py", line 883, in create_x509_certificate
    gc=True
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\backend.py", line 1009, in _create_x509_extensions
    handlers, extension
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\backend.py", line 1040, in _create_x509_extension
    ext_struct = encode(self, extension.value)
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\encode_asn1.py", line 359, in _encode_alt_name
    general_names = _encode_general_names(backend, san)
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\encode_asn1.py", line 351, in _encode_general_names
    gn = _encode_general_name(backend, name)
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\encode_asn1.py", line 387, in _encode_general_name
    value = _idna_encode(name.value)
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\openssl\encode_asn1.py", line 376, in _idna_encode
    return idna.encode(value)
  File "C:\Python27\Lib\site-packages\idna\core.py", line 355, in encode
    result.append(alabel(label))
  File "C:\Python27\Lib\site-packages\idna\core.py", line 276, in alabel
    check_label(label)
  File "C:\Python27\Lib\site-packages\idna\core.py", line 253, in check_label
    raise InvalidCodepoint('Codepoint {0} at position {1} of {2} not allowed'.format(_unot(cp_value), pos+1, repr(label)))
InvalidCodepoint: Codepoint U+003A at position 5 of u'2001:0:34a9:ece7:82e:1f23:abe7:dd15' not allowed
```

thezoggy:
@Safihre yeah.. its because crypto wants unicode argument for dns name so it can decode.. but doing so for v6 turns the colon  into unicode which then gets disallowed.. blah I'll just use the not so well known IPAddress type but then it involves having to convert the string to ipaddress type...  one sec. ill add 127.0.0.1 as well for ya

thezoggy:
so if I use x509.DNSName the apps are happy with 127.0.0.1/localhost/localip.. if I use x509.IPAddress for 127.0.0.1 or localip.. it doesnt validate...

thezoggy:
yeah, so you must pass a unicode string to crypto for dnsname.. but that name cant have a colon.. thus we wont be able to do the SAN for the v6 address.. we could add it under ipaddress type but as mentioned before.. its rarely used and in my case radarr/sonarr arent validating the certs off them (only dns name)..  so honestly those using v6 will have to just create their own cert since I cant find a way to sidestep crypto's requirement without monkey patching and I really dont want to go down this road

thezoggy:
alright pushed update, let me know when you want me to rebase+squash and force push. i just cant get v6 addresses to work in the dnsname.. due to the requirement of unicode string being passed but then it failing due to colons in it...

Safihre:
Oke looks good. I was only wondering why you leave out the `CN` completly? Why not put `127.0.0.1` or `localhost` there?

thezoggy:
if we include CN lots of app will check that and only that.. not including it forces them to use SAN..  CN is legacy way to validate.. 

per RFC 6125, published in '2011 the validator must check SAN first, and if SAN exists, then CN should not be checked. Now not all apps follow this, and some will use the CN if its there ignoring the SAN.. but if CN isnt there.. it forces it to use SAN. So this is to ensure things are checked correctly. Also I noticed that when I import the cert into the trusted root store that if I provide a CN it shows up as that.. if I dont provide CN it uses the issuer (sabnzbd) which looks much nicer.

Safihre:
Shouldn't the issuer have a `CN`? Not that it really matters of course..

thezoggy:
Pre 2011 yes, with our use (multiple domains/self-signed) CN would not be good..

Safihre:
No I mean `CN` for issuer, not subject.

thezoggy:
the CN is for the issuer.. which we purposely dont want to use.. because we want to use the SubjectAltName

thezoggy:
btw, the certgen code it makes the subject and the issuer the same which is the norm for self-signed certs... we could put a CN for the subject and not the issuer but honestly if we were really trying to create a legit cert we also need to put state/country/email/etc.. but all that stuff is moot since the CA issuer for self-sign certs can never be valid due to their nature.. thus it doesnt matter. the main goal was just to make other local apps that validate the cert to sab happier.. which this now does.

PR #824, opened 2017-02-18
Next steps

Safihre:
Many new things for the next major release, SABnzbd 2.0.0. 
Details in the commit titles/messages.

**Biggest:**
- Major cleanup of code by removing Secondary Skin
- Moved couple of items to Specials to simplify Config
- Set Certificate Validation to Strict for new servers (check out @sanderjo's new online SSL-checker that's linked from the error message! https://sabnzbd.org/certificate-errors) 
- Remove check for `ssl` module: the check was introduced to check for PyOpenSSL but regular `ssl` is a Python standard module.


interfaces/Config/templates/config_switches.tmpl
```diff
@@ -269,11 +269,6 @@
                     </div>
                     <!--#end if#-->
                     <div class="field-pair">
-                        <label class="config" for="enable_meta">$T('opt-enable_meta')</label>
```
shypike: Why move this option to Specials?
Shouldn't it be something the user is aware of?

interfaces/Config/templates/config_switches.tmpl
```diff
@@ -324,11 +324,6 @@
                         <span class="desc">$T('explain-rating_enable')</span>
```
shypike: I'm not sure why this option needs to go.
Even though someone uses an indexer, he/she may not necessarily want to provide rating feedback.

interfaces/Config/templates/config_switches.tmpl
```diff
@@ -321,11 +321,7 @@
                     <div class="field-pair">
```
shypike: As long as the rating host isn't communicated through the NZB, the host should be easy to set.

sabnzbd/interface.py
```diff
@@ -117,28 +117,26 @@ def ConvertSpecials(p):
     return p
```
shypike: Please explain in the commit comment why you do this. A rationale.

SABnzbd.py
```diff
@@ -130,24 +130,6 @@ def guard_loglevel():
     LOG_FLAG = True
```
shypike: Again. Please explain in the commit comment why this is a good idea.

sabnzbd/decoder.py
```diff
@@ -173,7 +173,7 @@ def run(self):
                         logging.info(logme)
```
shypike: Please summarize in the commit comment the rationale for this. Explain what happened.
The actual code changes seem to be more than is covered by the commit title.

cherrypy/_cperror.py
```diff
@@ -214,18 +214,7 @@ def __init__(self, urls, status=None, encoding=None):
         if isinstance(urls, text_or_bytes):
```
shypike: You should keep track of patches you apply to CherryPy for two reasons.
- You'll need to reapply for the next CP release.
- Some packagers assume they can use the official distribution of CP. If not they like to know what patches are needed.

interfaces/Config/templates/config_switches.tmpl
Safihre: With the changes in 1.1.0 we now have more indexers in the wild actually use the system, by sending the headers or using the fields in the NZB. 
But nowadays very few files have actual audio/video ratings of users, so the main purpose of this feature has evolved to automatically report back if the download failed on the server.

The text is also updated as such, that enabling it will send to the indexer if a download failed, only if it failed.

interfaces/Config/templates/config_switches.tmpl
Safihre: It always used to be in Specials. Moved it to Switches with the changes in 1.1.0, but turns out  indexers actually prefer to send their own custom rating URL instead of just a host and having to use the endpoint that was once hardcoded by OZnzb.

interfaces/Config/templates/config_switches.tmpl
Safihre: The question is more: why would a user ever turn it off? Using google and the forums, only once in all these years you suggested to a user to turn this off but that didn't even solve the problem.

So this is something that should just happen under the hood, so to say.

cherrypy/_cperror.py
Safihre: Luckily easy to see by checking the history of the `cherrypy` folder: https://github.com/sabnzbd/sabnzbd/commits/develop/cherrypy

We check very specifically for 8.1.2, so doubt that any packager will not use the included one.

interfaces/Config/templates/config_switches.tmpl
shypike: ok

Safihre:
@shypike Added texts to all commits!

Issue #831, opened 2017-02-23
SAB on Linux: Encoding and locale - Notes and Discussion


sanderjo:
This item is about SAB on Linux (not on Windows nor MacOS). It's meant for discussion and notes.

SABnzbd prints the encoding it is running under. If it says
`Preferred encoding = UTF-8`
everything is fine and SAB's unrar will safely unrar Unicoded rar files 

If it says something else, for example
`Preferred encoding = ANSI_X3.4-1968`
there will be problems if you have a download with unicoded filenames/directories in it. You can and should correct that setting at the OS level.

Suggestions for improvements in SABnzbd:
1) SAB on Linux should print a Warning or Error if the Preferred Encoding is something else than UTF-8
2) or SAB on Linux can print a Warning if Preferred Encoding != UTF-8, and the rar contains names other than pure ASCII.

And we could leave it at that. It's up to the user and his OS supplier to correct this OS locale settings

**Some background**

On my Ubuntu, default encoding is UTF-8:

```
$ locale
LANG=en_US.UTF-8
LANGUAGE=en_US
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=nl_NL.UTF-8
LC_TIME=nl_NL.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=nl_NL.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=nl_NL.UTF-8
LC_NAME=nl_NL.UTF-8
LC_ADDRESS=nl_NL.UTF-8
LC_TELEPHONE=nl_NL.UTF-8
LC_MEASUREMENT=nl_NL.UTF-8
LC_IDENTIFICATION=nl_NL.UTF-8
LC_ALL=
```

as confirmed by python:
```
$ python -c "import locale ; print locale.getpreferredencoding() "
UTF-8
```

With the "env LANG=..." before a command, you can set it temporarily:
```
$ env LANG=c python -c "import locale ; print locale.getpreferredencoding() "
ANSI_X3.4-1968
```

**Impact of LANG on unrar command**

**Preparation**

```
cd
mkdir we-love-unicode
cd we-love-unicode/
wget https://www.appelboor.com/dump/we-love-unicode/unicode-inside.rar
```

Then unrar with default locale LANG, which is UTF-8, which is correct:
```
sander@Streamer13:~/we-love-unicode$ unrar x unicode-inside.rar

[...]

PR #832, opened 2017-02-23
Warning in case of non-UTF-8 setting on Linux/Unix

sanderjo:
As discussed here https://github.com/sabnzbd/sabnzbd/issues/831

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1169,10 +1169,18 @@ def main():
         logging.info('Platform = %s', os.name)
     logging.info('Python-version = %s', sys.version)
     logging.info('Arguments = %s', sabnzbd.CMDLINE)
+
+    # Find encoding; relevant for unrar activities
     try:
-        logging.info('Preferred encoding = %s', locale.getpreferredencoding())
+        preferredencoding = locale.getpreferredencoding()
+        logging.info('Preferred encoding = %s', preferredencoding)
     except:
         logging.info('Preferred encoding = ERROR')
+        preferredencoding = ''
+    # On Linux/FreeBSD/Unix "UTF-8" is strongly, strongly adviced:
+    if 'UTF-8' not in preferredencoding and not sabnzbd.WIN32 and not sabnzbd.DARWIN:
+        logging.warning(T("Your operating system encoding is %s, but it should be UTF-8. Please correct this, or expect problems with Unicoded file and directory names in downloads. SABnzbd cannot solve this for you.") % preferredencoding)
+
 
 
     if sabnzbd.cfg.log_level() > 1:


SABnzbd.py
```diff
@@ -1169,10 +1169,18 @@ def main():
...
+    # Find encoding; relevant for unrar activities
     try:
-        logging.info('Preferred encoding = %s', locale.getpreferredencoding())
+        preferredencoding = locale.getpreferredencoding()
+        logging.info('Preferred encoding = %s', preferredencoding)
     except:
         logging.info('Preferred encoding = ERROR')
+        preferredencoding = ''
+    # On Linux/FreeBSD/Unix "UTF-8" is strongly, strongly adviced:
+    if 'UTF-8' not in preferredencoding and not sabnzbd.WIN32 and not sabnzbd.DARWIN:
```
Safihre: Is this really the only way it's reported? Not also `UTF8` or maybe lowercase?

SABnzbd.py
```diff
@@ -1169,10 +1169,18 @@ def main():
...
+        logging.warning(T("Your operating system encoding is %s, but it should be UTF-8. Please correct this, or expect problems with Unicoded file and directory names in downloads. SABnzbd cannot solve this for you.") % preferredencoding)
```
Safihre: I would avoid the "you" and it's not the OS encoding, it's the encoding the program was started with. How about the more general:
_SABnzbd was started with encoding %s, this should be UTF-8. Expect problems with Unicoded file and directory names in downloads._

SABnzbd.py
sanderjo: I searched for the possible return values, but it's not clear. It's probably also OS dependent.

How about: `preferredencoding.lower().find('utf') >= 0`

```
$ python
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
>>> preferredencoding = 'UTF-8'
>>> print preferredencoding.lower().find('utf') >= 0
True
>>> 
>>> preferredencoding = 'UTF-16'
>>> print preferredencoding.lower().find('utf') >= 0
True
>>> 
>>> preferredencoding = 'something with utf'
>>> print preferredencoding.lower().find('utf') >= 0
True
>>> 
>>> preferredencoding = 'ANSI XYZ'
>>> print preferredencoding.lower().find('utf') >= 0
False
>>> 
```

Safihre: Will UTF16 actually work?
If not, I would just say:
```python
if not ('utf' in preferredencoding.lower() and '8' in preferredencoding.lower()):
```

SABnzbd.py
thezoggy: you do realize that we already check for this?
https://github.com/sabnzbd/sabnzbd/blob/develop/SABnzbd.py#L23

Safihre: @thezoggy No we don't. All we do is try/catch the `setdefaultencoding`. But it turns out that even if it doesn't throw an exception it doesn't mean it actually succeeded in changing the encoding.

We have had multiple reports of this `ANSI.something` encoding in the forums and Reddit, despite the check.

thezoggy: looking at the forums for ansi, i see 1 post in the last several years... 
now putting unicode characters into a config field (like a folder path with one) ends up badly because of the way sab reads/writes it config..

thezoggy:
http://farmdev.com/talks/unicode/

SABnzbd.py
Safihre: I have some other ones via pm.
Users won't have any problems unless they download unicode stuff. 
Is there something against a warning when we detect something we know will cause problems?

thezoggy: Nope, but ideally we should instruct people to setup their env correctly at the start.. vs false hope that its going to work.. and then fail later. :)

sanderjo:
To avoid discussion around a piece of code, I had earlier opened this https://github.com/sabnzbd/sabnzbd/issues/831 to have a discussion there.

sanderjo:
> http://farmdev.com/talks/unicode/

@thezoggy What is your goal of dumping an URL?

Safihre:
Added! Thanks 👍

PR #846, opened 2017-03-07
PEP8 cleanup

thezoggy:
Normally I wouldn't try to mess with import cleanup since the IDE is untrustworthy due to our less than optimal import practice (circular imports/having to do some to make freeze/py2exe happy).. but found two that are safe.

* articlecache

Status was added under:
https://github.com/sabnzbd/sabnzbd/commit/a8788da698dd73f9529ccf42e214f3955a36bfc7

No longer needed due to:
https://github.com/sabnzbd/sabnzbd/commit/12a622a21fbaf69dac4e554b40a08494a983048b

* bpsmeter

floor import was last touched under:
https://github.com/sabnzbd/sabnzbd/commit/2811e5ae7de0075039295965ebafab2cc564746d

floor was removed when you cleaned it up under:
https://github.com/sabnzbd/sabnzbd/commit/5c77f9d9b3d759120edc6b0cb652312096d9fcde
> not sure if this the reason of the rounding error/difference from glitter vs actual?

### DIFF
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -281,7 +281,7 @@ class NzbQueue:
     @synchronized(NZBQUEUE_LOCK)
     def insert_future(self, future, filename, data, pp=None, script=None, cat=None, priority=NORMAL_PRIORITY, nzbname=None, nzo_info=None):
         """ Refresh a placeholder nzo with an actual nzo """
-        if 0: assert isinstance(future, NzbObject) # Assert only for debug purposes
+        if 0: assert isinstance(future, NzbObject) # Assert only for debug purposes @IgnorePep8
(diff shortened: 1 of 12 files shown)


sabnzbd/nzbqueue.py
```diff
@@ -441,6 +441,7 @@ def remove(self, nzo_id, add_to_history=True, save=True, cleanup=True, keep_basi
             nzo = self.__nzo_table.pop(nzo_id)
             nzo.deleted = True
             if cleanup and nzo.status not in (Status.COMPLETED, Status.FAILED):
+                # TODO: Use nzo.is_gone() ?
```
thezoggy: @Safihre pretty sure we should just use is_gone here, added note for ya to review later on..

PR #847, opened 2017-03-08
PEP8 cleanup

thezoggy:
Normally I wouldn't try to mess with import cleanup since the IDE is untrustworthy due to our less than optimal import practice (circular imports/having to do some to make freeze/py2exe happy).. but found two that are safe.

* articlecache

Status was added under:
https://github.com/sabnzbd/sabnzbd/commit/a8788da698dd73f9529ccf42e214f3955a36bfc7

No longer needed due to:
https://github.com/sabnzbd/sabnzbd/commit/12a622a21fbaf69dac4e554b40a08494a983048b

* bpsmeter

floor import was last touched under:
https://github.com/sabnzbd/sabnzbd/commit/2811e5ae7de0075039295965ebafab2cc564746d

floor was removed when you cleaned it up under:
https://github.com/sabnzbd/sabnzbd/commit/5c77f9d9b3d759120edc6b0cb652312096d9fcde
> not sure if this the reason of the rounding error/difference from glitter vs actual?

previous one had an issue, fixed and forced push.. but the pull request never updated only my repo.. not sure wtf is going on with that.. so just doing a new pull.
easier to review non-whitespace changes: https://github.com/sabnzbd/sabnzbd/pull/847/files?w=1

### DIFF
--- a/sabnzbd/config.py
+++ b/sabnzbd/config.py
@@ -93,6 +93,7 @@ class Option(object):
 
     def set_dict(self, dict):
         """ Set value based on dictionary """
+        # TODO: Review if `dict` is actually correct or typo
         if self.__protect:
             return False
         try:
@@ -377,7 +378,7 @@ class ConfigServer(object):
         self.password = OptionPassword(name, 'password', '', add=False)
         self.connections = OptionNumber(name, 'connections', 1, 0, 100, add=False)
         self.ssl = OptionBool(name, 'ssl', False, add=False)
-        self.ssl_verify = OptionNumber(name, 'ssl_verify', 2, add=False) # 0=No, 1=Normal, 2=Strict (hostname verification)
+        self.ssl_verify = OptionNumber(name, 'ssl_verify', 2, add=False)  # 0=No, 1=Normal, 2=Strict (hostname verification)
         self.enable = OptionBool(name, 'enable', True, add=False)
         self.optional = OptionBool(name, 'optional', False, add=False)
         self.retention = OptionNumber(name, 'retention', add=False)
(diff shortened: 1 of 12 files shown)


sabnzbd/nzbqueue.py
```diff
@@ -441,6 +441,7 @@ def remove(self, nzo_id, add_to_history=True, save=True, cleanup=True, keep_basi
             nzo = self.__nzo_table.pop(nzo_id)
             nzo.deleted = True
             if cleanup and nzo.status not in (Status.COMPLETED, Status.FAILED):
+                # TODO: Use nzo.is_gone() ?
```
thezoggy: @Safihre pretty sure we should just use is_gone here, added note for ya to review later on..

sabnzbd/config.py
```diff
@@ -93,6 +93,7 @@ def get_dict(self, safe=False):
 
     def set_dict(self, dict):
         """ Set value based on dictionary """
+        # TODO: Review if `dict` is actually correct or typo
```
Safihre: Why do you think this is wrong? It seems to be a prototype to be overwritten by the inherited classes that use it to easily submit config settings.

sabnzbd/nzbqueue.py
Safihre: Yes, indeed. You can change it to that.

sabnzbd/config.py
thezoggy: dict is a reserved word, and were passing an object to overwrite another object. so it just screams at me in my ide about it.. why i put a note to review :)

thezoggy:
will make the changes and force push update here in a day..

thezoggy:
rebased, updated, and force pushed.

Safihre:
Merged! But removed all the `@UnresolvedImport` and `@IgnorePep8`.

thezoggy:
@Safihre why not include them?

PR #850, opened 2017-03-09
hide password in General Config via use of HTML element type "password"

cdanderson:
Changed form text field to HTML type “password” for web interface password in General Config section. Password is currently viewable as it’s by default set to type “text.”


Safihre:
Hmm I understand what you are trying to do, but this change was made on purpose.
Because we got numerous reports of users where overly active browsers would detect this username/password combi and start auto-filling it in the Server password pages.

Also, the `input` now has both `type="text"` and `type="password"`...

cdanderson:
Alright, I will make the change locally for myself.

Issue #848, opened 2017-03-08
FAT: pro-active and re-active detection & reporting


sanderjo:
A SABnzbd Complete directory on a FAT file system, can cause problems: if the download is bigger than 4GB (which does happens with MKV-files), SAB / unrar will discover & reports problems at the **end** of the download process. If we can detect and report that ealier & more clearly, that would be useful for users.

So my suggestion is:
1)  pro-active detection of FAT: at startup of SAB, and at setting another Complete directory
2) re-active detection in the unrar output of FAT-caused problems

The above is not needed for the Temp/Incomplete directory, as rar and par2 files will never be 4GB
The above is also needed for Category folders.

Such problems will not occur with other, more modern file systems, like NTFS, ext3, ext4, etc

If we agree on this (especially @safihre), I can write the code.

Hellowlol:
Why on earth would anybody use fat? The only reason I can come up with is sd cards. So the next question is why would anybody save to a sd card :p

sanderjo:
> Why on earth would anybody use fat?

If you buy a 2TB USB-disk, it's FAT, isn't?

thezoggy:
sadly fat32 is still used by a lot of people due to it being so open to write to..  for example amazon's firetv supports fat32 for external harddrives.. which people use and end up having to rarsplit movie files just to get them to fit and load in kodi.

anyways, we cant really do much on the remote file system.. but locally we deff should detect if FAT12/16/32 is being used for the incomplete/complete folders and warn against it..   

another battle is case sensitive file systems.. but i'll leave that conversation for another day

Safihre:
It should be at the start of the post processor I think. So at the start of its `run()` method. This way it's run asynchronously during startup. 

Can you also integrate a check if all the output folders are writeable? :+1:

thezoggy:
sounds like we really should just make a `health` tab in sab config.. where we can do all these sort of warnings / optimization / tests.. and put the pystone/disk speed/network tests/setup tests at

PR #859, opened 2017-03-14
Check if Completed Download Folder is on FAT. If so, give a warning

sanderjo:
As discussed here: https://github.com/sabnzbd/sabnzbd/issues/848

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -100,6 +100,7 @@ from sabnzbd.encoding import unicoder, deunicode
 import sabnzbd.notifier as notifier
 import sabnzbd.zconfig
 import sabnzbd.utils.sslinfo
+import sabnzbd.utils.checkdir
 
 from threading import Thread
 
@@ -1253,6 +1254,13 @@ def main():
                 sabnzbd.LINUXTRAY = sabnzbd.sabtraylinux.StatusIcon()
             except:
                 logging.info("pygtk2 not found. No SysTray.")
+
+    complete_dir = sabnzbd.cfg.complete_dir.get_path()
+    if sabnzbd.utils.checkdir.isFAT(complete_dir):
+        logging.warning(T('Completed Download Folder %s is on FAT, limiting maximum file size to 4GB') % complete_dir)
+    else:
+        logging.info("OK: Completed Download Folder %s is not on FAT", complete_dir)
+
 
     # Find external programs
     sabnzbd.newsunpack.find_programs(sabnzbd.DIR_PROG)
--- a/sabnzbd/utils/checkdir.py
+++ b/sabnzbd/utils/checkdir.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+# Functions to do checking on a directory:
+# is directory on a FAT filesystem? 
+
+import sys, os, subprocess
+
+debug = False
+
+def isFAT(dir):
+
+# Check if "dir" is on FAT. FAT considered harmful (for big files)
+# Works for Linux, Windows
+# NB: On Windows, full path with drive letter is needed!
+
+    FAT = False    # default: not FAT
+    if 'linux' in sys.platform:
+        # On Linux:
+        # df -T /home/sander/weg
+
+        '''
+        Example output of a 500GB external USB drive formatted with FAT:
+        $ df -T /media/sander/INTENSO
+        Filesystem     Type 1K-blocks      Used Available Use% Mounted on
+        /dev/sda1      vfat 488263616 163545248 324718368  34% /media/sander/INTENSO
+        '''
+
+        cmd = "df -T " + dir + " 2>&1"
+        for thisline in os.popen(cmd).readlines():
+            #print thisline
+            if thisline.find('/')==0:
+                # Starts with /, so a real, local device
+                fstype = thisline.split()[1]
+                if debug: print "File system type:", fstype
+                if fstype.lower().find('fat') >= 0:
+                    FAT = True
+                    if debug: print "FAT found"
+                    break
+    elif 'win32' in sys.platform:
+        dir = dir.upper()    # input could be in lower case like "e:\", so to upper first
+        # On Windows:
+        '''
+        C:\>wmic logicaldisk get name,filesystem
+        FileSystem  Name
+        NTFS        C:
+                D:
+        FAT         E:
+        FAT         F:
+        NTFS        G:
+        NTFS        H:
+        NTFS        M:
+        NTFS        Q:
+        NTFS        X:
+
+        We now can see dir "E:\other\dir\something" is on FAT (just like F:\bla), 
+        and "Q:\dome\dir\bla" is not on FAT.
+        '''
+        windowscmd = "wmic logicaldisk get name,filesystem".split()
+        CREATE_NO_WINDOW = 0x08000000
+        for line in subprocess.check_output(windowscmd,creationflags=CREATE_NO_WINDOW).replace('\r','').split('\n'):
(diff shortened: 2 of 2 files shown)


Safihre:
Merged 👍

thezoggy:
could have used win32api GetVolumeInformation
http://docs.activestate.com/activepython/2.4/pywin32/win32api__GetVolumeInformation_meth.html

```
D:\temp>python -c "import win32api;print win32api.GetVolumeInformation('D:\\')"
('storage', 107913418, 255, 65470719, 'NTFS')
D:\temp>python -c "import win32api;print win32api.GetVolumeInformation('E:\\')"
('UUI', 452793712, 255, 6, 'FAT32')
```

sanderjo:
@thezoggy OK, interesting. Does it also work for (I have no Windows at hand, so I can't check myself):

win32api.GetVolumeInformation('E:\\somedir\\')
win32api.GetVolumeInformation(''\\\\?\\C:\\Media\\'')

Because 1) we have the full path and 2) it could be the format '\\?\C:\Media\...' ... So we can't be sure the first character is the drive letter. :-(

Safihre:
@sanderjo 
It needs to only be the disk-letter and it's fine with the extra `\\?\`:
```
>>> win32api.GetVolumeInformation('\\\\?\\E:\\')
('Files', -1531830029, 255, 65470719, 'NTFS')
```
But luckily python has a function for that:
```
>>> os.path.splitdrive('\\\\?\\E:\\Downloads\\')
('\\\\?\\E:', '\\Downloads\\')
```

Doesn't work for sub-folders (the `Downloads` definitely exists!):
```
>>> win32api.GetVolumeInformation('E:\\Downloads\\')
pywintypes.error: (144, 'GetVolumeInformation', 'The directory is not a subdirectory of the root directory.')
```

It also works for mapped network drives:
```
>>> win32api.GetVolumeInformation('Z:\\')
('PromisePegasus', 0, 255, 262278, 'NTFS')
```

But not for network-paths:
```
>>> win32api.GetVolumeInformation('\\\\192.168.37.22\\')
pywintypes.error: (161, 'GetVolumeInformation', 'The specified path is invalid.')
```

Safihre:
It's also really fast:
```
>>> a=time.time(); win32api.GetVolumeInformation('\\\\?\\E:\\'); print 1000*(time.time()-a)
('Files', -1531830029, 255, 65470719, 'NTFS')
3.99994850159
```
Compared to the ~400ms of the current method 👍

thezoggy:
1 loop, 1 run, see it constantly get said info in ~3 msec

```C:\>python -m timeit -n1 -r1 "import win32api;print win32api.GetVolumeInformation('C:\\')"
('', 807141970, 255, 65470719, 'NTFS')
1 loops, best of 1: 2.65 msec per loop```

**Comment by sanderjo** [comment]
@thezoggy If you want, can you create & send a PR? 

(I haven't got a Windows PC with admin rights nor win32api nor a sab-from-source environment)

**Comment by thezoggy** [comment]
note that this code falsely warns about exFAT, which does not have the 4GB file size limit...
`File size limit of 16 EiB − 1 byte`

```
C:\Program Files (x86)\sabnzbd_git\sabnzbd\utils>python checkdir.py E:\
E:\ is on FAT
```

```
C:\>python -c "import win32api;print win32api.GetVolumeInformation('E:\\')"
('ps4', 346629577, 255, 6, 'exFAT')
```

so yes we deff need to update the code and optimize

**Comment by thezoggy** [comment]
could do it in one line..
`FAT = win32api.GetVolumeInformation(os.path.splitdrive(drive)[0])[4].startswith("FAT")`
but having it cleanup the `\\?\` requires an additional line.. then added some debug lines and comment..

**Comment by Safihre** [comment]
@thezoggy You don't need to remove the `\\?\`, see my post above!

**Comment by thezoggy** [comment]
depending on the passed dir.. I had to at certain times (lack of trailing slash)..

```
C:\>python -c "import win32api;print win32api.GetVolumeInformation('\\?\E:')"
  File "<string>", line 1, in <module>
pywintypes.error: (123, 'GetVolumeInformation', 'The filename, directory name, or volume label syntax is incorrect.')

C:\>python -c "import win32api;print win32api.GetVolumeInformation('\\\\?\\E:\\')"
('ps4', 346629577, 255, 6, 'exFAT')
```

**Comment by thezoggy** [comment]
looks like we only check the completed folder..  why not the incomplete folder as well?
`2017-03-17 08:56:37,959::INFO::[postproc:174] Completed Download Folder C:\test\downloads\complete is not on FAT`

**Comment by sanderjo** [comment]
@thezoggy would this help to get from both Windows path notations to a drive letter:

`path.replace('\\','').replace('?','')[:2]`

Check:

```
>>> path = '\\\\?\\E:\\and\\more\\bla\\'
>>> path.replace('\\','').replace('?','')[:2]
'E:'
```

```
>>> path = 'E:\\and\\more\\bla\\'
>>> path.replace('\\','').replace('?','')[:2]
'E:'
```

So works for both cases.

**Comment by sanderjo** [comment]
> looks like we only check the completed folder.. why not the incomplete folder as well?

Because in **Incomplete** you will get typical file sizes of a 50 or 100 MB max. So well below the 4GB limit.

Our approach:
- first introduce and release this check for Complete
- if not problems: introduce it for Category destination folders.

**Comment by sanderjo** [comment]
Triggered by @thezoggy 's remark about exFAT, I verified for Linux, and nothing extra is needed in checkdir.py for exFAT on Linux:

Proof:
```
sander@Streamer13:~/git/sab-sanderjo-16maart2017$ python sabnzbd/utils/checkdir.py /media/sander/mijnExFat/somedir/
/media/sander/mijnExFat/somedir/ is not on FAT
```
Reason:
```
/dev/mmcblk1p1 on /media/sander/mijnExFat type fuseblk (rw,nosuid,nodev,relatime,user_id=0,group_id=0,default_permissions,allow_other,blksize=4096,uhelper=udisks2)
```
exFAT is mounted via fuse, so the filesystem is fuseblk. No "FAT" in that filesystem.

So ... everything is fine.

**Comment by sanderjo** [comment]
Hmmm, new Ubuntu (20.04) installed, which has Linux kernel 5.4, and suddenly SAB says
```
Completed Download Folder /media/sander/6665-6464/complete is on FAT file system, limiting maximum file size to 4GB
```
... for the same SD card has been not given that warning on Ubuntu 19.04.
Let's see:
```
$ df -T /media/sander/6665-6464/
Filesystem     Type  1K-blocks     Used Available Use% Mounted on
/dev/sdc1      exfat 124835840 82152448  42683392  66% /media/sander/6665-6464
```
Ouch, pure `exfat` and not mounted via fuse anymore.
I'll have to update my code. PR on it's way.

**Comment by sanderjo** [comment]
Windows was already OK:

```
c:\Users\sande\git\sjo-sab-RSS-fix>python
Python 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)] on win32
>>> from sabnzbd.utils.checkdir import isFAT
>>> isFAT("/home/sander/")
False
>>> isFAT("c:\\")
False
>>> isFAT("D:\\")
False
```

Checking internals:

```
>>> import win32api
>>> win32api.GetVolumeInformation()
TypeError: GetVolumeInformation() takes exactly 1 argument (0 given)
>>> win32api.GetVolumeInformation("D:\\")
('', 1717920868, 255, 131590, 'exFAT')
>>>
```
and isFAT() checks like this: `if result[4].startswith("FAT"):`, so OK.

PR #866, opened 2017-03-17
Update windows section to use win32api

thezoggy:


thezoggy:
note that the `\\?\` cleanup was done like we do it here: https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/misc.py#L1511

PR #868, opened 2017-03-18
listquote removal

thezoggy:
We only used 1 line from listquote, just copied over that line to eliminate the need for the superfluously import.


Safihre:
Don't want to also delete the actual listquote.py?

thezoggy:
We do

Issue #877, opened 2017-03-30
Consuming all Memory before auto-killed


astro7467:
Seen on RC 1 also. Not seen on beta. 
Tested on Ubuntu MATE 16.04.2 with 8GB, 16GB and 32GB of RAM. 8GB swap on all CONFIGS. 

[  840.726098] Out of memory: Kill process 2338 (sabnzbdplus) score 974 or sacrifice child
[  840.726148] Killed process 2338 (sabnzbdplus) total-vm:44634208kB, anon-rss:31951492kB, file-rss:0kB, shmem-rss:0kB
[  841.497082] oom_reaper: reaped process 2338 (sabnzbdplus), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB

Changing top-only = 1 then start_paused = 1 make no difference. 

Queue currently has approximately 396 items (via incomplete directory) - has been higher with previous versions. 

Using jcfp PPA. 

* will check back in about 16hrs to see if any specific data or actions requested. 

Thanks

Safihre:
Are you using SABYenc? And do you have some logs maybe?
If not: could you switch logging to `+Debug` in the Status Window and share the logs? 
If you use the `Show logging` it will anonymize your log.

astro7467:
Yes to SABYenc 
`python-sabyenc/xenial,now 2.8.0-0ubuntu1~jcfp1~xenial amd64 [installed])`
Unable to get to web interface.
Have done the following;
for `/etc/default/sabnzbdplus`:
`EXTRAOPTS="--log-all"`
for `sabnzbd.ini`:
`[logging]`
`log_level = 2`
While monitoring the log for a few restarts, I did see different results before the memory timed out, until I realised during one run it was building and passing an argument with every single password from my password file (which is approx 683MB) - I assumed it was just passing the file as a parameter plus forgotten I had grab this file before to try out when I was running into a lot of passworded file issues
Setting was active for Abort if Passworded - so not sure why it was doing this. I have changed it to Pause if Passworded - which still causes a spike up to about 22GB and the web interface to be come unresponsive - but eventually settles down.
I have subsequently changed `sabnzbd.ini` to point to a different file with just a couple of lines.
Will monitor over the next days to see if there is another trigger as the logs did not produce the same result everytime - but this seems to be a valid candidate and probably an edge case.

[sabnzbd.log.zip](https://github.com/sabnzbd/sabnzbd/files/884278/sabnzbd.log.zip)

[...]

PR #879, opened 2017-03-31
get_all_passwords_as_generator

Hellowlol:
This is untested but what about something like this?

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -19,6 +19,7 @@
 sabnzbd.assembler - threaded assembly/decoding of files
 """
 
+from itertools import islice
 import os
 import Queue
 import logging
@@ -326,14 +327,15 @@ def check_encrypted_and_unwanted_files(nzo, filepath):
                     if is_cloaked(filepath, zf.namelist()):
                         nzo.encrypted = 1
                         encrypted = True
-                    elif not sabnzbd.HAVE_CRYPTOGRAPHY and not passwords:
-                        # if no cryptography installed, only error when no password was set
+
+                    elif not sabnzbd.HAVE_CRYPTOGRAPHY and any(slice(passwords, 2)) is False:
+                        # if cryptography isn't installed, only emit log when no passwords exists.
                         logging.info(T('%s missing'), 'Python Cryptography')
                         nzo.encrypted = 1
                         encrypted = True
 
                     elif sabnzbd.HAVE_CRYPTOGRAPHY:
-                        # Lets test if any of the password work
+                        # Lets test if any of the passwords work
                         password_hit = False
 
                         for password in passwords:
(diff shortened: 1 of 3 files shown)


Hellowlol:
https://github.com/sabnzbd/sabnzbd/issues/877

Safihre:
Maybe test it? :P

Hellowlol:
I wouldn't find a nzb that was encrypted... lol

Hellowlol:
So i have downloaded over 50 random nzbs still no, password protected. I tried to upload one myself, but it does not work, dunno why it dont work. Does anybody have link to a password protected nzb so i can test it im happy to do it.

Safihre:
Here: [Test123{{nZb-by-USC}}.txt](https://github.com/sabnzbd/sabnzbd/files/888276/Test123.nZb-by-USC.txt) (have to rename the extension to `.nzb`)

Hellowlol:
Thanks! I just tested it and it seemed to work just nicely. Ran the same file on develop and it crashed after 10 sec on a memory error (250ish mb password file.)

sabnzbd/assembler.py
```diff
@@ -326,14 +326,14 @@ def check_encrypted_and_unwanted_files(nzo, filepath):
                     if is_cloaked(filepath, zf.namelist()):
                         nzo.encrypted = 1
                         encrypted = True
-                    elif not sabnzbd.HAVE_CRYPTOGRAPHY and not passwords:
```
Hellowlol: I removed this. I dont understand why this is needed. I dont understand the comment it makes no sense to me.

thezoggy: if you dont understand it.. why remove it.

Safihre:
@Hellowlol: thanks for taking the time. I will merge this after 2.0.0-Final so it can be in 2.0.1, but for now we will first probably have an RC3 due to some unforseen bugs. 
The implementation is indeed how I also would have done it and I will take a look again at the Cryptography warning, right now I also don't understand anymore why I did that but I remember it was for a good reason :P

Hellowlol:
@Safihre Thanks. No worries, i dont think the password file feature is used by many, its just simpler to download a new nzb imo. I squashed the pr.

sabnzbd/assembler.py
```diff
@@ -326,14 +327,15 @@ def check_encrypted_and_unwanted_files(nzo, filepath):
...
-                        # if no cryptography installed, only error when no password was set
+
+                    elif not sabnzbd.HAVE_CRYPTOGRAPHY and any(slice(passwords, 2)) is False:
```
Hellowlol: This should be isslice not slice.

Safihre:
This one is a bit messy, I do understand the beauty of a Python generator but this approach does not seem to really increase readability of our code.
I think I would rather implement a warning message if more than 50 passwords are detected. This will inform users that think unrar can handle such things easily.
What do you think?

Hellowlol:
Honestly I think it simpler then the code we have today. I still think this should be a generator it's up to you. I skip all password protected nzbs so I will never need this feature anyway. I just did it since it was a easy fix and I wanted contribution something to sab. If this isn't accepted I'm gonna find some typos to fix :D

PR #888, opened 2017-04-06
SABYenc: Reporting of Required and Found version. Plus better var naming

sanderjo:
Example Warning:
"SABYenc disabled: no correct version found! (Found v2.8.0, expecting v3.0.1)"


thezoggy:
may want to add yenc/sabyenc + version to the config root page?

PR #890, opened 2017-04-07
SABYenc: explicitly define "SABYENC_VERSION = None" if none installed

sanderjo:
Solves https://forums.sabnzbd.org/viewtopic.php?f=3&t=22567


Safihre:
I saw that but mistakenly thought that would not happen..

PR #895, opened 2017-04-11
Update fr.po

RalfvdEnden:
Fix %s mismatch error when running tools/make_mo.py


Safihre:
Doesn't work like this.
The translations come from launchpad (https://translations.launchpad.net/sabnzbd) and thus will overwrite any changes made in the repo.
On Launchpad this is already corrected, but not going to release a whole new SABnzbd version just for 2 French texts :P

Issue #910, opened 2017-05-01
URL grabber unhandled exception


ppslim:
I am am on Ubuntu 12.4 with SAB 2.0.0 Beta 1 from the PPA. Unsure if this is the latest, but I am aware my boxen is about 5 weeks since last update.

A NZB URL was inserted by Sickbeard and SAB crashed on downloading it. It's not stuck in a "Trying to fetch" state. More importantly, the URL grabber crashed error occurred.

```
Traceback (most recent call last):
  File "/usr/share/sabnzbdplus/sabnzbd/urlgrabber.py", line 215, in run
    data = fn.read()
  File "/usr/lib/python2.7/socket.py", line 351, in read
    data = self._sock.recv(rbufsize)
  File "/usr/lib/python2.7/httplib.py", line 598, in read
    s = self.fp.read(amt)
  File "/usr/lib/python2.7/socket.py", line 380, in read
    data = self._sock.recv(left)
error: [Errno 104] Connection reset by peer
```

Reviewing the file object in use in the grabber, I note it is as `socket._fileobject` coming from a `urllib2.HTTPResponse`. `urllib2` is unsafe for partially returned data, as the file object it constructs has already read the headers, which is enough to help the library assume it is working. If the connection collapses at any point after headers, exceptions should be expected.

As the error is coming out of `socket`, this needs handling rather than just the current `IncompleteRead` handler in urlgrabber.

Pull request amends this handling to capture IOError as well, which is a parent of `socket.error` in Python 2.6 and above. It also drops the exception captured in `e` as this is presently unused.

PR #911, opened 2017-05-01
URL Grabber crashes on unhandled socket exceptions

ppslim:
URL grabber handles `httplib.IncompleteRequest` exceptions.
This exception is not returned when the HTTP responses closes in the body, rather than the headers.

Handles problem as described un #910


sanderjo:
Hi @ppslim,

Thanks for your PR.
FYI: SAB maintainer @Safihre will not be close to his computer for one or two weeks, so please expect an answer after that.

Safihre:
Thanks!

PR #918, opened 2017-05-12
Fix bug with DateJS

### LINKED ISSUE #916: Calculation error when pausing for "13 hours"
Gets converted to 4320 mins instead of 780 mins.

ali-l:
Fix DateJS bug where sometimes >12 minutes/hours/days converts into the wrong timestamp. Removing the space makes DateJS work properly.

Fixes #916

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.main.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.main.js
@@ -471,6 +471,12 @@ function ViewModel() {
             return;
         }
 
+        // Fix DateJS bug where sometimes >12 minutes/hours/days (ex. "13 hours") converts into the wrong timestamp 
+        // Removing the space makes DateJS work properly
+        newValue = newValue.replace(/\s*hour|\s*minute|\s*min|\s*day/g, function(match) {
+            return match.trim()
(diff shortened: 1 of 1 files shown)


interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -471,6 +471,12 @@ function ViewModel() {
...
         }
 
+        // Fix DateJS bug where sometimes >12 minutes/hours/days (ex. "13 hours") converts into the wrong timestamp 
+        // Removing the space makes DateJS work properly
+        newValue = newValue.replace(/\s*hour|\s*minute|\s*day/g, function(match) {
+            return match.trim()
+        });
+      
         // Parse
         var pauseParsed = Date.parse(newValue);
```
thezoggy: or you could just do
`var pauseParsed = Date.parse(newValue.replace(/\s/g,''))`

ali-l: Just tested a few cases with this strategy. Turns out it doesn't work for the 'tomorrow 7pm' example. The space needs to be preserved for that one at least.

PR #923, opened 2017-05-17
panic - fixed ports

thezoggy:
Panic when attempting to bind to host:port when using fixed ports and its used.
This way user gets a webpage with the info and console shows the message vs the user having to try and guess what the traceback + IOError means.


Safihre:
Implemented a bit more general fix that keeps everything nicely in `panic.py`.

Issue #902, opened 2017-04-18
How to handle the new par2cmdline version


Safihre:
The new par2cmdline 0.7.0 is out. This one now also includes the OpenMP code, so it is also multi-threaded.

Now how are we going to incorporate this into SABnzbd? par2-tbb is still faster, so for the Windows/macOS releases it doesn't seem logical to completly switch to par2cmdline.
On Linux platforms we already detect if there is a multi-threaded par2 installed which we can use to our advantage.

So we could really make the `Multicore par2` switch force to only use 1 thread by adding the `-t` parameter to both par2cmdline (if multi-threaded is detected) and to par2-tbb.
In that case I would opt to only have a Special setting called `force_par2cmdline` and not offer this option to the user to not use par2-tbb on Windows/macOS?

thezoggy:
par2-tbb on windows uses an old par2cmdline (0.4).. which means its behind the times in features and bugs that have been fixed still exist on it...  now it would awesome if there was a new par2-tbb build so we get the best of both worlds...  i know you ran tests before and found par2-tbb significantly faster than par2cmdline 0.6.13...  can you run them again against 0.7.0 ?
https://github.com/Parchive/par2cmdline/issues/27#issuecomment-170940850

Safihre:
Good point. 
Test system: Core i5-4200 @ 2.30Ghz (2 cores, 4 threads), 12GB memory and all data on a fast SSD.

I used a dataset of 2GB with 42 missing articles throughout the download, what usually happens on usenet and ran the test twice for each program:

**par2cmdline 0.6.13 (only x86):**
1: `TotalSeconds      : 87.6838234`
2:  `TotalSeconds      : 86.2515236`

**par2cmdline 0.7.0 (x64):**
1: `TotalSeconds      : 61.612304`
2:  `TotalSeconds      : 60.4033289`

**par2tbb (x64):**
1: `TotalSeconds      : 49.0121549`
2:  `TotalSeconds      : 51.4033289`

**That's still pretty substantial.. But:**

### **Hypothesis:** par2cmdline 0.7.0 doesn't use multithreading to do verification, so the biggest difference is there.
Let's test verification only:

**par2cmdline 0.6.13 (only x86):**
1: `TotalSeconds      : 20.8311689`
2:  `TotalSeconds      : 20.9405503`

**par2cmdline 0.7.0 (x64):**
1: `TotalSeconds      : 19.0471533`
2:  `TotalSeconds      : 19.0547957`

[...]

PR #926, opened 2017-05-21
Multipar integration for Windows

Safihre:
In the recent tests I found [that it's so much faster](https://github.com/sabnzbd/sabnzbd/issues/902#issuecomment-295238757) (~2x) than the other implementations that it was worth finishing my implementation.

**Windows-only**

The compiled SAB.exe of this version is available here:
https://github.com/sabnzbd/sabbuild/releases/tag/2.1.0-multipar-test2


thezoggy:
this include multipar 1.2.9.9 ?

Safihre:
Yes :+1:

PR #936, opened 2017-06-03
Implement __slots__ to conserve memory

Safihre:
Objects such as `Article()` get created a lot. By using the `__slots__` property, Python will only reserve space for the given keywords instead of a whole dictionary. 
Testing showed that 1 job now takes between 2-3MB of memory (on WinX64)  when loaded, compared to 4MB before.

The problem: it requires people to do a Queue Repair because the old pickles are completly incompatible with the new objects. I tried to implement a conversion strategy, but due to the nesting of objects this was impossible.
I was thinking of putting it into the next 2.1.0 RC, or should we wait for 2.2.0?


thezoggy:
in the past, when we change the queue format it was a major version change.. so 0.5->0.6, then 0.6->0.7, then 0.7->0.8->1.0. so per that pattern it should be 3.0...  but now that we are at x.x.x format.. maybe 2.2.0 would be okay? @shypike

labrys:
@thezoggy if you're following semantic versioning, then a breaking change like that would generally still be a major release, a minor release would be for additional (non-breaking) functionality (e.g. adding a new api endpoint).

labrys:
Also as an fyi, in py3.6 dicts are a lot more compact (approx 25%).

Safihre:
I tried for another hour to convert the pickles, but it's just such a mess of code that a repair is safer.

Why can we not just trigger an automatic queue repair? 
All that would change would be the order of the downloads.

Safihre:
Answer: no, because it will also add those orphaned jobs.

But I came up with another way where we only repair the jobs in the Queue and in the order they were in the Queue. There is minimal data loss, for example when a file was half-downloaded that half part will be re-downloaded.

Safihre:
Correction: I cannot save the jobs that were still grabbing, so have to remove those..

labrys:
@Safihre Can you simply make two separate classes, one like the traditional, the second with __slots__?  If the unpickled class has slots then use it, if not then continue using the original class.  Should make it safe for current queue, but new items would be added with the new __slots__ based classes when pickled.

Safihre:
@labrys You are right, I can do that!
Although since the change was to the base-class `TryList` it would also require making 2 version of the 3 children classes, that's a bit too much.
But what I can do is only use the new class when the queue is empty on startup. Theoretically this could mean that some users never get upgraded if they never start SABnzbd with an empty queue, but it's much better than the current conversion logic that I was trying.

labrys:
@Safihre if the base classes all use __slots__, you should be able to implement the dict versions like so to minimize class duplication:
```python
class Dictful(Dictless):
    pass
```
This works since any subclass of slot-based classes that does not define `__slots__` adds a `__dict__` to the class.  To prevent this you have to specifically define `__slots__` to avoid promoting existing (with `__slots = ()` if no attributes change) to prevent dict-promotion.

The big question is if the new `Dictful` pickle is compatible with the original pickled objects.

Safihre:
The problem is not the `__dict__`, it's the switch from an 'old-style' class to a 'new-style' one. 
```
class TryList:
```
To:
```
class TryList(object):
```
Without the `object` as base, the `__slots__` doesn't do anything. But this causes Python to import the pickles differently, because it forces the call to `__init__` which it doesn't do on old classes.

labrys:
Ok, that makes sense, I didn't realize you were using old-style classes and I wouldn't have expected a difference pickling between `__slots__` and `__dict__`.  There are some work-arounds to make this possible if you're interested, but it may not be worth the effort.

Safihre:
I know the method with temporarily overwriting class definitions, the problem there is that there are nested  classes in each object with the same problem. So it would have to be a recursive operation.
Possible yes, messy, also.

labrys:
Actually you could implement a __new__ in the base class that attempts to create an instance of the primary class first but falls back to the secondary (old-style).  Then have a class-attribute defining what the secondary class is.  The only difference for subclasses would be to define what the secondary class is.

Safihre:
Yes, I tried exactly this. But it would mean having to duplicate the class and it's methods, because there is no way for the new-style class to inherit the methods from the old-style class without getting the `__dict__` added. If that get's added the whole memory gain that we get from `__slots__` is gone again.

This is the example code how we use it:

```python
import pprint

class TryList(object):
    def __new__(cls):
        if new_cls:
            return TryListNew
        else:
            return TryListOld


class TryListOld:
    def hello(self):
        print hello


class TryListNew(object):
    __slots__ = ('new','old','hello')
    new=True


class NZBFile(TryList):
    pass


new_cls = False
blaat1 = NZBFile()

pprint.pprint(dir(blaat1))


new_cls = True
blaat2 = NZBFile()


pprint.pprint(dir(blaat2))

```

labrys:
Have you tried deleting `__dict__` upon new-style class creation?

Safihre:
Yes, that works. But I am more concerned that we are adding all this overhead during the creation of new objects (these are created often) for the whole future of SABnzbd.
While the 1 time conversion where only URL's waiting to be grabbed are lost, would only be done once per installation. The loss of these URL's we would mention nice and big in the changelog.

labrys:
Yeah and there are other ways too, but to be honest, they probably aren't worth the trouble.  I think the safest bet personally, if you wanted to support a transitional period, would be to have a deprecation period where both classes exist (sure its duplication but only temporary) with a deprecation warning (probably accompanied with a UI warning) and then remove the old style completely in a later version.

labrys:
Two more questions for you.

Which protocol are you using for pickling?

And about this statement:
> If that get's added the whole memory gain that we get from __slots__ is gone again.
Have you checked sizes of a class with both __slots__ and __dict__?  An unused dict by itself should only run approx 300 bytes, so even if the class has a __dict__, if you don't actually store any mutable attributes in it, it should not increase in size beyond that.

sabnzbd/newsunpack.py
```diff
@@ -1302,7 +1302,7 @@ def PAR_Verify(parfile, parfile_nzf, nzo, setname, joinables, classic=False, sin
                 continue
             extra_par2_name = None
 
-            if line.startswith('Invalid option specified') or line.startswith('Cannot specify recovery file count'):
+            if line.startswith('Invalid option specified') or line.startswith('Invalid thread option') or line.startswith('Cannot specify recovery file count'):
```
labrys:
```python
if line.startswith(['Invalid option specified', 'Invalid thread option', 'Cannot specify recovery file count']):
```

Safihre: Cool, didn't know that it allows that.

Safihre:
Going to merge this in and see how the conversion and see if problems arise.. In my tests it works fine every time.

PR #939, opened 2017-06-04
Added support for systemd power controls

BrendanBall:
added systemd support to powersup.linux_shutdown, linux_hibernate, linux_standby.

### DIFF
--- a/sabnzbd/powersup.py
+++ b/sabnzbd/powersup.py
@@ -116,6 +116,10 @@ except ImportError:
     HAVE_DBUS = False
 
 
+_IS_NOT_INTERACTIVE = False
+_LOGIND_SUCCESSFUL_RESULT = 'yes'
+
+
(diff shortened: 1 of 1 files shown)


Safihre:
Trusting @sanderjo with this one to test! 
Might not merge straight away, because we just did a Release Candidate for 2.1.0, so this will go in the next release  :+1:

thezoggy:
@sanderjo sickrage has been around for a bit and used by a bunch of people, so i would suspect their scripts are safe to reference off of: https://github.com/SickRage/SickRage/tree/master/runscripts

Safihre:
@thezoggy That's not what this is for. Those are init scripts, this PR is about the power-options of SABnzbd being called from python. To restart/shutdown/etc the whole PC from SABnzbd.

thezoggy:
yeah just saw what it is, saw systemd and asumed

PR #943, opened 2017-06-10
Typo

cheese1:
hopefully this one is better
i'm currently starting using git :)


thezoggy:
should squash and force push.. so theres the only one commit.. so we keep history clean

Safihre:
I realized that changing it like that would break all existing translations in other languages (since they link to the original English text), so I used our overwrite translations file for English `en.po` for these kind of fixes.
Thanks for reporting! 👍

PR #959, opened 2017-07-03
Move To Top/Move To Bottom buttons

gwyden:
Added a move to top and move to bottom control for the files inside individual nzbs

### DIFF
--- a/interfaces/Glitter/templates/include_overlays.tmpl
+++ b/interfaces/Glitter/templates/include_overlays.tmpl
@@ -528,11 +528,11 @@
                                                 <span>
                                                     <span data-bind="truncatedTextCenter: filename"></span>
                                                     <small>(<span data-bind="text: file_age"></span> - <span data-bind="text: mb"></span> MB)</small>
-													<a href="#" class="buttonMoveToTop" title="Move To Top"><span class="glyphicon glyphicon-chevron-up"></span></a>
-													<a href="#" class="buttonMoveToBottom" title="Move To Bottom"><span class="glyphicon glyphicon-chevron-down"></span></a>
-												</span>
-											</div>
-										</div>
+                                                    <a href="#" class="buttonMoveToTop" title="Move To Top"><span class="glyphicon glyphicon-chevron-up"></span></a>
+                                                    <a href="#" class="buttonMoveToBottom" title="Move To Bottom"><span class="glyphicon glyphicon-chevron-down"></span></a>
+                                                </span>
+                                            </div>
+                                        </div>
                                     </td>
                                 </tr>
                             </tbody>
--- a/interfaces/Glitter/templates/include_overlays.tmpl
+++ b/interfaces/Glitter/templates/include_overlays.tmpl
@@ -528,8 +528,8 @@
                                                 <div class="fileDetails">
                                                     <span data-bind="truncatedTextCenter: filename"></span>
                                                     <div class="fileControls">
-                                                        <a href="#" class="hover-button buttonMoveToTop" title="$T('Glitter-MoveToTop')"><span class="glyphicon glyphicon-chevron-up"></span></a>
-                                                        <a href="#" class="hover-button buttonMoveToBottom" title="$T('Glitter-MoveToBottom')"><span class="glyphicon glyphicon-chevron-down"></span></a>
+                                                        <a href="#" data-bind="click: \$parent.filelist.moveButton" class="hover-button buttonMoveToTop" title="$T('Glitter-MoveToTop')"><span class="glyphicon glyphicon-chevron-up"></span></a>
+                                                        <a href="#" data-bind="click: \$parent.filelist.moveButton" class="hover-button buttonMoveToBottom" title="$T('Glitter-MoveToBottom')"><span class="glyphicon glyphicon-chevron-down"></span></a>
                                                     </div>
                                                     <small>(<span data-bind="text: file_age"></span> - <span data-bind="text: mb"></span> MB)</small>
                                                 </div>
--- a/interfaces/Glitter/templates/include_overlays.tmpl
+++ b/interfaces/Glitter/templates/include_overlays.tmpl
@@ -527,12 +527,12 @@
                                                 <strong data-bind="text: percentage"></strong>
                                                 <div class="fileDetails">
                                                     <span data-bind="truncatedTextCenter: filename"></span>
+                                                    <div class="fileControls">
+                                                        <a href="#" class="hover-button buttonMoveToTop" title="$T('Glitter-MoveToTop')"><span class="glyphicon glyphicon-chevron-up"></span></a>
+                                                        <a href="#" class="hover-button buttonMoveToBottom" title="$T('Glitter-MoveToBottom')"><span class="glyphicon glyphicon-chevron-down"></span></a>
+                                                    </div>
                                                     <small>(<span data-bind="text: file_age"></span> - <span data-bind="text: mb"></span> MB)</small>
                                                 </div>
-                                            </div>
-                                            <div class="fileControls">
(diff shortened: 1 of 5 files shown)


gwyden:
To provide you a sample of what it looks like. the buttons are always visible. 
This shows it when the entire filename is present
image

and this one shows when the filename is super long the buttons are still accessable 
image

The one problem I have with how it is now is that I would have loved to have them right aligned and leave the rest of the content the space to grow into things but that is a problem for future me. 


I didn't think of this as a bulk option and I can see some value but until I have a look at that implementation I don't know how difficult/easy it would be.

thezoggy:
Put the fixed width ui to the left of the dynamic sized width release... Will make it look better and more consistent ui experience for the user

gwyden:
I agree the fixed controls for bulk operations should stay on the left but this are individual controls.  putting them on the left actually takes them away from a majority of the unique identifiers that I would use to make the decision if i want to move to top/bottom.

thezoggy:
That would work, overlay the buttons on the far right when mouseover

gwyden:
ok so I think I got it but I had to make some changes I am not 100% comfortable with.  I had to move out of a span and into a div in order to get the positioning to work right. so some new classes some new styles needed applying as well.  Maybe someone with more UI experience can make it work a different way but this way works and doesn't appear to impact anything else.

image


I tested it with a 60 character long filename and it still works
image

thezoggy:
you should squash and rebase so the git history is clean

Safihre:
@thezoggy @gwyden, no worries, I can do that when merging it in. No need to rebase now.

Safihre:
Here's just a thought: how about making the buttons appear in the space where now the % completed is shown, when hovered? Because with the age there on the right, it's a bit out of style.

gwyden:
@Safihre I stand by the comment of the most useful location for the buttons is closest to the part of the filedetail that is used for making the decision. But since I am not the owner it is your call.   How would you want the behavior to be?  Mouse over and replace the % completed with the controls?  I made a quick mock-up to see how it would look.

image

Safihre:
Yes that's indeed what I imagine. Think it looks pretty good! 
But I wouldn't mind them on the right, just have to make the age/size font bigger and just lower the number of allowed characters in the filename lower to make it all fit. 
This way the style of the buttons matches better the rest.

gwyden:
What about replacing the age/size with the control instead? Keeping the width and character count and  keep the formatting of the not so critical details.  
image

Safihre:
Oh yeah, good idea!

gwyden:
alright let me work up something to match the mock.

gwyden:
Here is what i came up with.  I forced hover on the tr and the control to show they are activating as expected.  
image

Safihre:
Oke looks good! I would suggest a tiny bit of padding on the left.

But a bigger issue: it doesn't actually work. It doesn't send the right API call to move the file to the top or bottom. This isn't really specified in our API documentation, so you'll have to infer it from the other calls.
If it's too strange, I can fix this when I get back to my dev-PC next week.

gwyden:
strange what do you mean it doesn't work.  i moved 37 to the top and that file started downloading instead.  (image for reference)   

image

I am not rebuilding the call to the API to do the work I am manipulating the table and calling the update call that knockout is using.  so if the reorganization of the table works for manually dragging the rows around then the button click should work.

Safihre:
Oké I'll have to try that again later!

gwyden:
I just did a clean pull from my branch and the controls are working but I did miss adding the css element for the control margin so that will be committed shortly.   I see an issue with Glitter and night mode.  The colors wash out and are difficult to identify so i am working to change that up before i do the next commit.

gwyden:
Sample of what it looks like in night mode
image

Safihre:
Oke so now I see what is going wrong:
The changes just move the item all the way to the top and all the way to the bottom, but this is not correct. You can see it when you open and close the modal.
For example all the way to the bottom places it below the postponed par2 files, while actually it should be before. Same if you move it all the way up, it goes above the already finished ones.

So either there should be smart logic in the javascript, or it should reload the data from a fresh filelist-API call.

gwyden:
I am not seeing what you are saying about opening and closing the modal.  The idea was move to top moved it above all others and so that way pushing it all the way to to top even over those things already downloaded.  I will get to changing the logic to only put it below the items already downloaded.

the control visibility for non-interact-able files can be easily resolved as well so I will get to that this afternoon/evening.

Safihre:
You have the Hide-100%-completed files turned on, that's why you're not seeing it. Click on Tee little eye icon on top and then you will see how the file that's send to the top gets placed above the already finished files, when you open-close-open the model the file will then shift below the already completed ones.

gwyden:
Yeah i was taking the "move to x" literally :)   I will get the behavior changed for ya.

Safihre:
If you are bored, we could also use such functionality in the main queue :wink:

But there priorities should also be considered. I know people have complained there's no good way to move stuff from for example the second page of jobs to the first page, this could be a great solution.

gwyden:
@safihre I might be able to do that but it will be a different pull request so we don't mix the two.  Baby steps :)

gwyden:
OK so the items now move to the top and bottom of the "workable" queue.  The controls also only show on those items that can be moved. 

I did find an interesting flaw. Once the button is used and the TR is moved to the final location that button no longer worked anymore.  The cause appears to be the way the element is being cloned before being moved so it is losing all of the event triggers it needs so I moved to detach and insert detached object in the final resting place and all seems well.  I also found another edge case where if you try to modify the order of the files before anything was downloaded i was not getting success when finding a target so i just defaulted to no action until there is a proper target for the row. we can modify this behavior i guess but that is your call.

gwyden:
Unless there are other modifications i believe this is ready to be pulled.  I have taken a look @ the queue page and giving a move to top move to bottom should be a little easier if we need the control to be a literal "move to top" and bottom  it is just a matter of calling it in the right context and making the control fit with on the page.

gwyden:
and a little taste of what is to come
image

Safihre:
Will test it tomorrow :) 
Regarding the queue, I would place the buttons next to the other buttons and then the age. 
Will have to test that at many different screen sizes, I remember spending a lot of time optimizing widths for every size.

Safihre:
Found some time to test today! One small issue I still found: when a job has no files downloaded at all yet (so also no par2-files postponed), the moving doesn't work. I see no API-call is made so something in the JS logic is preventing it.

gwyden:
That was intentional.  Did you want the files to move top and bottom of the entire queue? basically i just short circuited the move like this
                if(targetRow.length){
                    row = row.detach();
                    row.insertAfter(targetRow);
                    tbody.sortable('option', 'update').call(tbody[0],null, { item: row });
                } // no target row so we stay put

Safihre:
Yes, why not? Would be weird if they did nothing just because nothing was downloaded yet. Drag-and-drop also works before any files are downloaded.

gwyden:
ok.  easily resolved.  commit incoming

gwyden:
That took a little while longer than promised but it gave me the opportunity to clean up the code a little and make it more consistent with the rest of the code base. 

- Move to top and bottom controls are only displayed on files that can be moved
- Move to top will move the file to the next available interactable location meaning if files have already been downloaded it will move to the first slot under the most recent downloaded file.
- Move to bottom will move the file to the last interactable location. meaning if there are par files that are not slated to be downloaded it will place the file in the slot right above them.

Safihre:
@gwyden Thanks! Works great and indeed now it fits much better with the rest of the coding style.
This is a very nice addition.

I did reverse some of the changes (see 77f3dc18b58bfef094e9f365810d3abfa026e990) because the filelist for Night became unreadable and the buttons invisible. I also used existing texts from the Plush skin, so it's directly available in all languages!

PR #962, opened 2017-07-09
added buttons and logic to move to top and bottom of download queue

gwyden:


Safihre:
This is not ready yet right? Tell me when it's ready for testing 👍

gwyden:
no it is ready.  I am using it to maximum effect :)

gwyden:
i was debating extending the control to more than top and bottom and add a next/previous page but that is a problem for future me and currently outside the scope of this feature.  

So we have this problem where multiple servers have different retentions.  
image

showing the active control
image

and moved to the top of the queue
image

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -128,6 +128,32 @@ function ViewModel() {
         return parseInt(self.nrWarnings()) + self.allMessages().length;
     })
 
+    // process the full queue
+    self.processFullQueue = function(response){
```
Safihre: Why is this extra function necessary?

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -388,6 +414,23 @@ function ViewModel() {
             self.isRestarting(1)
         }).always(self.setNextUpdate);
 
+        // this is a test of getting the entire queue
```
Safihre: Test, so not done yet?

interfaces/Glitter/templates/static/javascripts/glitter.main.js
gwyden: This was more a i want to guarentee i get the actual first and last element and kind of a preemptive attempt at finding the first and last that were not paused.  It is not 100% necessary but gives a place for logic to happen on queue position selection if necessary.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
gwyden: and you are right totally not a test.  :)  i can change the logic to be purely based on the numbers from the queue itself.  instead of a second call to the queue api if you think the future of the feature is not going to need need knowledge of the entire queue before making a decision on where to go.

thezoggy:
while i'm not debating the usefulness of the move to top/bottom as that is a real one off situation that is nicely addressed by this (especially when you have a large queue or your wanting to mess with an item in the middle of the queue and you have a small pagination amount.  

is it more flexible to just offer up sorting the queue by age/size/name/etc

gwyden:
@thezoggy something like this?
image

I think they address two different needs, if i cared to do it by age sure but sometimes a man just wants to get this one thing right now while leaving the rest of the queue intact.    I know the idea is to mainly leave the autodownloader doing its auto downloading things but for me this started as a "this one download is giving me headaches so I had to manually baby the beast and sorting all the elements was just not going to do it.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
Safihre: Indeed, in this case I would **not** want a second API call but really for it to be moved to the top. I think the internal logic of the backend should only move it to the top of the priority of the job, so that a Normal priority will still be below High priority ones (if I remember cotrecly, should test that).

gwyden: :welp: looks like it does muck with the priority for some reason.  if slot 0 is a high priority and an item is moved up to slot 0 it gets high priority and keeps it.   That has to be some API mumbo jumbo but does look like we need to know the full queue to even begin to figure out here to go if we need to take priority into account.  @Safihre i have a commit to remove the "test" verbiage but can hold off on it and refocus on the behavior of the move if you want that included in this pull..

Safihre: Sorry I think I caused confusion. The behavior of the API is definitely as expected, because if a user drags (or uses the numbered drop-downs in smpl skin) he wants the job to move to wherever he just chose. Not for it to then jump down again if the priority doesn't match, so it has to change priority like this. 
So my comment from a while ago that the Javascript should take care of the priority stuff is wrong, the API will take care of it. No need to be smart, just move all the way to the top/bottom!

gwyden: yeah.  ok so i will be removing all the crap code ;)  and generating a new commit.  busy day today so this will likely come tonight/tomorrow.

gwyden:
this is complete and ready for testing.

Safihre:
Thanks 👍 👍 👍

Safihre:
@gwyden Should have made this longg time ago, it's really useful 🎉

gwyden:
@Safihre I am glad I could contribute something worth while.  This went a lot faster than the individual files and yeah I find myself using it a lot more than i thought i would.

PR #964, opened 2017-07-13
Direct Unpack

Safihre:
I have to give total credit to @hugbug who came up with the idea for NZBGet of using UnRar's `-vp` switch to enable unpacking of downloads as soon as each individual file comes in.
While I was initially skeptical because it looked like an exotic feature to watch while downloading, I realized for big downloads like movies this could make a huge difference in total post-processing time for all users that have semi-fast drives. My poor Intel NUC with a small SSD and 5400rpm data drive would spend 20 minutes downloading and then another 4 minutes unpacking, now we can reduce this to no extra unpack-time at all.
It seemed appropriate to give it the same name as in NZBGet: Direct Unpack, instead of pretending it's not the exact same thing 😄 Although this version is less advanced than NZBGet's: I did not implement it such that it downloads 1 article of each file just to determine the correct filename of very obfuscated posts and sorts them correctly. If the `.r00` file happens to be the last file of an obfuscated post: too bad!

I implemented it such that only 1 UnRar will be unpacking at a time and a maximum of 4 UnRar's will be spawned. If something goes wrong, we abort and let regular post-processing handle it.

To see if a user's system can handle it, I want to run a disk-speed measure **once** and if the incomplete-folder speed is above 80MB/s (just a guess, maybe higher/lower?) we enable it, otherwise leave it disabled.
I still have to implement that check and I also still have to implement a visual Queue-indicator when a job has active DirectUnpacker.
Also only tested it so far on my Windows machine.


Safihre:
Please open an issue in case of problems!

PR #974, opened 2017-07-23
Feature: History retention

Safihre:
@thezoggy What do you think? 
I opted to let the setting only apply to Completed jobs, so that Failed jobs can still be retried.

Setting it to number of items/days will make the extra `input` appear.
Had some troubles formulating the options, maybe you have better suggestions?

Will add also a midnight check using the scheduler.


Safihre:
@thezoggy so what do you think of this one?

thezoggy:
havent tested it out, but from the looks of it.. should cover most things. only concern is the number of items.. does it handle negative amount/invalid number/larger than int...

PR #975, opened 2017-07-23
Feature: Server stats graphs

Safihre:
image


thezoggy:
assuming your graph has a little button to show day/month/etc view.. since the totals and whats displayed only match up with 'today' ?

Safihre:
So far I will only support month view because in the BPSmeter I am only saving it per day, not per hour etc.
Up to now we did not record the data per day, so there's no historical data.

thezoggy:
be nice if we showed all the servers in a stacked area graph, 
sab
image is from netdata -- https://my-netdata.io/  (they use dygraph for their graphs -- http://dygraphs.com/)

Safihre:
Let's first start with this 😉

PR #976, opened 2017-07-23
Dupe - Tag only

thezoggy:
User requested a way to track dupes but not have sab block/discard/pause the items (as he doesnt want to have to manually unpause). Figured we could add a 'Tag' switch for the Dupe detection. Works same way as Pause (shows duplicate tag in queue/adds warning/internally tags it) but does not pause the item.

Once the file is done downloading and makes it to the queue, you have no idea it was a duplicate.
You would have to use the 'Warning' and search for that.. and then assume the newer one was the dupe... we really should expose in the History that an item was a duplicate. Right now the 'search' box only looks at the name.. and I dont think we should be messing with the name to add duplicate there. I'd rather us just add a tag/flag whatever to notate this.. and can filter/sort on it. We also should do the same to the Queue as well.. since changing the name of the item to be `DUPLICATE / XXXXX` is a little jarring.

Now if we just exposed the dupe flag in the history, then the whole 'tag' option really isnt needed as it was just a means to an end.

### DIFF
--- a/interfaces/Config/templates/config_switches.tmpl
+++ b/interfaces/Config/templates/config_switches.tmpl
@@ -86,6 +86,7 @@
                         <label class="config" for="no_dupes">$T('opt-no_dupes')</label>
                         <select name="no_dupes" id="no_dupes">
                             <option value="0" <!--#if int($no_dupes) == 0 then 'selected="selected"' else ""#--> >$T('nodupes-off')</option>
+                            <option value="4" <!--#if int($no_dupes) == 4 then 'selected="selected"' else ""#--> >$T('nodupes-tag')</option>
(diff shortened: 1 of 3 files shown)


Safihre:
Agreed, we need a much better system where we work with labels so that a download can have multiple labels like Oversized/Duplicate/Encrypted.
I considered that many times, but it seems to me like a breaking-API change so I am waiting for 3.0.0 for that.

thezoggy:
heres a thought, we dont touch status but just add additional fields.
add `is_Encrypted`, `is_Duplicate`, `is_Oversized` etc. which to the api is just extra fields that people can use if they want. then for glitter.. we can then check these fields to see if present and set.. if so.. show the little tag. then expand search box to have options to filter to said tags.

PR #980, opened 2017-07-25
Zeroconfig

Hellowlol:
Let bonjour know if we are using http, and make it possible to use without binding the cherryhost.


Safihre:
@Hellowlol You tested this right?

Hellowlol:
Ofc.

You can test it by running:
py -3 -m netdisco

PR #1041, opened 2017-10-02
Added 'with' to list of lowercase words in titles

walgarch:
Change made to avoid duplicate folders in linux for some titles


Safihre:
Looks good to me.

PR #1045, opened 2017-10-08
New feature: Safari pinned tab icon

andofrjando:
This pull request adds the pinned tab icon in Safari. The colour of the icon is set with the `color` attribute in the `<link>` element. The colour in this pull request is the base SAB colour, but darkened 20% to improve legibility. 
### Preview
**Active tab:**
screen1

**Inactive Tab:**
screen2

### Documentation on Safari pinned tabs:
https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html


Safihre:
Sounds like a fine addition, but I personally find the chosen icon and color not aesthetically pleasing. 
Why not use a high contrast version similar to Github, like we also do in the macOS top bar? So just the arrow, or just the black arrow with the white letters (with a border) on top?

andofrjando:
I would admit colour selection is not my forte... it doesn't look great! 🙈

I've created alternatives from your suggestions, and can push what ever is decided upon. Personally, my favourite is the arrow with word outline, in grey. 
___
**Inactive:**
Arrow with word outline:


Arrow only:

___
**Two different active colours:**
The same grey that is used for the MacOS menu bar icons `#383F45`:



Black (same as the adjacent GitHub icon):
6bk
5bk

Safihre:
:+1: much better, let's use those! Arrow with word outline.

PR #1052, opened 2017-10-13
correct spelling of 'separate' in various contexts

OneCDOnly:
mnemonic: "_there's **a-rat** in **sep-a-rat-e**._" :D

### DIFF
--- a/cherrypy/lib/sessions.py
+++ b/cherrypy/lib/sessions.py
@@ -596,7 +596,7 @@ class MemcachedSession(Session):
     # Wrap all .get and .set operations in a single lock.
     mc_lock = threading.RLock()
 
-    # This is a seperate set of locks per session id.
+    # This is a separate set of locks per session id.
     locks = {}
 
     servers = ['127.0.0.1:11211']
--- a/po/main/da.po
+++ b/po/main/da.po
@@ -3730,7 +3730,7 @@ msgstr ""
 "findes i RSS-feed, medmindre du trykker på \"Force Download\"."
 
 #: sabnzbd/skintext.py [Config->RSS, placeholder (cannot be too long)]
-msgid "Seperate multiple URLs by a comma"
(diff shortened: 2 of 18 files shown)


Safihre:
You can't do it like this, the translation system automatically updates all the translations and changing texts `skintext.py` would break all the translations since they are mapped to the original text.

Instead you only need to do 1 thing:
Add to `en.po` at the end:
```
msgid "Seperate multiple URLs by a comma"
msgid "Separate multiple URLS with a comma"
```

OneCDOnly:
Ah.... should I go back and try again? ;)

Safihre:
Yeah 😉 changes to translatable texts should always be done via `en.po`

OneCDOnly:
No worries - I knew that bigger-picture stuff would catch me out eventually. I'll try again. The other files only need that word changed in the code comments. Should I re-submit them too?

Safihre:
Sounds good, everything that's not in `T(xxx)` or in `skintext.py` can be corrected directly.

OneCDOnly:
Just thinking about this - as the placeholder text is changed too, the automatic mapping would still work correctly, wouldn't it?

Safihre:
No it's unilateral, we upload the English text to Launchpad and then in a seperate system the translations come back. It only imports the English, not the other languages. Maybe it could, but we didn't set it up like that.

OneCDOnly:
And final question: the format in `en.po` seems to be:
```
msgid "---"
msgstr "---"
```
So, I should therefore append:
```
msgid "Seperate multiple URLs by a comma"
msgstr "Separate multiple URLS with a comma"
```
Correct?

Safihre:
Oops yes indeed that's the correct format, not what I wrote previously!  🙈

OneCDOnly:
No worries, thanks for your patience. :)

OneCDOnly:
Resubmitting as a new pull request so I'll close this one.

Issue #585, opened 2016-06-01
Password support for unzip


shypike:
Currently unpacking of ZIP files is without password support.
This should be added.
The unrar and un-7zip do have password support code, but it is currently implemented with copy/pasted code, which makes it hard to maintain.
The code should be refactored to supply identical password support for unrar/unzip/un-7zip.

Safihre:
Implemented.

PR #1064, opened 2017-10-29
When available, use 7zip instead of built-in unzip.

shypike:
7zip supports passwords, many old platform unzips do not.

I thought this would be the easiest way to implement passwords for unzip.
Many standard unzips are so old (or compiled without encryption) that they don't support passwords. Example: OSX.
Haven't tested it yet, must find some test case NZB :)
Should solve #585


Safihre:
@sanderjo Maybe you can upload a password-protected `.zip` and `.7z`?

sanderjo:
7z with password HELLO

https://nzbindex.com/search/?q=zipped-with-password-HELLO+reftestnzb+f7874b38383&age

Do you also need a zip file?

Safihre:
@sanderjo Oops, I think I wrote badly. We need something else:
Now it is rar-files with a `.zip` inside. But we need to have a direct `.7z` and `.zip` uploaded, not inside rar's but direct. Would that be possible? 👍

sanderjo:
OK, I posted a 100 kB zip file directly (without rar / par2). NZB: https://nzbindex.com/search/?q=bb77a14ce76a

FWIW: the 100 kB zip file will unpack into 100 MB file filled with zeros.

Safihre:
Cool, the branch works!
```
7-Zip (a) [32] 16.04 : Copyright (c) 1999-2016 Igor Pavlov : 2016-10-04

Scanning the drive for archives:
1 file, 101979 bytes (100 KiB)

Extracting archive: \\?\C:\SABnzbd\zero-100MB---with-password-HELLO.zip\zero-100MB---with-password-HELLO.zip
--
Path = \\?\C:\SABnzbd\zero-100MB---with-password-HELLO.zip\zero-100MB---with-password-HELLO.zip
Type = zip
Physical Size = 101979

Everything is Ok

Size:       104857600
Compressed: 101979
```

Safihre:
@shypike Removed the `unzip` from the Windows installer since we don't use it anymore. 
Looks good to me, also for you?
Put it in 2.3.2 or later?

PR #1076, opened 2017-11-24
Update all web Dependencies 

Fish2:
Not Sure if this is worth it or will help but I updated all web Dependencies. 

Note: Updated them to latest and in some cases to the newest without big changes/breaking sab.

Bootstrap Typeahead 3.1.0 To 4.0.2
JQuery Form Plugin 3.51.0 To 4.2.2
JQuery Qrcode 0.12.0 To 0.14.0
Moment 2.18.1 To  2.19.2
Six 1.10.0 To 1.11.0


Safihre:
While some modules can be updated without problem, for the big ones there's a reason they are older:
For example we use a quite heavily modified version of cherrypy. And updating jquery results in slower initial page loads in Glitter (queue will first flash empty before being filled, this is not the case on older versions). 
Same for bootstrap, the newer version the icons (glyphicons) are less sharp and therefore we use the old one.

Fish2:
Ok let me know i can always remove some of the updates

SABnzbd.py
```diff
@@ -54,9 +54,6 @@
     sys.exit(1)
 
 import cherrypy
-if [int(n) for n in cherrypy.__version__.split('.')] < [8, 1, 2]:
```
thezoggy: removing this check is not ideal. while we provide cherrypy with the source.. not everyone uses it. this test is to safeguard that outside versions have the features we require and work with the code we have. our version is just modified to reduce bugs.

Safihre:
I think we could use these:
Bootstrap Typeahead 3.1.0 To 4.0.2
JQuery Form Plugin 3.51.0 To 4.2.2
JQuery Qrcode 0.12.0 To 0.14.0
Moment 2.18.1 To 2.19.2
Six 1.10.0 To 1.11.0

Fish2:
@Safihre ok thanks would you like me to remove the other comments or create a new pull request with them updates ?

Safihre:
Rebase with those commits removed is good 👍

Fish2:
All Done Thanks

Safihre:
Looking good.
Now could you maybe only include `moment.js` locale's that we actually use and have them have the filenames identical to our language codes? 
To save some space and without right filename they don't get loaded :)
```
da
de
en
fi
fr
he
nb
nl
pl
pt-br
ro
ru
sr
sv
zh-cn
``

Fish2:
@thezoggy Yer the filebrowser works its all good

Fish2:
@Safihre Ok nice thats all cleaned up

Safihre:
If you could add `he.js` (new language) then it's perfect, I tested all the other stuff 👍

Safihre:
👍

Fish2:
@Safihre Nice Cheers. Yer I have also tested 👍

thezoggy:
should rebase the moment changes to keep history clean.

PR #1097, opened 2018-01-30
64-bit version of unrar (for MacOS)

sanderjo:


thezoggy:
should update the license to reflect the correct version (its still for 5.40).

Safihre:
Which license file? The `license.txt` file in that folder is not version specific as far as I can see.

thezoggy:
what i get for trying to look at git via the phone, yeah ignore me. were good on the license. but anyways do we want to offer 64bit and 32bit utilities like we do on windows so those that want can fall back if there is any issues.. or do we want to just drop all 32bit stuff and do 64bit only?

Safihre:
Since everything else is 64bit (SAB itself and external tools) I think it's safe to go 64bit only.

PR #1101, opened 2018-02-13
Change max priority to 100 to fix #1100

### LINKED ISSUE #1100: NNTP Server Priority Setting
NNTP Server Advanced setting.

Priority: 0 is highest priority, 100 is the lowest priority
If you enter 100 and save, error displays that value must be less than or equal to 99.

Fish2:


Safihre:
It has to be a 2 digit number. Because we use it for sorting purposes later. So that's why the max is 99.
Will fix the text instead :)

Fish2:
Ok that make more sense

Issue #1127, opened 2018-04-06
Let Chrome trust https://localhost:9090/sabnzbd/


sanderjo:
Chrome is getting pickier and pickier about security with both HTTP and HTTPS. 

With some work, Chrome does trust https://localhost:9090/sabnzbd/ and https://127.0.0.1:9090/sabnzbd/ and https://192.168.1.10:9090/sabnzbd/ and https://sammie-1710:9090/sabnzbd/ (for example).

I'll send a PR this weekend. Can you please test and review @jcfp and @Safihre . And I'm interested in results on other OSes en webbrowsers.

Steps:
1) put the desired names and (private) IP addresses into the SAN of SAB's cert. I have achieved that in SAB's certgen.py (the current code is incorrect)
2) In Chrome, go to https://localhost:9090/sabnzbd/, and export the cert
3) In your OS, import that cert
4) restart Chrome, and Chrome trusts https://localhost:9090/sabnzbd/

Maybe, if wanted, we can make 2 and 3 easier.

Rationale

1) Chrome considers HTTP as insecure
2) Chrome considers HTTPS to a self-signed certificate as insecure
3) A cert for an IP address is impossible, so insecure too

... at least ... I thought a cert for an IP address was impossible, but https://1.1.1.1/ proves it is possible. In the cert, there are these SANs:
```
DNS Name: *.cloudflare-dns.com
IP Address: 1.1.1.1
IP Address: 1.0.0.1
DNS Name: cloudflare-dns.com
IP Address: 2606:4700:4700::1111
IP Address: 2606:4700:4700::1001
```

Furthermore, https://leehblue.com/add-self-signed-ssl-google-chrome-ubuntu-16-04/ tells  how to import that into OS and Chrome. Result: Chrome trusts the site (read: SABnzbd) again.

The Cert with SANs:
![image](https://user-images.githubusercontent.com/1273502/38442336-4d511e78-39e8-11e8-8926-70947c366374.png)

```
$ file Downloads/sabnzbd-with-all.bin
Downloads/sabnzbd-with-all.bin: PEM certificate
```
Import on Linux:
`certutil -d sql:$HOME/.pki/nssdb -A -t "P,," -n Downloads/sabnzbd-with-all.bin -i Downloads/sabnzbd-with-all.bin`

Result:

![image](https://user-images.githubusercontent.com/1273502/38442669-55e8c382-39e9-11e8-90ce-8a48ed7924db.png)

[...]

PR #1128, opened 2018-04-08
Improved SAN in SAB's self signed cert.

sanderjo:
Name and IP addresses now (correct) as SANs in SAB's self signed cert.

See https://github.com/sabnzbd/sabnzbd/issues/1127


thezoggy:
Sanderjo, check over my comments in https://github.com/sabnzbd/sabnzbd/pull/822

PR #1132, opened 2018-04-21
accept cheetah version newer than 2.x

jcfp:
I'm going to update cheetah to v. 3.x in debian soonish. All seems to work fine with sab in a quick test, apart from this version check.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -44,7 +44,7 @@ import re
 
 try:
     import Cheetah
-    if Cheetah.Version[0] != '2':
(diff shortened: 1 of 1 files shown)


thezoggy:
Didn't cheeta break with Sab in newer versions when they added portmon and other dependency and segmented out chroot

Safihre:
I also used this newer version for my Python 3-conversion attempt, worked well.
Actually in Python 3 the API and templates are much much faster than under Python 2.

PR #1165, opened 2018-08-05
Fix aborting active direct unpack

riksmith:
Downloads with directunpack are sometimes aborting while they are not "hanging".

This happens when the total time needed to unpack is > 2 min and the time to unpack a single item is > 4 sec.


Safihre:
Hah, I did not think about that scenario indeed. My systems are too fast I guess.
Good catch 👍

Safihre:
@riksmith Did you read my Reddit post by any chance? In case you'd like to get more involved into SAB development! https://www.reddit.com/r/usenet/comments/918nxv/looking_for_paid_help_with_sabnzbd_development/

PR #1166, opened 2018-08-05
Fix deobfuscate script

riksmith:


Safihre:
Odd, now that I look at it I wonder how it worked before. I did test it 😨

Issue #933, opened 2017-06-01
Upgrade to Python 3


Safihre:
It seems Python 3.6 is really speeding up things and I wonder if we should consider re-writing SAB for Python 3?
https://www.youtube.com/watch?v=d65dCD3VH9Q

This is also inspired by Cheetah now being ported to Python 3.
https://github.com/CheetahTemplate3/cheetah3

I already tried running the `2to3` tool on SABnzbd, but it turns out to be a bit more complicated than that...
Anyone willing to help?

Hellowlol:
Wouldnt it be better to use http://python-future.org/futurize.html to port the code to python3 check that it works then remove the fixes that enables py2 support?

Or are we gonna do a hard switch?

Safihre:
Hard switch. I've seen projects supporting both, and it's a mess. 
We control the packages for Windows and Mac, and on Linux all distributions have some version of Python 3, so we should be good. 
I am aware that older Python 3 versions have problems to, so we will have to figure those out as we go.

Hellowlol:
What is the minimum supported? 3.4?

Safihre:
No clue yet, need to convert first.
Working now with 3.6.1.

thezoggy:
i personally would say we would rewrite sab from scratch for py3 not convert. there is so much legacy code and cruft... better to take the time to just redo things correctly. converting you are most likely going to introduce bugs.. and the fact sab has 0 unit tests... is a huge issue where you dont know what you break until after its too late. 

adding test sets to sab is far more important imho than converting it to another language

labrys:
3.4 is safe.  Personally I prefer a 3.6 baseline for cross-platform due to PEP 529: "Change Windows filesystem encoding to UTF-8", but 3.6 is not available on all platforms yet (for example latest debian has 3.4 and 3.6 is only in experimental).  Also I would highly recommend a rewrite for py3 as @thezoggy recommends.

thezoggy:
```
Python 2.0 - October 16, 2000

    Python 2.1 - April 17, 2001
    Python 2.2 - December 21, 2001
    Python 2.3 - July 29, 2003
    Python 2.4 - November 30, 2004
    Python 2.5 - September 19, 2006
    Python 2.6 - October 1, 2008
    Python 2.7 - July 3, 2010

Python 3.0 - December 3, 2008

    Python 3.1 - June 27, 2009
    Python 3.2 - February 20, 2011
    Python 3.3 - September 29, 2012
[...]

PR #1176, opened 2018-09-16
Fix locale compilation in Python 3

svms1:
This pull request fixes the locale compilation tools for Python 3, a small contribution towards the ongoing effort to make sabnzbd ready for Python 3 (issue #933).

Successfully tested with Python 3.6.5 on Linux with an UTF-8 system locale, so any verification on other operating systems (and locales) would be appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ SABnzbd*.dmg
 # WingIDE project files
 *.wp[ru]
 
+# PyCharm project files
+.idea/
+
 # Testing folders
 .cache
 .xprocess
--- a/tools/make_mo.py
+++ b/tools/make_mo.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python -OO
+#!/usr/bin/python3 -OO
 # -*- coding: utf-8 -*-
 # Copyright 2010-2018 The SABnzbd-Team <team@sabnzbd.org>
 #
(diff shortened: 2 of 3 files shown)


svms1:
What were the commands you used? I don't get any errors on Linux even if the paths contain whitespaces.

Safihre:
On Windows it needs to be a list: https://stackoverflow.com/questions/25655173/is-python-subprocess-popen-accept-space-in-path

Maybe you can get that Windows VM 😉 Microsoft offers them: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

I tried to do that, but got some other errors that could be due to my local setup.

svms1:
I did change it to use a list exactly for that reason - and because I usually find it more readable than building a string:

    cmd = [TOOL, '-o', mo_file,  fname]
            if extra != '':
                cmd.insert(1, extra)
    ret, output = run(cmd)

Did you run anything else than ```make_mo.py```? I only adapted this and ```msgfmt.py```, but I see ```extract_pot.py``` in your ouput. I didn't change anything in there, so maybe that's the culprit?

If it's not an obvious or straightforward fix I will focus on getting other parts of SABnzbd working on Linux for now since it is the environment I'm most comfortable with. If I make significant progress there I can always circle back and try to fix issues on other platforms.

Safihre:
Makes sense 👍

Safihre:
I pushed the Windows-fix :)

PR #1183, opened 2018-10-23
Add initial snap support

eberkund:
This PR adds support to build a snap package for sabnzbd.


SABnzbd.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python -OO
+#!/usr/bin/env python
```
Safihre: Can we leave this the same? We had many discussions about this and finally settled on this one.

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{yml,yaml}]
+indent_size = 2
```
Safihre: For now I don't see the need to add this file to the repository.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
...
+    daemon: simple
+    plugs: [network, network-bind]
+
+parts:
+  sabnzbd:
+    plugin: python
+    source: .
+    python-version: python2
+    python-packages: [cheetah3, cryptography, sabyenc]
+    stage-packages: [par2, unrar, p7zip-full, unzip]
```
Safihre: Recently we removed the need for `unzip`, when there's no `unzip` we just use `p7zip`.

How does it get these packages? Is there a way to add `par2-mt` from our own PPA of `par2`?

nafanz:
I'd like to see support for snap packages to simplify the installation process

SABnzbd.py
eberkund: Not sure, I changed it because it would not run as-is with `/usr/bin/python`. I believe this is because the path for Python is different under snapd.

snap/snapcraft.yaml
eberkund: Yeah I am happy to change the the dependencies around, I am unfamiliar with what the preferred versions and installation method is (apt vs pip).

eberkund:
> I talked to Canonical about this a while back, but we also had to setup some automated build process to get the snaps in the snap-store? Or with this people could only build this locally?

It can be built automatically using: https://snapcraft.io/build

.editorconfig
eberkund: Sure, something to consider though as it makes it easier to maintain consistent formatting throughout

snap/snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
...
+summary: SABnzbd
+description: The automated Usenet download tool
+grade: stable
+confinement: strict
+
+apps:
+  sabnzbd:
+    command: opt/sabnzbd/SABnzbd.py -f $SNAP_COMMON
+    daemon: simple
+    plugs: [network, network-bind]
```
diddledan:
```suggestion
    plugs:
    - network
    - network-bind
    - removable-media
```

You probably want to include the `removable-media` plug because it is likely that a user will want to store their downloads in a location other than `$SNAP_DATA` or `$SNAP_COMMON`.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
+name: sabnzbd
+version: git
+summary: SABnzbd
+description: The automated Usenet download tool
+grade: stable
+confinement: strict
+
+apps:
+  sabnzbd:
+    command: opt/sabnzbd/SABnzbd.py -f $SNAP_COMMON
```
diddledan:
```suggestion
    command: python $SNAP/opt/sabnzbd/SABnzbd.py -f $SNAP_COMMON
```

Combine this with reverting the `#!` line in `SABnzbd.py`.

snap/snapcraft.yaml
eberkund: It gets the packages from the Ubuntu repositories, I am not sure there is a good way to add a PPA unfortunately.

eberkund:
Changes applied!

Safihre:
Let me see how to test this, spinning up a vm.

Safihre:
@eberkund How do I test this? I have all the packages installed, what commands to run to build locally, install, and run?

eberkund:
@Safihre it should be as simple as

```
sudo apt install snapd
sudo snap install snapcraft --classic
```

Then from the sabnzbd project directory,

```
snapcraft
```

And then you should get a .snap file produced.

Safihre:
I did that, but how do I then run it? 
The `/snap/bin` doesn't have anything in it after the install.

eberkund:
Oh ya, also to install a locally built snap specify the file name

```
sudo snap install some_file_name_sabnzbd.snap --dangerous
```

The web server should be running after that on the default port http://127.0.0.1:8080

You can check the logs/output with

```
journalctl -u snap.sabnzbd.sabnzbd -f
```

Safihre:
Alright it seems to be working 👍 
Now for the finishing touches:
- We need to compile language-files, like it says on the first page of the wizard once you start it. Somewhere have to run `python tools/make_mo.py` once.
- Can we try to install `par2-mt` from our PPA? 
- @diddledan How do I now write files outside of `/snap`?

diddledan:
> * How do I now write files outside of `/snap`?

You need to run `snap connect sabnzbd:removable-media` to connect the interface. This will allow the snap to access locations under `/media`. 

There are two other locations that the snap can write to by default: `$SNAP_DATA` and `$SNAP_COMMON` which expand to (usually) `/var/snap/sabnzbd/$snap_revision/` and `/var/snap/sabnzbd/common` respectively. The difference between those two paths is any files in `$SNAP_DATA` are copied when a new snap revision is installed, allowing for roll-back. `$SNAP_COMMON` remains untouched when a new revision is installed.

eberkund:
>- We need to compile language-files, like it says on the first page of the wizard once you start it. Somewhere have to run python tools/make_mo.py once.

Can this be done at build time? Where do the resulting files get written to?

Maybe it's as simple as,

```
    override-build: |
      snapcraftctl build
      python tools/make_mo.py
      mkdir -p $SNAPCRAFT_PART_INSTALL/opt/sabnzbd
      cp -R $SNAPCRAFT_PART_BUILD/* $SNAPCRAFT_PART_INSTALL/opt/sabnzbd
```

Safihre:
It can definitely be done at build time, that's all that's needed.

All the ubuntu versions are there, this we use to install on all platforms and (LTS) versions. @jcfp is the expert.

eberkund:
Okay, that part is working for me now. But I'm finding it a bit tricky to add the PPA.

eberkund:
Okay @Safihre it now compiles the language files and uses par2 from the PPA.

jcfp:
First off: nice work!

The par2-mt package on the ppa is a transitional package, an empty shell that depends on "regular" par2 ever since the latter pulled in the mt patches - so you might as well use that directly. For recent Ubuntu releases (bionic/18.04 and newer), the par2 package in the standard repositories is already multithreading. Par2-tbb is ppa-only and faster, but based on very old par2 sources (ca. 2004) without any further bug fixes.

Couple of questions and ideas:
Might be nice to create "channels" following different sab git branches, i.e. a stable tracking master and something marked 'less-than-stable' for develop? That way, users could check if bugs are fixed by the latest commit without risking their usual setup (snap or otherwise).

How would one make other dirs/locations available to the snap? After all, /media tends to list only removable devices like usb drives. Not exactly the kind of storage where users typically hoard the Linux isos they grab from Usenet.

Any way for end users to pass options to the program being run inside the snap?

Why the separate stage-packages/build-packages stanza for arm64 and armhf? Why is openssl needed for those two but not the other archs ("stage-packages/else")?

eberkund:
> The par2-mt package on the ppa is a transitional package, an empty shell that depends on "regular" par2 ever since the latter pulled in the mt patches - so you might as well use that directly. For recent Ubuntu releases (bionic/18.04 and newer), the par2 package in the standard repositories is already multithreading. Par2-tbb is ppa-only and faster, but based on very old par2 sources (ca. 2004) without any further bug fixes.

Hmm, okay well the PR right now uses par2-tbb for x86 and par2 for ARM. Does that sound about right? Or should it just use the standard par2 and forget the PPA altogether?

> Might be nice to create "channels" following different sab git branches, i.e. a stable tracking master and something marked 'less-than-stable' for develop? That way, users could check if bugs are fixed by the latest commit without risking their usual setup (snap or otherwise).

Sure, I think that is something that is handled through the snap store though.

> How would one make other dirs/locations available to the snap? After all, /media tends to list only removable devices like usb drives. Not exactly the kind of storage where users typically hoard the Linux isos they grab from Usenet.

No possible as far as I know, unless we make it a "classic" snap rather than a "strict" one. This means less strict confinement but you would be able to save to arbitrary file paths.

> Why the separate stage-packages/build-packages stanza for arm64 and armhf? Why is openssl needed for those two but not the other archs ("stage-packages/else")?

I'll add openssl for x86 as well, it was just pulling in the version from the host system automatically so I didn't notice.

eberkund:
>It shouldn't need openssl, since that's baked into Python. Does it need those dev packages for building?
Yes it looks like it definitely needs those dev packages, at least for armhf and arm64 builds because they appear to build some of the python extensions from source. On amd64 and i386 it looks like it isn't necessary because it downloads prebuilt extensions. However I think it should be left in for consistency and if this behavior changes in the future then it will still work.

Safihre:
@diddledan it looks good to me, what do you think?

jcfp:
> Hmm, okay well the PR right now uses par2-tbb for x86 and par2 for ARM. Does that sound about right? Or should it just use the standard par2 and forget the PPA altogether?

Either will work, so that just boils down to a informed choice for you as the author of the snap.

Only remaining thing that came to mind looking at the latest diff, is the explicit handling of a limited number of hardware architectures:

- What happens if one tries to run this on another than the explicitly listed archs (arm64 armhf amd64 i386), now that there's no "else" anymore like in a previous version of this PR?

- And couldn't that stanza in case of "build-packages" be reduced to just a single line, since every arch wants exactly the same?

eberkund:
> Either will work, so that just boils down to an informed choice for you as the author of the snap.

Okay well I guess if par2-tbb is supported and it is faster, that's the better choice to include.

> And couldn't that stanza in case of "build-packages" be reduced to just a single line, since every arch wants exactly the same?

Unfortunately not, that only works if the builds are being created on the same architecture they are compiled for. But since snapcraft.io/build uses amd64 for all builds it is necessary to specify the target architecture like that. Ideally the snapcraft tool would automatically append the target architecture to the list of packages but that is not what it does.

> What happens if one tries to run this on another than the explicitly listed archs (arm64 armhf amd64 i386), now that there's no "else" anymore like in a previous version of this PR?

It would fail. I guess I could put back the else part, but keep in mind that would only be useful if someone wanted to compile the snap natively on say ppc64el themselves. If we want to support other architectures in the CI environment then it would require additional arch specific entries.

jcfp:
Thanks for the info about the build process. The only other arch supported by ubuntu (ppc64el and s390x) are too exotic for anyone to use that kind of hardware for usenet downloading anyway, so don't bother with those. Time to start beta testing this on real users  :+1:

eberkund:
Great, well it can be installed through: https://snapcraft.io/sabnzbd

Once it is merged in the main repo then I can switch it over to use this repo instead of my fork.

nafanz:
> Great, well it can be installed through: https://snapcraft.io/sabnzbd
> 
> Once it is merged in the main repo then I can switch it over to use this repo instead of my fork.

Thanks for the work you've done
Please add the application icon in snapcraft.io

eberkund:
Nevermind, found it. Icon added.

Safihre:
Great. Will test it asap. 

Wondering why on snapcraft.io it lists the version as 1.1.0RC4?
In any case I have to dig a bit to see how I can get it correctly parse the versions, so we can have a stable/beta/edge version. 
Our `master` branch always contains the latest stable, if that's any help.

Safihre:
Oke maybe I need some help, I tried to dig through the documentation but want to know if I'm on the right path:

- Can we set `grade: stable` if on the `master` branch
- And then use [this](https://github.com/snapcore/snapcraft/blob/master/snap/snapcraft.yaml#L67-L72) to set version to `grep -oP '(?<=^Version: ).*' PKG-INFO`
- On `develop` we set `grade: devel` and version to some `git` hash?

nsg:
To answer your question about:

> Wondering why on snapcraft.io it lists the version as 1.1.0RC4?

sabnzbd $ git describe --abbrev=0
1.1.0RC4

Snapcraft's `version: git` will pickup the latest annotated tag. If I understand it correctly this follows git best practice. The easiest way to fix this is either to manually (or with a script generate) the version, or use the relatively new "snapcraftctl set-version" command in a override-build step (`snapcraftctl set-version $(git describe --tags --abbrev=0)`), example at https://github.com/snapcrafters/yakyak/blob/master/snap/snapcraft.yaml#L44, notice the missing version, and the added adopt-info.

nsg:
This is probably a better example, this is when snapcraft them self moved from `version: git` to set-version and set-grade https://github.com/snapcore/snapcraft/pull/2230/files

I can also recommend this site https://snapdocs.labix.org/

eberkund:
What about

```yaml
override-pull: |
  version=$(grep -oP '(?<=^Version: ).*' PKG-INFO)
  [ $(git rev-parse --abbrev-ref HEAD) = "master" ] && grade=stable || grade=devel
  snapcraftctl set-version "$version"
  snapcraftctl set-grade "$grade"
```

eberkund:
> and the added adopt-info.
What does adopt info do?

Safihre:
`adopt-info` can use metadata files to extract version etc, but it seems we have none of the supported metadata-files in the repo. So we can't use that.

eberkund:
I updated it to take the version from PKG_INFO and set the stability based on if it is on master branch.

Safihre:
@eberkund Can you release the `sabnzbd` name so I can register it?

eberkund:
@Safihre there is no way in the UI, but I made a request on the forums: https://forum.snapcraft.io/t/request-to-remove-transfer-sabnzbd-snap/8216

popey:
@eberkund it's not possible to rename / delete snaps yourself. The store team on the forum will pick up your request so @Safihre can own it. @Safihre will need a store account to transfer the snap into.

Safihre:
I created an account on https://build.snapcraft.io, or is this not the right place?


Ps: maybe I'm ranting here but this whole snap process is just so frustratingly typically Linux.. Only for the experts, no easy happy flow, too many (undocumented) options. Guess I'm too much of a Windows user.

eberkund:
> I created an account on https://build.snapcraft.io, or is this not the right place?

Yeah that's correct.

> Ps: maybe I'm ranting here but this whole snap process is just so frustratingly typically Linux.. Only for the experts, no easy happy flow, too many (undocumented) options. Guess I'm too much of a Windows user.

I'd say there are definitely some things which are easier on Windows but also things which are easier on Linux. Snaps still feel very beta to me but progress is coming along.

popey:
@Safihre I appreciate the feedback. You should create an account at https://dashboard.snapcraft.io/

Safihre:
Account name is Safihre / safihre@sabnzbd.org

eberkund:
@Safihre the snap has now been transferred.

Safihre:
👍

PR #1186, opened 2018-10-29
Add Travis and AppVeyor badges

eberkund:
Add nice looking badge icons to the readme based on CI build status.

image


eberkund:
> Maybe we should link to develop branch? Since it's more interesting?

You're right since develop is the default branch. I'll also add the snap badge when it gets transferred over.

thezoggy:
can prob add these badges too:
http://isitmaintained.com/project/sabnzbd/sabnzbd

eberkund:
I would also like to add test coverage https://codecov.io/

PR #1193, opened 2018-11-22
Option require_modern_tls if you want TLS 1.2 or higher for NNTPS

sanderjo:
Option require_modern_tls if you want TLS 1.2 or higher for NNTPS

Rationale:

- All 200+ NNTPS servers with a correct TLS setup, speak TLS1.2 or higher. No need to speak TLS1.1 or lower.
- Avoid MITM protocol downgrade attacks
- TLS1 and TLS1.1 are old: 1999 resp 2006. They will depreciated in 2018/2019/2020 by the Big 4 browsers

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -265,6 +265,7 @@ no_penalties = OptionBool('misc', 'no_penalties', False)
 debug_log_decoding = OptionBool('misc', 'debug_log_decoding', False)
 ignore_empty_files = OptionBool('misc', 'ignore_empty_files', False)
 x_frame_options = OptionBool('misc', 'x_frame_options', True)
+require_modern_tls = OptionBool('misc', 'require_modern_tls', False)
 
 # Text values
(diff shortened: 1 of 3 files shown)


sanderjo:
PS:

SAB could self-detect if the underlying OS supports TLS1.2 by connecting to the self-test host, and see which TLS protocol is used. But what then ... ? If OS does not support TLS 1.2, and option require_modern_tls  is on ... what to do?

Safihre:
Since python includes it's own SSL, I doubt in Windows 7 there wouldn't be TLSv1.2. Its system independent.

sanderjo:
Let's do this:
- Now introduce it like this. 
- At SABnzbd 3.0, turn it on by default for new installations.
- Between now and SABnzbd 3.0, introduce a test against `selftest_host` if the client supports TLS1.2. And warn users if not.

Safihre:
I think we are fine on the client-side, but I want to add mostly:

- Clear warning to users if the server doesn't support a modern TLS version

PR #1201, opened 2018-12-19
Fix missing unrar, fix Python encoding

### LINKED ISSUE #1197: Snap issues (missing unrar binay, ANSI encoding, multicore par2)
Hi, 

I'm trying to install SABnzbd through snap on Ubuntu 18.10. Is there a guide somewhere on how to do this? I had file permission issues (downloads are stored under /media/storage, but with interface removable-media on sabnzbd still didn't have permission to make directories. Tried installing in "classic" mode and the permission issue is resolved: 
`snap install sabnzbd --classic`

After resolving file permission I still have 3 issues: 
1) unrar binary... NOT found! Downloads will not be unpacked. 
2) SABnzbd was started with encoding ANSI_X3.4-1968, this should be UTF-8. Expect problems with Unicoded file and directory names in downloads. 
3) Multicore par2 is missing? 

I thought maybe I had screwed with something by enabling the --classic flag so I tried uninstalling and installing again without the --classic flag, but I still have the same issues (only now the file permission issue is back). 

Screenshots from the --classic installation: 
screenshot from 2018-12-01 21-05-05
screenshot from 2018-12-01 21-24-01

eberkund:
Fixes #1197 

image

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -10,6 +10,8 @@ version-script: |
 
 apps:
   sabnzbd:
+    environment:
+      LC_CTYPE: C.UTF-8
     command: python $SNAP/opt/sabnzbd/SABnzbd.py -f $SNAP_COMMON
     daemon: simple
     plugs: [network, network-bind, removable-media]
@@ -38,5 +40,7 @@ parts:
     override-build: |
       snapcraftctl build
       python tools/make_mo.py
-      mkdir -p $SNAPCRAFT_PART_INSTALL/opt/sabnzbd
-      cp -R $SNAPCRAFT_PART_BUILD/* $SNAPCRAFT_PART_INSTALL/opt/sabnzbd
+      mkdir -p $SNAPCRAFT_PART_INSTALL/opt
+      cp -R $SNAPCRAFT_PART_BUILD $SNAPCRAFT_PART_INSTALL/opt/sabnzbd
+    organize:
+      usr/bin/unrar-nonfree: usr/bin/unrar


Safihre:
Thanks 👍 Didn't test it, but assuming you did ;)

Safihre:
I don't understand Snapcraft. I just pushed to `master` the new 2.3.6 release, so why isn't it building and publishing it to our `stable` channel? We added a line for it: https://github.com/sabnzbd/sabnzbd/blob/develop/snap/snapcraft.yaml#L38
@diddledan Maybe has insights? :)

eberkund:
Yeah that is very weird, I am not too sure. But that line looks like it works as intended. I don't know why the built snap is being marked as dev grade...

diddledan:
I think it's because the line starts with a square bracket - try replacing it with an "if" instead:

```yaml
    override-pull: |
      snapcraftctl pull
      if [ $(git rev-parse --abbrev-ref HEAD) = "master" ]; then
        GRADE=stable
      else
        GRADE=devel
      fi
      snapcraftctl set-grade "$GRADE"
```

You might be able to get it all onto one line but I'm not certain on the syntax requirements for that. Here is what I _think_ a single line would look like that doesn't begin with a square bracket:

```bash
if [ $(git rev-parse --abbrev-ref HEAD) = "master" ]; then GRADE=stable; else GRADE=devel; fi
```

diddledan:
As to why it is marked as `devel` for the grade, that is because you haven't explicitly added the `grade` keyword to your yaml:

```yaml
confinement: strict # I've copied this so that you see that grade is the same indentation level
grade: devel
```

diddledan:
I see you have separated build-packages and stage-packages by architecture, but you are using the same packages on each platform. If you are using the builder provided by build.snapcraft.io or launchpad.net, or otherwise building on systems of the target architecture rather than cross-compiling, then you can get away with doing:

```yaml
    stage-packages: [unrar, p7zip-full, par2 ]
    build-packages: [libffi-dev, python-dev, ibssl-dev]
```

Safihre:
I think @eberkund did this on purpose, he mentioned needing this to get true cross-compiled output. Not needed anymore?

Safihre:
It seems to just not trigger a build from `master`, can it only trigger it from the main branch or something?

eberkund:
Yeah I think the issue is it just triggers on the primary branch on GitHub which in this case is develop. I saw some threads about this googling about but I'm not sure what the solution is because obviously we can't change the whole project workflow.

Safihre:
It seems I can do it from the commandline, even on Windows with the Linux-on-Windows thing!
```
snapcraft release sabnzbd 113 stable
```

eberkund:
Great, glad you figured out a way.

1875cbc can be reverted as well

PR #1204, opened 2018-12-26
Add test-download shortcuts

Safihre:
Initiated by @sanderjo, we introduce 2 new buttons in the _Status and Interface settings_ window.
This allows users to quickly add a test NZB filled with a random `.bin` file of 100MB or 1000MB, split over the usual RAR files and par2 files.
With this they can check if their setup is fine, for users that complain that SABnzbd fails all their downloads so something must be wrong with SABnzbd (instead of just DMCA takedowns).
Or people can use it for speed testing their connection.

We host the 2 NZB's at https://sabnzbd.org/tests, so that we can always [update them easily](https://github.com/sabnzbd/sabnzbd.github.io/tree/master/tests) later.


sanderjo:
I hope this test_download feature will be useful for a lot of users! I already like it ... :+1: 

Buttons:
sabnzbd test download buttons


Result:
image

PR #1217, opened 2019-01-17
retry nscript

daimoniac:
When using sabnzbd inside a docker container it ocassionally happens that external scripts cannot be run because of some resource pressure inside docker. It terminates then with exit code -11, which indicates a segmentation fault. you cannot catch that error with a wrapper script.

Retrying the call from within notifier.py succeeds, though.

### DIFF
--- a/sabnzbd/notifier.py
+++ b/sabnzbd/notifier.py
@@ -540,7 +540,13 @@ def send_nscript(title, msg, gtype, force=False, test=None):
     if force or check_classes(gtype, 'nscript'):
         script_path = make_script_path(script)
         if script_path:
-            output, ret = external_script(script_path, gtype, title, msg, parameters)
+            # retry calling the script up to 5 times
+            for i in range(5):
+                output, ret = external_script(script_path, gtype, title, msg, parameters)
+                if ret:
+                    continue
+                else:
+                    break
             if ret:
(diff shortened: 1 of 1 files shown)


Safihre:
Hmmm, this seems like a bad idea. What if the script failed for a proper reason? For example a server not allowing for API calls?

daimoniac:
You could, perhaps, react specifically to error -11 (which does not ever occur in any script if not deliberately thrown or a true segmentation fault) and only retry in that case. Would that be satisfactory?

Safihre:
Could you not also catch that in your own script?
By calling a `subprocess.POpen`-style? Because that's also what we do.

daimoniac:
The error eludes me completely, I even tried it with an empty script in bash (only shebang line). Even then the error -11 sometimes happens. Therefore, no logic inside the external script can catch the error.

I tried with a docker container (Dockerfile attached) locally, in a vm, on AWS, GKE and Digitalocean Kubernetes - the error happens in about .1% of executions.

Safihre:
I'm sorry but I'll have to close this, since it will just create too many problems for others where they wouldn't want the script to be run 5x on failure!

Hellowlol:
Can you just have a script that calls the real script 5 times on your own computer?

PR #1219, opened 2019-01-18
Py3

kitten77:
Added basic testing for sabnzbd/misc
```python
python -m unittest tests.test_misc.MiscTest
```


Safihre:
Was looking into this today and I think we should use `pytest`, since it builds upon `unittest` and adds a lot of neat extras. Also very widely used :)
Specifically because this way we can create a `fixture` for the `cfg` module, so that we can controle the outcomes. 
https://docs.pytest.org/en/latest/fixture.html

kitten77:
I agree on that one, ``` pytest ``` has a lot of advantages. And it is still being developed so there is tons of plugins that can be useful later on. Also been wondering that it might be better to build a test suite against the development branch, so that we can have a set of test that is both working on py3.x and the old py2.x branch. When we need to change to python3.x it will become clear where and why something is failing?

thezoggy:
sab needs a whole slew of tests. api/ui (config/settings/etc) besides the actual logic of its downloader/assembler/etc.  then sabyenc probably could use its own as well.

Safihre:
Sabyenc already has quite a lot of tests, they are in the `test` folder.
The PR is open now to add basic tests, we'll work on it there.

Safihre:
@kitten77 / @OutOfGears I fixed our `cfg` problem with a nice decorator, which I borrowed from `cherrypy` workflow.
Allows us to set a number of variables per-test, which return to default after that.
It's in the `py3` branch! Including your work 👍

kitten77:
I won't have time to do work on this now, might get back to it later on when my schedule allows me.

PR #1221, opened 2019-01-21
Still trying to fix LGTM

Safihre:


Safihre:
Dear @xcorail, I am trying to get LGTM setup since it looks like it gives some nice pointers.
But I am having a lot of difficulties getting this set up correctly.
Could you please help us? I find the documentation a bit short..
I want to ignore multiple directories, but why is this not working? It seems to ignore the other `exclude`'s.
Where can I see the status of the analysis of the main Github branch? I can see if it's processing PR's, but not when it's evaluating commits to `develop`.
```yml
extraction:
  javascript:
    index:
      exclude: interfaces/smpl
      exclude: interfaces/Plush
      exclude: "**/*min*"
  python:
    index:
      exclude: cherrypy
      exclude: gntp
```

xcorail:
Oh right ... `interfaces/Plush` and `interfaces/smpl` are not working either (they are still included), but `gntp` is excluded. Let me dig further and get back to you.

xcorail:
After checking with my colleagues, they suggest to use [file classification](https://lgtm.com/help/lgtm/customizing-file-classification) rather than file exclusion. File classification is designed to do exactly what you want, while exclusion is designed to cope with problems during the extraction process (when some builds are too complex, too slow ...). 
So please try with a `lgtm.yml` file as follows:
```
path_classifiers:
  library:
    - interfaces/smpl
    - interfaces/Plush
    - cherrypy
    - gntp
    - "**/*min*"
```
Let me know how it goes
Cheers

Safihre:
@xcorail Oke that seemed to have fixed it.
But I hope you agree that classifying stuff you don't want analysed as a "library" isn't very intuitive.
A simple global "exclude"-key would probably be easier, not just for me but others too?

xcorail:
@Safihre 
My bad, I put everything under `library` to give you an example, but failed to explain the exact use. 
In fact, these classifiers allows you to add into this configuration file the reason **why** a file is excluded. For example, we have some built-in classifiers such as `library` (some files that are coming from a 3rd party lib), `test`, `doc`, `samples`,`generated`.  And you could put any new one you like, like `minified`, or `interfaces` ... and you can also add a classifier `excluded` if you like :) Just don't use `exclude` which is a reserved keyword. So the below examples will work
```
path_classifiers:
  excluded:
    - interfaces/smpl
    - interfaces/Plush
    - cherrypy
    - gntp
    - "**/*min*"
```

```
path_classifiers:
  interfaces:
    - interfaces/smpl
    - interfaces/Plush
  excluded:
    - cherrypy
  library:
    - gntp
  minified:
    - "**/*min*"
```

Safihre:
Thanks, I will update it like that!

PR #1260, opened 2019-03-21
Py3: Should we use Python function annotations

Safihre:
@thezoggy @sanderjo @Hellowlol and anyone else that has an opinion 👍 

Python 3.5+ supports a way to be more clear about what kind of parameters functions expect, just like almost every other programming language.
More background information can be found here:
https://dev.to/dstarner/using-pythons-type-annotations-4cfe

For this discussion I ran the `pyannotate` tool and applied it's automatically deducted types to `misc.py` and `encoding.py`.
Is this something we would like to start using?

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -14,7 +14,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
+from pyannotate_runtime import collect_types
+collect_types.init_types_collection()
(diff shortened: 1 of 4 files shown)


sanderjo:
I find the comments in https://dev.to/dstarner/using-pythons-type-annotations-4cfe interesting (and agree with them).

So:
- python itself doesn't care
- is there a tool at all that you can run to check the annotations?

With annotations I can write this very confusing code, but it still works ...

```
def myfunction(a: str) -> int:
    print(a*10)
    return "Yes, OK"

myfunction(111)
```
No problem 
```
$ python3 wrong_annotations.py 
1110
```

PR #1266, opened 2019-04-07
Added new tests for utils

sauravk90:
Added test for newsunpack


tests/test_newsunpack.py
```diff
@@ -0,0 +1,34 @@
...
+import subprocess
+from sabnzbd.newsunpack import *
+from tests.testhelper import *
+
+
+class TestNewsUnpack():
+    def test_list_to_cmd(self):
+        """ Test to convert list to a cmd.exe-compatible command string """
+        lst = ['cmd1', 9, 'cmd3']
+        res = list2cmdline(lst)
```
Safihre: Should also test what happens with `None`, `0` and `""`,

tests/test_utils/test_cert_gen.py
```diff
@@ -0,0 +1,39 @@
...
+
+from sabnzbd.utils.certgen import generate_key, generate_local_cert
+
+
+class TestCertGen:
+    def test_generate_key(self):
+        # Generate private key
+        private_key = generate_key()
+
+        assert private_key.key_size == 2048
```
Safihre: Should also check if the output file is written to disk at `outfile` location.
Should also check if the `key_size` of the result changes if specified differently when calling the function.

tests/test_utils/test_cert_gen.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_generate_local_cert(self):
+        # Generate private key
+        private_key = generate_key()
+        # Generate local certificate using private key
+        local_cert = generate_local_cert(private_key)
```
Safihre: Should also check the actual file.
Should do some checks on the resulting certificate, like expiration date and SAN-list.

tests/test_utils/test_diskspeed.py
```diff
@@ -0,0 +1,39 @@
...
+            dir = "/media/sander/"
+
+        elif "win32" in sys.platform:
+            dir = "D:/"
+
+        elif "darwin" in sys.platform:
+            dir = "/Volumes/CARTUNES/"
+
+        speed = diskspeedmeasure(dir)
+        # TODO : Provide a valid directory
```
Safihre: You can just use 
```python
import os
cwd = os.getcwd()
```
Directory doesn't really matter, since it should just be able to run anywhere.
After the test you should check if the `outputTESTING.txt` file is not present in the directory.

tests/test_utils/test_internetspeed.py
```diff
@@ -0,0 +1,45 @@
...
+class TestInternetSpeed:
+    """ This class contains tests to measure internet speed with an active and inactive connection """
+    def test_measurespeed_invalid_url(self):
+        speed = measurespeed("www.fake-url-9999999.xyz")
+
+        assert isinstance(speed, float)
+        assert speed == 0.0
+
+    def test_measurespeed_valid_url(self):
+        speed = measurespeed("https://sabnzbd.org/tests/internetspeed/5MB.bin")
```
Safihre: Maybe use 
```python
from sabnzbd.utils.internetspeed import internetspeed, measurespeed, SizeUrlList
speed = measurespeed(SizeUrlList[0])
```
this way the URL only has to be changed in 1 location.

sauravk90: thanks, will accommodate all the above comments.

tests/test_newsunpack.py
sauravk90: Added more scenarios (parametrized). Also, one of the test with input as `["", "cmd1", "5"]` fails. I think there must be a check for `len(arg)>0 `in `list2cmdline`.

tests/test_utils/test_cert_gen.py
sauravk90: Certificate validation added.

tests/test_utils/test_diskspeed.py
sauravk90: Changes done.

tests/test_utils/test_internetspeed.py
sauravk90: Thanks for the suggestion, I have used the `SizeUrlList` variable for test now.

tests/test_utils/test_cert_gen.py
sauravk90: Added new test with multiple scenarios.

PR #1292, opened 2019-05-22
Reduce CPU load from downloader thread

puzzledsab:
Reduce CPU load from downloader thread
See https://forums.sabnzbd.org/viewtopic.php?f=11&t=24206 for more info.


Safihre:
Interesting!
So let's also do a thought experiment, how would we design the perfect queue system if we could start from scratch?
Starting with the main requirements:
- It needs to support multiple servers at the same or different priority settings.
- The queue should be sortable at any time.
- We can't hold every article in memory, it's too costly.

From there we could go on and consider the actual downloading. How do we manage the status of each connection? How do we handle decoding of the Yenc articles? For that I can already say we have `sabyenc` which can do the decoding outside of the GIL.
I looked into many things like `asyncio` and `twisted` but never managed to fully understand it.

But let's consider these requirements above first :)

puzzledsab:
Regarding this commit, I have reconsidered some of the numbers slightly. I think 4 seconds on line 439 is too long, it's probably better to make it 2. If the server is semi idle it doesn't matter that it has to wait an extra second. 1 might be too low for servers with only 1 connection, because it would have no busy threads fairly often.

If transfer speed is reduced then try idle_count > 1 or higher on line 517. Slow CPUs may also benefit noticeably from skipping unused servers for more than 1 second (the tested_time test). Maybe a configurable number for aggressiveness could be added and used to adjust all these values.

When it comes to a rewrite, I don't understand the complexity well enough to add much of use. I have however noticed that in all the cases where things are really slow and it's not caused by external programs, almost all the load is created by one or more of these: NzbObject.get_article, NzbFile.get_article, Article.get_article, NzbObject.server_in_try_list and the related __getstate__ and  __setstate__ functions.

This may not be very popular in an object oriented environment, but I suspect that you could get a huge speed increase by using a multidimensional dictionary which contained all the data. Then one function could iterate over everything with less overhead than always having to go through all the layers of object abstraction.

Particularly for tasks like marking all articles in one nzb as tested if it is too old, but it may also make it faster to fetch several articles in a row instead of one at a time when there are several idle connections. Of course, I realize that because of my very limited understanding of the queue system I may be completely wrong.

puzzledsab:
I will do some changes and resubmit.

PR #1294, opened 2019-05-30
Optimization changes

puzzledsab:
Sorry, I haven't used git for a while and had forgotten that I need to put commits in separat branches to make separate pull requests.

First commit avoids the 9 second wait for each file if the temp drive is separate from the final destination.

Regarding the downloader changes:
- The decoder if rearrangement is just nitpicking. It will now do the most time consuming test last. Returns available queue spots so the server retention loop can avoid filling it and trigger delayed mode. It can stall downloading for a long time.
- Info logging of "Job %s too old for..." was removed because it can be triggered many times if the queue is filled.
- I added a test for the loop time to avoid sleeping if the rest of the loop takes too long (> 0.0012 sec). This should reduce the chance of impacting the download speed.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -174,10 +174,6 @@ class Downloader(Thread):
         cfg.bandwidth_max.callback(self.speed_set)
         self.speed_set()
 
-        # Used to see if we can add a slowdown to the Downloader-loop
-        self.can_be_slowed = None
-        self.can_be_slowed_timer = 0
-
         self.postproc = False
 
         self.shutdown = False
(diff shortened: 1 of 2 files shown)


thezoggy:
Wouldn't this break Sab for most as lots of people run Sab on a different is than where the final media ends up (Nas)

puzzledsab:
I'm not sure what you mean? That is the kind of setup that the rename patch is intended to help. The files will be moved like before, by using copy and delete. Currently it will first try rename three times with a 3 second delay after each failed attempt, which can't work if the target dir is on a separate volume. It causes a 9 second delay for each file that needs to be moved to the NAS, even if the files are just 1 KB each.

jeffgeorge:
Right, I've not tested it personally as I don't run SABnzbd on Windows, but that first commit looks like it will make life better explicitly *for* people who run SABnzbd with their local cache directory on a different mounted drive than their completed directory, as every file-complete move operation will skip straight to a move rather than always having to spin 3 times trying to do a rename and failing.

puzzledsab:
I removed the can_be_slowed stuff because it's not necessary with the sleep patch. Also I have a hypothesis that it may be the cause of why some have seen the speed rise to maximum and then go down. Sab currently polls the sockets very frequently. If the network delivers data in batches, for instance because of a large transfer window, the 16KB buffer in sockets will fill up and be emptied very quickly. Then there will be a period in which no data is delivered, and the 0.01 second delay triggers. During that time the buffers will refill and not be emptied, which causes the network to slow down the transfer speed.

Safihre:
I'll have to review that, but I'm pretty sure it's not the reason why people see the speed go up and and then down. That's just simply the article and/or decoder cache being filled up and the `delay` being triggered.

PR #1297, opened 2019-06-04
Don't try move between different disks on Windows

puzzledsab:
Adapted to develop because three people in this thread said they moved to nzbget because of slow unrar using SABnzbd: https://www.reddit.com/r/usenet/comments/bwff5a/sabnzbd_or_nzbget/

Unrar should be the same so they probably mean moving.

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -1128,30 +1128,41 @@ def renamer(old, new):
     logging.debug('Renaming "%s" to "%s"', old, new)
     if sabnzbd.WIN32:
         retries = 15
-        while retries > 0:
-            # First we try 3 times with os.rename
-            if retries > 12:
-                try:
-                    os.rename(old, new)
-                    return
-                except:
-                    retries -= 1
-                    time.sleep(3)
-                    continue
 
-            # Now we try the back-up method
-            logging.debug('Could not rename, trying move for %s to %s', old, new)
+        # splitdrive only supports drive letters and splitunc only supports UNC mount points in Python 2
+        old_disk = os.path.splitdrive(old)[0]
+        new_disk = os.path.splitdrive(new)[0]
+        if not old_disk:
+            old_disk = os.path.splitunc(old)[0]
+        if not new_disk:
+            new_disk = os.path.splitunc(new)[0]
+        if old_disk != new_disk:
+            logging.debug('Different disks ("%s" != "%s"), skipping rename', old_disk, new_disk)
+            retries = 12
+
+        while retries > 0:
             try:
-                shutil.move(old, new)
-                return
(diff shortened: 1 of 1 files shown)


Safihre:
Maybe instead we should just revert https://github.com/sabnzbd/sabnzbd/commit/b8c4f1a09a44c86fc39a1c8c79e2c8e7f0b34c3c.
I understand that `shutil.move` just uses `os.rename` [anyway](https://stackoverflow.com/questions/8858008/how-to-move-a-file-in-python) I am not sure really why I made this change in the first place..

Safihre:
Seems my reasoning was that on Windows the file can be locked by virusscanners etc (happens often), resulting in the same general `OSError` result as other errors. This will then cause it to immediately switch to the slow copy operation.

puzzledsab:
I think the best option is to keep your patch and add mine, then.

Safihre:
How about this?
Taking into account the requirements:
- Only retry when the file is busy
- Don't wait when it's different disk problem
- Try to not have too much extra logic

```python
    logging.debug('Renaming "%s" to "%s"', old, new)
    if sabnzbd.WIN32:
        retries = 15
        while retries > 0:
            try:
                # First we try 3 times with os.rename
                if retries > 12:
                    os.rename(old, new)
                else:
                    # Now we try the back-up method
                    logging.debug('Could not rename, trying move for %s to %s', old, new)
                    shutil.move(old, new)
            except WindowsError, err:
                logging.debug('Error renaming "%s" to "%s" <%s>', old, new, err)
                if err[0] == 17:
                    # Error 17 - Rename can't move to different disk
                    # Jump to moving with shutil.move
                    retries -= 3
                elif err[0] == 32:
                    # Error 32 - Used by another process
                    logging.debug('File busy, retrying rename %s to %s', old, new)
                    retries -= 1
                    # Wait for the other process to finish
                    time.sleep(2)
                else:
                    break
        raise WindowsError(err)
    else:
        shutil.move(old, new)
```

puzzledsab:
I can't test the specific code because I don't have Python 2.7 on Windows, but I assume it will do the same. You could probably do more retries with rename now that it's a caused by a genuine error.

Safihre:
Force pushed it on your commit! Nice to figure this out :)

Safihre:
Oh, seems this fails for unicode downloads.. At least on Py3. 
Oops: https://ci.appveyor.com/project/Safihre/sabnzbd/builds/25137223/job/ynsc5iqa65l42hyp
`os.rename()` can't handle those it seems in long-path notation.

Safihre:
Oops, I'm just a bit stupid. Forgot the `break` after the rename was finished 😨 Fixing.

PR #1298, opened 2019-06-06
Fetch multiple articles and put them in a queue on the server object

puzzledsab:
This potentially uses less CPU time than fetching one at a time. The sleep time for the sleep between each download poll is made configurable. 0.001 sec does not seem to do any harm on my Rpi 3 at 6.5 MB/s, but it needs to be higher for the cpu usage to go down compared to the current py3 version. Some users with high bandwidth may benefit from setting it to 0.

The method for purging the queue and resetting the try_list at shutdown needs improvement. Regarding rearranging the nzb queue, it will currently finish the articles in the article queue and use the updated order when it fetches new articles, similar to before. Alternatively purge_article_queue could be called after a change of the order. I'm not sure where that should be triggered.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -179,6 +179,7 @@ def sig_handler(signum=None, frame=None):
     if signum is not None:
         logging.warning(T("Signal %s caught, saving and exiting..."), signum)
     try:
+        Downloader.do.set_paused_state(True)
         save_state()
         sabnzbd.zconfig.remove_server()
     finally:
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -84,6 +84,7 @@ class Server:
 
         self.busy_threads = []
         self.idle_threads = []
+        self.article_queue = []
         self.active = True
         self.bad_cons = 0
         self.errormsg = ''
@@ -401,9 +402,11 @@ class Downloader(Thread):
         # See if there's space left in cache, pause otherwise
         # But do allow some articles to enter queue, in case of full cache
         qsize = self.decoder_queue.qsize()
+
         if (qsize > LIMIT_DECODE_QUEUE) or (lines and qsize > MAX_DECODE_QUEUE and not ArticleCache.do.reserve_space(lines)):
             sabnzbd.downloader.Downloader.do.delay()
             return 0
+
         return LIMIT_DECODE_QUEUE - qsize
 
     def run(self):
@@ -441,8 +444,8 @@ class Downloader(Thread):
                 if server.busy_threads:
                     last_busy[serverid] = time.time()
 
-                # Skip this server if idle for 2 seconds and has already been tested less than 2 second ago
-                if last_busy.get(serverid, 0) + 2 < time.time() and tested_time.get(serverid, 0) + 2 > time.time():
+                # Skip this server if idle for 4 seconds and has already been tested less than 1 second ago
+                if not server.article_queue and last_busy.get(serverid, 0) + 2 < time.time() and tested_time.get(serverid, 0) + 2 > time.time():
                     continue
                 tested_time[serverid] = time.time()
 
@@ -469,6 +472,7 @@ class Downloader(Thread):
                         # Restart pending, don't add new articles
                         continue
 
+
                 if not server.idle_threads or server.restart or self.is_paused() or self.shutdown or self.delayed or self.postproc:
                     continue
 
@@ -476,6 +480,9 @@ class Downloader(Thread):
                     continue
 
                 for nw in server.idle_threads[:]:
+                    if self.shutdown or self.is_paused():
+                        break
+
                     if nw.timeout:
                         if time.time() < nw.timeout:
                             continue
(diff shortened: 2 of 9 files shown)


Safihre:
Could you maybe remove the commits not related to multi-article-fetch? :)
Then I can give it a proper test with all the edge cases I can think off.

puzzledsab:
I have reverted a couple of the separate commits, but the rest is kind of integrated because each commit builds on the previous. I realize there are actually 3 changes, the one skipping unused servers, the sleep and the multifetch. The sleep can be cancelled by commenting out the two lines (if and sleep in downloader:run). I hope you can test it like this.

thezoggy:
may want to squash + rebase and force push to rewrite history so its much easier to review.

Safihre:
Will run this code as my main SAB instance and let's see.
The code still feels a bit out of my comfort zone, but we have to start somewhere.

sabnzbd/__init__.py
```diff
@@ -179,6 +179,7 @@ def sig_handler(signum=None, frame=None):
     if signum is not None:
         logging.warning(T("Signal %s caught, saving and exiting..."), signum)
     try:
+        Downloader.do.set_paused_state(True)
```
puzzledsab: This is done to make the article queue get purged and stop it from adding new articles to the queue.

sabnzbd/downloader.py
```diff
@@ -401,8 +402,12 @@ def decode(self, article, lines, raw_data):
         # See if there's space left in cache, pause otherwise
         # But do allow some articles to enter queue, in case of full cache
         qsize = self.decoder_queue.qsize()
-        if (not ArticleCache.do.reserve_space(lines) and qsize > MAX_DECODE_QUEUE) or (qsize > LIMIT_DECODE_QUEUE):
+
+        if (qsize > LIMIT_DECODE_QUEUE) or (lines and qsize > MAX_DECODE_QUEUE and not ArticleCache.do.reserve_space(lines)):
             sabnzbd.downloader.Downloader.do.delay()
+            return 0
+
+        return LIMIT_DECODE_QUEUE - qsize
```
puzzledsab: decode is rewritten to let the caller know if the queue is filled to fast, particularly related to articles that are out of retention. The if statement is rearranged to avoid calling ArticleCache.do.reserve_space if it isn't necessary.

sabnzbd/downloader.py
```diff
@@ -420,8 +425,30 @@ def run(self):
...
         while 1:
+            idle_count += 1
             for server in self.servers:
+                serverid = server.id
+                if server.busy_threads:
+                    last_busy[serverid] = time.time()
+
+                # Skip this server if idle for 4 seconds and has already been tested less than 1 second ago
+                if not server.article_queue and last_busy.get(serverid, 0) + 2 < time.time() and tested_time.get(serverid, 0) + 2 > time.time():
+                    continue
```
puzzledsab: Sorry, just noticed the comment is out of date. It will test a server for new articles once every other second if it hasn't been active (found any articles to download) for the last 2 seconds.

sabnzbd/downloader.py
```diff
@@ -465,18 +496,25 @@ def run(self):
                             request_server_info(server)
                         break
 
-                    article = sabnzbd.nzbqueue.NzbQueue.do.get_article(server, self.servers)
+                    if not server.article_queue:
+                        fetch_amount = server.threads * 2
+                        if fetch_amount > 50:
+                            fetch_amount = 50
+                        server.article_queue = sabnzbd.nzbqueue.NzbQueue.do.get_articles(server, self.servers, fetch_amount)
```
puzzledsab: I don't really know if there is a good number of articles to fetch. The benefit is reduced by fetching too few but 50 may be excessive.

sabnzbd/downloader.py
```diff
@@ -465,18 +496,25 @@ def run(self):
...
                     if server.retention and article.nzf.nzo.avg_stamp < time.time() - server.retention:
-                        # Let's get rid of all the articles for this server at once
-                        logging.info('Job %s too old for %s, moving on', article.nzf.nzo.final_name, server.host)
-                        while article:
-                            self.decode(article, None, None)
-                            article = article.nzf.nzo.get_article(server, self.servers)
-                        break
+                        if self.decode(article, None, None) < LIMIT_DECODE_QUEUE * 0.5:
+                            break # Try to avoid delayed mode
+                        continue
```
puzzledsab: This part is rewritten mostly because triggering delayed mode can slow down the downloading significantly for a long time, but also because the queue makes it unnecessary to use article.nzf.nzo.get_article. Unfortunately it also makes it messier to keep the "Job %s too old" info message. It would require a test to check if the server has decoded any parts of that nzb previously, otherwise it will be logged may times.

sabnzbd/downloader.py
```diff
@@ -529,24 +576,6 @@ def run(self):
...
-                    if not self.can_be_slowed_timer and BPSMeter.do.get_stable_speed(timespan=10):
-                        self.can_be_slowed_timer = time.time()
-
-                    # Check 10 seconds after enabling slowdown
-                    if self.can_be_slowed_timer and time.time() > self.can_be_slowed_timer + 10:
-                        # Now let's check if it was stable in the last 10 seconds
-                        self.can_be_slowed = BPSMeter.do.get_stable_speed(timespan=10)
-                        self.can_be_slowed_timer = 0
-                        logging.debug('Downloader-slowdown: %r', self.can_be_slowed)
-
```
puzzledsab: Removed because the sleep on idle_count should be enough. From what I understand this part is not necessary for the download throttling to work.

sabnzbd/downloader.py
```diff
@@ -912,6 +942,19 @@ def stop(self):
...
+        for server in self.servers:
+            if server.article_queue:
+                for article in server.article_queue:
+                    article.fetcher = None
+                    article.reset_try_list()
+                    article.nzf.reset_try_list()
+                    article.nzf.nzo.reset_try_list()
+                    article.tries -= 1
+                server.article_queue = []
+                logging.debug('article_queue wiped for %s', server.id)
```
puzzledsab: This function is the part I'm most in doubt about. I'm not sure if it is sufficient to avoid missing articles after shutdown, server restart or similar in all cases, or if it should be called more or different places than it currently is.

puzzledsab:
> may want to squash + rebase and force push to rewrite history so its much easier to review.

I have very basic Git skills so I'm afraid I'll break something and make it hard to merge if I try that. Are you thinking of making everything look like a single commit? Are there any advantages to that compared to using the Files Changed on this pull request to see the changes?

I have been running this version continuously on my Windows server since I started creating it and I haven't had any problems lately. I've also done some testing on Raspberry Pi 3 using Python 3.5.

Regarding the get_articles functions they will only fetch available articles inside the same file. Things got messier when I tried span multiple files or nzbs and the benefit didn't seem very big.

Sorry about all the comments, I thought they would only show up on https://github.com/sabnzbd/sabnzbd/pull/1298/files.

PR #1327, opened 2019-09-08
JSON response types are mismatched

ohrodr:
JSON provides types for int and float.  Changing these values to a string before returning them requires the user/consumer of the API to write conversions.  Seems counter-intuitive.  I think it makes more sense to actually respond with the type expected.  It'll make consumers in static typed languages easier to implement.

Also direct_unpack would sometimes be a string, and other times be an integer.  This makes for more code to static type language consumers.

I looked through the code, and didn't see any reason this would break, except for other API consumers who can now rid themselves of any conversion code.

There are a few more, but I haven't begun consuming them yet so I didn't fix them.

One could also make a good argument that the size should be the numbers, and the mb value should be the string, but that's not a hill to die on.

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1340,13 +1340,13 @@ def build_queue(start=0, limit=0, trans=False, output=None, search=None):
         slot['filename'] = converter(pnfo.filename)
         slot['password'] = converter(pnfo.password) if pnfo.password else ''
         slot['cat'] = converter(pnfo.category) if pnfo.category else 'None'
-        slot['mbleft'] = "%.2f" % mbleft
-        slot['mb'] = "%.2f" % mb
+        slot['mbleft'] = round(mbleft, 2)
+        slot['mb'] = round(mb, 2)
         slot['size'] = format_bytes(bytes)
         slot['sizeleft'] = format_bytes(bytesleft)
-        slot['percentage'] = "%s" % (int(((mb - mbleft) / mb) * 100)) if mb != mbleft else '0'
-        slot['mbmissing'] = "%.2f" % (pnfo.bytes_missing / MEBI)
-        slot['direct_unpack'] = pnfo.direct_unpack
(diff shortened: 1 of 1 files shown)


thezoggy:
Changing the API responses now would break apps/gui and other scripts/plugins. This really shouldn't be changed on 2.x.  the reason why stuff was done in this way was because back in the day Sab existed before jaon and output was text. Then xml format was added then. Later on json. It was normalized to be all text so they worked the same.... In v3 we should probably drop everything but json.

ohrodr:
I get that, and have been using sab for quite awhile.  While writing an automation in a static typed language where specifications matter, I had to jump through all sorts of reflection and type assertion and value conversions that just add unnecessary weight to the code.  A scripting language like python would care a lot less by the type conversions, but with static typed languages the encoding/serialization method should be as conformant as possible to ensure portability.

Getting rid of plain text isn't really the problem.  The problem is incorrect type assertion within the serialization output.

There are also instances where a value will be a string in one occurrence, and an integer in another which poses problems as well, and isn't aligned with the intent of all output values being typecast as strings.  Those have work arounds, or don't matter to dynamic type consumers.

How far along is the v3 work?  Do you want me to fix this in totality based upon the premise that XML/text is being deprecated?  Point me to the branch to work off and it can be resolved relatively quickly.

Safihre:
Although I think the effect will be minimal, indeed this should be part of the v3 version. Could you move your PR to the `py3` branch? As the name suggest, the v3 version also only runs on Python 3 so might have to do a quick compatibility check of the code. 

We're pretty far with the v3 version, but it's a bit stuck at the moment due to lac of motivation to fix all those annoying little last issues.. Working in it!

thezoggy:
Back when Sab added json output we used our own serializer as it wasn't native in python just yet... But obviously in py3 native json lib should be used

Safihre:
@ohrodr Could you update the PR to use the new `develop` branch that's now Python 3 based? 👍

Issue #1319, opened 2019-08-15
py3: suggestion "3" in logo? Or some other 3 indicator


sanderjo:
Sometimes it's unclear to me which SAB version (py2 or py3) I'm running. That will also be the case in the transition period of users going from py2 to py3.

So, how about something on the frontpage that makes that clear?

As an example, I hacked a "3" into the SVG icon on the frontpage (`inkscape interfaces/Config/templates/staticcfg/images/logo-small.svg`):

![image](https://user-images.githubusercontent.com/1273502/63104558-81be6080-bf7f-11e9-9605-0073ab8ee81e.png)

There will probably better solutions than this. 

Or white-yellow font color:

![image](https://user-images.githubusercontent.com/1273502/63214193-638a6900-c115-11e9-8225-da21dfb41b08.png)

sanderjo:
PS: @thezoggy did you create the SVG? If so, I really look forward to your feedback.

jeffgeorge:
Nope, that was me. Rather than tweaking the SVG, I think it might be a bit easier (and easier to roll back) if we did this in css. I'll take a look at what that might take to implement.

thezoggy:
I think i did the .ai files but gave them to jeff to process/convert.

sanderjo:
... and how about a "2" in to-be-released SABnzbd 2.x versions?

Issue #521, opened 2016-03-24
Modernize SABnzbd's Branding & Logos


jeffgeorge:
(Picking up where we left off in #408, around [here](https://github.com/sabnzbd/sabnzbd/issues/408#issuecomment-180875625).)

Goals:
- [x] New Arrow icon to use, with an original flat design
- [x] New Overall logo (so, arrow icon + type)
- [x] Vector assets for use on a new SABnzbd.org homepage and the Glitter theme
- [ ] Primary (orange) .ico assets to use throughout the application itself (
- [ ] Secondary (green) .ico assets to use for .nzb file associations

I've been working with a friend to iterate on where we got in the previous issue, and have ended up with this for the first two items:

https://resources.sabnzbd.org/logo/sabnzbd_arrow_sharp_3.ai

![sabnzbd_logo_composited](https://cloud.githubusercontent.com/assets/806112/14032006/594bd610-f1e6-11e5-81e2-a648deaf059e.png)

Some notes:
- @Safihre recommended in the previous thread to only use "sharp" corners in the strokes. Turns out that's basically impossible in Photoshop. If you do an outside stroke it's always rounded. If you do an inside stroke it's sharp, but then you're eating into the inside dimensions, so the "stripes" going to the upper corners no longer go straight to the corner. Someone may be able to fix this in Illustrator or some other vector-based tool that can read .ai's, but I'm both terrible with Illustrator and don't have access to a copy.
- I went back and forth on colors (various [Material colors](https://www.google.com/design/spec/style/color.html#color-color-palette) out of the "amber" segment) and number of segments (3 vs 5) multiple times. I landed on Amber 400 (#FFCA28) / Amber 600 (#FFB300), and 3 segments. I'm totally open to fiddling with exactly which Ambers to use, but think it should stay at 3 segments.

Next Steps:
- **Feedback**! Please provide it! Please tell me why this sucks!
- Once we like something, drag in someone competent with Illustrator to composite some new svgs with strokes and such
- Once we've got the Amber version we like, we need to make a green version
- Once we've done all that, we get to compile new icons, which is the worst and most thankless job in the world
- Then, finally, after all that, we can open a PR to implement all these new icons throughout Glitter and the file build process

shypike:
Looks very nice.

Safihre:
Yes looks good! :+1: 
[...]

PR #1333, opened 2019-09-24
"3" in SABnzbd logo on frontpage

sanderjo:
as discussed https://github.com/sabnzbd/sabnzbd/issues/1319

We can always improve later on, but I need it now in testing.


thezoggy:
perhaps the `3` should be done in the same font that 'sab' is in?
I believe it is `FFFFREED.TTF` if you want me to share it

sanderjo:
Oh, the text "SAB" is a _font_ ? I thought a patient person had drawn something.

Anyway ... as you seem to be the expert on this, maybe quicker if you do it, including PR?

thezoggy:
I can, but @Safihre / @inpheaux  actually messed with it the most and may be able to get to this before I can: https://github.com/sabnzbd/sabnzbd/issues/521

jeffgeorge:
For reference: https://www.dafont.com/fff-freedom.font

Not sure what license the font is available under, but it's freely available from dafont.

Issue #1203, opened 2018-12-25
Python 3 conversion status


Safihre:
It seems I will need to process every single file function by function.. So just here to keep track.

- [x] `cherrypy` - Replace by `requirements.txt`
- [x] `gntp` - Replace by `requirements.txt`
- [ ] `interfaces`
  - [x] Removed smpl, references in other files still remain.
- [ ] `licenses` (needs to be updated after all this is done)
- [ ] `sabnzbd`
  - [ ] `utils` 
    - [x] `__init__.py` (deleted for now)
    - [x] `certgen.py`
    - [ ] `checkdir.py`
    - [x] `diskspeed.py`
    - [x] `getperformance.py`
    - [ ] `happyeyeballs.py`
    - [x] `kronos.py`
    - [x] `pathbrowser.py`
    - [x] `pybonjour.py`
    - [x] `pystone.py`
    - [x] `rarfile.py`
    - [x] `rsslib.py` - Removed RSS-output capabilities in 2.3.x
    - [x] `servertests.py`
    - [x] `systrayiconthread.py`
    - [x] `upload.py`
  - [ ] `__init__.py`
  - [ ] `api.py`
  - [x] `articlecache.py`
  - [ ] `assembler.py`
  - [x] `bpsmeter.py`
  - [x] `cfg.py`
  - [x] `config.py`
  - [ ] `constants.py`
  - [ ] `database.py`
  - [ ] `decoder.py`
  - [ ] `decorators.py`
  - [ ] `directunpacker.py`
  - [ ] `dirscanner.py`
  - [ ] `downloader.py`
  - [ ] `emailer.py`
  - [ ] `encoding.py`
  - [ ] `filesystem.py` - New
  - [ ] `getipaddress.py`
  - [ ] `interface.py`
  - [x] `lang.py`
  - [ ] `misc.py`
  - [ ] `newsunpack.py`
  - [ ] `newswrapper.py`
  - [ ] `notifier.py`
  - [x] `nzbparser.py` - New
  - [ ] `nzbqueue.py`
  - [ ] `nzbstuff.py`
  - [ ] `osxmenu.py`
  - [ ] `panic.py`
  - [x] `par2file.py`
  - [ ] `postproc.py`
  - [ ] `powersup.py`
  - [ ] `rating.py`
  - [ ] `rss.py`
  - [ ] `sabtraylinux.py`
  - [x] `sabtray.py`
  - [x] `scheduler.py`
[...]

PR #1334, opened 2019-09-28
Ongoing code cleanup for Python 3 conversion

pl77:
Just answering the "help wanted" badge on https://github.com/sabnzbd/sabnzbd/issues/1203.

I'll open this pull request to track my adjustments to the develop branch as I work through some of the functions and to prevent our work from overlapping.  

Please let me know if I'm breaking sabnzbd code conventions or you prefer me to focus on something else.  I'm doing my best not to change any of the intent of the underlying code, only clean it up for Python 3.5. 

Also, please let me know if Python 3.5 is still the target.  I could really clean up string formatting if I could use f-strings from 3.6!

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -795,7 +795,6 @@ if sabnzbd.WIN32:
         import win32api as w32
     except ImportError:
         w32 = None
-        pass
 
 
     def diskspace_base(_dir):
@@ -830,8 +829,7 @@ else:
             except OSError:
                 return 0.0, 0.0
 
-    except OSError:
-
+    except ImportError:
         def diskspace_base(_dir):
             return 20.0, 10.0
 
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -791,19 +791,21 @@ if sabnzbd.WIN32:
     # windows diskfree
     try:
         # Careful here, because win32api test hasn't been done yet!
-        import win32api as w32
+        import win32api
+
+        def diskspace_base(_dir):
+            """ Return amount of free and used diskspace in GBytes """
+            _dir = find_dir(_dir)
+            try:
+                available, disk_size, total_free = win32api.GetDiskFreeSpaceEx(_dir)
+                return disk_size / GIGI, available / GIGI
+            except (win32api.error, AttributeError):
+                return 0.0, 0.0
     except ImportError:
-        w32 = None
+        win32api = None  # code inspector wants this declared if the exception is an ImportError.
 
-    def diskspace_base(_dir):
-        """ Return amount of free and used diskspace in GBytes """
-        _dir = find_dir(_dir)
-        try:
-            available, disk_size, total_free = w32.GetDiskFreeSpaceEx(_dir)
-            return disk_size / GIGI, available / GIGI
-        except (w32.error, ModuleNotFoundError, AttributeError):
+        def diskspace_base(_dir):
             return 0.0, 0.0
-
 
 else:
     try:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -39,7 +39,7 @@ def get_ext(filename):
     """ Return lowercased file extension """
     try:
         return os.path.splitext(filename)[1].lower()
-    except:
+    except (OSError, IndexError):
         return ""
 
 
@@ -47,7 +47,7 @@ def get_filename(path):
     """ Return path without the file extension """
     try:
         return os.path.split(path)[1]
-    except:
+    except (OSError, IndexError):
         return ""
 
 
@@ -122,7 +122,7 @@ def has_win_device(p):
 
 CH_ILLEGAL = "/"
 CH_LEGAL = "+"
-CH_ILLEGAL_WIN = '\/<>?*|"\t:'
+CH_ILLEGAL_WIN = '\\/<>?*|"\t:'  # Clean up escape sequence
 CH_LEGAL_WIN = "++{}!@#'+-"
 
 
@@ -143,7 +143,7 @@ def sanitize_filename(name):
 
     if ":" in name and sabnzbd.DARWIN:
         # Compensate for the foolish way par2 on OSX handles a colon character
-        name = name[name.rfind(":") + 1 :]
+        name = name[name.rfind(":") + 1:]
 
     lst = []
     for ch in name.strip():
@@ -242,8 +242,8 @@ def sanitize_files_in_folder(folder):
                 try:
                     os.rename(path, new_path)
                     path = new_path
-                except:
-                    logging.debug("Cannot rename %s to %s", path, new_path)
+                except OSError:
+                    logging.debug("Cannot rename {} to {}".format(path, new_path))
             lst.append(path)
     return lst
 
@@ -300,16 +300,16 @@ def create_real_path(name, loc, path, umask=False, writable=True):
     if path:
         my_dir = real_path(loc, path)
         if not os.path.exists(my_dir):
-            logging.info("%s directory: %s does not exist, try to create it", name, my_dir)
+            logging.info("{} directory: {} does not exist, try to create it".format(name, my_dir))
             if not create_all_dirs(my_dir, umask):
-                logging.error(T("Cannot create directory %s"), clip_path(my_dir))
+                logging.error("Cannot create directory {}".format(clip_path(my_dir)))
                 return False, my_dir
 
         checks = (os.W_OK + os.R_OK) if writable else os.R_OK
         if os.access(my_dir, checks):
             return True, my_dir
         else:
-            logging.error(T("%s directory: %s error accessing"), name, clip_path(my_dir))
+            logging.error("{} directory: {} error accessing".format(name, clip_path(my_dir)))
             return False, my_dir
     else:
         return False, ""
@@ -337,7 +337,7 @@ def same_file(a, b):
         if os.path.samefile(a, b) is True:
             return 1
         return is_subfolder
-    except:
+    except OSError:
         if int(a == b):
             return 1
         else:
@@ -359,7 +359,7 @@ def check_mount(path):
         for n in range(sabnzbd.cfg.wait_ext_drive() or 1):
             if os.path.exists(m.group(1)):
                 return True
-            logging.debug("Waiting for %s to come online", m.group(1))
+            logging.debug("Waiting for {} to come online".format(m.group(1)))
             time.sleep(1)
     return not m
 
@@ -413,8 +413,8 @@ def fix_unix_encoding(folder):
                 if name != new_name:
                     try:
                         renamer(os.path.join(root, name), os.path.join(root, new_name))
-                    except:
-                        logging.info("Cannot correct name of %s", os.path.join(root, name))
+                    except OSError:
+                        logging.info("Cannot correct name of {}".format(os.path.join(root, name)))
 
 
 def make_script_path(script):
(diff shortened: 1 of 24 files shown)


thezoggy:
A lot of these changes break Sab. The T variable is used for translation (language) support.

sabnzbd/filesystem.py
```diff
@@ -792,23 +792,25 @@ def find_dir(p):
...
             return disk_size / GIGI, available / GIGI
-        except:
+        except (w32.error, ModuleNotFoundError, AttributeError):
             return 0.0, 0.0
 
 
 else:
     try:
-        os.statvfs
+        # os.statvfs                # commented this because it does not appear to do anything.
```
Safihre: What this does is just to see if `statvfs` is available, if not the `try` will catch it and we skip this function.

sabnzbd/filesystem.py
```diff
@@ -792,23 +792,25 @@ def find_dir(p):
     # windows diskfree
     try:
         # Careful here, because win32api test hasn't been done yet!
-        import win32api
-    except:
-        pass
+        import win32api as w32
+    except ImportError:
+        w32 = None
```
Safihre: Interesting approach, is this the recommended way?

sabnzbd/filesystem.py
```diff
@@ -300,16 +300,16 @@ def create_real_path(name, loc, path, umask=False, writable=True):
     if path:
         my_dir = real_path(loc, path)
         if not os.path.exists(my_dir):
-            logging.info("%s directory: %s does not exist, try to create it", name, my_dir)
+            logging.info("{} directory: {} does not exist, try to create it".format(name, my_dir))
```
Safihre: I enjoy the new `.format()`, but in the case of logging is it really more readable than using the parameters to the `info`/`debug`/etc functions?
That's also the way how we do it in all the other files :)

sabnzbd/filesystem.py
```diff
@@ -300,16 +300,16 @@ def create_real_path(name, loc, path, umask=False, writable=True):
...
             if not create_all_dirs(my_dir, umask):
-                logging.error(T("Cannot create directory %s"), clip_path(my_dir))
```
Safihre: As @thezoggy mentioned, the `T` function is patched in `lang.py` so it's available everywhere as a more general way to translate instead of Python's weird `_` function. So we really need it!

sabnzbd/filesystem.py
```diff
@@ -557,17 +557,17 @@ def get_unique_filename(path):
...
-        fname = "%s.%d%s" % (name, num, ext)
+        fname = "{}.{:d}{}".format(name, num, ext)
         num += 1
         path = os.path.join(new_path, fname)
     return path
 
 
 @synchronized(DIR_LOCK)
-def recursive_listdir(dir):
+def recursive_listdir(target_dir):   # renamed to avoid shadowing out of scope variable
```
Safihre: Great 💯

sabnzbd/filesystem.py
```diff
@@ -648,11 +648,11 @@ def get_filepath(path, nzo, filename):
         for n in range(200):
             new_dirname = dirname
             if n:
-                new_dirname += "." + str(n)
+                new_dirname += ".{}".format(n)
```
Safihre: Here I like the use of `format`!

sabnzbd/utils/certgen.py
```diff
@@ -57,17 +58,17 @@ def generate_local_cert(private_key, days_valid=3560, output_file="cert.cert", L
...
+    try:
         san_list.append(x509.IPAddress(ipaddress.IPv4Address("127.0.0.1")))
         san_list.append(x509.IPAddress(ipaddress.IPv6Address("::1")))
 
         # append local v4 ip
         mylocalipv4 = localipv4()
         if mylocalipv4:
             san_list.append(x509.IPAddress(ipaddress.IPv4Address(str(mylocalipv4))))
-    except:
+    except (crypto_exceptions.InternalError, ipaddress.AddressValueError, ipaddress.NetmaskValueError):
```
Safihre: Are you sure these are all that can occur?

sabnzbd/utils/checkdir.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
Safihre: 👍

sabnzbd/utils/checkdir.py
```diff
@@ -9,13 +9,16 @@
 
 debug = False
 
+# TODO: Rename function isFAT and variable FAT to PEP8-compliant 'is_fat32' and 'fat32. Check for any API impacts first'
```
Safihre: Yes, we should do this 👍 
Just check for it's usages throughout the code.

sabnzbd/utils/checkdir.py
```diff
@@ -54,7 +57,7 @@ def isFAT(check_dir):
                     print(result)
                 if result[4].startswith("FAT"):
                     FAT = True
-            except:
+            except (OSError, SystemError, IndexError, win32api.error):                  # Eliminate broad exception
```
Safihre: Good change, the comment isn't needed.

sabnzbd/utils/diskspeed.py
```diff
@@ -50,22 +50,22 @@ def diskspeedmeasure(dirname):
     print("Let's go")
 
     if len(sys.argv) >= 2:
-        dirname = sys.argv[1]
-        if not os.path.isdir(dirname):
+        directoryname = sys.argv[1]
```
Safihre: 👍

Safihre:
Thanks for the PR! I left some comments in the code.

While the CI/CD pipeline isn't super stable, it does give some feedback. For example we use the `black` code-formatter, so it's good to run that on all the listed files:
https://travis-ci.org/sabnzbd/sabnzbd/jobs/590686688#L421

sabnzbd/filesystem.py
pl77: > A lot of these changes break Sab. The T variable is used for translation (language) support.

Thanks @thezoggy and @Safihre, I will return the T variable back to its original position, lol.  I swear I did search and couldn't find a reference anywhere for it, but that makes way more sense.  

Ignore this update for now and I'll take your feedback into account and add a new commit that deals with this.

sabnzbd/utils/certgen.py
pl77: These were the ones that popped up when I ran a simple test, but you're absolutely right: there are likely more that can occur.  What I'll try and do is add some broader exceptions, but not leaving a bare exception, like BufferError or LookupErrors, that capture more without interfering with real bugs.

sabnzbd/filesystem.py
pl77: Thanks!  I was wondering about that! I will drop the # and add a comment to that effect.

sabnzbd/filesystem.py
pl77: I'm still trying to figure out the recommended way.  Normally an ImportError that isn't dealt with ends the whole script.  Since this keeps going I need to define the variable and catch it when the module isn't there.  

I changed the name just out of habit in case win32api was being called elsewhere to avoid clobbering it.  I mean, it shouldn't be called elsewhere, but I'm still working my way through this codebase and am very unfamiliar with it.  Until I'm more confident I understand it I was playing it safe.  Just like the "T" confusion I caused earlier!

pl77:
> For example we use the `black` code-formatter, so it's good to run that on all the listed files:

Thanks again for the feedback, it was greatly appreciated.  I installed black and incorporated it into my PyCharm.  I also adjusted my line endings to 88 to match it's requirements and set the "T" builtin to be ignored in my own reviews.  These files seem to be passing tests, so I thought I'd do a quick commit to see if I'm on the right track and if the integration tools agree with me!  

Thanks again, this is a fantastic project and you folks are doing a great job.

sabnzbd/filesystem.py
pl77: You are absolutely right, so I removed the 'format' from all logging functions.  As you stated elsewhere, I left it in for the string modifications.  It's a bit more robust when handling integers, etc., so there is some value in using it.

sabnzbd/filesystem.py
pl77: I modified this to a straight function call to ensure Python actually calls it and doesn't ignore it with lazy-loading or whatever happens under the hood.  It's a simple call to the cwd now to make sure it's actually doing something.  It can always be reverted later if it causes a slowdown or other issues.

Safihre:
So we actually use `-l120` (so line-length of 120) since otherwise it feels a bit too short and it makes these weird cvhanges.
Could you run `black` again with `-l120`? :)

pl77:
> So we actually use `-l120` (so line-length of 120) since otherwise it feels a bit too short and it makes these weird changes.
> Could you run `black` again with `-l120`? :)

Oh thank goodness, 88 was really short!  I will rerun and recommit.  Thanks for your patience with me!

pl77:
I'm going to close this mess and spend a bit of time taking a different approach.  It's getting a little unwieldy.

Safihre:
It was going pretty well :) looking forward to the next PR!

Have to say that I'm going on vacation to Africa Friday for 3 weeks, so it will be hard for me to review stuff then!

pl77:
Don't worry, @Safihre, I'm still submitting my changes! 

After spending a few days working on it I feel a bit more comfortable with how the program is structured.  In some ways, it kind of felt like I had jumped into the middle of the lake and then tried to figure out which way I needed to swim.

In particular, I am spending a bit more time with the translation tooling scripts to make sure I don't irreversibly break something while I tweak, and to make sure the tests keep up with my changes.  I'm learning so much from this code base, it's a lot of fun!

My fork is still at the same link (https://github.com/pl77/sabnzbd), but I backed up my old commits to a local repo and started with a fresh clone.  I'm spending some time getting the testing suite fully operational on my dev machine before I commit any of my old code.  

**One question for you: when you do local development do you spin up docker instances to test new builds or just run the test suite in a local environment?** 

I'll put a note in my calendar to make sure a submit a new PR when you're back. I hope you enjoy your vacation!  It sounds amazing!  Africa is one of the countries I haven't visited yet and I'd love to some day.

thezoggy:
@pl77 If i recall correctly you just need to be in sab root dir when you execute it.
For building the translations you need to do:
https://github.com/sabnzbd/sabnzbd/blob/develop/INSTALL.txt#L92

pl77:
Hahaha, you're totally right.  I had to set the proper location in PyCharm for the tests, as well as download a chromedriver.exe that matched my installed chrome, and add some environmental variables.  But now the tests are passing!   Thanks!

Safihre:
I personally created a Powershell script that just changes the directory to the root dir, sets the environment variables and then runs `pytest`. 👍

pl77:
That's a good idea.  I was only trying to get them to run in PyCharm so I could run pause the tests at certain spots with the debugger.  Here's the configuration that ended up working for me, in case anyone reading this is struggling with their own IDE.  

Note the warning about no pytest runner in the selected interpreter.  I'm not sure why it's saying that, because pytest was installed in that environment and it had no problems running the tests... 

The test pane and menu for changing configurations:
PyCharm pytest pane

The configuration screen:
PyCharm Configuration Screen

Regardless, so far I've gotten "test_functional_config.py", "test_functional_misc.py", and "test_functional_downloads.py" to all pass on my local machine, which gives me much more confidence that I haven't broken anything significant when I go to submit my next PR.

pl77:
One quick question: I noticed an assortment of files that had unix LF line endings and others with the windows CRLF line endings.  Do you have a preference as to which I standardize on?

Safihre:
> Do you have a preference as to which I standardize on?

Not really :)

thezoggy:
main thing is to make sure git is not set to do autocrlf correction, leave it as is.
we ran into this forever ago where devs on different os would keep checking all the line endings and producing unneeded changes.

pl77:
Excellent, thanks guys.  @thezoggy that was the main reason I was asking; if you were all working on CRLF and I kept changing it to LF every time I committed I figured it'd drive you all nuts.  But it sounds like it's not a big deal regardless, so I won't make it one.

The only reason I was asking was in one type or another had been identified in the past as messing with any of the automated build systems.  I doubt it is a big problem, but I've been caught off guard in the past where running a CRLF file in a Linux environment caused all sorts of hassles that were difficult to debug.

PR #1341, opened 2019-11-06
Fix ImportError when win32api fails to load dll

### LINKED ISSUE #1340: ImportError: DLL load failed while importing win32api
SAB fails to start with import error with the following configuration:
* Windows 10
* Python 3.8
* pywin32 225

labrys:
importing pywintypes before win32api fixes #1340

See https://github.com/mhammond/pywin32/issues/1327 for more details

### DIFF
--- a/SABHelper.py
+++ b/SABHelper.py
@@ -25,13 +25,13 @@ import subprocess
 
 
 try:
+    import pywintypes
     import win32api
     import win32file
     import win32serviceutil
     import win32evtlogutil
     import win32event
     import win32service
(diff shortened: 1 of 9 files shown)


CristiFati:
This seems to be just a workaround. Check https://github.com/mhammond/pywin32/issues/1327#issuecomment-552714457 (2  *URL*s).

labrys:
Since pywin32 226 is also broken, closing this in favor of pinning dependency to pywin32 >= 227

PR #1346, opened 2019-11-28
docs(special-options): put wiki explanations directly to web-ui

casaper:
- Instead of pointing to the wiki, explain directly in the Web-UI
- Benefit is direct explanation where the user would change it
- Adds possibility to translate the explanations with po files
- If the documentation between different sabnzbd versions differs, that can be directly reflected

I've put the same text as https://sabnzbd.org/wiki/configuration/2.3/special has, but inside the app.

### DIFF
--- a/interfaces/Config/templates/config_special.tmpl
+++ b/interfaces/Config/templates/config_special.tmpl
@@ -21,7 +21,7 @@
                                 <!--#if $option[1] != $option[2] then '<span class="glyphicon glyphicon-asterisk"></span>' else ''#-->
                             </label>
                             <input type="checkbox" name="$option[0]" id="$option[0]" value="1" <!--#if int($option[1]) > 0 then 'checked="checked"' else ""#--> />
-                            <span class="desc"></span>
+                            <span class="desc">$T('explain-' + $option[0])</span>
                         </div>
                     <!--#end for#-->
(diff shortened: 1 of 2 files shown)


interfaces/Config/templates/config_special.tmpl
```diff
@@ -51,7 +52,25 @@
...
+                <h3>$T('title-config_lock') <a href="$helpuri$help_uri" target="_blank"><span class="glyphicon glyphicon-question-sign"></span></a></h3>
+                <p>$T('lead-config_lock')</p>
+            </div><!-- /col2 -->
+            <div class="col1">
+                <fieldset>
+                    <div class="field-pair">
+                        <label class="config wide" for="config_lock">
+                            config_lock ( <span class="path">off</span> )
+                        </label>
+                        <span class="desc">$T('explain-config_lock')</span>
```
Safihre: Why does `config_lock` need to be separate?

casaper: Because unlike all the other settings it can can't be set here at all. So this is just for documentary purposes.

Safihre: Let's leave it out then :)

Safihre:
Closing due to inactivity.

PR #1349, opened 2019-12-10
convert linux systray to python3 and gi

jcfp:


requirements.txt
```diff
@@ -8,6 +8,7 @@ cherrypy
 portend
 chardet
 gntp
+pygobject
```
Safihre: This needs to be conditional, on Windows this can't be installed.

jcfp: Good catch, I forget about windoze all the time :grin:

Also added a minimum version to ensure proper threading support, [3.10.2](https://ftp.acc.umu.se/pub/GNOME/sources/pygobject/3.10/pygobject-3.10.2.changes) is from 2013 so that shouldn't be a problem.

Safihre:
Seems MacOS also doesn't like it: https://travis-ci.org/sabnzbd/sabnzbd/jobs/625219418#L152

Unfortunately the Travis build is unstable for Linux, Selenium is just so unreliable and the errors so vague.

jcfp:
That happens because it needs cairo headers files to build these python bindings (pycairo). The same is true on other operating systems, see [pycairo docs](https://pycairo.readthedocs.io/en/latest/getting_started.html). On Ubuntu, things fail in a similar fashion if libcairo2-dev (etc.) isn't installed.

I guess pypi and pip aren't equipped to handle non-Python dependencies? If so, it might be better to drop pygobject from requirements.txt - the systray is very much an optional feature, after all.

Safihre:
I think it's probably best to exclude it from the requirements, since it's optional and does require a lot of depencies.
Plus on Mac, we already include our specialized Mac icon with Mac specific actions. I don't think the general Linux one can replace the Mac functionality? Or can it?

jcfp:
Don't think so. Commented out gobject, with a note about the pycairo dependency.

Safihre:
Trying to get this work on Ubuntu 18, but I can't be able to get the icon to appear anywhere. Had to manually disabled the check for `dbus`, so at least it would start running the icon. And I see it calls the main `run()` function.. But I don't see it anywhere.

jcfp:
I use xubuntu (xfce) for my desktop, notification icons such as the one from sab show up in a "notification area" on the panel, which itself may be added or removed by the user. Not entirely sure if it's there by default or how that works in gnome and ubuntu's own gui experiment. 

Anyway, if you can get the py2 systray to show, so will the py3 version.

Safihre:
Let's give it a go!

Safihre:
@jcfp What do we actually support on Ubuntu/Debian for notifications?
The code seems ancient, do we even support NotifyOSD?

sanderjo:
> @jcfp What do we actually support on Ubuntu/Debian for notifications?

Ubuntu user here ... I've never seen a SABnzbd tray bar or notification on my Ubuntu.

jcfp:
@Safihre the py2 code used pynotify, which did work afaik as long as python-notify is installed. In the ubuntu package, these were set as very weak ("suggested") dependencies only, because they do pull in a load of gui-related deps which is useless in a server environment.

Python-notify doesn't have py3 support and has been replaced with [notify2](https://pypi.org/project/notify2/) ([upstream](https://bitbucket.org/takluyver/pynotify2)). Too bad I forgot about the whole notification thing or I would have taken a look while working on the systray conversion.

thezoggy:
We should really push to just suggest https://github.com/caronc/apprise

Safihre:
@thezoggy we have been doing for a long time, check the Notification Script section of the wiki: https://sabnzbd.org/wiki/configuration/2.3/notifications
It's a wrapper around apprise basically, from the same author.

Safihre:
@jcfp do you think you'll have time to port it to notify2?

jcfp:
@Safihre #1366

Issue #1356, opened 2020-01-11
certain (small) NZBs: Badly formed yEnc article


zaadstra:
Hello,

There is a weird issue with some nzb's created by index sites. They want to grab very small files.
This is reproduceable with other nzb's created of the same or older csv files.
The nzb's created download correctly on Newsleecher on an old pc.

This is the error message of one of the failed downloads:
![2020-01-11 100701](https://user-images.githubusercontent.com/29779266/72201943-2757fc00-345a-11ea-9100-8aa6e208b643.png)

Name
ctp new NZBindex
Completed
2020-01-11 09:58
Status
Failed
Size
4.4 MB
Category
Default
Path
D:\DownloadNews\Incomplete\ctp new NZBindex
Source
ctp new NZBindex.nzb
Download
Downloaded in 2 seconds at an average of 2.1 MB/s
Age: 4d
4 articles were malformed
4 articles were missing
Aborted, cannot be completed - https://sabnzbd.org/not-complete
Servers
xsnews=2.2 MB, cheapnews=1.9 MB, ssl-eu.astraweb.com=375 KB

The nzb's of Binsearch and NZBindex show the same behaviour, I'll attach them. The files look ok to me if I look inside.

[ctp new NZBindex.nzb.txt](https://github.com/sabnzbd/sabnzbd/files/4049137/ctp.new.NZBindex.nzb.txt)
[NEW-CTP-update binsearch.nzb.txt](https://github.com/sabnzbd/sabnzbd/files/4049138/NEW-CTP-update.binsearch.nzb.txt)

Links to create the nzb's for these files:
http://binsearch.info/?q=NEW-CTP-update&max=100&adv_age=1100&server=2
https://nzbindex.nl/?q=ctp&max=25&minage=&maxage=&hidespam=1&hidepassword=0&sort=agedesc&minsize=&maxsize=&complete=0&hidecross=0&g%5B%5D=294&hasNFO=0&poster=

Download tested on SABnzbd version: 2.3.9 on Windows and on Synology.

thezoggy:
tried the first nzb:
```
Servers
UNS-US=2.2 MB, SuperNews=2.2 MB, NGD=2.2 MB, UsenetExpress=2.2 MB, Astraweb (US)=2.2 MB, UNS-EU=2.2 MB, Usenet.farm=2.2 MB, Astraweb (EU)=2.2 MB, XSusenet (free)=2.2 MB
```

[...]

PR #1358, opened 2020-01-26
Detect, notify and abort uuencoded posts

sanderjo:
as discussed https://github.com/sabnzbd/sabnzbd/issues/1356

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -172,6 +172,14 @@ class Decoder(Thread):
                     elif not killed and not found:
                         logme = T('Badly formed yEnc article in %s') % art_id
                         logging.info(logme)
+                        # bad yEnc article, so ... let's check if it's uuencode, to inform user and stop download
+                        for line in data_to_check:
+                            lline = ubtou(line).lower()
+                            if lline.find('\nbegin ') >= 0:
+                                logme = T('UUencode detected, only yEnc encoding is supported [%s]') % nzo.final_name
+                                logging.error(logme)
+                                sabnzbd.nzbqueue.NzbQueue.do.end_job(nzo)
+                                break
 
                     if not found or killed:
                         new_server_found = self.search_new_server(article)


sabnzbd/decoder.py
```diff
@@ -172,6 +172,14 @@ def run(self):
                     elif not killed and not found:
                         logme = T('Badly formed yEnc article in %s') % art_id
                         logging.info(logme)
+                        # bad yEnc article, so ... let's check if it's uuencode, to inform user and stop download
+                        for line in data_to_check:
```
jcfp: the only line this code uses for detecting uuencoding is always at the start of the article body. Do you really want to loop over the entire article here?

sanderjo: I thought so, but it is not; the first line is a header part, then one or two newlines, then \nbegin.

I did some debugging: the find was often in position 36.

And I do a break after finding it.

jcfp: Hence why I wrote 'start of the article body' - i.e. the part directly following the headers. Thing is, if the problem isn't uuencoding the loop continues over all data of every affected article, for every configured server.

Even imposing a very basic limit on the number of lines to check (like the [old code](https://github.com/sabnzbd/sabnzbd/blob/2.3.x/sabnzbd/decoder.py#L249)) would avoid a lot of futile work. Or maybe check the ending instead, that ```\n`\nend\n.``` stuff?

sanderjo: OK. I can add a maxline-checker. Or: How about only inspecting the first 1000 bytes?

sabnzbd/decoder.py
```diff
@@ -172,6 +172,14 @@ def run(self):
...
+                            lline = ubtou(line).lower()
+                            if lline.find('\nbegin ') >= 0:
+                                logme = T('UUencode detected, only yEnc encoding is supported [%s]') % nzo.final_name
+                                logging.error(logme)
+                                sabnzbd.nzbqueue.NzbQueue.do.end_job(nzo)
```
Safihre: This won't fail the job.
Change this to:
```python
                                logme = T('UUencode detected, only yEnc encoding is supported [%s]') % nzo.final_name
                                logging.error(logme)
                                nzo.fail_msg = logme
                                sabnzbd.nzbqueue.NzbQueue.do.end_job(nzo)
```

sabnzbd/decoder.py
Safihre: @jcfp/@sanderjo in the old code the data was already split in lines, but here we still have raw data from the socket. So here the `data_to_check` will be much smaller. I'd recommend maybe only checking the first 2 or 3.

This reminds me that I need to remove the `nw.lines` stuff, since from Py3 we only work with the raw data.

Safihre: I updated `develop`, have to use `raw_data` now instead of `data_to_check`.

sanderjo:
Cool, that `nzo.fail_msg = logme` indeed aborts, with a clear message.

image

sanderjo:
I've sent a new PR with a more KISS style approach https://github.com/sabnzbd/sabnzbd/pull/1361

PR #1362, opened 2020-02-25
Py3: Replace xrange with range

gaaf:
https://docs.python.org/3.0/whatsnew/3.0.html#views-and-iterators-instead-of-lists says:

- range() now behaves like xrange() used to behave, except it works with values of arbitrary size. The latter no longer exists.

### DIFF
--- a/interfaces/Plush/templates/nzo.tmpl
+++ b/interfaces/Plush/templates/nzo.tmpl
@@ -11,7 +11,7 @@
 
   <div>
   <select name="index"><optgroup label="$T('order')">
-  <!--#for $i in xrange($noofslots)#-->
+  <!--#for $i in range($noofslots)#-->
     <option value=$i <!--#if $i == $index then "selected" else ""#-->>$i</option>
   <!--#end for#-->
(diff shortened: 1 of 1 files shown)


sanderjo:
Oh, indeed, xrange() does not exist anymore (see below).

But ... it's in template Plush ... @gaaf do you use Plush, with SAB-py3?

```
sander@brixit:~$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
>>> range()
TypeError: range expected 1 arguments, got 0
>>> xrange()
NameError: name 'xrange' is not defined
>>> 

```

Safihre:
Forgot about Plush 🍭 
But indeed @gaaf, you really still use that?

gaaf:
> But indeed @gaaf, you really still use that?

Yes, for years. Never noticed is was deprecated nor that there was another skin. I hardly ever look at the config screens, had to change something yesterday and noticed 2 bugs.

Just took a look at Glitter, not sure I like it...

PR #1363, opened 2020-02-25
config: Don't try to create already existing paths

gaaf:
After pressing save, sabnzbd tries to recreate all folders. This will fail if the user has not enough permissions in the parent folder.

### DIFF
--- a/sabnzbd/config.py
+++ b/sabnzbd/config.py
@@ -239,7 +239,7 @@ class OptionDir(Option):
             if self.__validation:
                 error, value = self.__validation(self.__root, value, super().default())
             if not error:
-                if value and (self.__create or create):
+                if value and (self.__create or create) and not os.path.exists(value):
                     res, path = create_real_path(
                         self.ident()[1], self.__root, value, self.__apply_umask, self.__writable
                     )


Safihre:
The `OptionDir.set()` calls `create_real_path()` which does a `os.path.exists(my_dir)`, so it will skip directory creation it if the path exists.
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/filesystem.py#L302
So can you give examples of paths that would fail?
In your case, what are the values of `name`, `loc` and `path` when `create_real_path` is called?

gaaf:
In my situation, I have a ```complete_dir``` folder ```/home/media```. It, as well as its parent, are not writeable by sabnzbd. Below that I have folder for Movies, TV, etc, which are writeable.

It went wrong when I wanted to add a script folder. Even though I did not change ```complete_dir```, I got the failure message ```Cannot create complete_dir folder /home/media```. Now that I look at the code for ```create_real_path()```, the error message is misleading. That function returns an error even if the folder exists. The error is returned because the folder is not writeable. But it doesn't have to be.

Because of the check for ```create``` in https://github.com/sabnzbd/sabnzbd/blob/e7efb3f804146989eae8b12971e240daf85e4507/sabnzbd/config.py#L237 the attempt to create the folder is not skipped, even though the value has not changed. I could have changed that, but chose for the existence check.

Maybe it is better to just skip unchanged values completely. Something like this:
```diff
diff --git a/sabnzbd/config.py b/sabnzbd/config.py
index fc202269..342b9cc7 100644
--- a/sabnzbd/config.py
+++ b/sabnzbd/config.py
@@ -234,7 +234,7 @@ class OptionDir(Option):
             'create' means try to create (but don't set permanent create flag)
         """
         error = None
-        if value is not None and (create or value != self.get()):
+        if value is not None and value != self.get():
             value = value.strip()
             if self.__validation:
                 error, value = self.__validation(self.__root, value, super().default())
```
History for that line is a bit unreadable due to broken line-endings in the past, so I don't know whey the check for ```create``` was introduced.

Safihre:
> But it doesn't have to be.
But it kind of does, because we can't know that the user has specified sub-folders via Categories or RSS. Since the `complete_dir` is one of the core-directory of SABnzbd we always want to make sure we can access it.
It was introduced specifically for that purpose: https://github.com/sabnzbd/sabnzbd/commit/fa2a6c673c8349ce2feb096b9b4b71bcda59f1cb#diff-41a0c3c79ae5e9c34a5d5ea52e1b2aecL1115
We could probably adapt it such that it only creates if the value changed.

gaaf:
> But it kind of does, because we can't know that the user has specified sub-folders via Categories or RSS.

That would show an error when the user configures that, won't it? If the error message is not as wrong as the one I encountered, it would point the user to the change he made outside of sabnzbd.


> We could probably adapt it such that it only creates if the value changed.

That's what I tried to propose with the patch in my previous comment.

Safihre:
> That would show an error when the user configures that, won't it? 

No, if they just set "Videos" for Folder/Path setting it will be created relative to the `download_dir`. But only when the first download finishes that has to go to "Videos".
This function is really doing what it is supposed to: make sure all the base-folders exist and are writeable. The error could be improved though, so let's fix that.

PR #1366, opened 2020-03-01
Switch from pynotify to notify2

jcfp:
For basic notifications it's pretty much a plug-in replacement, so code changes are minimal. Got rid of the warnings handling as that stuff caused problems with dbus.

Noticed and removed some gntp leftovers while at it.


Safihre:
Just for my basic understanding, what extra `apt-get`'s do I need to do to get it working in my plain Ubuntu? 👷

jcfp:
Installing _python3-notify2_ should do the job on any reasonably modern desktop environment.

jcfp:
Found an issue with my changes, when sab is started as a daemon but shut down from the web interface on the same system. Will look for a fix later and ping you once it's done.

jcfp:
@Safihre ^^ that should do the job.

PR #1377, opened 2020-03-19
Docker check: Show whether SAB is running inside docker

sanderjo:
In case of problems or user's help requests, it's relevant to know whether SAB inside a docker container. So SAB now detects and reports that in the logging at startup.

This PR puts the info into sabnzbd.IN_DOCKER so that it can be used elsewhere, for example the SAB GUI

Background info from https://stackoverflow.com/a/20012536

> The most reliable way is to check /proc/1/cgroup. It will tell you the control groups of the init process, and when you are not in a container, that will be / for all hierarchies. When you are inside a container, you will see the name of the anchor point. With LXC/Docker containers, it will be something like /lxc/<containerid> or /docker/<containerid> respectively.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1106,6 +1106,11 @@ def main():
     else:
         logging.info('Platform = %s', os.name)
     logging.info('Python-version = %s', sys.version)
+    sabnzbd.IN_DOCKER = sabnzbd.in_docker()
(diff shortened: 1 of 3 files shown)


thezoggy:
looks like stale commit got added? need to rebase?

PR #1378, opened 2020-03-20
Detect and reports whether running inside Docker

sanderjo:
In case of problems or user's help requests, it's relevant to know whether SAB is inside a docker container. So SAB now detects and reports that in the logging at startup.

This PR puts the info into sabnzbd.DOCKER so that it can be used elsewhere, for example the SAB GUI

Background info from https://stackoverflow.com/a/20012536

> The most reliable way is to check /proc/1/cgroup. It will tell you the control groups of the init process, and when you are not in a container, that will be / for all hierarchies. When you are inside a container, you will see the name of the anchor point. With LXC/Docker containers, it will be something like /lxc/<containerid> or /docker/<containerid> respectively.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1107,6 +1107,11 @@ def main():
         logging.info('Platform = %s', os.name)
     logging.info('Python-version = %s', sys.version)
     logging.info('Arguments = %s', sabnzbd.CMDLINE)
+    sabnzbd.IN_DOCKER = sabnzbd.in_docker()
+    if sabnzbd.IN_DOCKER:
+        logging.info("Running inside a docker container")
+    else:
+        logging.info("Not inside a docker container")
 
     # Find encoding; relevant for external processing activities
     logging.info('Preferred encoding = %s', sabnzbd.encoding.CODEPAGE)
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1107,8 +1107,7 @@ def main():
         logging.info('Platform = %s', os.name)
     logging.info('Python-version = %s', sys.version)
     logging.info('Arguments = %s', sabnzbd.CMDLINE)
-    sabnzbd.IN_DOCKER = sabnzbd.in_docker()
-    if sabnzbd.IN_DOCKER:
+    if sabnzbd.DOCKER:
         logging.info("Running inside a docker container")
     else:
         logging.info("Not inside a docker container")
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -36,7 +36,7 @@ from threading import Lock, Thread
 ##############################################################################
 # Determine platform flags
 ##############################################################################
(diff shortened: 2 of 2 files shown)


sanderjo:
Note to self: Never forget to run `python3 -m black -l120  sabnzbd/__init__.py` (or other changed files) ...

thezoggy:
to add a comment, looks like this bit of code might could use some update although purely cosmetic

Discord_2023-05-18_15-20-00

sanderjo:
Thanks, @thezoggy 

So ... the detection method does not work on Ubuntu Jammy aka 22.04?

I checked on Ubuntu Jammy, and I can confirm:

```
sander@brixit:~$ docker run -it ubuntu:22.04 /bin/bash

root@fe30ec6b62ce:/# cat "/proc/1/cgroup"
0::/
```

No "docker". Weird.

In the same stackoverflow thread, https://stackoverflow.com/a/72136877/1492917 gives more informtion: "As of 2022, with lxd v4.0+, none of the answers so far work for both docker and lxc."

`grep 'kthreadd' /proc/2/status` seems the methods that works with Ubuntu

Within Docker on Ubuntu 22.04:

```
sander@brixit:~$ docker run -it ubuntu:22.04 /bin/bash
root@ab9190d7823a:/# grep 'kthreadd' /proc/2/status
grep: /proc/2/status: No such file or directory

```

On plain Ubuntu 22.04, no docker:
```
sander@brixit:~$ grep 'kthreadd' /proc/2/status
Name:   kthreadd
```

On Docker on Ubuntu 18.04:

```
sander@brixit:~$ docker run -it ubuntu:18.04 /bin/bash
root@55777a9ee1ba:/# grep 'kthreadd' /proc/2/status
grep: /proc/2/status: No such file or directory
```

sanderjo:
So, PoC:

```
#!/usr/bin/env python3
try:
    with open("/proc/2/status", "rt") as ifh:
        DOCKER = not("kthreadd" in ifh.read())
except:
    DOCKER = True
    pass
print(DOCKER)

```

On plain Ubuntu 22.04:

```
sander@brixit:~$ ./docker_or_not.py
False
```

Docker on Ubuntu 22.04:
```
root@ab9190d7823a:/# ./docker_or_not.py
True
```
Docker on Ubuntu 18.04
```
root@0a81effbd96e:/# ./docker_or_not.py
True
```

thezoggy:
@sanderjo hey there, I was working and saw that popped up in discord from the lsio guys and figured I'd share it real quick. Glad you were able to take a look.

sanderjo:
Hmmm, not good on Linux on ChromeOS: That is not Docker, but the new check says it is Docker ...

TBC

```
sander@penguin:~$ python3 docker_bla.py 
True
sander@penguin:~$ ll /proc/2/
ls: cannot access '/proc/2/': No such file or directory
```

The old check is OK:

`INFO::[SABnzbd:1187] Dockerized = False`

PR #1383, opened 2020-04-03
develop -- Upgrade unrar to 5.90

thezoggy:


Safihre:
The MacOS version of `unrar` in this PR is actually broken. But it doesn't show up on the test-suite because the download-check isn't run on PR's since it contains passwords.
Will update it from my MacBook and revert the commit for the 2.3.x branch.

thezoggy:
i grabbed the file from: https://www.rarlab.com/rar/rarosx-5.9.0.tar.gz
(now granted i did do this from a windows machine as i was updating the binaries over there)

Issue #1389, opened 2020-04-08
py 2 working, py 3 no downloading/stalls


mgaulton:
Opensuse Tumbleweed
Sab info 
Version: | 3.0.0-develop [unknown]
(latest checkout of develop currently)
/opt/.sabnzbd/sabnzbd.ini
/opt/sabnzbd/SABnzbd.py -d -f /opt/.sabnzbd/sabnzbd.ini
3.8.2 (default, Mar 05 2020, 18:58:42) [GCC] [UTF-8]
OpenSSL 1.1.1f 31 Mar 2020
 

I've been struggling with something for a while where when running python2.x, SAB runs full out all day, every day.
I switch to the develop branch or py3 branch, run it with python3.x and nothing happens.  Doesn't seem to be downloading anything. I do see RSS checks and some other housekeeping but actual download, nothing.
I've left it as long as 3 days to see if it just got stuck on something.
I've also rebuilt the queue multiple times.
I keep the config separate, so i've even nuked the program folder and recloned it, no change.
Any suggestions would be helpful.

The only meaningful log entry in errors, I believe is related to a empty nzb message I get in the console.
`
AttributeError: 'list' object has no attribute 'startswith'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/opt/sabnzbd/sabnzbd/decoder.py", line 149, in run
    if nzo.precheck and raw_data and raw_data.startswith(b'223 '):
AttributeError: 'list' object has no attribute 'startswith'

`

sanderjo:
That traceback is from this part of the code:
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/decoder.py#L145-L151

```
                except (BadYenc, ValueError):
                    # Handles precheck and badly formed articles
                    killed = False
                    found = False
                    if nzo.precheck and raw_data and raw_data.startswith(b'223 '):
                        # STAT was used, so we only get a status code
                        found = True
```

So:
* SAB got an exception earlier on. That is bad. While running SAB-py3, what does `sabnzbd.log` say about `SABYenc module`
* do you have pre-check aka "check before download"  on? See config/switches/#pre_check  in the SAB-gui
[...]

PR #1390, opened 2020-04-09
Make Pre-Check Work Again

sanderjo:
raw_data is now a list, so we need the first element raw_data[0].

Example
`raw_data is [b'223 0 <OuIgBdRbTmOrXuAyFcSjDgTh-1584818153948@nyuu>\r\n']`

As discussed https://github.com/sabnzbd/sabnzbd/issues/1389


Safihre:
Thanks!

PR #1402, opened 2020-04-25
Update text files for 3.0.0 Alpha 1

Safihre:
I am preparing the 3.0.0 Alpha 1 release, but I have worked on so many lines of code and there are so many commits that I kind of lost track of all the things that have changed.

Could you guys check if you agree with my changes?
Specifically `INSTALL.txt`, `README.md` and 'README.mkd' (the releasenotes).

Thanks!

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -1,4 +1,4 @@
-                            SABnzbd 2.3.8
+                            SABnzbd 3.0.0
 
 -------------------------------------------------------------------------------
 0) LICENSE
@@ -52,38 +52,11 @@ Specific guides to install from source are available for Windows and macOS:
     https://sabnzbd.org/wiki/installation/install-macos
     https://sabnzbd.org/wiki/installation/install-from-source-windows
 
-You need to have Python installed plus some non-standard Python modules
-and a few tools.
+Only Python 3.5 and above is supported. All comands below asume
+Python 3 is the only installed version of Python.
 
-All platforms
-    Python-2.7.latest       http://www.python.org (2.7.9+ recommended)
-
-Windows
-    PyWin32                 use "pip install pypiwin32"
-
-Essential modules
-    cheetah-2.0.1+          use "pip install cheetah"
-    par2cmdline >= 0.4      https://github.com/Parchive/par2cmdline/releases
-                            See also: https://sabnzbd.org/wiki/installation/multicore-par2
-    unrar >= 5.00+          http://www.rarlab.com/rar_add.htm
-    openssl >= 1.0.0        http://www.openssl.org/
-
-Optional modules
-    unzip >= 6.00           http://www.info-zip.org/
-    7zip >= 9.20            http://www.7zip.org/
-    sabyenc == 3.3.1        use "pip install sabyenc"
-                            More information: https://sabnzbd.org/sabyenc
-    cryptography >= 1.0     use "pip install cryptography"
-                            Enables certificate generation and detection of encrypted RAR-files
-
-Optional modules Linux
-    notify2                 Should be part of GTK for Python support on Debian/Ubuntu
-                            If not, you cannot use the NotifyOSD feature.
-    python-dbus             Enable option to Shutdown/Restart/Standby PC on queue finish.
-
-Embedded modules (preferably use the included version)
-    CherryPy-8.1.2 with patches http://www.cherrypy.org
-
+Install all required packages by running:
+    pip install -r requirements.txt
 
 Unpack the ZIP-file containing the SABnzbd sources to any folder of your liking.
 
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -55,7 +55,7 @@ Specific guides to install from source are available for Windows and macOS:
 Only Python 3.5 and above is supported. Replace python3 with python in the commands
 below if Python 3 is the only version of Python on your system.
 
-Install all required packages by running:
+Install all required Python packages by running:
     python3 -m pip install -r requirements.txt
 
 Unpack the ZIP-file containing the SABnzbd sources to any folder of your liking.
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -52,20 +52,20 @@ Specific guides to install from source are available for Windows and macOS:
     https://sabnzbd.org/wiki/installation/install-macos
     https://sabnzbd.org/wiki/installation/install-from-source-windows
 
-Only Python 3.5 and above is supported. All comands below asume
-Python 3 is the only installed version of Python.
+Only Python 3.5 and above is supported. Replace python3 with python in the commands
+below if Python 3 is the only version of Python on your system.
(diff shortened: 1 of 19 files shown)


INSTALL.txt
```diff
@@ -52,38 +52,11 @@ Specific guides to install from source are available for Windows and macOS:
...
-                            If not, you cannot use the NotifyOSD feature.
-    python-dbus             Enable option to Shutdown/Restart/Standby PC on queue finish.
-
-Embedded modules (preferably use the included version)
-    CherryPy-8.1.2 with patches http://www.cherrypy.org
+Only Python 3.5 and above is supported. All comands below asume
+Python 3 is the only installed version of Python.
 
+Install all required packages by running:
+    pip install -r requirements.txt
```
sanderjo, line 59: Safer to say `python3 -m pip`
At least on Ubuntu: with python/python2 not installed, there is no pip either. With python3, it's called pip3. To avoid confusion (among distro's), with python3 installed (as defined 3 lines earlier), python3 -m pip will always work.

INSTALL.txt
```diff
@@ -52,38 +52,11 @@ Specific guides to install from source are available for Windows and macOS:
...
-
-Optional modules Linux
-    notify2                 Should be part of GTK for Python support on Debian/Ubuntu
-                            If not, you cannot use the NotifyOSD feature.
-    python-dbus             Enable option to Shutdown/Restart/Standby PC on queue finish.
-
-Embedded modules (preferably use the included version)
-    CherryPy-8.1.2 with patches http://www.cherrypy.org
+Only Python 3.5 and above is supported. All comands below asume
+Python 3 is the only installed version of Python.
```
jcfp, line 56: I get the point you're trying to make, but this could be read as "sab conflicts with having python2 installed". Probably better to just use 'python3' in the commands, afaik that would work on every mainstream distro.

README.md
```diff
@@ -17,16 +17,14 @@ If you want to know more you can head over to our website: https://sabnzbd.org.
 
 SABnzbd has a good deal of dependencies you'll need before you can get running. If you've previously run SABnzbd from one of the various Linux packages, then you likely already have all the needed dependencies. If not, here's what you're looking for:
 
-- `python` (SABnzbd version 2: only python 2.7.x. SABnzbd version 3: python 3.5.x and higher, often called python3)
-- `python-cheetah`
+- `python` (Python 3.5 and higher, often called python3)
 - `par2` (Multi-threaded par2 installation guide can be found [here](https://sabnzbd.org/wiki/installation/multicore-par2))
```
jcfp, line 21: Info on the required python modules is missing. Rather than list any modules in README.md I suggest adding a reference to requirements.txt, to avoid keeping this info in multiple places where it inevitably gets out of sync. In that file, the modules can be sorted into 'required' and 'optional/nice to have' categories with short comments on why where appropriate.

README.md
```diff
@@ -17,16 +17,14 @@ If you want to know more you can head over to our website: https://sabnzbd.org.
...
 - `unrar` (Make sure you get the "official" non-free version of unrar)
 - `sabyenc` (installation guide can be found [here](https://sabnzbd.org/sabyenc))
 
 Optional:
```
jcfp, line 25: The optional modules stuff should be removed, see my prior comment.

README.md
Safihre: True! Listed them and added a reference for `pygobject` with note to check `requirements.txt` for more info.

README.md
Safihre: Hmm but how would we list something like `python-dbus` in `requirements.txt`?

jcfp: Something like this?
```
# optional, for system power management support
dbus-python; sys_platform != 'win32' and sys_platform != 'darwin'
```
Should be enough for anyone running from source as well as packagers to base their decisions on. If too many issues were to arise with dbus-python, we could simply comment it out by default later.

Safihre: That doesn't seem to work:
https://travis-ci.org/github/sabnzbd/sabnzbd/jobs/679656500#L219
Will mention that that it requires `dbus` to be installed, that seems to be the problem right?

The PyPi index also shows we could use the GObjects for it: https://pypi.org/project/dbus-python/

jcfp: The actual problem is this one:
> configure: error: Package requirements (dbus-1 >= 1.8) were not met:
>  No package 'dbus-1' found

Forgot the python3-dbus is an extension rather than pure-python. Probably needs `libdbus-1-dev` to build.

INSTALL.txt
```diff
@@ -52,38 +52,11 @@ Specific guides to install from source are available for Windows and macOS:
...
-                            Enables certificate generation and detection of encrypted RAR-files
-
-Optional modules Linux
-    notify2                 Should be part of GTK for Python support on Debian/Ubuntu
-                            If not, you cannot use the NotifyOSD feature.
-    python-dbus             Enable option to Shutdown/Restart/Standby PC on queue finish.
-
-Embedded modules (preferably use the included version)
-    CherryPy-8.1.2 with patches http://www.cherrypy.org
+Only Python 3.5 and above is supported. All comands below asume
```
albino1: Just a small spelling correction.  This:

`All comands below asume`

Should be:

`All commands below assume`

README.mkd
```diff
@@ -1,30 +1,36 @@
...
-- When upgrading from 2.2.0 or older the queue will be converted. Job order,
-  settings and data will be preserved, but all jobs will be unpaused and
-  URL's that did not finish fetching before the upgrade will be lost.
-- The organization of the download queue is different from 0.7.x releases.
-  This version will not see the 0.7.x queue, but you can restore the jobs
-  by going to Status page and using Queue Repair.
+- When upgrading from 2.x.x or older the queue will be converted. Job order,
+  settings and data will be preserved, but if you decide to go back to 2.x.x
+  your queue cannot be downgraded again. But you can restore the jobs by going
+  to the Status page and running Queue Repair.
```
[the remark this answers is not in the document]
Safihre: The pickles from 3.0.0 will fail to load in 2.x, so you loose the queue. That's why I added the notice about downgrading the queue.

Safihre:
Pushed updates, let me know if I missed anything else! @sabnzbd/contributers

README.mkd
```diff
@@ -1,30 +1,36 @@
...
+- Work is still in progress on Windows Service support.
+
+## Big changes
+- Python 3.5 and above are the only supported versions of Python.
+- Fully obfuscated RAR-sets with no verification files are detected and extracted.
+- SABYenc, par2 and unrar are now required to start downloading.
+- The Windows installer is 64-bit only, for 32-bit use the standalone package.
+- Growl-support was removed.
+- macOS features such as the menu and notifications now use native code.
+
```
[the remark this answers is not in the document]
Cpuroast: Typo, "bandwidth" is the correct spelling.

PR #1406, opened 2020-04-28
remove cruft from licenses dir

jcfp:
Copies of cherrypy, configobj, feedparser, six aren't kept in git or supplied with source releases anymore. Rsslib is no longer part of sab.


Safihre:
👍

Issue #1425, opened 2020-05-05
[dev-mode] ResourceWarning: unclosed socket.socket


sanderjo:
With IPv6 disabled I get the below.
I will solve, although I don't understand the relevance of this function anymore.

```
2020-05-05 11:40:43,655::DEBUG::[__init__:1098] Test IPv6: Cannot reach IPv6 test host. Disabling IPv6
/home/sander/git/sab-5mei2020/sabnzbd/downloader.py:410: ResourceWarning: unclosed <socket.socket fd=15, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::', 0, 0, 0)>
  sabnzbd.EXTERNAL_IPV6 = sabnzbd.test_ipv6()
Object allocated at (most recent call last):
  File "/usr/lib/python3.8/threading.py", lineno 890
    self._bootstrap_inner()
  File "/usr/lib/python3.8/threading.py", lineno 932
    self.run()
  File "/home/sander/git/sab-5mei2020/sabnzbd/downloader.py", lineno 410
    sabnzbd.EXTERNAL_IPV6 = sabnzbd.test_ipv6()
  File "/home/sander/git/sab-5mei2020/sabnzbd/__init__.py", lineno 1091
    sock = socket.socket(af, socktype, proto)
2020-05-05 11:40:43,672::DEBUG::[downloader:411] External IPv6 test result: False
```

Safihre:
We use it to see if we should connect to IPv6 servers or not when returned by `getaddrinfo`.
You made that 😉 
https://github.com/sabnzbd/sabnzbd/blame/develop/sabnzbd/newswrapper.py#L80

sanderjo:
AFAIK this is much older code. 

But anyway: then I'll certainly send the PR! :-)

PR #1426, opened 2020-05-05
test_ipv6(): solve a ResourceWarning: unclosed socket.socket

sanderjo:
Closes https://github.com/sabnzbd/sabnzbd/issues/1425

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -1090,7 +1090,6 @@ def test_ipv6():
         with socket.socket(af, socktype, proto) as sock:
             sock.settimeout(2)  # 2 second timeout
             sock.connect(sa[0:2])
-            sock.close()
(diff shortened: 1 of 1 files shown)


sabnzbd/__init__.py
```diff
@@ -1087,10 +1087,10 @@ def test_ipv6():
...
     try:
         af, socktype, proto, canonname, sa = info[0]
-        sock = socket.socket(af, socktype, proto)
-        sock.settimeout(2)  # 2 second timeout
-        sock.connect(sa[0:2])
-        sock.close()
+        with socket.socket(af, socktype, proto) as sock:
+            sock.settimeout(2)  # 2 second timeout
+            sock.connect(sa[0:2])
+            sock.close()
```
Safihre, line 1093: This line needs to be removed, since you use the `with`.

sanderjo: Thanks. Solved with additional commit.

Issue #1416, opened 2020-05-01
[dev-mode] directunpacker.py:81: ResourceWarning: subprocess still running Popen() & unclosed file 


sanderjo:
After a download. As this is class stuff (`self.active_instance = Popen(...)` ) it's unclear to if it's solvable with a `with ... as ... ` ... as I don't see where/when the Popen can be closed.

```
2020-05-01 07:39:18,685::DEBUG::[directunpacker:284] DirectUnpack Unrar output 
/usr/lib/python3.8/subprocess.py:942: ResourceWarning: subprocess 118967 is still running
  _warn("subprocess %s is still running" % self.pid,
Object allocated at (most recent call last):
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/nzbstuff.py", lineno 1216
    self.direct_unpacker.add(nzf)
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/decorators.py", lineno 36
    return f(*args, **kw)
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/directunpacker.py", lineno 139
    self.create_unrar_instance()
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/decorators.py", lineno 36
    return f(*args, **kw)
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/directunpacker.py", lineno 367
    self.active_instance = Popen(command, shell=False, stdin=subprocess.PIPE,
/home/sander/git/sjo-sab-1mei2020/sabnzbd/directunpacker.py:81: ResourceWarning: unclosed file <_io.FileIO name=18 mode='wb' closefd=True>
  self.active_instance = None
Object allocated at (most recent call last):
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/decorators.py", lineno 36
    return f(*args, **kw)
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/directunpacker.py", lineno 139
    self.create_unrar_instance()
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/decorators.py", lineno 36
    return f(*args, **kw)
  File "/home/sander/git/sjo-sab-1mei2020/sabnzbd/directunpacker.py", lineno 367
    self.active_instance = Popen(command, shell=False, stdin=subprocess.PIPE,
  File "/usr/lib/python3.8/subprocess.py", lineno 838
[...]

PR #1428, opened 2020-05-05
directunpacker: process and filehandles are closed nicely

sanderjo:
Closes https://github.com/sabnzbd/sabnzbd/issues/1416
So solves ResourceWarning: subprocess still running Popen() & unclosed file when running with `python3 -X dev -X tracemalloc=5 ./SABnzbd.py`

### DIFF
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -78,6 +78,15 @@ class DirectUnpacker(threading.Thread):
         pass
 
     def reset_active(self):
+        # make sure the process and filehandles are closed nicely:
+        try:
+            # Creation was done via "self.active_instance = Popen()", so:
+            self.active_instance.stdout.close()
+            self.active_instance.stdin.close()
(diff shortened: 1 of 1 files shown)


sabnzbd/directunpacker.py
```diff
@@ -78,6 +78,15 @@ def save(self):
...
 
     def reset_active(self):
+        # make sure the process and filehandles are closed nicely:
+        try:
+            # Creation was done via "self.active_instance = Popen()", so:
+            self.active_instance.stdout.close()
+            self.active_instance.stdin.close()
+            self.active_instance.wait(timeout=2)
+        except:
+            logging.debug("Exception in reset_active()")
```
Safihre, line 88: I would suggest to include the traceback:
```python
logging.debug("Exception in reset_active()", exc_info=True)
```

sanderjo: OK, I'll do.

PR #1434, opened 2020-05-09
Article Cache Flusher

Safihre:
SABnzbd will keep decoded articles in the article cache, until it has all articles of a file. But when when the article cache gets full, SABnzbd starts to behave poorly. It tries to write the cache to disk, while fetching other cached items back from the disk, slowing the downloading down, etc, it just makes SABnzbd look super slow and bumpy.
As an example: if I speed limit my old i3 system to 30MB/s, it does so happily for hours. If I remove the rate limit, it downloads at 90MB/s for a minute or so, and then drops to 15MB/s while it is messing with the cache. It never recovers to a stable speed.

Therefore, I came up with a rather simple fix: let's just write all the articles we already have decoded to disk. Not for _every_ article, but in a batch approach. Since most articles in the wild have a size of around 350-700KB, I opted for every 35 articles (12-24MB).

Still working on it, have to fix backwards-compatibility with existing jobs in the queue. 
If anyone wants to inspect the code and give feedback, feel free to do so 💯


sanderjo:
Great! I'm using it now on my download box. It feels better: more steady, more responsive.

To see if I'm running the actual code, I grep for the strings below. Correct?

```
$ cat .sabnzbd/logs/sabnzbd.log | grep -e "Decoding part of" -e "Decoding finished"

2020-05-10 09:44:20,370::DEBUG::[assembler:84] Decoding part of /home/sander/Downloads/incomplete/test_download_100MB/sometestfile-100MB.part4.rar
2020-05-10 09:44:20,427::INFO::[assembler:109] Decoding finished /home/sander/Downloads/incomplete/test_download_100MB/sometestfile-100MB.part4.rar
2020-05-10 09:44:20,722::DEBUG::[assembler:84] Decoding part of /home/sander/Downloads/incomplete/test_download_100MB/sometestfile-100MB.part5.rar
2020-05-10 09:44:20,776::INFO::[assembler:109] Decoding finished /home/sander/Downloads/incomplete/test_download_100MB/sometestfile-100MB.part5.rar
2020-05-10 09:44:21,554::DEBUG::[assembler:84] Decoding part of /home/sander/Downloads/incomplete/test_download_100MB/sometestfile-100MB.part6.rar
2020-05-10 09:44:21,609::INFO::[assembler:109] Decoding finished /home/sander/Downloads/incomplete/test_download_100MB/sometestfile-100MB.part6.rar
```

jcfp:
So it writes partial files (for example, partial rar segments) every 35 articles to spread out disk I/O, as a saving grace for systems with fast connections but (comparatively) slow disks? Laptops on wired gbit, raspberries and the likes.

What results do you get on your i3 without speed limit on a sizeable job? Does it still reach the point where the cache is 'full' (albeit later), or not at all?

danepowell:
This solved a major performance bottleneck that I was facing. I can't comment on the technical quality, but performance wise it seems to work.

Safihre:
Indeed it spreads the writes out. 

Testing on my i3 system showed that actually the bottleneck for that system is deeper in Python. It just seems that Python can't write faster than 45MB/s while the disk in the system can actually handle up to 70MB/s-ish. Will think a bit more if I can do something smart about that..

sanderjo:
Note: I have to **turn off Unwanted Extensions** (and probably Encrypted Rar), otherwise I get tracebacks & slower downloads. With that turned off:

The download speed & steadiness: on my Celeron (!), I get an **incredible** 81.9 MB/s on a big 10GB download. Proof:

`Downloaded in 2 mins 10 seconds at an average of 81.9 MB/s`

Probably my fiber (750 / 750 Mbps) is the limit, not SAB / CPU / ...

At the end, the download speeds goes down, and SAB is saying

```
2020-05-13 21:13:13,361::DEBUG::[downloader:287] Delaying
2020-05-13 21:13:13,362::DEBUG::[downloader:287] Delaying
2020-05-13 21:13:13,363::DEBUG::[downloader:287] Delaying
2020-05-13 21:13:13,363::DEBUG::[downloader:287] Delaying
```

So far very impressive results, @Safihre 

image

Cpuroast:
Looks good, didn't notice much difference, 450Mbit connection, -1 article cache(so 4GB, since this is a 64-bit OS), USB3 2.5" 5400rpm temp folder, maybe 450Mbit isn't enough to cause a problem, I get a steady 53MB/sec, I don't use direct-unpack as it falls flat on its face more often then not even in 3.0.0 Alpha2 and I use "Pause Downloading During Post-Processing".

Safihre:
Yeah, in the end I realize that this only helps in 2 cases:

- Uploads that have 1 giant file inside the NZB instead of separate rar-files, which used to clog up SABnzbd. People still upload stuff like that unfortunately. 
- When connections are unstable and there's a lot of files with only 1 article left to fetch, which normally would all be kept in the cache.

So not much of a speedbump is really to be expected. If the disk is too slow, this won't make it much faster.

jcfp:
Jobs with a single giant file should be easy to detect directly from any nzb file. In fact, cache skipping would probably be beneficial for any job with at least one file that's too big for the configured cache size (minus a bit of a margin for the inevitable overlap with subsequent files or jobs).

For other jobs, most benefits would seem to come from avoiding the ugly disk i/o side effects of a full cache, akin to that higher overall download performance through speed limiting on your i3.

Any real life scenarios where spreading out disk i/o is a negative? Very high end systems that can keep jobs entirely in memory could use a memdisk for incomplete...

PR #1450, opened 2020-05-18
svg icons for sabtraylinux

jcfp:
Better quality, scalable icons at 1/10th the filesize; kept the colours the same as the ico files. Ico files not deleted since these are still used by the windows tray.


Safihre:
Strange we didn't have these anywhere else yet.

jcfp:
I was surprised too, even asked around on irc. All I could find were those ico files, dating all the way back to the introduction of the windoze trayicon in 2011.

PR #1452, opened 2020-05-19
If running SAB from git repo, find and give info on branch and commit

sanderjo:
shows in sabnzbd.log and on GUI

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1183,8 +1183,6 @@ def main():
 
     sabnzbd.initialize(pause, clean_up, evalSched=True, repair=repair)
 
-    os.chdir(sabnzbd.DIR_PROG)
-
     sabnzbd.WEB_DIR = identify_web_template(sabnzbd.cfg.web_dir, DEF_STDINTF, fix_webname(web_dir))
     sabnzbd.WEB_DIR_CONFIG = identify_web_template(None, DEF_STDCONFIG, '')
     sabnzbd.WIZARD_DIR = os.path.join(sabnzbd.DIR_INTERFACES, 'wizard')
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1089,14 +1089,14 @@ def main():
     os.chdir(sabnzbd.DIR_PROG)
 
     logging.info('--------------------------------')
-    # If git, find and give info on branch and commit
+    # If running from git repo, find and give info on branch and commit
     try:
-        mygitcmd = 'git log --oneline --decorate -1  --no-color'
-        with os.popen(mygitcmd) as p:
-            gitinfo = p.readline().rstrip()
-        # Todo: which way is the correct way:
-        sabnzbd.__baseline__ = gitinfo
-        sabnzbd.version.__baseline__ = gitinfo
+        mygitcmd = 'git log -1 --oneline --decorate'.split(' ')
+        gitinfo = sabnzbd.newsunpack.run_simple(mygitcmd).split('\n')[0]
+        if len(gitinfo) >= 5:
+            # We found git info, so store it:
+            sabnzbd.__baseline__ = gitinfo
+            sabnzbd.version.__baseline__ = gitinfo
     except:
         # No git info
         pass
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1086,9 +1086,24 @@ def main():
         if no_file_log:
             logging.info('Console logging only')
 
+    os.chdir(sabnzbd.DIR_PROG)
+
     logging.info('--------------------------------')
+    # If git, find and give info on branch and commit
+    try:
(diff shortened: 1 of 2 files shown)


SABnzbd.py
```diff
@@ -1086,9 +1086,24 @@ def main():
         if no_file_log:
             logging.info('Console logging only')
 
+    os.chdir(sabnzbd.DIR_PROG)
+
     logging.info('--------------------------------')
+    # If running from git repo, find and give info on branch and commit
+    try:
+        mygitcmd = 'git log -1 --oneline --decorate'.split(' ')
```
jcfp, line 1094: Careful there, "git" may not be what you expect in the user's context. Consider find_on_path() or similar to build the cmd.

Do you really need more than just the revision? If not, you could go with `git rev-parse HEAD` instead. That would fit in nicely with existing code and thus also make the edit in interface.py redundant, which has the undesirable side effect of displaying a long commit hash in tagged revisions (i.e., releases).

SABnzbd.py
```diff
@@ -1086,9 +1086,24 @@ def main():
...
+        gitinfo = sabnzbd.newsunpack.run_simple(mygitcmd).split('\n')[0]
+        if len(gitinfo) >= 5:
+            # We found git info, so store it:
+            sabnzbd.__baseline__ = gitinfo
+            sabnzbd.version.__baseline__ = gitinfo
```
jcfp, line 1099: Don't these refer to exactly the same object?

SABnzbd.py
```diff
@@ -1086,9 +1086,24 @@ def main():
...
+    except:
```
jcfp, line 1100: Might want to change this to `except Exception:` to prevent catching base exceptions (user interrupts and the likes).

SABnzbd.py
```diff
@@ -1086,9 +1086,24 @@ def main():
         if no_file_log:
             logging.info('Console logging only')
 
+    os.chdir(sabnzbd.DIR_PROG)
+
     logging.info('--------------------------------')
+    # If running from git repo, find and give info on branch and commit
+    try:
+        mygitcmd = 'git log -1 --oneline --decorate'.split(' ')
+        gitinfo = sabnzbd.newsunpack.run_simple(mygitcmd).split('\n')[0]
```
jcfp, line 1095: `rstrip()`?

SABnzbd.py
sanderjo: My goal is to on which branche I am, and which commit:

```
sander@brixit:~/git/sabnzbd$ git log -1 --oneline --decorate
640814d4 (HEAD -> feature/smarter-cache) Add debug logging when delaying the download

```

This is too little info for me:

```
sander@brixit:~/git/sabnzbd$ git rev-parse HEAD
640814d44c952675bed35b4f7df4d2f927bd3b5a
```

SABnzbd.py
sanderjo: Yes, no. Hopefully @Safihre can explaing ... something to do with pythom modules.

SABnzbd.py
sanderjo: Ah, yes. I first had a multiline output there.

sanderjo:
See below how it looks like. 
@Safihre please comment in this thread how you want it. 

```
$ ./SABnzbd.py 
2020-05-20 20:05:55,270::INFO::[SABnzbd:1091] --------------------------------
2020-05-20 20:05:55,296::INFO::[SABnzbd:1104] SABnzbd.py-3.0.0-develop (rev=7e86752f (HEAD -> develop, origin/develop, origin/HEAD) git repo info: better result on non-git env)
2020-05-20 20:05:55,296::INFO::[SABnzbd:1105] Full executable path = /home/sander/git/sjo-19mei2020/SABnzbd.py
2020-05-20 20:05:55,296::INFO::[SABnzbd:1116] Platform = posix
```

image

Safihre:
It's a bit long, but does contain all the info we would want also from a user to let us know where they are if running from git.. So maybe it's good. What do you think @jcfp?

jcfp:
Could put the commit msg on a line of its own (or even ditch it altogether) to avoid excessive line length?

sanderjo:
> Could put the commit msg on a line of its own (or even ditch it altogether) to avoid excessive line length?

My goal is see these details easily in the GUI so that I can easily see which git version / branch / commit a SAB version is running on yet another SAB instance.

So I do want to see the _text_ of the commit. That does say something to me (a hex code does not)

If others think that is not needed, I will retract this PR. @Safihre please decide.

sanderjo:
ok ... I think this functionality is too much a niche, for myself (with a lot of boxes, and testing of different branches and commits), so not useful enough for others.

albino1:
Why not just show the version number like normal, and then put a `(show more)` link that expands to display the rest of the git info and whatever else you think might be useful?

PR #1457, opened 2020-05-22
correct regex in filesystem.check_mount

jcfp:
The extra forward slash required by the regexp can cause non-existent paths located directly under the mountpoint to be misreported:
```
# /mnt/foo doesn't exist
>>> filesystem.check_mount("/mnt/foo/")
False
>>> filesystem.check_mount("/mnt/foo")
True
>>> filesystem.check_mount("/mnt/foo/bar/")
False
>>> filesystem.check_mount("/mnt/foo/bar")
False
```

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -351,11 +351,11 @@ def check_mount(path):
         Retry 6 times with an interval of 1 sec.
     """
     if sabnzbd.DARWIN:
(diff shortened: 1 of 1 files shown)


Safihre:
Could you add unit tests, since we already have a section to test filesystem.py?

jcfp:
I came across this while working on expanding the filesystem tests. That is still a WIP though covering much more than just check_mount, hence the separate PR.

Safihre:
Great 😎

Issue #1456, opened 2020-05-22
Crash in RAR Renamer


Safihre:
Testing a super-obfuscated NZB (that I can't share here), the renamer crashed. I think because the NZB is a bit messed up and contains some files twice. 
But NZBGet is able to process it, so I hope we can too 👍 
Shared the original NZB with @sanderjo for debugging!
```
2020-05-21 09:39:16,411::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-0.0
2020-05-21 09:39:16,411::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-0.1.0
2020-05-21 09:39:16,412::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-0.2.0
2020-05-21 09:39:16,412::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-1.0
2020-05-21 09:39:16,413::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-1.1
2020-05-21 09:39:16,413::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-1.1.0
2020-05-21 09:39:16,414::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-1.1.1
2020-05-21 09:39:16,414::DEBUG::[postproc:907] No RAR-volume-number found in \\?\C:\Users\<USERNAME>\Downloads\a b\YellowPeople.S16E20.1080p.BluRay.x264\165099-1.2.0
[...]

PR #1459, opened 2020-05-23
find and use obfuscated SFV files

sanderjo:
If there are no plain *.sfv files, find obfuscated SFV files, and use them.

As discussed https://github.com/sabnzbd/sabnzbd/issues/1456#issuecomment-632567752


sabnzbd/newsunpack.py
```diff
@@ -2210,6 +2210,51 @@ def par2_mt_check(par2_path):
...
+
+    hexcounter = 0
+
+    for line in lines:
+        # handle non-comment, non-empty lines:
+        if not line.startswith(";") and len(line) > 0:
+            lastword = line.rstrip().split(" ")[-1]
+            try:
+                # check if it's a hexadecimal, and 8-digit:
+                if int(lastword, 16) and len(lastword) == 8:
```
jcfp, line 2245: Lines that are not valid entries in an sfv file can pass these checks, for example `line = '12345678'` or `'Aaaaaaaa'` (and variations thereof with whitespaces on either end).

Any reason for not using a regular expression here? Something like `^[^;].*\..+?\ +[A-Fa-f0-9]{8}$` (disclaimer: I didn't look up sfv spec and assumed every filename comes with an extension) could efficiently combine your current checks with a sanity check of the rest of the line that would prevent the above from getting through. [Regex101](https://regex101.com/) can be useful for testing.

sanderjo: Invalid fake-SFV will cause an error with renaming ... which is fine for me.

Because I like KISS. And I find regexps not KISS.

sabnzbd/newsunpack.py
```diff
@@ -2210,6 +2210,51 @@ def par2_mt_check(par2_path):
...
+
+    # first check if it's ASCII, or non-ASCII
+    try:
+        with open(myfile, "rb") as f:
+            # get first 10000 bytes to check
+            myblock = f.read(10000)
+            if is_ascii_string(myblock):
+                # ASCII, so store lines for further inspection
+                try:
+                    lines = myblock.decode("utf-8").split("\n")
```
Safihre, line 2227: Use the `ubtou` function from `sabnzbd.encoding`, to keep things uniform.
```python
lines = ubtou(myblock).split("\n")
```

sabnzbd/newsunpack.py
jcfp: The standard string manipulation methods may look familiar to you, but relying solely on those to do pattern matching is complicated and error prone, as demonstrated. Regular expressions _are_ the simple and effective way of doing that.

sanderjo: This is a `cksfv` generated so valid SFV:
```
; Generated by cksfv v1.3.14 on 2020-05-25 at 17:49.33
; Project web site: http://www.iki.fi/shd/foss/cksfv/
;
;     10485760  17:47.45 2020-05-25 normalascanbe.bin
;     10485760  17:44.36 2020-05-25 schöne_Türen_öffnen.bin
;     10485760  17:44.05 2020-05-25 this is a file with spaces.bin
;     10485760  17:47.30 2020-05-25 你好世界
normalascanbe.bin 681718CC
schöne_Türen_öffnen.bin 788E9541
this is a file with spaces.bin 24041228
你好世界 78938E11

```
So the left side can be about anything.

The only forbidden thing is
```1234abcd
 abcdabcd
```
as there must be something, then one or more spaces, and then a 8-digit hex number.
So
`I live at Highway 12345678` should be considered as valid SFV.

Agree?

sanderjo: Going further:
```
; this is a line
; and this another
```
is a valid SFV file, but has no info, so I would not want to count this as SFV file; I really want at least one line with useful and valid SFV info.

That means that per line, we have to determine:
valid and useful info => increment counter
only comment => continue
non-valid SFV info => break because not a SFV file

jcfp: > only comment => continue

Indeed. Probably easiest to put this check first, by looking at the first character like your code already does.

If the line makes it past that first check, see if there's a match with what you expect to find in an sfv file, i.e. your `<some non-space(s)> <space(s)> <8-digit hex>` which means the regexp simplifies to something like this:
`.*\S.*\ +[A-Fa-f0-9]{8}$`

where:
`.*\S.*` matches anything that has at least one non-whitespace character (\S) in it.
`\ +` matches 'one or more spaces' (and only spaces, not tabs or other whitespaces)
`[A-Fa-f0-9]{8}$` matches hex string, exactly 8 digit long, at the end of the line

> valid and useful info => increment counter
> non-valid SFV info => break because not a SFV file

That sounds exactly right.

PR #1466, opened 2020-05-27
Retry CRC errors on different server

Safihre:
We used to retry all CRC errors on all the specified newsservers, but then [in 2015](https://github.com/sabnzbd/sabnzbd/commit/64ca554746f4fba6e80a833a4730c83cce639f7d#diff-e6f32f1b180bd79e271944d23cb61c9c) this was removed because we made the assumption that a CRC-error is always the result of a bad upload.

Nowadays this doesn't always seem to be the case anymore, illustrated by some Reddit and forum posts. Other usenet clients, such as NZBGet and Newsleecher, do retry CRC errors on other servers. So I think we should also start doing this again.

I took this opportunity to rewrite the decoder failure handling and wanted to ask if you guys think it's more clear this way?
@sabnzbd/contributers


jcfp:
I've seen failures in recent times too because invalid article data didn't trigger retries on other servers. These jobs would complete without issue once retried with the offending server disabled (their entire platform in beta it wasn't very hard to guess the troublemaker either). So returning to the crc handling of the old days is definitely a good thing.

Code looks cleaner this way, haven't yet had the chance to put it to use though.

thezoggy:
i wonder if its 'better' to assume the article is not found and then mark it when it is. since that way it seems like it would be more likely to handle future dmca/issues/reasons why the article is missing. vs assuming that the article is there and trying to figure out why its not.

Safihre:
@thezoggy, agreed. This is a remnant when there was also `article_killed`. The actually name should be `article_found_and_processed`, because the code in the `except` to handle pre check is really about finding the article.

PR #1469, opened 2020-05-29
pytest: postproc -> rar_renamer()

sanderjo:
New test in SABnzbd pytest collection: verify that rar_renamer() does its deobfuscate work right. 
Three test cases

* a single obfuscated rar set, no SFV
* two obfuscated rar sets, no SFV
* something obfucated, not a rar

The `assert` is done on the number of renamed files (which is returned by rar_renamer()).

FWIW: 
I introduced a `nzo = mock.Mock()` thanks to @Safihre 's tip
I've kept the rar files very small: 15k each. 

```
sander@witte2004:~/git/sjo-sab$ pytest tests/test_postproc.py 

======================================================================== test session starts =========================================================================
platform linux -- Python 3.8.2, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /home/sander/git/sjo-sab
collected 1 item                                                                                                                                                     

tests/test_postproc.py .                                                                                                                                       [100%]

========================================================================= 1 passed in 0.04s ==========================================================================
```


sanderjo:
@safihire does this pytest work on Windows? I do use forward-slashes in the pathes, likes this:

```
"tests/data/obfuscated_single_rar_set"
```

https://www.pitt.edu/~naraehan/python3/file_path_cwd.html seems to tell it's OK on Windows, but can you verify?

sanderjo:
@safihire does this pytest work on Windows? I do use forward-slashes in the pathes, likes this:

```
"tests/data/obfuscated_single_rar_set"
```

https://www.pitt.edu/~naraehan/python3/file_path_cwd.html seems to tell it's OK on Windows, but can you verify?

Safihre:
> @safihire does this pytest work on Windows? I do use forward-slashes in the pathes, likes this:
> 
> ```
> "tests/data/obfuscated_single_rar_set"
> ```
> 
> https://www.pitt.edu/~naraehan/python3/file_path_cwd.html seems to tell it's OK on Windows, but can you verify?

AppVeyor [says it works](https://ci.appveyor.com/project/Safihre/sabnzbd/builds/33203507)! 💯  
Python handles this. If you want to make it perfect you would use
```python
os.path.join('test','data','obfuscated_single_rar_set')
```
But I also don't do that.. So not needed, if it works like this.

tests/test_postproc.py
```diff
@@ -0,0 +1,56 @@
...
+            # if workingdir still there from previous run, remove it:
+            try:
+                shutil.rmtree(workingdir)
+            except:
+                pass
+            # create a fresh copy
+            try:
+                shutil.copytree(sourcedir, workingdir)
+            except:
+                assert False
```
Safihre, line 38: Pytest has some nice tools for this to make the errors more clear. You can say this:
```python
pytest.fail("Could not create copy of files for rar_renamer")
```
This way it will display a better error instead of just an assertion-failure.

tests/test_postproc.py
```diff
@@ -0,0 +1,56 @@
...
+
+            # And now let the magic happen:
+            nzo = mock.Mock()
+            nzo.final_name = "some-download-name"
+            number_renamed_files = rar_renamer(nzo, workingdir)
+
+            # Remove workingdir again
```
Safihre, line 45: If here we could verify the exact names of the renamed files on the filesystem, it would make this test 100% perfect 🥇 
I would suggest a simple:
```python
expected_files = ['bla.part01.rar', 'bla.part02.rar']
files_renamed = os.listdir('output_folder')
for test_file in expected_files:
    assert test_file in files_renamed 
```
The `expected_files` can be an input to your `deobfuscate_dir` function.

sanderjo: Beautiful, but dangerous: in case of >1 rarsets, it is non-deterministic aka non-predictable which rarset gets which prefix ("A", or "B", or ...) 
And IMHO the `expected_files` would become big.

Alternatives:
* Check if there is one `*part009.rar` and two `*part007.rar`. That is a very strong indication rar_renamer() has correctly reached the end of both rar sets.
* change the prefix method rar_renamer() itself: base prefix on the biggest file inside the first rar (could be meaningful in regular usage / debugging), or a hash of it. But this would make rar_renamer() even more complex (if we take into account edge cases), and it's already complex now.


```
A--some-download-name.part001.rar
A--some-download-name.part002.rar
A--some-download-name.part003.rar
A--some-download-name.part004.rar
A--some-download-name.part005.rar
A--some-download-name.part006.rar
A--some-download-name.part007.rar
A--some-download-name.part008.rar
A--some-download-name.part009.rar
B--some-download-name.part001.rar
B--some-download-name.part002.rar
B--some-download-name.part003.rar
B--some-download-name.part004.rar
B--some-download-name.part005.rar
B--some-download-name.part006.rar
B--some-download-name.part007.rar
```

tests/test_postproc.py
sanderjo: :+1:

sanderjo:
@Safihre Can you help:

AppVeyor finds all well on Windows Python 3.5 3.6 3.7, but fails with 3.8:

image

It goes wrong in the shutil.copytree():

`E           shutil.Error: [('C:\\projects\\sabnzbd\\tests\\data\\obfuscated_single_rar_set\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\0f32e539-81a6-4549-9695-50603ba11f14', 'C:\\projects\\sabnzbd\\tests\\data\\obfuscated_single_rar_set\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\0f32e539-81a6-4549-9695-50603ba11f14', "[Errno 2] No such file or directory: 'C:\\\\projects\\\\sabnzbd\\\\tests\\\\data\\\\obfuscated_single_rar_set\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\working-dir\\\\0f32e539-81a6-4549-9695-50603ba11f14'"), ('C:\\projects\\sabnzbd\\tests\\data\\obfuscated_single_rar_set\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\working-dir\\work`

Where do all those working-dir subdirs come from, and why does that happen only on Windows python3.8?

https://docs.python.org/3/library/shutil.html#shutil.copytree only says

> Changed in version 3.8: Platform-specific fast-copy syscalls may be used internally in order to copy the file more efficiently. See Platform-dependent efficient copy operations section.

But I don't think that is relevant?

I tried a mini script with python3.8 and `shutil.copytree(src, dst)` , and no problem

tests/test_postproc.py
```diff
@@ -0,0 +1,97 @@
...
+            #print("workingdir is", workingdir)
+            # if workingdir is still there from previous run, remove it:
+            if os.path.isdir(workingdir):
+                try:
+                    shutil.rmtree(workingdir)
+                except PermissionError:
+                    pytest.fail(
+                        "Could not remove existing workingdir %s for rar_renamer",
+                        workingdir,
+                    )
```
Safihre, line 39: You can't use it like this. You are giving `workingdir` as the second argument for `fail` instead of as a print-input. Only `logging.info` etc work like this.
It should be
```python
 pytest.fail("Could not remove existing workingdir %s for rar_renamer" % workingdir)

sanderjo: Thanks! Solved.

sanderjo:
@Safihre I think I'm done now. Can you final check, and then (hopefully) merge?

PR #1481, opened 2020-05-30
extend filesystem tests

jcfp:
After they saved me from a mistake recently, I figured it was time to give these tests some tlc. Unittests for issue #1457 are included.

Also made some modifications in sabnzbd/filesystem.py:
- is_obfuscated_filename() looks for the presence of an extension to determine obfuscation but would clear filenames that merely ended in a dot. Modified to require the extension part to be at least strlen 2 (the dot itself plus one, i.e. an actual extension).
- set_permissions() didn't strip special bits (setuid, setgid, sticky) from files. Changed umask_file to safeguard against unintentional stupid umask settings. Handling of directories with special bits is unchanged.

Also came across some other things in _sabnzbd/filesystem.py_ that might be worth a look:
- get_unique_path() calls itself without upper limit on the number of attempts.
- around line 818/838: `try ... except ImportError` but no code that would raise that error when defining the posix version of diskspace_base(). Leftover debugging code with an except that never triggers?

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -252,7 +252,7 @@ def is_obfuscated_filename(filename):
     """ Check if this file has an extension, if not, it's
         probably obfuscated and we don't use it
     """
-    return os.path.splitext(filename)[1] == ""
+    return len(os.path.splitext(filename)[1]) < 2
 
 
 def real_path(loc, path):
@@ -471,8 +471,8 @@ def set_permissions(path, recursive=True):
             umask = int("0777", 8) & (sabnzbd.ORG_UMASK ^ int("0777", 8))
             report = False
 
-        # Remove X bits for files
-        umask_file = umask & int("7666", 8)
+        # Remove executable and special permissions for files
+        umask_file = umask & int("0666", 8)
 
         if os.path.isdir(path):
             if recursive:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -252,7 +252,7 @@ def is_obfuscated_filename(filename):
     """ Check if this file has an extension, if not, it's
         probably obfuscated and we don't use it
     """
-    return len(os.path.splitext(filename)[1]) < 2
+    return len(get_ext(filename)) < 2
 
 
 def real_path(loc, path):
--- a/tests/test_filesystem.py
+++ b/tests/test_filesystem.py
@@ -451,11 +451,12 @@ class TestTrimWinPath:
     def test_short_path(self):
         assert filesystem.trim_win_path(r"C:\short\path") == r"C:\short\path"
 
-    # @set_platform("win32")
-    # def test_long_path_short_segments(self):
-    #     test_path = "C:\\" + "A" * 20 + "\\" + "B" * 20 + "\\" + "C" * 20  # Strlen 65
-    #     # Current code causes the path to end up with strlen 70 rather than 69
-    #     assert filesystem.trim_win_path(test_path + "\\" + ("D" * 20)) == test_path + "\\" + "D" * 4
+    @pytest.mark.xfail(sys.platform == "win32", reason="Bug in trim_win_path")
+    @set_platform("win32")
+    def test_long_path_short_segments(self):
+        test_path = "C:\\" + "A" * 20 + "\\" + "B" * 20 + "\\" + "C" * 20  # Strlen 65
+        # Current code causes the path to end up with strlen 70 rather than 69 on Windows
+        assert filesystem.trim_win_path(test_path + "\\" + ("D" * 20)) == test_path + "\\" + "D" * 3
 
 
 @pytest.mark.skipif(sys.platform.startswith("win"), reason="Broken on Windows")
--- a/tests/test_filesystem.py
+++ b/tests/test_filesystem.py
@@ -185,15 +185,17 @@ class TestFileFolderNameSanitizer:
 
 
 class TestSameFile:
-    def test_nothing_in_common(self):
+    def test_nothing_in_common_win_paths(self):
         assert 0 == filesystem.same_file("C:\\", "D:\\")
         assert 0 == filesystem.same_file("C:\\", "/home/test")
+
+    def test_nothing_in_common_unix_paths(self):
         assert 0 == filesystem.same_file("/home/", "/data/test")
         assert 0 == filesystem.same_file("/test/home/test", "/home/")
         assert 0 == filesystem.same_file("/test/../home", "/test")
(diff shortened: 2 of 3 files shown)


jcfp:
Looks like I still have some debugging to do. :grimacing:

Safihre:
@jcfp we run black with `-l120` so it doesn't cut off everything at 88 chars.
Probably it will make many of the lines look better (although now you'll have to manually restore the ugly lines before running black again.

jcfp:
Thanks, I did notice but just as quickly forgot about that option again. At least black does reverse much of its own changes if you run with -l120 after first clipping things too short.

tests/test_filesystem.py
```diff
@@ -32,17 +42,30 @@ def test_empty(self):
...
+        assert filesystem.sanitize_filename(":aftertest") == "aftertest"
+        assert filesystem.sanitize_filename("::aftertest") == "aftertest"
+        assert filesystem.sanitize_filename(":after:test") == "test"
+        # Empty after sanitising with darwin colon handling
+        assert filesystem.sanitize_filename(":") == "unknown"
+        assert filesystem.sanitize_filename("test:") == "unknown"
+        assert filesystem.sanitize_filename("test: ") == "unknown"
 
     @set_platform("linux")
+    @set_config({"sanitize_safe": False})
```
Safihre, line 63: Nice addition! 👍

sabnzbd/filesystem.py
```diff
@@ -252,7 +252,7 @@ def is_obfuscated_filename(filename):
     """ Check if this file has an extension, if not, it's
         probably obfuscated and we don't use it
     """
-    return os.path.splitext(filename)[1] == ""
+    return len(os.path.splitext(filename)[1]) < 2
```
Safihre, line 255: Sounds good! Maybe we should use the `get_ext` function here?

tests/test_filesystem.py
```diff
@@ -61,21 +85,138 @@ def test_win_devices_not_win(self):
...
+    @set_platform("linux")
+    @set_config({"sanitize_safe": False})
+    def test_folder_illegal_chars_linux(self):
+        assert filesystem.sanitize_foldername('test"aftertest') == "test'aftertest"
+        assert filesystem.sanitize_foldername("test:") == "test-"
+        assert filesystem.sanitize_foldername("test<>?*|aftertest") == "test<>?*|aftertest"
+
+    def test_char_collections(self):
+        assert len(filesystem.CH_ILLEGAL) == len(filesystem.CH_LEGAL)
+        assert len(filesystem.CH_ILLEGAL_WIN) == len(filesystem.CH_LEGAL_WIN)
```
Safihre, line 109: A simple check, but usefull 👍

tests/test_filesystem.py
```diff
@@ -84,3 +225,491 @@ def test_capitalization(self):
         assert 1 == filesystem.same_file("D:\\", "d:\\")
         assert 2 == filesystem.same_file("\\\\?\\c:\\", "\\\\?\\C:\\Users\\")
         assert 2 == filesystem.same_file("/HOME/test123", "/home/test123/sub")
+
+    @pytest.mark.skipif(sys.platform.startswith("win"), reason="Requires a case-sensitive filesystem")
+    @pytest.mark.skipif(sys.platform.startswith("darwin"), reason="Requires a case-sensitive filesystem")
+    @set_platform("linux")
```
Safihre, line 231: Can be combined into 1, don't need the `set_platform` anymore also:
```python
@pytest.mark.skipif(sys.platform.startswith(("win", "darwin")), reason="Requires a case-sensitive filesystem")
```

tests/test_filesystem.py
```diff
@@ -84,3 +225,491 @@ def test_capitalization(self):
...
+    def test_long_path_non_win(self):
+        # Shouldn't have any effect on platforms other than Windows
+        assert filesystem.long_path(r"\\?\UNC\test") == r"\\?\UNC\test"
+        assert filesystem.long_path(r"\\?\F:\test") == r"\\?\F:\test"
+        assert filesystem.long_path(r"\\test") == r"\\test"
+        assert filesystem.long_path(r"F:\test") == r"F:\test"
+        assert filesystem.long_path("/test/dir") == "/test/dir"
+
+
+class TestCheckMountLinux(ffs.TestCase):
```
Safihre, line 305: Nice util!

Safihre:
Something's wrong on Windows, the [AppVeyor builds hang](https://ci.appveyor.com/project/Safihre/sabnzbd/builds/33222991/job/obindqgqav3ksvdf).

Safihre:
I already copied the "Remove executable and special permissions for files" bit to develop, so I can build a Beta2 that includes it.

jcfp:
I've gotten around to setting up some windoze test vm. Although the os is even slower than I remembered, it did (eventually) manage to reproduce the appveyor results. Will fix when I find time to work on that, no ETA for now though.

Safihre:
@jcfp If you use the Windows-images provided by Microsoft, just let them sit for a while so they can do their security/updates thing. Otherwise indeed it's terribly slow for the first few hours of usage.

jcfp:
Much of the appveyor issue was trying to use a unix style fake filesystem on a windows host which apparently doesn't work very well. Marked the affected test classes as non-windows and added equivalent tests using windows-style fake filesystems where useful.

The test failures in `trim_win_path()` exposed some more real issues:
- the path length is limited to 70 (rather than the 'below 70' promised in the docstring);
- when supplied with a path where the part prior to the last segment exceeds 69 chars (e.g. c:\\__70+ chars here__\lastsegment), `maxlen` goes negative and `folder[:maxlen]` may thus return almost the entire last path segment. This runs contrary to the concept of trimming the last part to reduce path length.

Given my limited knowledge of windows path length restrictions and related considerations, I didn't try to fix the function and instead removed or commented out a few test that would trigger the problems described above. These do need to be fixed though.

tests/test_filesystem.py
```diff
@@ -61,26 +85,761 @@ def test_win_devices_not_win(self):
...
+        # Outside the local filesystem
+        assert filesystem.check_mount("//test/that/") is True
+
+
+class TestTrimWinPath:
+    @set_platform("win32")
+    def test_short_path(self):
+        assert filesystem.trim_win_path(r"C:\short\path") == r"C:\short\path"
+
+    # @set_platform("win32")
```
Safihre, line 454: Great find, we need to fix it.
Let's instead mark it `@pytest.mark.xfail(sys.platform == "win32", reason="Bug in trim_win_path")`

jcfp: Done

Issue #1467, opened 2020-05-27
remarks about INSTALL.txt


sanderjo:
I read INSTALL.txt and have some remarks:

https://github.com/sabnzbd/sabnzbd/blob/develop/INSTALL.txt#L55-L56

Is that true? For example, on Ubuntu 20.04, with only python3 installed, no need to replace `python3 `with `python `in the commands. Just use `python3`, which is always correct and safe.

In https://github.com/sabnzbd/sabnzbd/blob/develop/INSTALL.txt#L58-L68 the needed OS packages are missing, which you need before you run the pip3 requirements:

For ubuntu that's
```
git # only needed when getting from github
par2 unrar python3-setuptools python3-pip
libffi-dev libssl-dev # needed if pip starts compiling ...?
```
Those last two packages names are probably Ubuntu/Debian specific names. @jcfp do you know the general name for libssl-dev? https://packages.ubuntu.com/search?keywords=libssl-dev&searchon=names&suite=eoan&section=all says 

Port issue:

https://github.com/sabnzbd/sabnzbd/blob/develop/INSTALL.txt#L74-L83 is that still what we want to advice?

jcfp:
Generic names for that libwhatever-dev stuff would simply be "_development headers for [openssl/ffi/etc.]_". If you paste any package name into the _Debian source package_ field at [packages.qa.debian.org](https://packages.qa.debian.org/common/index.html), it will tell you what source package produced it and provide a link to the homepage of that project.

Using `python3` everywhere should be fine; and even if it were to not work on some obscure system it's self-explanatory enough.

sanderjo:
@jcfp Additonal question: I wrote

`libffi-dev libssl-dev # needed if pip starts compiling ...?`

I want to change that to:

`libffi-dev libssl-dev # needed on non-X86 platforms to compile the python cryptography module`

OK?

Reason: On ARM (so: non-X86?), **without** those libs, I get:

```
Collecting cryptography
  Downloading cryptography-2.9.2.tar.gz (517 kB)
     |████████████████████████████████| 517 kB 5.2 MB/s
  Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
[...]

PR #1485, opened 2020-06-02
INSTALL.txt: updated details installation from source

sanderjo:
Closes https://github.com/sabnzbd/sabnzbd/issues/1467

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -52,19 +52,25 @@ Specific guides to install from source are available for Windows and macOS:
     https://sabnzbd.org/wiki/installation/install-macos
     https://sabnzbd.org/wiki/installation/install-from-source-windows
 
-Only Python 3.5 and above is supported. Replace python3 with python in the commands
-below if Python 3 is the only version of Python on your system.
+Only Python 3.5 and above is supported.
 
-Install all required Python packages by running:
-    python3 -m pip install -r requirements.txt
+You need to install
+    par2 unrar unzip python3-setuptools python3-pip
+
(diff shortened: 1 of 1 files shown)


INSTALL.txt
```diff
@@ -52,19 +52,25 @@ Specific guides to install from source are available for Windows and macOS:
...
     https://sabnzbd.org/wiki/installation/install-from-source-windows
 
-Only Python 3.5 and above is supported. Replace python3 with python in the commands
-below if Python 3 is the only version of Python on your system.
+Only Python 3.5 and above is supported.
 
-Install all required Python packages by running:
-    python3 -m pip install -r requirements.txt
+You need to install
+    par2 unrar unzip python3-setuptools python3-pip
```
Safihre, line 58: This is only for non-Windows/macOS systems:
`On Linux systems you need to install:`

INSTALL.txt
```diff
@@ -52,19 +52,25 @@ Specific guides to install from source are available for Windows and macOS:
...
+
+On non-X86 platforms (for which pypi does not provide all pre-compiled packages),
```
Safihre, line 60: `PyPI`

Safihre: And maybe remove the `()` in this line.

INSTALL.txt
sanderjo: We're talking here about "4) INSTALL with only sources", not installing from the packages. So, on Windows and MacOS ... don't you need to install unrar etc?

Safihre: No, these are included here in the repo and when you install Python on Windows it already includes setuptools and pip.

PR #1490, opened 2020-06-06
fix urllib deprecation warning, base64 encoding in _build_request()

### LINKED ISSUE #1415: [dev-mode] urlgrabber: DeprecationWarning
I think I can solve this one myself.
```
2020-05-01 07:38:30,155::INFO::[urlgrabber:126] Grabbing URL https://sabnzbd.org/tests/test_download_100MB.nzb
/home/sander/git/sjo-sab-1mei2020/sabnzbd/urlgrabber.py:328: DeprecationWarning: urllib.parse.splittype() is deprecated as of 3.8, use urllib.parse.urlparse() instead
  urltype, rest = urllib.parse.splittype(url)
/home/sander/git/sjo-sab-1mei2020/sabnzbd/urlgrabber.py:329: DeprecationWarning: urllib.parse.splithost() is deprecated as of 3.8, use urllib.parse.urlparse() instead
  realhost, rest = urllib.parse.splithost(rest)
/home/sander/git/sjo-sab-1mei2020/sabnzbd/urlgrabber.py:331: DeprecationWarning: urllib.parse.splituser() is deprecated as of 3.8, use urllib.parse.urlparse() instead
  user_passwd, realhost = urllib.parse.splituser(realhost)
```

jcfp:
Removed the deprecation warning by using urllib.parse.urlparse() and tried to keep url handling the same as the old code.

Came across another bug while testing some urls: `str.encode("base64")` is py2-only and results in `LookupError: 'base64' is not a text encoding; use codecs.encode() to handle arbitrary codecs` with py3. Solution looks a bit ugly with the double string/bytes conversion but that seems unavoidable.

Closes #1415.

### DIFF
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -29,6 +29,7 @@ import urllib.error
 import urllib.parse
 from http.client import IncompleteRead
 from threading import Thread
+import base64
 
 import sabnzbd
 from sabnzbd.constants import DEF_TIMEOUT, FUTURE_Q_FOLDER, VALID_NZB_FILES, Status
@@ -40,6 +41,7 @@ from sabnzbd.postproc import PostProcessor
 import sabnzbd.cfg as cfg
 import sabnzbd.emailer as emailer
 import sabnzbd.notifier as notifier
+from sabnzbd.encoding import ubtou, utob
 
 
 _BAD_GZ_HOSTS = ('.zip', 'nzbsa.co.za', 'newshost.za.net')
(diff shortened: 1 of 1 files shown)


sabnzbd/urlgrabber.py
```diff
@@ -325,12 +327,12 @@ def _build_request(url):
...
     user_passwd = None
-    urltype, rest = urllib.parse.splittype(url)
-    realhost, rest = urllib.parse.splithost(rest)
-    if realhost:
-        user_passwd, realhost = urllib.parse.splituser(realhost)
-        if user_passwd:
-            url = '%s://%s%s' % (urltype, realhost, rest)
+    u = urllib.parse.urlparse(url)
+    if u.username or u.password:
+        user_passwd, host_port = u.netloc.split('@')
```
Safihre, line 332: This would fail for `"http://user@gmail.com:password@usenet.nl"`.
Maybe we can even make some unittest for this function.
Probably better to use the `u.username` and `u.password` to build the input for `base64.b64encode`.

jcfp: Damn, how did I miss that...

Anyway, from the http basic auth specs it seems that both user and pass are required, so I'll modify the code to only send basic auth if both are present in the url, and otherwise discard (if, for example, only a user is set but no pass). The old code did pass such fringe cases along though.

The thought of unittests did pass my mind when running some test urls, but without a clear idea of what to accept and what not I didn't yet write any.

Safihre:
Thanks!

PR #1498, opened 2020-06-10
show active connections by default

jcfp:
Show active connections by default under in Glitter's _Status and interface options // Connections_. It seems to me that's the main reason anyone would take a look at that tab, yet the checkbox is unchecked by default.


jcfp:
> Well, I didn't do that because the updating is quite cpu intensive. Both for the server, but also for the client. Chrome showed large CPU usage. But this was back in 2015!

I have to admit I didn't check that, since there wasn't any noticeable slowdown or other negative effects with Firefox since I made that change locally a few weeks ago. Just now tried Chrome though and CPU use about triples :flushed: when activating that checkbox... and that's without any active downloads. :vomiting_face:

Safihre:
It's also because we use knockoutJS and already in 2015 it was an old-skool framework, probably if we used something newer like Angular it wouldn't be so bad..

PR #1502, opened 2020-06-12
Refactor the adding of NZB's

Safihre:
What started with me trying to fix the Retry-button, turned into a whole giant refactor of the way we add things to the queue. It was such a big giant mess of old pre-2010 code, I kept digging deeper into the mud 💩 .

I would really like to add some form of tests for this, because there are a lot of ways this can go wrong. Anyone have any ideas @sabnzbd/contributers ?

- All is handled by sabnzbd.add_nzbfile
- Moved the actual file-processing to nzbparser
- We always support gzip in URLGrabber
- Remove upload.py, all handled by add_nzbfile
- Rework the dirscanner and urlgrabber to the new reality
- Retry job was broken if you added a file


Safihre:
I now found `pytest-httpserver`, which uses `werkzeug` (already included because use `pytest-httpbin`) to host a small local webserver. I will look into using this to add more tests.

Issue #1513, opened 2020-06-18
Check for unwanted extensions in the NZB


Safihre:
Supported by: https://forums.sabnzbd.org/viewtopic.php?f=3&t=24804

We only check inside the rar-files if there are unwanted extensions.. but what if they are just as a file inside the main NZB? We should detect those to.

sanderjo:
can you give a hint:
- where is the loading of a NZB done?
- where is (de)selecting of a specific file in GUI done?

sanderjo:
WIP: Detection is working, and failing works too, but in an ugly way ...

![image](https://user-images.githubusercontent.com/1273502/85043458-ab8e7800-b18c-11ea-9c38-09ee1bc32da7.png)

sanderjo:
With the same code, and a fabricated with EXE in plain sight, and the rest just in plain RARs (NZB here: https://raw.githubusercontent.com/sanderjo/NZBs/master/reftestnzb_small_post_with_EXE_as_plain_file.nzb ), the code works quite nicely:

- On NZB import, the EXE file is skipped, with a warning.
- The rest is imported, and succesfully downloaded.

@Safihre I just skip the EXE (and other unwanted extensions). No pausing. I see that as the only feasible to do, without too much code overhauling. 

![image](https://user-images.githubusercontent.com/1273502/85130930-2069bc80-b236-11ea-93cb-0d16655ff16c.png)

PR #1515, opened 2020-06-19
unwanted extension: do not import such a file in an NZB

sanderjo:
code prototype. It works, but feels a bit hacked together: for example: first the if for the exception case (unwanted ext), then the mainstream ... 
Feedback very welcome @Safihre @jcfp 

Closes https://github.com/sabnzbd/sabnzbd/issues/1513

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -871,7 +871,7 @@ class NzbObject(TryList):
                 and get_ext(nzf.filename).replace(".", "") in cfg.unwanted_extensions()
             ):
                 # So we found an unwanted extension
-                logging.warning("Unwanted Extension in file %s (%s)" % (nzf.filename, self.final_name))
+                logging.warning(T("Unwanted Extension in file %s (%s)") % (nzf.filename, self.final_name))
                 # Pause, or Abort:
                 if cfg.action_on_unwanted_extensions() == 1:
                     logging.debug("Unwanted extension ... pausing")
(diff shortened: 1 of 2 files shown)


sabnzbd/nzbstuff.py
```diff
@@ -864,6 +864,24 @@ def __init__(
...
                 self.priority = NORMAL_PRIORITY
 
+        # Check if there are any unwanted extension in plain sight in the NZB itself
+        for nzf in self.files:
+            if (
+                cfg.action_on_unwanted_extensions() >= 1
+                and get_ext(nzf.filename).replace(".", "") in cfg.unwanted_extensions()
+            ):
+                # So we found an unwanted extension
+                logging.warning(T("Unwanted Extension in file %s (%s)") % (nzf.filename, self.final_name))
```
Safihre, line 874: Code-style (just how we do it everywhere else)
```
logging.warning(T("Unwanted Extension in file %s (%s)"), nzf.filename, self.final_name)
```

sabnzbd/nzbstuff.py
```diff
@@ -864,6 +864,24 @@ def __init__(
...
+                # Pause, or Abort:
+                if cfg.action_on_unwanted_extensions() == 1:
+                    logging.debug("Unwanted extension ... pausing")
+                    self.unwanted_ext = 1
+                    self.pause()
+                if cfg.action_on_unwanted_extensions() == 2:
+                    logging.debug("Unwanted extension ... aborting")
+                    self.fail_msg = T("Unwanted extension ... aborting")
```
Safihre, line 882: I would use `T("Aborted, unwanted extension detected")`, it's a better history-status and it's already an available translated-texts from `assembler.py`.

PR #1562, opened 2020-07-29
Also find passwords which are not at the end of the file

kaiffeetasse:
sabnzbd only finds passwords which are at the end of the file, e.g. `myawesomenzbfile{{password}}.nzb`. 
Often the filename contains some numbers or other characters after the password, but then sabnzbd isn't able to extract the password. (e.g. `Sleepy.Hollow.S03E06.GERMAN.1080P.WEB.H264.INTERNAL-THUGANOMiCS{{HuNmiLfjne5zCalIoBhcE1}} [15877906376368]`)

With this pull request sabnzbd will be able to extract those passwords. If more than one is found, sabnzbd will use the first one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,4 +68,3 @@ Conditions:
 - Bugfixes created specifically for a release branch are done there (because they are specific, they're not cherry-picked to `develop`).
 - Bugfixes done on `develop` may be cherry-picked to a release branch.
 - We will not release a 1.0.2 if a 1.1.0 has already been released.
-- Code must be formatted with [black code formatter](https://github.com/psf/black): `black sabnzbd --line-length=120 --target-version=py35`
--- a/README.md
+++ b/README.md
@@ -68,3 +68,4 @@ Conditions:
 - Bugfixes created specifically for a release branch are done there (because they are specific, they're not cherry-picked to `develop`).
 - Bugfixes done on `develop` may be cherry-picked to a release branch.
 - We will not release a 1.0.2 if a 1.1.0 has already been released.
+- Code must be formatted with [black code formatter](https://github.com/psf/black): `black sabnzbd --line-length=120 --target-version=py35`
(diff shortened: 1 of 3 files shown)


Safihre:
Thanks! 👍 sounds like a good addition.
I apologize for not providing good instructions how to run `black`. 
To fix things, revert your last commit c7b16a1 and the run `black sabnzbd --line-length=120 --target-version=py35`

What would be amazing if we could add unittest for the `scan_password` function. But as these do not exist at the moment, we can merge this without them.
If you feel up to the task: check the `tests` folder and create a new file `test_nzbstuff.py` where you add tests for `scan_password`.

kaiffeetasse:
Hi! No problem. Maybe you could add those instructions to the README for future contributors.

I formatted the file according to your conventions and added tests for scan_password in nzbstuff.py.

README.md
```diff
@@ -68,3 +68,4 @@ Conditions:
 - Bugfixes created specifically for a release branch are done there (because they are specific, they're not cherry-picked to `develop`).
 - Bugfixes done on `develop` may be cherry-picked to a release branch.
 - We will not release a 1.0.2 if a 1.1.0 has already been released.
+- Code must be formatted with [black code formatter](https://github.com/psf/black): `black sabnzbd --line-length=120 --target-version=py35`
```
Safihre, line 71: I will create a Github pull request template for this, I think that would be a better place 😊 this specific section is about how we handle our branches.

tests/test_nzbstuff.py
```diff
@@ -0,0 +1,19 @@
...
+    def test_scan_passwords(self):
+        file_names = {
+            "my_awesome_nzb_file{{password}}": "password",
+            "file_with_text_after_pw{{passw0rd}}_test": "passw0rd",
+            "file_without_pw": None,
+            "file_with_multiple_pw{{first-pw}}_{{second-pw}}": "first-pw",
+        }
+
+        for file_name, password in file_names.items():
+            assert nzbstuff.scan_password(file_name)[1] == password
```
Safihre, line 19: Great you were able to add tests! 
I think it's also essential to not only check the extracted password, but also the filename.
What if it extracts the password, but produces nonsense for the filename?

kaiffeetasse: done 👍

PR #1568, opened 2020-08-07
Permissions were not set correctly when creating directories

Safihre:
Restores changes made in d2e0ebe

@jcfp I tried to add some unittest, but I'm not very good at this permissions stuff.
It needs maybe some more permissions testing, but writing these tests already took me almost a day 🙃


jcfp:
I don't think it necessary to add more testing than this; after all the error you really want to catch here is not setting the perms on the entire tree.

PR #1603, opened 2020-09-08
Deobfuscate final files: more intelligence, default obfuscated = True, 

sanderjo:
plus much more unit testing

### DIFF
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -82,7 +82,7 @@ def entropy(string):
     return entropy
 
 
-def is_probably_obfuscated(myinputfilename) -> object:
+def is_probably_obfuscated(myinputfilename):
     """Returns boolean if filename is likely obfuscated. Default: True
     myinputfilename can be a plain file name, or a full path"""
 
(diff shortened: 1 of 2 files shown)


sabnzbd/deobfuscate_filenames.py
```diff
@@ -97,38 +97,49 @@ def is_probably_obfuscated(myinputfilename):
...
     # little entropy in the filebasename is a sign of useless names
     if entropy(filebasename) < 3.5:
         logging.debug("Obfuscated: entropy < 3.5")
         return True
+
     # high entropy in the filebasename is a sign of useful name, so not obfuscated
     if entropy(filebasename) > 4.0:
         logging.debug("Not obfuscated: entropy > 4.0")
         return False
+    """
```
Safihre, line 138: If this code is not needed, please remove. For history we have Git to always recover code if we need.

Safihre (changes requested):

sabnzbd/deobfuscate_filenames.py
```diff
@@ -97,38 +97,49 @@ def is_probably_obfuscated(myinputfilename):
...
-    if spacesdots > 3:
+    # Example: "this is a download"
+    if spacesdots >= 3:
         # useful signs in filebasename, so not obfuscated
-        logging.debug("Not obfuscated: spacesdots > 3")
+        logging.debug("Not obfuscated: spacesdots >= 3")
         return False
 
-    if decimals > 3 and spacesdots > 1:
+    # Example: "Beast 2020"
```
jcfp, line 117: This example doesn't meet the `spacedots > 1` requirement of the check (and is only considered not obfuscated as a result of the later 'Catullus' check).

sabnzbd/deobfuscate_filenames.py
sanderjo: I'll do!

sabnzbd/deobfuscate_filenames.py
sanderjo: Nice find! Thanks. I'll take care of it.

PR #1612, opened 2020-09-14
Deobfuscate newfiles

sanderjo:
based on `newfiles` list instead of `workdir_complete`


sabnzbd/deobfuscate_filenames.py
```diff
@@ -118,11 +118,21 @@ def is_probably_obfuscated(myinputfilename):
...
 
 
-def deobfuscate(workingdirectory, usefulname):
-    """ In workingdirectory, check all filenames, and if wanted, rename """
+def deobfuscate_dir(workingdirectory, usefulname):
+    """ In workingdirectory, check all filenames, and if wanted, deobfuscate """
+
+    list_of_files = []
+    for (dirpath, dirnames, filenames) in os.walk(workingdirectory):
+        list_of_files += [os.path.join(dirpath, file) for file in filenames]
```
Safihre, line 126: Let's just use the `globber_full` function, it does exactly this and is already imported! See deleted line below. Just leave out the pattern and it should list then all.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -118,11 +118,21 @@ def is_probably_obfuscated(myinputfilename):
...
+    deobfuscate_list(list_of_files, usefulname)
+
+
+def deobfuscate_list(filelist, usefulname):
```
Safihre, line 130: Might just have 1 function `deobfuscate` since it's so easy to use `globber_full` as the input in case a directory listing needs to be used.

tests/test_deobfuscate_filenames.py
```diff
@@ -54,29 +60,83 @@ def test_is_probably_obfuscated(self):
...
+        # Full test: a filelist with a non-useful named file in it: Test that deobfuscate() works and renames it
+        # ... but not the files that are NOT in the filelist (although in same directory)
+
+        # Create directory (with a random directory name)
+        dirname = os.path.join(SAB_DATA_DIR, "testdir" + str(random.randint(10000, 99999)))
+        os.mkdir(dirname)
+
+        # Create a big enough file with a non-useful filename
+        output_file1 = os.path.join(dirname, "111c1c9e2bdfb5114044bf25152b7eaa.bla")
+        create_big_file(output_file1)
```
Safihre, line 111: Is this test different than the one above? Adds to my feeling we can do with just 1 function and leave the directory listing out :)

Safihre (changes requested):

sabnzbd/deobfuscate_filenames.py
sanderjo: Indeed, I had already done that.

tests/test_deobfuscate_filenames.py
sanderjo: Already removie in the commit deobfuscate: removed deobfuscate_dir()

sabnzbd/deobfuscate_filenames.py
```diff
@@ -118,11 +118,11 @@ def is_probably_obfuscated(myinputfilename):
     return True  # default not obfuscated
 
 
-def deobfuscate(workingdirectory, usefulname):
-    """ In workingdirectory, check all filenames, and if wanted, rename """
+def deobfuscate_list(filelist, usefulname):
+    """ Check all files in filelist, and if wanted, deobfuscate """
```
Safihre, line 123: Just to be sure, I would add here at the start:
```python
filelist = [f for f in filelist if os.path.exists(f)]
```
Since all code below would crash the post-processor.
It could be possible that the paths in `newfiles` are invalid, since we parse them from unrar-output. Which could be ugly when running on Windows with Unicode.

Safihre (changes requested):

sabnzbd/deobfuscate_filenames.py
sanderjo: thanks

PR #1614, opened 2020-09-19
Add typing hints to some SABnzbd-specific objects and general functions

Safihre:
With IDE's like PyCharm these type-hints turn out to be a very usefull tool. It will highlight when you try to supply variables to functions that don't match the specifications. Additionally, the auto-complete actually knows what kind of objects you are working with.
Also includes fixes that I found because I added these type hints!

I think I also want to work on changing a bit of the structure, so we no longer need these `NzbQueue.do.xxx` structures, which are very weird.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1232,7 +1232,7 @@ def main():
     if autobrowser is not None:
         sabnzbd.cfg.autobrowser.set(autobrowser)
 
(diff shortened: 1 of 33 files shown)


Safihre:
Will re-open when I fixed the import issues on Python 3.6!

PR #1628, opened 2020-10-04
Correctly handle reference counting of SABnzbd-objects

### LINKED ISSUE #1472: memory consumption
Hi,

From a clean start, sab3 idles at around 250mb(v) &45mb (r). When downloading this peaks at approx 490/270; however, when sab is idle again rami usage is 280/63. While this is significantly better than previous iterations of Sab I wondered if this is to be expected or is there a leak? 

Please see attached log
[sabnzbd.log](https://github.com/sabnzbd/sabnzbd/files/4703509/sabnzbd.log)

Safihre:
Triggered by #1472 I investigated the observed behavior that memory wasn't cleared after jobs were finished.
The root cause turned out to be treefold:

1. In all `Thread`'s, like the `Assembler`, `Decoder`, `PostProcessor` we use queues. We were using code like this: `nzo, nzf, file_done = self.queue.get()`. The result is that the reference to `nzo` and `nzf` is only cleared when the next job is picked up by the thread. For example in the case of the `PostProcessor` this means that the `NzbObject` (and all the `NzbFile`'s and `Article`'s in it) is kept alive until the next job is post-processed. This can be quite a while!
This is resolved by resetting these variables to `None` just before the `self.queue.get()`. IDE's like PyCharm indicate this as a useless statement that should be removed, so I added comments to all these resets.
2. Additionally, we start the `DirectUnpacker` in it's own thread and have a circular reference to/from the `NzbObject`. This resulted in the `NzbObject` staying alive because it was references from the dead `DirectUnpacker`. This was resolved by setting the reference to `None`.
3. In the `DecoderWorker` the raw and decoded data was preserved until the next article was offered to be decoded. Since we have 4 `DecoderWorker`-threads, this resulted in another 5MB of extra memory.

To prevent this in the future, I added a special `api?mode=gc_stats` that lists all the SABnzbd specific-objects that are alive (`NzbObject`/`NzbFile`/`Article`).
By checking after every functional download test if no objects are left alive we can make sure this is not reintroduced in the future!  

Using that test I actually found that also in the `URLGrabber` we had an unresolved reference.

On my Windows machine, this resulted in:

Action | `develop` | `bugfix/memusage`
------------ | ------------ | -------------
Start | 52MB | 52MB
After download 100MB | 61MB | 53MB

~So, there is still some memory increase, but less. I imagine it has something also to do with `cherrypy`. But that's for another day.~
This already took me almost a full weekend of manual debugging using the `gc` module (there are no packages that can handle a big multi-threaded application like SABnzbd).

I even wrote code like this below, but it actually turned out **not** to be needed after the 2 solutions mentioned above were implemented!
```python
    def clear_object_references(self):
        """Clear all object references to child objects so we
        all be removed by the garbage collector (see #1472)"""
        # Remove (circular) references from NzbFile objects
        # Sub-lists need to be cleared individually
        for setname in self.extrapars:
            self.extrapars[setname].clear()
        self.extrapars.clear()
        self.partable.clear()
        self.files.clear()
        self.files_table.clear()
        self.finished_files.clear()

        # Clear articles, as they have references to NzbFile objects
        self.saved_articles.clear()

        # Need to clear the DirectUnpacker-NzbObject reference. If we don't,
        # the NzbObject is kept in memory due to a reference from the stopped Thread.
        self.direct_unpacker = None
```

Closes #1472

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -22,6 +22,7 @@ sabnzbd.api - api
 import os
 import logging
 import re
+import gc
 import datetime
 import time
 import json
@@ -902,6 +903,14 @@ def _api_server_stats(name, output, kwargs):
     return report(output, keyword="", data=stats)
 
 
+def _api_gc_stats(name, output, kwargs):
+    """Function only intended for internal testing of the memory handling"""
+    # Collect before we check
+    gc.collect()
+    # We cannot create any lists/dicts, as they would create a reference
+    return report(output, data=[str(obj) for obj in gc.get_objects() if isinstance(obj, sabnzbd.nzbstuff.TryList)])
+
+
 ##############################################################################
 _api_table = {
     "server_stats": (_api_server_stats, 2),
@@ -938,6 +947,7 @@ _api_table = {
     "restart_repair": (_api_restart_repair, 2),
     "disconnect": (_api_disconnect, 2),
     "osx_icon": (_api_osx_icon, 3),
+    "gc_stats": (_api_gc_stats, 3),
     "rescan": (_api_rescan, 2),
     "eval_sort": (_api_eval_sort, 2),
     "watched_now": (_api_watched_now, 2),
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -56,7 +56,7 @@ class Assembler(Thread):
     def run(self):
         while 1:
             # Set NzbObject and NzbFile objects to None so references
-            # from this thread do not keep the objects alive (see #1472)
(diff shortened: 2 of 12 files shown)


sanderjo:
> To prevent this in the future, I added a special `api?mode=gc_stats` that lists all the SABnzbd specific-objects that are alive (`NzbObject`/`NzbFile`/`Article`).

Cool!!!

Safihre:
Found the last problem!

Safihre:
Also in `Downloader` a reference to `Article` was stored, which was found in #1749.

PR #1651, opened 2020-10-22
only replace the first occurence of "script_"

jcfp:
Use of str.replace() without a count replaces all occurences. As a result, scripts with filenames such as "my_script_for_sab.py" would be mangled when trying to set them as action on queue completion.


jcfp:
Updated to also modify the check of var. action, which suffered from a similar problem.

PR #1655, opened 2020-10-23
randomize age for generated nzb files

jcfp:
Useful for testing queue sorting function of the api. Timestamp values are randomly chosen between september '93 and now.


Safihre:
@jcfp Can you make black happy? 💯

Safihre:
😁

PR #1659, opened 2020-10-26
Fix nzbstuff scan_password, expand tests

jcfp:
Came across some weird results testing the api job rename function, traced back to issues in scan_password.

Bug fixes:
* ensure `{{` and `}}` appear in the correct order; necessary now that #1562 removed the requirement for the password to appear at the end;
* guard against returning an empty filename in favour of a non-empty password, i.e. consider the entire input as the password when a delimiter is found at the start (affects all delimiter styles);
* no longer strip a char off the end of the name if it starts with `/`;
* for multiple occurrences of `{{...}}` do a greedy match (restoring old behaviour) instead of picking the first.

Tests:
* Rename to match the function name;
* Collapse functions into a single one and parametrise;
* Add a bunch of tests for corner cases, Unicode input, empty output.

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -2086,9 +2086,11 @@ def scan_password(name: str) -> Tuple[str, Optional[str]]:
     if "http://" in name or "https://" in name:
         return name, None
 
-    braces = name[1:].find("{{") + 1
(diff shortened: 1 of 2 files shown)


Safihre:
Nice

PR #1666, opened 2020-10-31
fix order for sorting queue by avg_age

jcfp:
Doing an api call to sort the queue in ascending order of avg_age actually sorts it in descending order, and vice versa. The other sorting methods (size and name) are not affected. The root cause is `nzbqueue.sort_queue()` not accounting for job ages being stored in the form of unix timestamps, where higher values equal lower age.

This is easily fixed by reversing the 'reverse' variable when for sorting by avg_age. In the interface templates, sorting options were correctly described towards the user in terms of oldest->newest but the underlying javascript vars embed the incorrect sort order in their names. As a result, the changes make for some weird reading (variables with "asc" in their names now set "desc"), but everything still does what it says on the tin and changes to the translations are avoided.

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.queue.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.queue.js
@@ -228,11 +228,11 @@ function QueueListModel(parent) {
         switch($(event.currentTarget).data('action')) {
             case 'sortAgeAsc':
                 sort = 'avg_age';
-                dir = 'asc';
+                dir = 'desc';
                 break;
             case 'sortAgeDesc':
                 sort = 'avg_age';
-                dir = 'desc';
+                dir = 'asc';
                 break;
             case 'sortNameAsc':
                 sort = 'name';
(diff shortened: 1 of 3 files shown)


Safihre:
I would expect only change in 1 spot is needed, either JS or Python? Why both are needed?

jcfp:
> I would expect only change in 1 spot is needed, either JS or Python? Why both are needed?

The api call feeds directly into the python function, which sorted incorrectly. Hence: fixing that function fixes the results of the api call.

The interface scripts on the other hand had correct sorting _with the incorrect python function_. The skintexts clearly demonstrate how the javascript was build around the incorrect order returned by the sorting function: `"Glitter-sortAgeAsc": TT("Sort by Age <small>Oldest&rarr;Newest</small>")` - the variable name says ascending, but "oldest &rarr; newest" is the very definition of descending age. Hence, fixing the underlying sorting function requires edits to the javascript too (so it keeps getting things right).

Safihre:
Makes sense.
Are you testing all the API calls?

jcfp:
Yup, see #1668

Safihre:
https://forums.sabnzbd.org/viewtopic.php?f=3&t=25237
Did we do it wrong or the user was wrong?

jcfp:
Reproduces as described on the forums. Found the problem: another piece of code designed to work with the broken sorting function. I'll file a PR once I've tested the fix.

Issue #1650, opened 2020-10-22
change_complete_action api call should verify input


jcfp:
The [api call](https://sabnzbd.org/wiki/advanced/api#change_complete_action) accepts a number of predefined actions as well as scripts in case the supplied _value_ starts with "script_". Unfortunately, the value of the script is not subject to any further checks. As a result, not only do invalid script values get accepted, but, using relative paths (to the script dir), any command on the system can be called. The setting should be limited to only whatever scripts actually exist in the script dir.

Safihre:
Hmm, it's not just this call but all the calls that have `script` as input.
Such as changing the script of a job, category, RSS-feed, etc.

jcfp:
> Hmm, it's not just this call but all the calls that have `script` as input.
Yeah, I should have mentioned not checking any of the others; change_completion_action just happened to be the one I came across while writing tests.

PR #1668, opened 2020-11-02
Api tests

jcfp:
Add tests for the api, plus a few bug fixes and enhancements:
- basic tests for api_handler;
- functional tests for most api calls not already part of other functional tests;
- verification (using tavern) of output format, presence of fields and expected value types for api calls that return bigger data structures (queue, history, stats, files), for all output styles (json, text, xml);
- ~~fix deprecation warning in a sabnews regex;~~ [reverted: broke functional downloads test]
- enable output styles other than json from testhelper.get_api_result().


tests/data/tavern/api_get_files_format.yaml
```diff
@@ -0,0 +1,107 @@
...
+        content-type: !re_search "charset=(UTF|utf)-8"
+        pragma: "no-cache"
+        access-control-allow-origin: "*"
+      verify_response_with:
+        function: tavern.testutils.helpers:validate_regex
+        extra_kwargs:
+          # Verify response parameters with a regex using positive
+          # looksaheads, so parameters match regardless of their order
+          # of appearance in the text output.
+          expression: "^\\[{{(?=.*'filename': )(?=.*'mbleft': )(?=.*'mb': )(?=.*'bytes': )(?=.*'age': )(?=.*'nzf_id': 'SABnzbd_nzf_.*')(?=.*'status': '(finished|active|queued)').*}}.*\\]$"
```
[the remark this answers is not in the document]
Safihre: Should we maybe leave out the tests for `text` output? Since they seem quite ugly and easy to brake..

tests/test_api.py
```diff
@@ -0,0 +1,67 @@
...
+            assert api.api_handler({"mode": []}).strip() == expected_error
+            assert api.api_handler({"mode": ""}).strip() == expected_error
+            assert api.api_handler({"mode": None}).strip() == expected_error
+
+    def test_version(self):
+        assert api.api_handler({"mode": "version"}).strip() == sabnzbd.__version__
+
+    @set_config({"disable_key": False})
+    def test_auth(self):
+        assert api.api_handler({"mode": "auth"}).strip() == "apikey"
```
Safihre, line 51: Should we just add `.strip` to the `api_handler`?

tests/test_functional_api.py
```diff
@@ -0,0 +1,1251 @@
...
+            # Mock time when calling add_history_db() to randomize completion times
+            almost_time = mock.Mock(return_value=time.time() - randint(0, 10 ** 8))
+            with mock.patch("time.time", almost_time):
+                self.add_history_db(
+                    nzo,
+                    storage=os.path.join(os.path.dirname(db.HistoryDB.db_path), "placeholder_workdir"),
+                    postproc_time=randint(1, 10 ** 3),
+                    script_output="",
+                    script_line="",
+                )
```
Safihre, line 121: Very nice 👍

tests/test_functional_api.py
```diff
@@ -0,0 +1,1251 @@
...
+            ("daemon_version", sabnzbd.__version__),
+            ("daemon_apikey", self.daemon_apikey),
+        ]
+        if extra_vars:
+            vars.append(extra_vars)
+        if self.api_server_parameters["add_fake_history"]:
+            vars.append(("daemon_history_size", self.history_size))
+
+        result = run(
+            os.path.join(os.path.dirname(__file__), "data", "tavern", test_name + ".yaml"),
```
Safihre, line 183: Use `SAB_DATA_DIR`.

tests/test_functional_api.py
```diff
@@ -0,0 +1,1251 @@
...
+        delete_these = [slot["nzo_id"] for slot in json["history"]["slots"]]
+        assert len(delete_these) == len(json["history"]["slots"])
+
+        # Kill 'm with fire!
+        json = self._get_api_history({"name": "delete", "value": ",".join(delete_these)})
+        assert json["status"] is True
+
+        # Make sure nothing survived
+        json = self._get_api_history()
+        assert json["history"]["noofslots"] is int(0)
```
Safihre, line 1246: `== 0`?

tests/test_functional_api.py
```diff
@@ -0,0 +1,1251 @@
...
+            ("", "PassWord", None, "PassWord", False),
+            (None, None, None, None, False),
+            (None, "PassWord", None, "PassWord", False),
+            ("Job}}Name{{FTW", None, "Job}}Name{{FTW", None, True),  # Issue #1659
+            (".{{PasswordOnly}}", None, ".{{PasswordOnly}}", None, True),  # Issue #1659
+            # Supplying password through value3 should leave any {{...}} in value2 alone
+            ("Foo{{Bar}}", "PassFromValue3", "Foo{{Bar}}", "PassFromValue3", True),
+        ],
+    )
+    def test_api_queue_change_job_name(self, value2, value3, expected_name, expected_password, should_work):
```
Safihre, line 951: Nice!

tests/test_functional_api.py
```diff
@@ -0,0 +1,1251 @@
...
+                warn("Failed to remove %s" % job_dir)
+
+    def _purge_queue(self, del_files=0):
+        """ Clear the entire queue """
+        self._get_api_json("queue", extra_args={"name": "purge", "del_files": del_files})
+        assert len(self._get_api_json("queue")["queue"]["slots"]) is int(0)
+
+
+@pytest.fixture(scope="class")
+def api_server(request):
```
Safihre, line 257: Hmmm, seems similar to the code in `conftest.py`. Nothing we can reuse to generalize this?

tests/test_functional_api.py
jcfp: Done, also removed all other occurences of the same.

tests/test_functional_api.py
jcfp: I refrained from modifying start_sabnzbd_and_selenium() because I'd end up introducing bugs in the existing functional tests while making my own life harder. Now that the api tests are mostly complete and given the amount of overlap, the api_server() code should be integrated in conftest.py though. I'll get to work on that.

tests/test_api.py
jcfp: The api code explicitly adds those newlines, so they may well serve some purpose. Maybe simply that the receiving end can be sure they read the entire line, rather than partial output?

Safihre: Aaah, ok!

tests/data/tavern/api_get_files_format.yaml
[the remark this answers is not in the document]
jcfp: I agree these checks for the text version look fugly, if only because they can't be split over multiple lines and need escaping for both regex and tavern (e.g. the double `{{`). But once in place, I don't think they are particularly risky nor hard to maintain. Most of it is just a clearly delimited list of parameters being checked, easily expanded/modified/removed as needed.

tests/test_functional_api.py
jcfp: Scope differences don't help with this. I wrote api_server() using a class scope so the server config can easily be tuned from the test class and a fake history db added at startup. As the api tests require just sab itself, there wasn't much of a penalty.

For start_sabnzbd_and_selenium() on the other hand, scope is set to session, probably in order to avoid having to start sab, chrome, and sabnews multiple times during a pytest run. Mixing scopes (or dynamically setting them) gets ugly real quick, and moving everything to scope class isn't exactly without penalty, so I'll need to change api_server() to use session scope and make it trigger stop+start of the sab instance whenever a "refresh" of the fake history db is needed.

Safihre:
Nice these last code changes.
I do wonder why the tests take 15min on Windows (AppVeyor) and only 3-4 minutes on Linux/macOS:
https://ci.appveyor.com/project/Safihre/sabnzbd/builds/36140643

jcfp:
Looks like I did find the casue of the slow test runs on windoze. Unfortunately replacing localhost with the ip address broke something else. Looks like the wizard comes up instead of the standard interface with queue and history in test_functional_downloads.

Safihre:
I fixed the functional problem.
I saw 2 more things:
Is this a bug in SABnzbd we should fix?
```
Exception in thread Thread-51:
  File "C:\Python38-x64\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\Python38-x64\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
TypeError: run_script() missing 1 required positional argument: 'script'
```
And:
`scripts/my_script_for_sab.py` is not removed after testing.
Maybe that file can be also placed in the cache-directory so it's cleaned up with the rest?

jcfp:
> TypeError: run_script() missing 1 required positional argument: 'script'

This is almost certainly triggered by one of the api tests trying to set a script in a way that result in the scriptname ending up empty, but sab marching on anyway. The bug tracking the lack of validation for this input is #1650.

> `scripts/my_script_for_sab.py` is not removed after testing.

We could simply copy the scripts dir into cache, and add this one in there as well.

PR #1671, opened 2020-11-05
Separate email message lines are with CRLF

exizak42:
SMTP protocol dictates that all lines are supposed to be separated
with CRLF and not LF (even on LF-based systems). This change ensures
that even if the original byte string message is using `\n` for line
separators, the SMTP protocol will still work properly.

This resolves sabnzbd#1669

### DIFF
--- a/sabnzbd/emailer.py
+++ b/sabnzbd/emailer.py
@@ -27,6 +27,7 @@ import glob
 
 from Cheetah.Template import Template
 from email.message import EmailMessage
+from email import policy
 
 from sabnzbd.constants import *
 import sabnzbd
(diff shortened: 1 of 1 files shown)


Safihre:
Lol, the Python documention sounds a bit stupid.. https://docs.python.org/3/library/email.policy.html#email.policy.Policy.linesep
"We know it should be this but we won't do it by default" 😄

Safihre:
Could you fix the `black` error? The problem is that you should be using `"` instead of `'`

exizak42:
> Could you fix the `black` error? The problem is that you should be using `"` instead of `'`

Fixed - I think I just copy-pasted the email policy from the documentation.

Issue #1675, opened 2020-11-14
"Version:	3.2.0-develop [fatal: ]"


sanderjo:
Weird version tag: "Version:	3.2.0-develop [**fatal:** ]"

Cause: the git command is not run from the git directory, as SAB is run with `git/sabnzbd/SABnzbd.py`, so with a path.

Proof, from home dir:
```
sander@witte2004:~$ git rev-parse --short HEAD
fatal: not a git repository (or any of the parent directories): .git
```

This does work: specify git-dir, with .git:
Relative path:
```
sander@witte2004:~$ git --git-dir=git/sabnzbd/.git  rev-parse --short HEAD
575fbc06
```
Absolute path
```
sander@witte2004:~$ git --git-dir=/home/sander/git/sabnzbd/.git  rev-parse --short HEAD
575fbc06
```
So I have a fix in mind

```
sander@witte2004:~$ git/sabnzbd/SABnzbd.py -b1
2020-11-14 20:17:39,847::INFO::[SABnzbd:1142] --------------------------------
2020-11-14 20:17:39,847::INFO::[SABnzbd:1143] SABnzbd.py-3.2.0-develop
2020-11-14 20:17:39,876::INFO::[misc:992] [sabnzbd.misc.run_command] Running external command: ['git', 'rev-parse', '--short', 'HEAD']
2020-11-14 20:17:39,876::DEBUG::[misc:993] Popen arguments: {'stdin': -1, 'stdout': -1, 'stderr': -2, 'startupinfo': None, 'creationflags': 0}
2020-11-14 20:17:39,882::INFO::[SABnzbd:1151] Commit: fatal: not a git repository (or any of the parent directories): .git
2020-11-14 20:17:39,882::INFO::[SABnzbd:1152] Full executable path = /home/sander/git/sabnzbd/SABnzbd.py
```

![image](https://user-images.githubusercontent.com/1273502/99155248-827c0000-26b6-11eb-97dc-a2bbe19f5ef2.png)

sanderjo:
Solved:

```
sander@witte2004:~$ git/sjo-sab/SABnzbd.py -b0
2020-11-14 20:44:55,192::INFO::[SABnzbd:1142] --------------------------------
2020-11-14 20:44:55,192::INFO::[SABnzbd:1143] SABnzbd.py-3.2.0-develop
[...]

PR #1676, opened 2020-11-14
correct git id when not run from SAB directory itself

sanderjo:
Solves #1675


Safihre:
Used a different way :)

PR #1678, opened 2020-11-15
diskspeed: follow pylint's advice, and more pytesting

sanderjo:


sabnzbd/utils/diskspeed.py
```diff
@@ -8,39 +10,36 @@
...
-            os.close(fp)
-            os.remove(filename)
-        except:
-            pass
-    except:
-        # No succesful measurement, so ... report None
-        return None
+        os.remove(filename)
+    except (PermissionError, NotADirectoryError, FileNotFoundError):
+        # Could not write, so ... report None
```
Safihre, line 41: Not `None` anymore 😉

tests/test_utils/test_diskspeed.py
```diff
@@ -19,15 +19,38 @@
...
+        Only useful on Linux, and only if not-writable
+        """
+        non_writable_dir = "/usr"
+        if os.path.isdir(non_writable_dir) and not os.access(non_writable_dir, os.W_OK):
+            speed = diskspeedmeasure(non_writable_dir)
+            assert speed == 0
+
+    def test_file_not_dir_specified(self):
+        """ testing a file should result in 0"""
+        speed = diskspeedmeasure("tests/data/somefile.txt")
```
Safihre, line 55: Instead of including this file, let's use a temporary file:
```python
with tempfile.NamedTemporaryFile() as temp_file:
    speed = diskspeedmeasure(temp_file.name)
```

Safihre (changes requested):

Safihre:
Could you remove the testfile? We don't need it anymore now we use the temp file.
Then we can merge.

sanderjo:
> Could you remove the testfile? We don't need it anymore now we use the temp file.
> Then we can merge.

Which testfile?

PR #1684, opened 2020-11-26
explicitly set mode for gzip.GzipFile()

jcfp:
Currently defaults to the mode of the open() just above, but:
```
sabnzbd/__init__.py:634: FutureWarning: GzipFile was opened for writing, but this will change in future Python releases.  Specify the mode argument for opening it for writing.
    f = gzip.GzipFile(filename, fileobj=tgz_file)
```

[Docs](https://docs.python.org/3/library/gzip.html):
> The mode argument can be any of 'r', 'rb', 'a', 'ab', 'w', 'wb', 'x', or 'xb', depending on whether the file will be read or written. The default is the mode of fileobj if discernible; otherwise, the default is 'rb'. In future Python releases the mode of fileobj will not be used. It is better to always specify mode for writing.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -611,7 +611,7 @@ def save_compressed(folder: str, filename: str, data: AnyStr):
     try:
         # Have to get around the path being put inside the tgz
         with open(os.path.join(folder, filename), "wb") as tgz_file:
-            f = gzip.GzipFile(filename, fileobj=tgz_file)
+            f = gzip.GzipFile(filename, fileobj=tgz_file, mode="wb")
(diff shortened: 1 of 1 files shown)


Safihre:
When uploading a `.nzb.gz` file I think it also throws this error. Let's include it there as well?

jcfp:
Doesn't show up when uploading an nzb.gz from the web interface with python3.9. The only other reference to GzipFile is at sabnzbd/nzbparser.py:289; the subsequent call of nzb_reader_handler already sets a mode?

Safihre:
Aah guess I fixed it before!

Issue #1687, opened 2020-11-30
Crash in rar-renamer when there is no first rar-file


Safihre:
Triggered by:

- https://www.reddit.com/r/SABnzbd/comments/k2syxt/please_help_download_failed/
- https://forums.sabnzbd.org/viewtopic.php?f=2&t=25018

If a rar-set is fully obfuscated, but is missing the first rar-file (`part001`), post-processing wil crash.
Reproducers:
- Delete the files `tests\data\obfuscated_two_rar_sets\3b7d77bc-b95d-45b9-bb25-22b156d4ffef` and `tests\data\obfuscated_two_rar_setsaaffd55d-5282-4497-a52e-894181dd7531` and run the `test_rar_renamer` unit-test.
- [reftestnzb_100MB_obfuscated_rars_rar-only_missing_first_rar_bf1664007a71.zip](https://github.com/sabnzbd/sabnzbd/files/5615547/reftestnzb_100MB_obfuscated_rars_rar-only_missing_first_rar_bf1664007a71.zip)

sanderjo:
I'll write a nice rar-set pre-checker, that avoids tracebacks, and gives a nice warning if the rar-set is incorrect.

jcfp:
> If rar # 3 is missing completely, can SAB repair stuff? I think not.

Why not? Par2 can handle missing files.

sanderjo:
> Why not? Par2 can handle missing files.

It depends on the redundancy in the par2 files. Anyway: I've written the code that in case of one rar set, I do renaming, and so SAB can try to par2 and then unrar.

Safihre:
@jcfp the Rar renamer code is only executed when the full NZB is obfuscated and we can't find any par2 files (which we search for by first bytes signature).

PR #1688, opened 2020-11-30
obfuscated rar sets: better handling missing rars

sanderjo:
As described in https://github.com/sabnzbd/sabnzbd/issues/1687

And added two unit test for these cases.


sanderjo:
Closes #1687 , Black and CI are happy, so ... @Safihre ... are you willing to accept?

sabnzbd/postproc.py
```diff
@@ -948,47 +952,77 @@ def rar_renamer(nzo: NzbObject, workdir):
...
+
+    # Sanity check of the rar set
+    highest_rar = sorted(rarvolnr.keys())[-1]
+    """
+    # First an easy check
+    if highest_rar != len(rarvolnr):
+        # there is a missing rar file, so
+        logging.warning("Incomplete rar set, so I can't deobfuscate")
+        return 0
+    """
```
Safihre, line 967: Please remove unused code.

sabnzbd/postproc.py
```diff
@@ -948,47 +952,77 @@ def rar_renamer(nzo: NzbObject, workdir):
...
+    # More intelligent check aka staircase check: number of rarsets should no go up, but stay the same or go down
+    how_many_previous = 1000  # 1000 rarset mixed ... should be enough ...
+    for rar_set_number in range(1, highest_rar + 1):
+        try:
+            how_many_here = len(rarvolnr[rar_set_number])
+        except:
+            # rarset does not exist at all
+            logging.warning("rarset %s is missing completely, so I can't deobfuscate.", rar_set_number)
```
Safihre, line 975: So this works well if one rar-set has 10 files and the other 15? Almost looks like it requires all sets to have same amount of rar's?

sanderjo: Yes, that works. Please note that this was existing code (with just indentation changed)

sabnzbd/postproc.py
sanderjo: OK, done

Issue #1682, opened 2020-11-18
Can't deobfuscate certain names


Diniboy1123:
Good day,

I often pull linux isos from a specific indexer that's known to obfuscate their releases in an interesting format which results in having abc.xyz.EXT filenames in the resulting folders.

I looked around inside the nzb file itself and I came across the fact that these abc.xyz names are nowhere mentioned actually. Also, these files are not packed to rars or anything, therefore its a bit unclear for me where this naming comes from. However the subject field contains the right filename:

`<file poster="<redacted>&gt;" date="<redacted>" subject="[PRiVATE]-[GROUPNAME]-[INSTALLER.ISO]-[3/82] - &quot;&quot; yEnc  100000 (1/1)">`

I obviously want the file to have the INSTALLER.ISO name in this case. I have tried to enable the "Deobfuscate final filenames
" feature, but that seemingly didn't do anything with the files. I also tried the movie sorting with the %dn, but that just placed each abc.xyz.EXT file to the root download folder.

Is there a way I can extract titles from the subject with a regex a script or whatsoever? Also, where does this abc.xyz gets fetched from?

For investigating purposes I can provide nzbs but due to the nature of their contents I can only do so in email.

thezoggy:
you can look at sab logs (debug) for more info.
but most of the time its just the rarset filename or the parset data to rename. ideally you should have the deobfuscated name from the nzbjob name unless the indexer is indexing the obfuscated name..

Diniboy1123:
The nzb doesn't contain abc.xyz namings anywhere, there is no par in this specific nzb and the files are not in rar containers or anything, just raw uploaded. The name still gets picked up from somewhere when I start the download however. If I keep it in paused state it still has the names from the subject which is my goal. But as soon as it starts to download, I can see all of them changing to this abc.xyz thing.

thezoggy:
check logs, it will show where it is extracted with the name or renamed to it.

Diniboy1123:
It mentions: `[nzbstuff:1697] Detected filename from yenc: ` Is there a way to disable that?

thezoggy:
its doing that because of no pars. but this is just gathering data that can be used later on.

Diniboy1123:
Nope, it is actually the one line that renames the files from the subject field to the abc.xyz ones. I wanna get rid of that feature completely. Since in this indexer's case it just makes everything obfuscated.

There are no par files. that's right.

[...]

PR #1693, opened 2020-12-06
Smarter extraction of filenames from NZB-subject

Safihre:
Relates to #1682.
The change [here](https://github.com/sabnzbd/sabnzbd/pull/1688/files#diff-1b31e79dc5cd1653ba3e1d1ec7c3aa52ea25430f489c4bedb56441aaa1135db9L1694) resulted in unwanted side effects: 
For obfuscated yEnc-names we kept the NZB-subject, but since the NZB-subject was nonsense the resulting files were even more garbage than before. Because of those bad filenames, "Obfuscate final filenames" wouldn't give them pretty names anymore.

Solution: try and see if we can extract the filename from the subject a bit smarter than just whatever is between nice quotes and has a nice file extensions. 
If I missed some cases, please let me know. I did change the default behavior of SABnzbd that has been around for many years, so a couple of extra eyes checking this change is needed!

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -85,11 +85,11 @@ import sabnzbd.cfg as cfg
 import sabnzbd.nzbparser
 from sabnzbd.downloader import Server
 from sabnzbd.database import HistoryDB
-from sabnzbd.deobfuscate_filenames import *
+from sabnzbd.deobfuscate_filenames import is_probably_obfuscated
 
 # Name patterns
-SUBJECT_FN_MATCHER = re.compile(r'"([^"]*)"')
-RE_NORMAL_NAME = re.compile(r"\.\w{1,5}$")  # Test reasonably sized extension at the end
+RE_SUBJECT_FILENAME_QUOTES = re.compile(r'"([^"]*)"')  # In the subject, we expect the filename within double quotes
+RE_SUBJECT_BASIC_FILENAME = re.compile(r"([\w\-+()'\s.,]*\.\w{2,4})")  # Otherwise something that looks like a filename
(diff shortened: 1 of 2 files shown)


Safihre:
Next step would be to change this code to check both names (subject-based-filename and yenc-based-filename) and see which one is not-obfuscated using @sanderjo's detection. Preferring the yEnc name if both are equal.
```python
        if (
            yenc_filename
            and yenc_filename != nzf.filename
            and not is_probably_obfuscated(yenc_filename)
            and not nzf.filename.endswith(".par2")
        ):
```

jcfp:
I'm unfamiliar with the variety of subject lines found in the wild, but, that said, I don't really see major issues with this change.

For "nicely wrapped" filenames, the regex drops the requirement for an extension in the filename but -looking at the tests- that change seems intentional. So as long as the results from the "basic search" are no worse than returning the entire subject, the risk of regressions seems limited.

- The tests don't cover any subjects where no filename should be detected (and the entire input gets returned).

sanderjo:
Tested it with real downloads: too hard cut off on ... embedded HTML codes like `&#039;` in the subject filename.

Workaround: `html.unescape(subject)`

... and allow `'` (and other special characters?) in the regexp `RE_SUBJECT_BASIC_FILENAME = re.compile(r"([\w\-+\(\)\s.]*\.\w{2,4})")`

jcfp:
If it's common for subject lines to be encoded, html might not be the style used. For email (a rather close relative of usenet), there's rfc1342 to deal with representing non-ascii chars in headers (incl. subjects) as either base64 or quoted-printable.

sanderjo (commented):
> If it's common for subject lines to be encoded, html might not be the style used. For email (a rather close relative of usenet), there's rfc1342 to deal with representing non-ascii chars in headers (incl. subjects) as either base64 or quoted-printable.

I don't understand what you mean. 
" html might not be the style used." ... do you mean the `&#039` is not HTML ... ?

PR #1695, opened 2020-12-09
prevent repetition of unwanted extension warnings

### LINKED ISSUE #1686: 'unwanted extension' warning repeats after overriding
With _Action when unwanted extension detected_ set to _Pause_, after hitting an unwanted extension a warning is printed and the job gets paused. So far so good.

However, when the users then unpauses the job, new extension warnings keep popping up whenever the directunpacker hits a new file segment (and finds the same extension yet again). This shouldn't happen; instead, the warning should be silenced after the user decides to override the configured pause action.

Note that the issue is only the repeated warning messages; the pause action itself doesn't get repeated.

jcfp:
Fixes #1686

A minor disadvantage of this change may be that there'll be no further warnings in case a single jobs packs multiple unwanted extensions and the user has overridden the pause triggered by whatever extension caused the first hit. That was already the case for the pause itself though.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -135,11 +135,13 @@ class Assembler(Thread):
                                 sabnzbd.NzbQueue.end_job(nzo)
 
                         if unwanted_file:
-                            logging.warning(
-                                T('In "%s" unwanted extension in RAR file. Unwanted file is %s '),
-                                nzo.final_name,
-                                unwanted_file,
-                            )
+                            # Don't repeat the warning after a user override of an unwanted extension pause
+                            if nzo.unwanted_ext == 0:
+                                logging.warning(
+                                    T('In "%s" unwanted extension in RAR file. Unwanted file is %s '),
+                                    nzo.final_name,
+                                    unwanted_file,
+                                )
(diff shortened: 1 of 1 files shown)


sabnzbd/assembler.py
```diff
@@ -135,11 +135,13 @@ def run(self):
                                 sabnzbd.NzbQueue.end_job(nzo)
 
                         if unwanted_file:
-                            logging.warning(
-                                T('In "%s" unwanted extension in RAR file. Unwanted file is %s '),
-                                nzo.final_name,
-                                unwanted_file,
-                            )
+                            # Don't repeat the warning after a user override of an unwanted extension pause
+                            if nzo.unwanted_ext == 0:
```
Safihre, line 139: Shouldn't it be `2`?
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/nzbstuff.py#L1383-L1384

jcfp: The only way I see for nzo.unwanted_ext to have a non-zero value at this point is after a resume, so whether one checks for `== 0` or `!= 2` shouldn't make a difference. I used the first for consistency with the check a few lines down.

PR #1696, opened 2020-12-09
Only check idle servers for new articles twice per second

puzzledsab:
Currently all servers will be checked on every run loop, even if only one is in use. This patch makes idle servers search for new articles twice per second, saving CPU time.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -443,19 +443,20 @@ class Downloader(Thread):
         last_busy = {}
 
         while 1:
+            now = time.time()
             for server in self.servers:
                 serverid = server.id
                 if server.busy_threads:
-                    last_busy[serverid] = time.time()
+                    last_busy[serverid] = now
 
                 # Skip this server if idle for 1 second and it has already been searched less than 0.5 seconds ago
-                if last_busy.get(serverid, 0) + 1 < time.time() and last_searched.get(serverid, 0) + 0.5 > time.time():
+                if last_busy.get(serverid, 0) + 1 < now and last_searched.get(serverid, 0) + 0.5 > now:
                     continue
 
-                last_searched[serverid] = time.time()
+                last_searched[serverid] = now
 
                 for nw in server.busy_threads[:]:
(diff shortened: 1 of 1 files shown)


puzzledsab:
I'm unable to find out what the formatting problem is, can anyone tell me? I thought it might be the line length, but there is another line in downloader.py that is the same length.

jcfp:
> I'm unable to find out what the formatting problem is, can anyone tell me?

It's whatever [black](https://github.com/sabnzbd/sabnzbd/blob/develop/.github/workflows/black.yml) thinks is wrong. You can run the same checks locally to see what changes: `python3 -m black --line-length 120 --target-version=py36 FILE`.

sabnzbd/downloader.py
```diff
@@ -436,8 +436,24 @@ def run(self):
...
+        last_searched = {}
+
+        # Store when each server last downloaded anything or found an article
+        last_busy = {}
+
         while 1:
             for server in self.servers:
+                serverid = server.id
+                if server.busy_threads:
+                    last_busy[serverid] = time.time()
```
Safihre, line 449: Sounds like a good approach.
Just looking through this new code it seems we call `time.time()` a lot, maybe we should only call it once per main-`while` loop and store it in a variable?
Not just for your new code, but also the rest of the code in the main loop.

puzzledsab: I decided to test it. It turns out `time.time()` is incredibly fast, so I don't think using a variable is necessary. On my Rpi 3, doing
`x = time.time()`
1M times took 2.19s, while
`x += 1`
took 1.33s. The difference is even smaller with Windows/Intel.

puzzledsab: Actually, maybe it would be a bit more useful than I first thought. When I subtract the time the loop itself takes and remove the assignment then calling time.time() takes about 5 times longer than calling a variable containing a float.

Safihre: 👍

Safihre:
It looks good to me, let me know when you are done and want it merged 💯

puzzledsab:
Great, I think it's ok now.

PR #1697, opened 2020-12-09
Make sleep more fine grained and add short sleep when no processing is done

puzzledsab:
Add a short sleep when no servers has added any new articles for the last two run loops. This can reduce the CPU load significantly, particularly when the user has not configured a maximum download speed. 0.001 second is the default because it is unlikely to affect the download speed and it is the shortest sleep that Windows can handle.

Having a longer sleep period will rarely reduce the CPU load noticeably compared to multiple shorter periods. A longer sleep increases the risk of stalling the download speed because of full buffers.

This makes the old slow down method redundant, so I have removed it.

The sleep_time special variable can be set to 0 to disable sleeping.


Safihre:
Interesting approach, would like to test this on my low level devices and see what happens.
I would like to suggest actually to add an attribute `self.__sleep_time` to the `Downloader` and calculating the time only once. Because getting the value of a Config setting is "expensive" and since it's a Special setting its fine if it requires a restart to take effect.

puzzledsab:
Yes, calculating sleep time more seldom would be an advantage. I wasn't sure how to do it so that it gets recalculated if it is changed without requiring a restart.

sabnzbd/downloader.py
```diff
@@ -428,7 +428,7 @@ def decode(self, article, raw_data):
                     sabnzbd.Assembler.queue.qsize(),
                 )
                 logged = True
-            time.sleep(self.sleep_time)
+            time.sleep(self.sleep_time + 0.001)
```
Safihre, line 431: I would like to suggest to keep the hardcoded values here, unless there's a real good reason to make it flexible?

puzzledsab: I don't really know, but my hypothesis was that on a fast connection and a fast CPU a short sleep is desirable, while on a slower CPU which uses more time on decoding and thread switching it is better with a higher value. 0.05 seems too long to me as it limits the speed to 320 KB/s (16KB buffer / 0.05) per connection. This may make the transfer rate choppy and slower than necessary. I don't mind hard coding it, but I would suggest have a lower value than 0.05. How about 0.005? Do you know how much time it takes in all to decode one article or clear for instance 10% of the decoder buffer on various systems? That might be a good starting point.

thezoggy:
The logging probably is more impactful (disk io/file access) than the sleep

sabnzbd/downloader.py
Safihre: Testing this on a raspberry pi and my super slow old Synology NAS I actually found it's rarely in this loop for too long. Maybe a few times per second. The slow CPU in general will just prevent things from going too fast. 
So I'm quite confident that the hard-coded value is a decent one :)

puzzledsab: I don't think I'm explaining this well, or maybe I'm just wrong. Anyway, I think the problem will be with fast CPUs and fast connections, not the slow ones. The NewsWrapper recv_chunk sets the buffer to 16KB for SSL. When it is full, I assume the server will quickly have to stop sending data, and the speed will slow down.

On a fast connection this takes a very short time. At 1Gbps, even with 50 connections you would fill all the 16KB buffers in about 0.006 seconds (total buffer = 50 connections * 16384 bytes * 8  = 6.25 Mbit. 6.25/1000 = 0.00625). If the CPU is just a little bit too slow for this speed then sleeping 0.05 seconds means it's waiting much longer than it needs to to keep up.

Safihre: Luckily the decoding is super fast and has its own internal queue, so on those systems it never hits this code. I can confirm with my i7 on a gigabit connection.

Safihre: Oh wait I see you referenced the buffer in newswrapper for receiving the chucks. Actually that's not effected here, since this code only checks for the decoder and assembler queues. The chunk buffer is only dependent on the main while loop and the select in there.

puzzledsab: The decode function which does the sleep is called from the downloader main while loop, so when you sleep in decode you pause the main loop for the same amount of time. All the packages that are received after the buffers have filled will be dropped, and the sender will assume it's sending too fast.

It's no problem if the CPU is fast enough, but it is when it's not. If it can decode 900 Mbps then it will start sleeping in decode. However, when you wait as much as 0.05s before clearing of the receive buffers then that's the equivalent of limiting the bandwidth to about 125 Mbps with 50 connections.

On my CPU it seems that one thread can decode one 800 KB article in about 0.008 seconds (total processing time in the DecoderWorker loop, not just the decode). That's about the maximum time it will take before there is a free spot in the queue. If it was almost done with the previous article it will take less time. That means it could frequently get back to checking the receive buffer in a much shorter time than 0.05s.

Safihre: Yes, that is if you assume every decode input will trigger a 0.05 delay, but that's not the case. When we wait 0.05 it will have decoded maybe 10 articles, so the sleep won't be called for the next 10 articles.
This is just a theoretical discussion, all I wanted to say is that the sleep here and in the speed limiter have a different impact than the one in the main loop because their purpose is different. So that's why I don't want to use the same constant.

puzzledsab: How about a compromise?
time.sleep(self.sleep_time * 10 + 0.001)

That defaults so 0.011s, which won't trigger the loops a lot more than now. It's much higher than my current one but gives some leeway in adjusting it both up and down for end users. That way those who experience slow downs can check if this is the problem.

From my own testing a with a 4MB/s transfer rate using 1 connection, a 0.05s sleep can start slowing it down if it triggers more often than 1 per second.

thezoggy:
https://stackoverflow.com/questions/1133857/how-accurate-is-pythons-time-sleep
trying to do .001 with variable and multiplying other values.. your prob just better off leaving things simple since the precision wont be there.

puzzledsab:
On Windows the shortest possible sleep period is between 1 and 2ms. My Raspberry Pi 3 is within 0.1ms. I think it is accurate enough for this purpose. The reason why I think it should be configurable is that there are a lot of platforms out there. It will help the debugging process if it's possible for end users to change it.

There was a guy on the forum that was able to increase the download speed from 30 MB/s to 80 MB/s by reducing the slow down timer from 0.01s to 0.001s: [https://forums.sabnzbd.org/viewtopic.php?f=2&t=24612](url)

I have seen others have issues where I suspect that might be the case, but currently most users are not able to test it.

thezoggy:
well there was a known issue with binhex's docker. so not sure that issue applies to everyone. but anyways i dont disagree with some sleep but making it more complicated than it needs to be with additional logging in my eyes is a step backwards. Rather have less logging / simpler logic - or smarter logic that doesnt just do some fixed value but more adaptive to the amount of articles backloged or something.

puzzledsab:
logging.debug is called once before the loop starts. Are you thinking of some other kind of logging?

The main logic is the 4 new lines in the while loop, which makes it sleep if no servers have added new articles for two rounds. It is much simpler than the can_be_slowed/get_stable_speed system it replaces. Hopefully it works better and gives additional benefits without causing harm.

The size of the backlog of articles is not important for the sleeping periods, only the download speed. The goal is to sleep as much as possible without getting slow downs caused by dropped packages. I think the safest way to do that is to have a very short interruption between every poll of the buffers, instead of mostly no interruptions and then a long one.

These days a single nzb can be a lot bigger than 10GB. Even if a RAM disk would have helped I don't think it's a better alternative.

puzzledsab:
As far as I can tell it wasn't my last commit that caused the test to fail. Anyone know what happened?

Safihre:
That download test is sometimes randomly unstable, not a problem.

Safihre:
I pushed some changes, what do you think about them?
I did revert the change to the sleep for the speedlimit because I feel if it ain't broken, don't fix it 😄

puzzledsab:
Thanks for testing it!

I noticed you have effectively increased the default to 0.002s and the minimum to 0.001s. I tested downloading with 1 connection on Windows. With no delay at all I get ~4.3MB/s, 0.001s = ~4.2MB/s and with 0.002s I get ~3.4MB/s. Even with 0.001s it will usually use less or the same amount CPU as the get_stable_speed system, so I think it would be safer to keep the default at 0.001s and have the option to reduce it to 0.

I'm a bit worried that with all the potential additional sleeping from sleep_time, the 0.05s sleep could cause more problems than it currently does. How about not adding 0.001 to the base sleep_time, and setting the bandwidth limiter to self.sleep_time * 49 + 0.001? That way it will be 0.05s by default. If anyone has problems it will be easy to try a lower value. The calculation is quite simple so it won't take a noticeable amount of time compared to the 0.05s it is sleeping.

Safihre:
I know the SABnzbd code isn't the prettiest, but I would still like to avoid code that doesn't make clear what it is trying to do.
So preferably not `self.sleep_time * 49 + 0.001`.
But you make a good point about this maximum speed, so I just removed the `+0.001` in `sleep_time_set`. Since the speed-limiter doesn't use `sleep_time` anymore it's not a problem. It's the user's own problem if they mess with this setting and break the delaying in the `decode` function (we should mention it in the documentation though). It's not like SABnzbd will break down without the delaying, it just gets less optimal.

This looks all good to me now, ready to merge. I like this improvement a lot over the `get_stable_speed` that I made ages ago, since that was also not very clean code 😄

puzzledsab:
Sorry about all the nit picking. I think there should always be some sleep in decode and the bps limiter in order to give run time to the other Python threads. If you don't want a calculation then I would prefer a hard coded value over just sleep_time. I suggest 0.01, but 0.05 will probably work too. Other than that it's ok to merge.

Safihre:
@puzzledsab Check out this: https://forums.sabnzbd.org/viewtopic.php?f=11&t=25138

Issue #1150, opened 2018-05-31
[Feature request] Show per-server speed


Safihre:
See here:
https://www.reddit.com/r/usenet/comments/8n77eb/see_download_speed_of_individual_servers_sabnzbd/

I can image this being usefull, if we can find a fast way to do so. Maybe less accurate than the regular BPSMeter.

Safihre:
I was actually thinking more of a counter in the Status and Interface settings window on the Severs page.

atg22896:
Any progress on this?

Safihre:
Closed by #1704

PR #1704, opened 2020-12-14
Show per-server speed (issue #1150)

puzzledsab:
I'm on pretty shaky ground here, so this is on a "seems to work" level. The speed is shown in the Glitter Interface popup window and the server config page.


sanderjo:
While downloading, I see the speed in the Server page. But what do you mean with "Glitter Interface popup window"?

image

sanderjo:
Oh wait, do you mean Wrench -> Status and interface options -> Connections? I see it there:

image

Safihre:
Nice change 🏅 
I don't want to put in the Config, there we place static information while this is dynamic information. The pop-up is the right place for it 👍

sabnzbd/bpsmeter.py
```diff
@@ -243,6 +244,25 @@ def update(self, server=None, amount=0):
             self.bps = (self.bps * (self.last_update - self.start_time) + amount) / (t - self.start_time)
         except:
             self.bps = 0.0
+            self.server_bps = {}
```
Safihre, line 247: Is the reset needed?

sabnzbd/bpsmeter.py
```diff
@@ -243,6 +244,25 @@ def update(self, server=None, amount=0):
...
+
+        for s in self.server_bps:
+            if s == server:
+                continue
+            try:
+                self.server_bps[s] = (self.server_bps.get(s, 0.0) * (self.last_update - self.start_time) + 0) / (
+                    t - self.start_time
+                )
+            except:
+                self.server_bps[s] = 0.0
```
Safihre, line 265: It feels like we are doing the same thing here twice, I suggest could combine:
```python
#        if server:
#            try:
#                self.server_bps[server] = (
#                    self.server_bps.get(server, 0.0) * (self.last_update - self.start_time) + amount
#                ) / (t - self.start_time)
#            except:
#                self.server_bps[server] = 0.0

        for server_to_update in self.server_bps:
            try:
                self.server_bps[server_to_update] = (self.server_bps.get(server_to_update, 0.0) * (self.last_update - self.start_time) + amount * (server_to_update == server) / (
                    t - self.start_time
                )
            except:
                self.server_bps[server_to_update] = 0.0
```
The `amount * (server_to_update == server)` is a bit not-so-nice-code-style, but it saves us making an extra if.

sabnzbd/api.py
```diff
@@ -1273,6 +1273,7 @@ def build_status(skip_dashboard=False, output=None):
                 "servererror": server.errormsg,
                 "serverpriority": server.priority,
                 "serveroptional": server.optional,
+                "serverbps": to_units(sabnzbd.BPSMeter.server_bps.get(server.id, 0), "B/s"),
```
Safihre, line 1276: Let's drop the `, "B/s"` and do that in the front-end.

interfaces/Glitter/templates/include_overlays.tmpl
```diff
@@ -210,6 +210,10 @@
                                         <span class="glyphicon glyphicon-ok"></span> <span data-bind="text: serversslinfo"></span>
                                     </div>
                                 </div>
+                                <div class="row">
+                                    <div class="col-sm-6">$T('speed')</div>
+                                    <div class="col-sm-6" data-bind="text: serverbps"></div>
```
Safihre, line 215:
```html
<div class="col-sm-6"><span data-bind="text: serverbps"></span>B/s</div>
```

interfaces/Config/templates/config_server.tmpl
```diff
@@ -243,7 +243,8 @@
                             $T('today'): $(server['amounts'][3])B<br/>
                             $T('thisWeek'): $(server['amounts'][2])B<br/>
                             $T('thisMonth'): $(server['amounts'][1])B<br/>
-                            $T('custom'): <span id="server-data-value-${cur}"></span>
+                            $T('custom'): <span id="server-data-value-${cur}"></span><br/>
+                            $T('speed'): $(server['amounts'][5])B/s
```
Safihre, line 247: As mentioned, let's not put it here.

sabnzbd/interface.py
```diff
@@ -1555,9 +1555,9 @@ def index(self, **kwargs):
         )
         for svr in server_names:
             new.append(servers[svr].get_dict(safe=True))
-            t, m, w, d, timeline = sabnzbd.BPSMeter.amounts(svr)
+            t, m, w, d, timeline, bps = sabnzbd.BPSMeter.amounts(svr)
```
Safihre, line 1558: We can ignore the output here `t, m, w, d, timeline, _ = sabnzbd.BPSMeter.amounts(svr)`

Safihre (changes requested):

sanderjo:
I would like to also see it in the speedometer on the front page. One color per server.

image

interfaces/Glitter/templates/include_overlays.tmpl
```diff
@@ -210,6 +210,10 @@
...
+                                </div>
```
Safihre, line 216: Let's add the speed below the `# Connections`.

sabnzbd/bpsmeter.py
puzzledsab: My assumption was that if the total bps should be reset to 0, then so should the bps for each server. Otherwise the sum of the servers and the total won't be in sync. It was easier to set it to {} than setting each server to 0.

puzzledsab:
> I don't want to put in the Config, there we place static information while this is dynamic information.

My thought was that then it wouldn't be visible anywhere in other skins than Glitter, but maybe nobody uses them. I just tried Plush and it seems like it makes downloads stop downloading, no matter if they are paused or not.

sabnzbd/bpsmeter.py
puzzledsab: This change alone doesn't work because then the server won't be added to self.server_bps, so it will always be empty.

The + 0 is just silly, I should have removed it after debugging.

interfaces/Glitter/templates/include_overlays.tmpl
puzzledsab: I've already changed it. However, now it will come between the # connections and the list of connections if displaying those is enabled.

puzzledsab:
> I would like to also see it in the speedometer on the front page. One color per server.

Sorry, my CSS and Javascript skills are close to non-existent so I probably won't be attempting this for a while.

Safihre:
Yes I think that's out of scope for now.

Safihre:
Pushed some small changes.
However, `BSPMeter.update()` is called _a lot_, so I wonder of the performance impact and if it's worth the extra information this provides.. Bit hard to tell.

puzzledsab:
I tested it with Yappi. The update function takes 11.2% longer with the server BPS code. Without the BPS code and using SSL transfer, the BPS update function uses 6.0% of the total CPU time used by the downloader run function so it adds about 0.66% CPU load to it.

Safihre:
It also depends on the number of servers and the number of servers in use (so the size of the `server_bps` dictionary). Did you take that into account?
Not to be negative, I'm just curious :)

puzzledsab:
I don't mind at all. I have tested this quite extensively now and my first numbers were probably not accurate. I ended up making a system for testing several configurations many times automatically. It starts Sab, downloads the 1000MB sample (no bandwidth limit) and shuts down to get the yappi numbers. These are the results for calls to update():

New version with sleep_time, 1 or 3 servers, with or without the server calculation, using wall clock:
1-noserver Samples: 11 calls: 126667 time: 1.28s avg time:10.10μs
1-server Samples: 9 calls: 124754 time: 1.57s avg time:12.60μs
3-noserver Samples: 10 calls: 201522 time: 1.25s avg time:6.22μs
3-server Samples: 9 calls: 202271 time: 1.84s avg time:9.12μs

Old version without sleep_time using CPU clock.
1-noserver Samples: 8 calls: 538252 time: 4.21s avg time:7.86μs
1-server Samples: 15 calls: 537851 time: 4.97s avg time:9.30μs
3-noserver Samples: 8 calls: 188078 time: 1.80s avg time:9.58μs
3-server Samples: 12 calls: 189422 time: 2.08s avg time:10.97μs

Samples = number of download runs, calls = average number of calls to update(), time = the average total time spent in update() for one run and avg time = the average time one call to update() took.

I had to switch to wall clock because the sleep seems to mess up the timing for yappi. The average became less than 1μs. I assume the difference in average time between 1-noserver and 3-noserver is an indication that the timings are not very accurate. The numbers for the same amount of servers using the same clock timings are probably comparable, though.

TLDR: It always takes a measurable amount of time. If you think it's too much then perhaps an alternative could be a config switch to enable it, or a switch in the wrench window that shows connections? Not sure how to do the last alternative but I'm trying to read up on Knockout JS.

Safihre:
Alright, it seems in general not too matter too much.

Safihre:
If you feel up for a big challenge: can we write a proof of concept that uses asyncio to handle the download (and subsequent decoding)?
Each server thread could be its own asyncio thing that only comes to life when it can get an article. Or maybe there's a whole better way to approach it. If we just ignore the whole monolith of Sabnzbd, but only a proof of concept with multiple servers, how would it work? Maybe also `sabyenc` needs to be modified in a later stage to be asyncio, I can look into that too.

puzzledsab:
I've actually tried writing a very basic asyncio based usenet article downloader a while ago. Just fetching multiple articles concurrently, nothing else. It worked and had a low CPU load, but the problem is that so much in Sab is written around looping, async polling and using the results of those polls on each iteration. Switching to asyncio would require some big rewrites. You should probably freeze the development for everything except vital fixes in the mean time because lots of it won't be reusable.

I can clean it up a little and post it somewhere if you're interested. Currently it doesn't even support login because I used the free IPv6 server.

PR #1708, opened 2020-12-18
Build binary release for each commit and pull request

Safihre:
Because TravisCI limited the number of build minutes, I moved our CI-pipeline to GitHub actions. Turned out to be super-super easy.
This inspired me to also see if it would be easy to move our release pipeline here.

With this change, every commit and PR will produce binaries that can be tested.

@Cpuroast I was wondering on what macOS versions the binary produced this way will run?
Currently the binary is not signed yet, so it needs the right-click workaround.
To download the release, click Details next to the Build-binaries check below here. Then select the Artifacts dropdown!


Safihre:
Updated to include codesign for MacOS, super easy with GitHub Actions. I am impressed!
Decided to skip notarisation, as it can easily take 10 minutes before Apple servers finish the process.

puzzledsab:
That's pretty neat. Now anyone can test the changes once they're committed. Could there be security issues regarding pull requests, though? Not sure how it will work, but it might be easier to add a pull request and give the impression that it is verified by you if it builds the binary.

Safihre:
The secrets are not available for pull requests when the branch is not from `sabnzbd/sabnzbd`. So then the signing step is skipped.

Cpuroast:
@Safihre Tested the artifacts, runs on 10.14, 10.15 and 11 only.

These builds exclude 10.9 through 10.13.

Safihre:
Hmmm that is unfortunate..

Cpuroast:
> Hmmm that is unfortunate..

I mean that all depends on if you want to cut off all officially un-supported versions of macOS with 3.2.0.

Current versions are 10.14, 10.15 and 11.
10.9 through 10.13 are end of life as far as Apple is concerned.

Safihre:
@Cpuroast I build now using older Xcode SDK, could you check if these new artifacts maybe do work on the older macOS versions?

Cpuroast:
> @Cpuroast I build now using older Xcode SDK, could you check if these new artifacts maybe do work on the older macOS versions?

Maybe I'm not pulling he right artifact.

Just tried this one: https://github.com/sabnzbd/sabnzbd/actions/runs/435605454

And it doesn't run even on 10.13

I don't get an error message, just a launch animation and then nothing.
Doesn't even list as crashed in Console.

Safihre:
@Cpuroast could you try this release? https://github.com/sabnzbd/sabnzbd/suites/1712069080/artifacts/32315280
It runs on my Big Sur machine.

Cpuroast:
> @Cpuroast could you try this release? https://github.com/sabnzbd/sabnzbd/suites/1712069080/artifacts/32315280
> It runs on my Big Sur machine.

Yup, exact same problem as the other artifacts, works on Mojave, Catalina and Big Sur, but not on anything older.
Something about the build process is restricting it to Mojave and up.

Safihre:
Alright, it seems I cannot go down to Xcode 10.3, it won't compile the python packages.

Safihre:
Will keep Travis around then! 
Can't start any builds yet on Travis, seems only after 27th of december my credits will be below 10.000 again.

PR #1714, opened 2020-12-22
Add multiple extra columns and option for unlimited width

### LINKED ISSUE #1474: [FR] More infos in Queue rows
Right now its only possible to add ONE more info to each row.

As a heavy use user my most wanted is cat and age.
In the history cat and speed.

I would like to request 1 more info shown or all infos shown.

All infos would also be nice with #1473

puzzledsab:
Fixes #1474. I couldn't make pureComputed work with the showColumn function because I couldn't make it use the $data variable. Not sure if it's necessary, the fields update if they are changed. Also I'm pretty weak at CSS, so there are probably better ways to do it. I tested it in Firefox and Microsoft Edge in lots of different widths, and it usually looks ok. I don't have other types of screens, though. I used multi select because I couldn't make it work with more complex data structures in Knockout.


interfaces/Glitter/templates/static/stylesheets/glitter.css
```diff
@@ -1840,6 +1845,15 @@ input[name="nzbURL"] {
     padding-top: 5px;
 }
 
+.container-unlimited-width .container {
+    width: calc(100% - 60px);
+}
+
+.queue-table .table-header-progress-smaller {
+    width: 25%;
+}
```
Safihre, line 1854: This should not be here, because we only support extra columns when the screen size is >1200pixels, otherwise everything is smashed together.
That's why it's also listed below in the `Dynamic sizing` section.

sabnzbd/skintext.py
```diff
@@ -908,6 +908,7 @@
     "Glitter-glitterTips": TT("Glitter has some (new) features you might like!"),
     "Glitter-custom": TT("Custom"),
     "Glitter-displayCompact": TT("Compact layout"),
+    "Glitter-displayUnlimitedWidth": TT("Unlimited width"),
```
Safihre, line 911: I would call it `Full width`

interfaces/Glitter/templates/static/stylesheets/glitter.css
```diff
@@ -1735,7 +1735,9 @@ input[name="nzbURL"] {
 .container-compact .search-box .form-control {
     height: 28px;
 }
-
+.container-compact .form-control-multiselect {
+    height: 60px;
```
Safihre, line 1739: Remove here, also listed few lines below.

interfaces/Glitter/templates/include_queue.tmpl
```diff
@@ -67,8 +67,10 @@
             <tr>
                 <th style="width: 25px;"></th>
                 <th></th>
-                <th class="table-header-extra" data-bind="css: { 'table-extra-header-visible' : extraQueueColumn }"></th>
-                <th class="table-header-progress" data-bind="css: { 'table-header-progress-smaller' : extraQueueColumn }"></th>
+                <!-- ko foreach: extraQueueColumns -->
+                <th style="width:7%"></th>
```
Safihre, line 71: I would suggest maybe `width: 5%`, 7 seems a lot.

Safihre (changes requested):
This looks pretty good, just a few minor comments 🥇 

Maybe we should also implement this for the History right away?

interfaces/Glitter/templates/include_queue.tmpl
puzzledsab: 5% seemed very narrow in my tests so I set it to 6. Other than that I think I've done the changes you suggested.

puzzledsab:
I forgot about the History part, I'll look into it.

Safihre:
Pushed 2 extra fixes.

interfaces/Glitter/templates/include_queue.tmpl
```diff
@@ -78,7 +78,7 @@
         <!-- ko if: !hasQueue() -->
         <tbody class="no-downloads">
             <tr>
-                <td colspan="6">
+                <td colspan="6" data-bind="attr: { 'colspan': 6 + extraQueueColumns().length }">
```
puzzledsab, line 81: I missed that part. Shouldn't this be 5 + extraQueueColumns().length, because the old extra column is removed?

Safihre: Good catched, I fixed it.

Safihre:
Looks good to me, also to you?

puzzledsab:
I think it's ok now.

Safihre (approved):

Issue #1713, opened 2020-12-22
Add a warning/disclaimer to "Enable accessing the interface from a HTTPS address."


sanderjo:
Proposal:

Add a warning/disclaimer to "Enable accessing the interface from a HTTPS address." like

"Note: Modern web browsers and other clients will not accept self-signed certificates and will give a warning and/or won't connect at all. So only activate HTTPS if you know what you're doing."

Rationale: more and more users are reporting problems with Chrome and Edge. Furthermore, Radarr 3.0.0.x by default does not accept self-signed, and throws an error.

![image](https://user-images.githubusercontent.com/1273502/102886310-2a72af00-4455-11eb-88aa-2f3f9bd95a84.png)

Safihre:
Sounds good, go ahead!

sanderjo:
@shypike How about this?

![image](https://user-images.githubusercontent.com/1273502/102915677-7556eb80-4482-11eb-96e0-7baeeca82d38.png)

Safihre:
Sure!

PR #1715, opened 2020-12-22
 Warning for enabling HTTPS

sanderjo:
Closes https://github.com/sabnzbd/sabnzbd/issues/1713

### DIFF
--- a/interfaces/Config/templates/config_general.tmpl
+++ b/interfaces/Config/templates/config_general.tmpl
@@ -31,7 +31,9 @@
                 <div class="field-pair">
                     <label class="config" for="enable_https">$T('opt-enable_https')</label>
                     <input type="checkbox" name="enable_https" id="enable_https" value="1" <!--#if int($enable_https) > 0 then 'checked="checked" data-original="1"' else ""#-->/>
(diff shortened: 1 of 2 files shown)


Safihre:
That's not the way to do this. Should introduce a new variable in `skintext.py`, for example `warn-https` and add the new text. Then add this new variable in the template. This way also the translations won't be broken.

interfaces/Config/templates/config_general.tmpl
```diff
@@ -32,6 +32,7 @@
                     <label class="config" for="enable_https">$T('opt-enable_https')</label>
                     <input type="checkbox" name="enable_https" id="enable_https" value="1" <!--#if int($enable_https) > 0 then 'checked="checked" data-original="1"' else ""#-->/>
                     <span class="desc">$T('explain-enable_https')</span>
+                    <span class="desc"><span class="label label-warning">WARNING</span>$T('explain-enable_https_warning')</span>
```
Safihre, line 35: Almost there! It should be:
```html
<span class="desc"><span class="label label-warning">$T('warning').upper()</span> $T('explain-enable_https_warning')</span>
```

PR #1718, opened 2020-12-24
add selecting by nzo_id(s) for api queue and history output

jcfp:
This change adds an _nzo_ids_ option to the queue and history api calls that limits output to jobs with matching ids, along the lines of what the _search_ parameter does based on a job name. Unlike job names, nzo_ids are unique and not subject to user changes which makes them more reliable as a job identifier.

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -54,7 +54,7 @@ jobs:
         name: Windows installer
     - name: Import codesign certificates
       uses: apple-actions/import-codesign-certs@v1
-      if: runner.os == 'macOS'
+      if: runner.os == 'macOS' && github.event_name == 'push'
       with:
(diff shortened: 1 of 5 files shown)


Safihre:
Looks good to me, just out of curiosity: was there a specific reason to add this functionality?
And will you also update the API documentation after the merge? 
I might take this PR as an opportunity to fix the binary building for PR's :)

jcfp:
> Looks good to me, just out of curiosity: was there a specific reason to add this functionality?

For specific jobs (whose nzo_id was known from the returned value when adding) there was no reliable way to get their info other than fetching all of queue and/or history, and selecting the correct jobs client-side. Found that annoying and inefficient, esp. over slower and metered internet connections.

> And will you also update the API documentation after the merge?

Of course, https://github.com/sabnzbd/sabnzbd.github.io/commit/47b2b6ece39eaef7b894801a1ffa3af897a57895

PR #1720, opened 2020-12-25
Match lline with x- instead of X-

puzzledsab:
Right? Because lline is lowered.


thezoggy:
per rfc spec it should be X-  ?
https://tools.ietf.org/html/rfc3977#section-9.4.1

Safihre:
You are right. But I wonder if this code is even applicable, it's been there for ages and broken for a while it seems and nobody complained..

Safihre:
Seems probably to be still relavant enough.

PR #1721, opened 2020-12-27
Filename limit in sanitize_filename() to avoid traceback

sanderjo:
I got tracebacks in urlgrabber with strange, 255+ char, non-ascii filenames, resulting in a no-download.
This patch solves that by brute-force removing non-ascii chars, and limiting to 250 chars. 

```
2020-12-26 22:30:51,461::ERROR::[urlgrabber:299] URLGRABBER CRASHED
  File "/usr/share/sabnzbdplus/sabnzbd/urlgrabber.py", line 261, in run
    with open(path, "wb") as temp_nzb:
OSError: [Errno 36] File name too long: "/home/sander/.sabnzbd/admin/future/[GuodongSubs]
```

### DIFF
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -40,7 +40,7 @@ ANFO = namedtuple("ANFO", "article_sum cache_size cache_limit")
 # Leave some space for "_UNPACK_" which we append during post-proc
 # Or, when extra ".1", ".2" etc. are added for identically named jobs
 DEF_FOLDER_MAX = 256 - 10
-DEF_FILE_MAX = 255 - 6  # max filename length on modern filesystems, minus 6 positions
+DEF_FILE_MAX = 255 - 10  # max filename length on modern filesystems, minus some room for extra chars later on
 
 GIGI = float(2 ** 30)
 MEBI = float(2 ** 20)
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -40,7 +40,7 @@ ANFO = namedtuple("ANFO", "article_sum cache_size cache_limit")
 # Leave some space for "_UNPACK_" which we append during post-proc
 # Or, when extra ".1", ".2" etc. are added for identically named jobs
 DEF_FOLDER_MAX = 256 - 10
-DEF_FILE_MAX = 255 - 6 # max filename length on modern filesystems, minus 6 positions
+DEF_FILE_MAX = 255 - 6  # max filename length on modern filesystems, minus 6 positions
 
 GIGI = float(2 ** 30)
 MEBI = float(2 ** 20)
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -40,6 +40,7 @@ ANFO = namedtuple("ANFO", "article_sum cache_size cache_limit")
 # Leave some space for "_UNPACK_" which we append during post-proc
 # Or, when extra ".1", ".2" etc. are added for identically named jobs
 DEF_FOLDER_MAX = 256 - 10
+DEF_FILE_MAX = 256 - 10
 
 GIGI = float(2 ** 30)
 MEBI = float(2 ** 20)
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -40,7 +40,7 @@ ANFO = namedtuple("ANFO", "article_sum cache_size cache_limit")
 # Leave some space for "_UNPACK_" which we append during post-proc
 # Or, when extra ".1", ".2" etc. are added for identically named jobs
 DEF_FOLDER_MAX = 256 - 10
-DEF_FILE_MAX = 255 # max filename length on modern filesystems
+DEF_FILE_MAX = 255 - 6 # max filename length on modern filesystems, minus 6 positions
 
 GIGI = float(2 ** 30)
 MEBI = float(2 ** 20)
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -40,7 +40,7 @@ ANFO = namedtuple("ANFO", "article_sum cache_size cache_limit")
 # Leave some space for "_UNPACK_" which we append during post-proc
 # Or, when extra ".1", ".2" etc. are added for identically named jobs
 DEF_FOLDER_MAX = 256 - 10
-DEF_FILE_MAX = 256 - 10
+DEF_FILE_MAX = 255 # max filename length on modern filesystems
 
 GIGI = float(2 ** 30)
 MEBI = float(2 ** 20)
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -177,7 +177,7 @@ def sanitize_filename(name: str) -> str:
         name = str(name.encode("ascii", "ignore"), "utf-8")
         if len(name) + len(ext) > DEF_FILE_MAX:
             # still too long, limit the extension
-            maxextlength = 100 # max length of an extension
+            maxextlength = 100  # max length of an extension
             if len(ext) > maxextlength:
                 # allow first <maxextlength> chars, including the starting dot
                 ext = ext[:maxextlength]
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -169,7 +169,8 @@ def sanitize_filename(name: str) -> str:
     # now split name into name, ext
     name, ext = os.path.splitext(name)
 
-    # If filename is too long (more than DEF_FILE_MAX bytes), brute-force truncate it.
+    # If filename is too long (more than DEF_FILE_MAX bytes), brute-force truncate it,
+    # preserving the extension (max ext length 20)
     # Note: some filesystem can handle up to 255 UTF chars (which is more than 255 bytes) in the filename,
     # but we stay on the safe side: max DEF_FILE_MAX bytes
     if len(name.encode("utf8")) + len(ext.encode("utf8")) > DEF_FILE_MAX:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -169,15 +169,22 @@ def sanitize_filename(name: str) -> str:
     # now split name into name, ext
     name, ext = os.path.splitext(name)
 
-    # If filename is too long, truncate it:
-    if len(name) + len(ext) > 222:
+    # If filename is too long, brute-force truncate it:
+    maxlength = 222
+    if len(name) + len(ext) > maxlength:
         logging.debug("Filename %s is too long, so truncating", name + ext)
         # Too long filenames are often caused by incorrect non-ascii chars,
-        # so brute-force remove those non-ascii chars, and only keep 222 chars
-        # ... keeping in place the original extension
+        # so brute-force remove those non-ascii chars
         name = str(name.encode("ascii", "ignore"), "utf-8")
-        name = ''.join( c for c in name if  ord(c) >= 32 and ord(c)<128 )
-        name = name[: 222 - len(ext)]
+        if len(name) + len(ext) > maxlength:
+            # still too long, limit the extension
+            maxextlength = 100
+            if len(ext) > maxextlength:
+                # allow first 150 chars, including the starting dot
+                ext = ext[:maxextlength]
+            if len(name) + len(ext) > maxlength:
+                # Still too long, limit the basename
+                name = name[: maxlength - len(ext)]
 
     lowext = ext.lower()
     if lowext == ".par2" and lowext != ext:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -169,8 +169,10 @@ def sanitize_filename(name: str) -> str:
     # now split name into name, ext
     name, ext = os.path.splitext(name)
 
-    # If filename is too long (in bytes), brute-force truncate it:
-    if len(str.encode(name)) + len(str.encode(ext)) > DEF_FILE_MAX:
+    # If filename is too long (more than DEF_FILE_MAX bytes), brute-force truncate it.
+    # Note: some filesystem can handle up to 255 UTF chars (which is more than 255 bytes) in the filename,
+    # but we stay on the safe side: max DEF_FILE_MAX bytes
+    if len(name.encode("utf8")) + len(ext.encode("utf8")) > DEF_FILE_MAX:
         logging.debug("Filename %s is too long, so truncating", name + ext)
         # Too long filenames are often caused by incorrect non-ascii chars,
         # so brute-force remove those non-ascii chars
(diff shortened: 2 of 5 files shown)


Safihre:
@sanderjo As discussed, we should fix this in `sanitize_filename`.
We can make that smarter by first removing the extension and if it has a sensible length `<=5` deduct that length from the maxium, so we can re-attach  the extension after shortening the filename.
For example:
```
very<etc>longgggname.txt`
```
Becomes
```
very<etc>longg.txt`
```

sanderjo:
OK. 

Test-URL:
```http://www.appelboor.com/dump/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.longext```

sabnzbd.log:

```
2021-01-15 21:06:15,205::DEBUG::[filesystem:175] Filename aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.longext is too long, so truncating
2021-01-15 21:06:30,029::DEBUG::[filesystem:175] Filename _UNPACK_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa is too long, so truncating
```

**Comment by Safihre** [comment]
Great. Let's add some unit tests!

File: sabnzbd/filesystem.py [hunk 1]
```diff
@@ -166,7 +166,19 @@ def sanitize_filename(name: str) -> str:
...
+
+    # If filename is too long, truncate it:
+    if len(name) + len(ext) > 222:
+        logging.debug("Filename %s is too long, so truncating", name + ext)
+        # Too long filenames are often caused by incorrect non-ascii chars,
+        # so brute-force remove those non-ascii chars, and only keep 222 chars
+        # ... keeping in place the original extension
+        name = str(name.encode("ascii", "ignore"), "utf-8")
+        name = ''.join( c for c in name if  ord(c) >= 32 and ord(c)<128 )
+        name = name[: 222 - len(ext)]
```
jcfp, line 180: The length of the extension isn't sanity-checked anywhere. With os.path.splitext simply returning everything after the last dot, you're in for unexpected results for the value `name` once `222 - len(ext)` goes negative - both longer and (much) shorter than intended.

It would make sense to add tests guarding against such issues as well.

sabnzbd/filesystem.py
```diff
@@ -166,7 +166,19 @@ def sanitize_filename(name: str) -> str:
...
     name, ext = os.path.splitext(name)
+
+    # If filename is too long, truncate it:
+    if len(name) + len(ext) > 222:
+        logging.debug("Filename %s is too long, so truncating", name + ext)
+        # Too long filenames are often caused by incorrect non-ascii chars,
+        # so brute-force remove those non-ascii chars, and only keep 222 chars
+        # ... keeping in place the original extension
+        name = str(name.encode("ascii", "ignore"), "utf-8")
+        name = ''.join( c for c in name if  ord(c) >= 32 and ord(c)<128 )
```
jcfp, line 179: Unless I'm mistaken, this may significantly shorten the length of name to the point that the whole thing already fits inside the 222 limit. Even so, the code would still proceed to (unnecessarily) shorten the name by len(ext).

Safihre: I don't fully get this part. Why do we have the ascii replace and then another check using the ord()?

sanderjo (commented):
thanks for the feedback. Solved now, IMHO

sabnzbd/filesystem.py
```diff
@@ -41,6 +41,7 @@
 from sabnzbd.constants import FUTURE_Q_FOLDER, JOB_ADMIN, GIGI
 from sabnzbd.encoding import correct_unknown_encoding
 from sabnzbd.utils import rarfile
+from sabnzbd.constants import DEF_FILE_MAX
```
Safihre, line 44: Can you add it to the import 2 lines above?

sabnzbd/filesystem.py
```diff
@@ -166,7 +167,26 @@ def sanitize_filename(name: str) -> str:
     if not name:
         name = "unknown"
 
+    # now split name into name, ext
     name, ext = os.path.splitext(name)
+
+    # If filename is too long, brute-force truncate it:
+    maxlength = DEF_FILE_MAX
+    if len(name) + len(ext) > maxlength:
```
Safihre, line 175: No need to introduce extra variable, just use DEF_FILE_MAX everywhere.

Safihre (changes requested):
Looks good to me, maybe @jcfp has some more things? :)

sabnzbd/filesystem.py
sanderjo: Done

sabnzbd/filesystem.py
sanderjo: Done

sabnzbd/constants.py
```diff
@@ -40,6 +40,7 @@
 # Leave some space for "_UNPACK_" which we append during post-proc
 # Or, when extra ".1", ".2" etc. are added for identically named jobs
 DEF_FOLDER_MAX = 256 - 10
+DEF_FILE_MAX = 256 - 10
```
jcfp, line 43: Why subtract 10 for the filename? Any risk of lengthy tags like those applies to dirs when unpacking?

And IIRC the actual filename length limit on mainstream filesystems is 255.

sabnzbd/filesystem.py
```diff
@@ -166,7 +166,25 @@ def sanitize_filename(name: str) -> str:
...
+
+    # If filename is too long, brute-force truncate it:
+    if len(name) + len(ext) > DEF_FILE_MAX:
+        logging.debug("Filename %s is too long, so truncating", name + ext)
+        # Too long filenames are often caused by incorrect non-ascii chars,
+        # so brute-force remove those non-ascii chars
+        name = str(name.encode("ascii", "ignore"), "utf-8")
+        if len(name) + len(ext) > DEF_FILE_MAX:
+            # still too long, limit the extension
+            maxextlength = 100  # max length of an extension
```
jcfp, line 180: Maybe go with a more realistic limit here? After all, we only really care about the extension because it may provide useful information; in all other cases it's just a random dot somewhere in the filename.

sanderjo: OK. I checked my files, and the longest extension I had was "releaseinfo". 

So ... limit to ... 20?

jcfp: If anything, that would still leave plenty of room. But definitely better than 100.

sabnzbd/constants.py
sanderjo: Just trying to stay on the safe side, without too much thinking because of this edge case. I would consider 255 too risky and/or too much work to think about and to test. But ... I can let the unittest take care of the testing and let it write the actual file to see if it works.

... playing with that

>>> outF = open("A"*255, "w")
>>> 
>>> outF = open("A"*256, "w")
OSError: [Errno 36] File name too long: 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
>>> 

... OK, good, but then I discovered something bad:

```
>>> name = "你"*86
>>> len(name)
86
>>> outF = open(name, "w")
OSError: [Errno 36] File name too long: '你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你你'
```

Ouch: 86 UTF-8 is too much! Which is logical if you look at the real bytes:

```
>>> str.encode(name)
b'\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0\xe4\xbd\xa0'
>>> len(str.encode(name))
258
```
So > 255.

... I have some work to do in the code. Thanks!

jcfp: Nice find on the encoding.

For the record: I didn't check whether sab appends a `.1 .2 ... [.n]` style suffix as a workaround for duplicate filenames. If it does, some margin is obviously needed, although reserving 10 chars (i.e. room for 10<sup>9</sup> duplicates of any single filename) just for that seems excessive.

Safihre: This works:
```
>>> name = "你"*86
>>> len(name.encode("utf8"))
258
```

sanderjo: I've set it to 5.

sanderjo:
Ah, interesting / FWIW

As a test, I set `DEF_FILE_MAX = 255` and put http://www.appelboor.com/dump/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.longext  (so 254 chars) into SABnzbd ... resulting in the traceback below.

Cause: https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/__init__.py#L604-L610 extends the filename, without sanitize(). And running sanize() is dangerous because of the double extension `.nzb.gz`, so the `.nzb` is not part of the extension, and would get cut off ...

```
2021-01-19 21:02:17,414::INFO::[__init__:621] Traceback: 
  File "/home/sander/git/sjo-sab/sabnzbd/__init__.py", line 614, in save_compressed
    with open(os.path.join(folder, filename), "wb") as tgz_file:
OSError: [Errno 36] File name too long: '/home/sander/Downloads/incomplete/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/__ADMIN__/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.longe.nzb.gz'
```

PR #1726, opened 2020-12-28
docs: fix simple typo, sturct -> struct

timgates42:
There is a small typo in sabnzbd/utils/rarfile.py.

Should read `struct` rather than `sturct`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


Safihre:
We just use a modified copy, so should also check if this needs to be fixed in the source: https://github.com/markokr/rarfile

PR #1731, opened 2020-12-31
Don't crash if nzf is gone

### LINKED ISSUE #1728: Exception when calling article.nzf.nzo.update_download_stats because article.nzf was deleted
This has happened at least a couple of times today. It was a coincidence that I noticed because it wasn't logged, but I have a long scroll back. It caused a crash and restart. Both times it was right after a download was finished. I assume it had time to delete it while the bandwidth limiter was sleeping. Not sure what a good way to fix it would be. Maybe do the update before the loop, or put the bandwidth loop after the `for selected in read:` loop?

Exception in thread Thread-5:
  File "C:\Users\t\AppData\Local\Programs\Python\Python38\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\git\sabnzbd\sabnzbd\downloader.py", line 639, in run
    article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, bytes_received)
AttributeError: 'NoneType' object has no attribute 'nzf'

This is the line:
https://github.com/sabnzbd/sabnzbd/blob/5df4a76367fba92756c2c3dba50c3c5126f5dfb0/sabnzbd/downloader.py#L639

puzzledsab:
Should fix #1728. I considered logging the error instead, but it's not very important as long as it happens rarely and doesn't cause a crash. Putting the update before the loop should reduce the chance of it happening to ~0.

The reason why it has started happening is probably that until 3 months ago a reference to the nzo was fetched before the loop started.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -629,6 +629,10 @@ class Downloader(Thread):
                     continue
 
                 else:
+                    try:
+                        article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, bytes_received)
+                    except:
+                        pass
                     if self.bandwidth_limit:
                         limit = self.bandwidth_limit
                         if bytes_received + sabnzbd.BPSMeter.bps > limit:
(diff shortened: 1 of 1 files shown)


sabnzbd/downloader.py
```diff
@@ -629,14 +629,17 @@ def run(self):
                     continue
 
                 else:
+                    try:
+                        article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, bytes_received)
+                    except:
```
Safihre, line 634: Let's catch `AttributeError` specifically and add a comment why we do it.

Safihre (changes requested):

PR #1735, opened 2021-01-02
Add Disk class for checking free space in a directory

puzzledsab:
This was just a test. The variable names could definitely be better, but the concept of de-referencing the volume and making links to directories works pretty well. I also prefer the timeout to be configurable by the caller. Let me know if you want it and what changes to make.

I used it in assembler.py by importing disk and calling for instance `disk.free(cfg.download_dir.get_path())`. Not sure if I did that part right. What I wanted was to share the data between all callers, like cfg.


Safihre:
Good start, feel free to submit improved version :)
The code in the free function looked indeed nice, don't know if the Disk class is even needed.

Issue #1709, opened 2020-12-19
Discussion about asyncio rewrite


puzzledsab:
Continuation of the asyncio discussion from #1704

I've actually tried writing a very basic asyncio based usenet article downloader a while ago. Just fetching multiple articles concurrently, nothing else. It worked and had a low CPU load, but the problem is that so much in Sab is written around looping, async polling and using the results of those polls on each iteration. Switching to asyncio would require some big rewrites. You should probably freeze the development for everything except vital fixes in the mean time because lots of it won't be reusable.

I can clean it up a little and post it somewhere if you're interested. Currently it doesn't even support login because I used the free IPv6 server.

Safihre:
Yes, absolutely a big rewrite is needed. But that doesn't have to be a problem, for Python 3 conversion we also didn't release for a long time to get things right.
The current code is based on principles from back in Python 2.5 or even earlier times. You can see it when you do a git blame, so much code from 10 years ago. So a big refresh is worth it.

puzzledsab:
I made it work with SSL and login. For me it doesn't seem to use less CPU than Sab currently does, even with no decoding and extra logic. Maybe there's nothing to gain from it? Here's the test downloader I wrote: https://gist.github.com/puzzledsab/3605240541ee7cf2bb0eae962ba91182

It reads a list of articles from articles.txt, like this one: https://gist.github.com/puzzledsab/88fdf61c1d9a6607b24cae89e92db578

Safihre:
Does it run concurrently?
I would imagine it can show its strength when decoding is included, since this way each server thread can also be a decoder thread, so many more at once can be active.

puzzledsab:
Yes, you can set the number of connections using the threads variable. I think the decoding would run as separate jobs using asyncio, to avoid delays between requests for new articles. Regarding decoding, couldn't there be more decoder threads with the current version as well? I assume CPUs with many cores could benefit from it.

mnightingale:
> I made it work with SSL and login. For me it doesn't seem to use less CPU than Sab currently does, even with no decoding and extra logic. Maybe there's nothing to gain from it? Here's the test downloader I wrote: https://gist.github.com/puzzledsab/3605240541ee7cf2bb0eae962ba91182
> 
> It reads a list of articles from articles.txt, like this one: https://gist.github.com/puzzledsab/88fdf61c1d9a6607b24cae89e92db578 You need to configure a server and login in main().

[...]

PR #1745, opened 2021-01-08
Try sleeping less often at higher speeds

puzzledsab:


Safihre:
You'll have to explain a bit more the logic 😅
The speed devided by 50M relates to the idle counter?

puzzledsab:
Currently it's just meant as a test, but my hope is that polling more times between each sleep at higher speeds will mitigate the speed reduction that is discussed in the forum. At lower speeds it can easily be compensated by increasing the number of connections, but apparently there are diminishing returns at higher speeds. With this change it will poll at least 5 times instead of 2 times between sleeps at 25 MB/s, and 10 times at 50 MB/s. At 100 MB/s there will hardly be any sleeping at all unless the CPU is very fast.

JackDandy:
Just a thought, be aware that the purpose of `sleep` is to signal processors to conduct other work. Without sleep some processors don't get a chance to switch/process workload to another task. Therefore, this change _may_ create the effect of sending some processors longer into 100% and prevent anything other than the core task at hand which could actually make the general OS sluggish, and slow down what you are trying to speed up by reducing/removing the sleep (aka breathing) periods. Testing at higher speeds, and with direct unpack may present the issue, and would give confidence to if this change _actually_ makes an improvement or is detrimental compared to the existing sabnzbd logic.

thezoggy:
when you pause matters. imagine if you forced a marathon runner to stop and take a breathe after every x steps, it would be quite inefficient and slow them down considerably

JackDandy:
and to breathe too little, the runner would likely collapse, so yeah, to sleep matters a great deal.

puzzledsab:
Yes, you're right, I did consider removing the reset of the idle counter after an article is found. That way one could be sure that there would be some sleeping even at high speeds. On the other hand, the Python code is locked to one core. I assume the unpack process will run in a separate core, so it shouldn't be strongly impacted.

There will be less time for some of the other threads, though, but I hope we'll get some feedback with this method first.

@thezoggy: This isn't quite comparable because the CPU can pick up full speed again immediately when the pause ends, and it costs no extra energy to do so. I still think it's better to have several short pauses than one long, like the get_stable_speed did. With short pauses less data will pile up before the CPU starts reading it again. It's very limited how much data can be buffered before it has to be read or dropped.

thezoggy:
you are assuming that the cpu isnt busy with another task, but in most cases I would guess that sab is not cpu bound but io bound. waiting on sockets handling (netio) / disk writing (diskio) etc. when sab is d/ling at 100M+ I see <4% cpu. historically a lot of people complained about sab speed are ones that run it on traditional hdd or have bad workflow where they are moving files from one place to another and that transfer (disk read / network transfer) kills their performance.

JackDandy:
> when sab is d/ling at 100M+ I see <4% cpu

i'm guessing that is with the existing `sleep` logic. does that 4% change/improve with the proposed change?

puzzledsab:
The problem is that SAB isn't waiting for the sockets, it has to poll them constantly to see if there is new data. That's what the loop we're delaying is doing. The reason why it's not using one full core all the time is that it tries to figure out what the max speed is, and regularly add a 0.05s sleep. For some use cases this is too long, so it makes the speed drop back down again. It also doesn't sleep as much as it could.

A rewrite to using asyncio, as discused in https://github.com/sabnzbd/sabnzbd/issues/1709, might solve this. Then it would be waiting for the buffers to fill instead of polling them constantly. Unfortunately it's a very big job to do it because so much of SAB is written around the polling. I'm also not sure how much it would actually help. The test I posted uses surprisingly much CPU on my system.

Safihre:
@puzzledsab I think the big difference with 3.1.x is that the new code uses the article-selection to determine if there should be delayed, while the old code use the outcome of `select.select` to see if it was checking the sockets too often. So the code had a very different purpose..
For some reason I totally missed that during the PR.

Safihre:
To add to that: the previous code with the `check_stable_speed` was so weird because we can't just simply check the output of `select.select` and if it's number of available `read`'s is lower than a certain threshold, pause. I tried this and it turned out that on people running SABnzbd on a virtual machine with virtual network adapters and on some VPN's, that `select.select` would only output something when requested. So this would destroy their speeds.
https://www.google.com/search?q=sabnzbd+slow+vpn

puzzledsab:
I have done some more testing with this and the old sleep method using no speed limit, and the results are a bit confusing. Even with no sleep at all, my version will not use more than 6% CPU. That's 30% of one core on my I5 at 3.8GHz. It used to be 25%, or 100% of one core. Maybe that's changed with the newer Python versions. I'm using 3.8 because I had problems with 3.9. With a idle_count > 1 it uses around 1%, > 10 uses just over 2%, and >20 uses 3-4%.

The master branch uses 6%, although once in a while it falls to 3-4%. I guess that means that the sleep rarely kicks in at all with the check_stable_speed method on my system.

Unfortunately it doesn't seem like anyone who currently has problems will test this change. You could try a new beta using Python 3.8 and this patch or a fixed idle_count threshold of for instance 10 or higher. If not then I guess the only alternative is to revert to the old system.

Safihre:
I'm not sure I understand what you mean with the cpu percentages?
Did you also check with SSL turned on? Due to the small chunk sizes the `select` needs to be called much more often.

puzzledsab:
I mean the CPU percentage that the Windows task manager says the python process is using when SAB is running. I'm always using SSL.

Safihre:
What is was trying to say is that I was expecting the idle counter maybe to also be lowered/reset when there's (a lot of) results from the `select.select`?
Now it's soly based on if new articles were assigned to the server.

puzzledsab:
Oh, I see. That might work well. In my tests I always only get one result, except after a sleep, but then I have a fast CPU and a slow connection.

I'm not sure if the buffer has to be full to be returned, though. In some other tests I did a while ago the buffer was sometimes only 50% full so it may return several results even if it could have slept. On the other hand it would reduce the chance of polling too rarely, and some sleeping is better than none.

I think it would be fairly conservative to set an idle threshold of 5, and reduce the counter by 1 if there is more than one read result, or maybe more than 5-10% of the number of active connections, if that can be easily determined. The counter should be reset after a sleep.

If you're OK with this I'll update the PR.

Safihre:
I checked your changes (up to edc4367) on my slow Pentium N3700 NUC and your changes seem to have worked, I can also confirm the bad results of 3.2.0Beta1 that users were reporting:
image

puzzledsab:
I've tested some more variations, and it seems that the problem with high CPU load only appears when I use few connections. With 4 connections I will get 1 result in read almost every time. When I increase it to 20 connections, for some reason it starts waiting a lot longer for `select.select()`. I get 19 or 20 results every time, and the CPU load falls to about 1%. If anyone has any idea why this happens I'd really like to know.

Anyway, the current PR should only add sleep if there is a lot of polling. I added the connection counter because otherwise idle_count would be reset every time before reaching the threshold when there are many connections.

puzzledsab:
Actually, it would still be reset every time if the CPU manages to poll quickly enough to get only 1 result from read and there are many connections. Either the threshold would have to be lowered to avoid it, or the reset after adding a new article would have to be removed. Neither is ideal so I'll just remove the connection counter to save a few CPU cycles.

Safihre:
In a way, it's now the same code that I made long ago and that forced me to introduce the `get_stable_speed` due to the virtual machie and VPN having a problem with it (https://www.google.com/search?q=sabnzbd+slow+vpn).
My old code [here](https://github.com/sabnzbd/sabnzbd/blame/04fa3c31158649b13c5b5cfeb0f84ac89147ebf4/sabnzbd/downloader.py#L509-L511
).
Only difference is that this code is only applied after 6 consecutive `<2`, so less often.
I guess it's worth a try..

puzzledsab:
It's certainly the same concept. The old code would kick in at least 5 times as often and sleep 5-10 times as long, though. If you wait a little longer I'll try to come up with something that works better with many connections.

puzzledsab:
Ok, if this doesn't work I'm giving up. Subtracting when there are lots of connections should reduce or eliminate the number of wrong sleeps, but at the same time not stopping it completely from sleeping when there are many connections like resetting idle_count to 0 would. Inc by one is quite slow in Python so `len(server.busy_threads)` will be much faster when there are many.

Safihre:
The `readkeys` have the number of active connections. But do we really need this extra step based on number of connections? Without it it also worked fine right?

puzzledsab:
`readkeys` includes all the open connections, not just the ones that is actively downloading. It can be a much higher number than `connection_count` for a while if one article is missing and there are several backup servers. I'm just guessing, but I think that the reason why it works without it is that you're testing on a system where it waits for all or most of results at each poll, like it did for me when I used many connections. In those cases the sleep shouldn't kick in at all.

If you are using 50 connections and you're mostly getting 5 or fewer results from each poll I think it's still safe to add the sleep once in a while. It won't be if you are using 5 connections in total. In that case it shouldn't sleep if there is more than 1 result from each poll. The divisor could be increased to 15 or 20 to be even safer, though.

Safihre:
Hmmm, I find the new code not very pretty, I know that's subjective of course but it's hard to see what it is actually doing and why.
It definitely needs comments for the if-statements.

thezoggy:
Do we have any real world benchmarks or synthetic tests with it before/after?

puzzledsab:
@Safihre : Was the problem you were trying to fix that `select.select()` would always only return one result with some setups, no matter how much it slept? If so I think it's safer to test for that and disable the sleeping.

Safihre:
@puzzledsab That's exactly indeed what I found, it would always report 1 to 3 results even if you slept for 1 second on a fast connection.
On slow connections or with few connections it can still be "normal" to only have 1 to 3 results, so how to check for that?

puzzledsab:
Maybe trigger a one time test when the downloading reaches a certain speed, do a longer pause and disable the sleeping if too few results are returned compared to the number of connections. If the speed is low a 0.001s sleep once in a while won't matter, and the user will probably try to increase the number of connections if the speed is lower than expected.

Safihre:
That is kind of what the old test did, apply the sleep and check if it affects the speed. Because "a certain" speed to trigger the test can be quite different: 100MB/s is easy for an Intel i7 but 20MB/s can already be hard for a raspberry..

thezoggy:
Maybe test for the issue rather than the symptom? Check if cpu is high for sab process and sleep rather than try and guess using bandwidth

puzzledsab:
@Safihre : I guess we've come almost full circle. I realize now that I didn't fully understand the old system, and the reason why it never worked properly for me was that I usually use less than 8 connections. I still think it would be an advantage to count the number of connections and adapt to it instead of setting a minimum of 8 connections, and to use a 0.001s sleep instead of 0.01. The increase in CPU usage would be minimal on most systems, and the chance of slowing down the transfer should be less. How about you just revert this commit? If you want to keep it the way it is I will stop nagging you about it, or I can submit a PR with modifications I mentioned.

@thezoggy : Then one would have to determine if the high CPU usage is caused by a slow CPU or because it was polling more than necessary.

puzzledsab:
Either way I suggest you just revert all the changes and use the old system for the next release so it doesn't delay it further.

Safihre:
We're not in a rush, and I never liked the old system either. It was indeed just born out of trial and error.
I am in favor of indeed using a low sleep value instead, with a smarter threshold.

puzzledsab:
Great. I'll make a new PR with a mix of the old and new system. Not sure if there will be conflicts if you try to revert it directly.

puzzledsab:
Replacing this with a new PR.

PR #1748, opened 2021-01-13
can_be_slowed with some modifications

puzzledsab:
I kept the configurable sleep time so that it can be set to 0 in case of problems. I also made the required number of reads depend on the number of active connections, so it will trigger somewhat differently than before. It will start working for systems that use less than 8 connections, but there might still be problems if anyone are using 40-50 connections and never can get more than 1-3 results. Hopefully the speed test will disable the sleep in those cases. I also reduced the BPS limit sleep to 0.01 in order to work better with the same cases.

### DIFF
--- a/sabnzbd/bpsmeter.py
+++ b/sabnzbd/bpsmeter.py
@@ -372,6 +372,32 @@ class BPSMeter:
         # We record every second, but display at the user's refresh-rate
         return self.bps_list[::refresh_rate]
 
+    def get_stable_speed(self, timespan=10):
+        """See if there is a stable speed the last <timespan> seconds
+        None: indicates it can't determine yet
+        False: the speed was not stable during <timespan>
+        """
+        if len(self.bps_list) < timespan:
+            return None
+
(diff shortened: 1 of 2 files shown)


Safihre:
Why the `1 + connection_count / 10`? Could just trust all on the speed-test?

puzzledsab:
It's intended to increase the chance that it sleeps when it can but not when it shouldn't. In my view it's a separate issue from what the speed test is trying to solve.

With a fixed limit it would be the same for 5 and 50 connections. No matter what configuration the user is running I don't think sleeps should be added if there are 4 connections and 2 results because that's 50% of the buffer capacity, but it's probably not a problem if there are 4 results from 50 connections.

Another related issue is that I noticed that the number of results would increase when other stuff was done in addition to the downloading, like unpacking. With a fixed limit of 2 or 3 results I think this could increase the risk that the speed test works because there is little load in the beginning. After a while it could be slowed by the combination of the sleep and the additional load.

PR #1749, opened 2021-01-13
Simplify delay if no articles were found for a server

Safihre:
@puzzledsab I changed a bit the skipping of servers that have no articles.
What do you think of this version?

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1376,7 +1376,7 @@ def build_queue(start=0, limit=0, trans=False, output=None, search=None, nzo_ids
 
         if (
             sabnzbd.Downloader.paused
-            or sabnzbd.Downloader.postproc
+            or sabnzbd.Downloader.paused_for_postproc
             or is_propagating
             or status not in (Status.DOWNLOADING, Status.FETCHING, Status.QUEUED)
         ) and priority != FORCE_PRIORITY:
@@ -1627,7 +1627,7 @@ def build_header(webdir="", output=None, trans_functions=True):
         header["new_release"], header["new_rel_url"] = sabnzbd.NEW_VERSION
 
     header["version"] = sabnzbd.__version__
-    header["paused"] = bool(sabnzbd.Downloader.paused or sabnzbd.Downloader.postproc)
+    header["paused"] = bool(sabnzbd.Downloader.paused or sabnzbd.Downloader.paused_for_postproc)
     header["pause_int"] = sabnzbd.Scheduler.pause_int()
     header["paused_all"] = sabnzbd.PAUSED_ALL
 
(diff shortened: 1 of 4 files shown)


puzzledsab:
Not sure why you `int()` the time but the logic is definitely an improvement.

Safihre:
Did I maybe miss something? Because you specifically implemented `last_busy` and `last_searched`, so did I miss some cases here?

The `int` is because the comparison between float and int is slower. Maybe it's too much of a micro-optimization...

puzzledsab:
My solution distinguished between the time since it last looked for an article and when it was last busy downloading. That way, if there were several levels of backup servers and the articles were trickling down slowly, it wouldn't stop searching too quickly between each new article and make it even slower.

It was an issue because I initially had longer delays between each time it searched when it was idle. Now that it's reduced to 1 second between searches it probably doesn't matter.

I love micro-optimization :) As far as I can tell, both `nw.timeout` and `nzo.avg_stamp` are float, though? `next_article_search` can what you chose. Also, according to my tests with timeit, you need to avoid 5 int/float comparisons to save more than the cost of calling `int()` once.

Safihre:
> I love micro-optimization :) As far as I can tell, both nw.timeout and nzo.avg_stamp are float, though? next_article_search can what you chose. Also, according to my tests with timeit, you need to avoid 5 int/float comparisons to save more than the cost of calling int() once.

Ah, thanks for testing 😁 Then it's not worth it and it can just stay a float.

Have to see why the test are failing though, somehow the `TryList` objects are kept a few seconds longer in memory due to this change. I could just add a `time.sleep(10)` to the tests. But want to make sure there's no bigger underlying issue.

puzzledsab:
Try this:
`if server.next_article_search > now_int and not server.busy_threads:`

Other differences are that my version would continue polling for 1 second after there were no busy_threads and wait 0.5s between each server check instead of 1s in idle mode.

Safihre:
Found the issue to be that the tests didn't auto-disconnect after finishing, so the cleanup of threads would be delayed with 1 second due to the pause.

Safihre:
@puzzledsab Due to your code I finally took the time to look into this whole `read_fds` and `write_fds` stuff.. Looking at the Git Blame much of this code was from before SABnzbd was on Github, so 13+ years old 🤣
I was surprised to see the `read_fds` also contained dormant connections, so I updated the code to remove non-active connections. This way we can use `readkeys` for the improved sleep-timer.

I also found that `write_fds` is useless, because it is only needed if a `socket.connect` is performed in non-blocking mode. However, we only turn on non-blocking-mode _after_ the `socket.connect`. So I removed all of that.

Refactor <3

Let me know if you (or anyone else) spots any flaws in my changes.

puzzledsab:
I never did understand that part of the code. I'm happy to see that it's at least partly because it didn't make sense.

PR #1750, opened 2021-01-16
ssdp: make the time.sleep() less blocking

sanderjo:
Solves https://forums.sabnzbd.org/viewtopic.php?f=11&t=25148

### DIFF
--- a/sabnzbd/utils/ssdp.py
+++ b/sabnzbd/utils/ssdp.py
@@ -111,6 +111,7 @@ OPT: "http://schemas.upnp.org/upnp/1/0/"; ns=01
         MCAST_PORT = 1900
         MULTICAST_TTL = 2
 
+        # TODO: is that 1 needed?
         while 1 and not self.__stop:
             # Do network stuff
             # Use self.__host, self.__url, self.__server_name to do stuff!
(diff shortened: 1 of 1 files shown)


Safihre:
I think it's better and easier to just use `.kill()` instead of `.join()` in the shutdown. There is no vital information in this thread anyway.

sanderjo:
> I think it's better and easier to just use `.kill()` instead of `.join()` in the shutdown. There is no vital information in this thread anyway.

`.kill()`? Is that a Thread call? I can't find it

```
>>> from threading import Thread
>>> dir(Thread)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_bootstrap', '_bootstrap_inner', '_delete', '_initialized', '_reset_internal_locks', '_set_ident', '_set_native_id', '_set_tstate_lock', '_stop', '_wait_for_tstate_lock', 'daemon', 'getName', 'ident', 'isAlive', 'isDaemon', 'is_alive', 'join', 'name', 'native_id', 'run', 'setDaemon', 'setName', 'start']
>>> 
```

and

```
  File "/home/sander/git/sjo-sab/sabnzbd/utils/ssdp.py", line 159, in stop_ssdp
    __SSDP.kill()
AttributeError: 'SSDP' object has no attribute 'kill'
```

sanderjo:
Just no .join() solves it.

Safihre:
Not using `join` could result in leftover threads after shutdown.
I fixed this problem in 30e692cefeed6a5b5c4c54bc9fa7f934ea0f9133 using `Condition(Lock())`. No `sleep`'s needed :)

PR #1756, opened 2021-01-18
Sort server list and take advantage of it in loops

puzzledsab:
Not sure if this is safe or if it would require locking. Maybe the `sabnzbd.Downloader.init_server(serv, serv)` call from scheduler requires some kind of change?

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -284,10 +284,16 @@ class Downloader(Thread):
                 retention,
             )
             self.servers.append(server)
+            self.sort_servers()
             self.server_dict[newserver] = server
 
         # Update server-count
         self.server_nr = len(self.servers)
+
+    def sort_servers(self):
+        """ Sort by active (active first), priority and id """
+        # Active flag is switched (True = 0, False = 1) for sorting
+        self.servers.sort(key=lambda srv: "%d%02d%s" % ((srv.active + 1) % 2, srv.priority, srv.id))
 
     def add_socket(self, fileno: int, nw: NewsWrapper):
         """ Add a socket ready to be used to the list """
@@ -387,10 +393,7 @@ class Downloader(Thread):
         """Return True when this server has the highest priority of the active ones
         0 is the highest priority
         """
-        for server in self.servers:
-            if server is not me and server.active and server.priority < me.priority:
-                return False
-        return True
+        return me.priority == self.servers[0].priority
 
     def nzo_servers(self, nzo):
         return list(filter(nzo.server_in_try_list, self.servers))
@@ -408,6 +411,7 @@ class Downloader(Thread):
             # Not fully the same as the code below for optional servers
             server.bad_cons = 0
             server.active = False
+            self.sort_servers()
             self.plan_server(server, _PENALTY_TIMEOUT)
 
         # Optional and active server had too many problems.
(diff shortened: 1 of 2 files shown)


sabnzbd/downloader.py
```diff
@@ -387,10 +393,7 @@ def highest_server(self, me: Server):
         """Return True when this server has the highest priority of the active ones
         0 is the highest priority
         """
-        for server in self.servers:
-            if server is not me and server.active and server.priority < me.priority:
-                return False
-        return True
+        return me.priority == self.servers[0].priority
```
Safihre, line 396: I don't fully get it, why just comparing to the first one is enough?
Does that work when there's 3 levels of priority?

sabnzbd/nzbstuff.py
```diff
@@ -197,17 +197,18 @@ def get_article(self, server: Server, servers: List[Server]):
...
+                            break
+                        if not self.server_in_try_list(server_check):
+                            if log:
+                                logging.debug(
+                                    "Article %s | Server: %s | setting found priority to %s",
+                                    self.article,
+                                    server.host,
+                                    server_check.priority,
+                                )
+                            found_priority = server_check.priority
```
Safihre, line 210: Wouldn't be possible to maybe write unit tests? If we could do that we can test all kinds of situations.
Should be possible to construct list of Server's and see how the Article proceeds depending on their settings.

sabnzbd/downloader.py
puzzledsab: The way I understand this function, it should return false if there is a higher level server. It doesn't matter how much higher. It is used for stopping lower level servers from running ahead of the top level servers and loading articles from disk that they won't be allowed to try to download.

If it is a top level server and the list is unsorted it has to go through the whole list to know that there are no higher servers. When it's sorted, a top level server will be placed in index 0, so it must have the same priority as that one. If the priority number is different, it can't be a top level server. It doesn't matter how many levels there are because all the servers that aren't top level should return false.

sabnzbd/nzbstuff.py
puzzledsab: Probably, but I have no knowledge of how to do it. Are there any existing tests that does a similar thing that I could look at?

sabnzbd/downloader.py
Safihre: You are right, I forgot what this function does. Just postpone the loading of articles to at least wait when we tried the first level servers.

sabnzbd/nzbstuff.py
Safihre: No, it would have to be made from scratch.

Safihre:
It will probably require locking, see the documentation:

> CPython implementation detail: While a list is being sorted, the effect of attempting to mutate, or even inspect, the list is undefined. The C implementation of Python makes the list appear empty for the duration, and raises ValueError if it can detect that the list has been mutated during a sort.

Because I think the list is read from outside the Downloader loop, or maybe I'm mistaken..

puzzledsab:
After a lot of testing I've found that there is probably nothing to be gained from this change. The affected loops are called too rarely and rewriting it to be thread safe is messy.

Safihre:
Too bad, because indeed it looked more elegant.. But thread safety is a killer indeed..

Issue #973, opened 2017-07-23
Write article database back to disk when not used


Safihre:
To preserve memory, we should write the article database back to disk when it's not used for a while. 
For example when all jobs in the queue are out of retention of the first server, it will have loaded all articles into memory even though the second server might take ages to reach it. 

- [x] `TryList` can currently not be pickled, it has `Server` objects which in  turn have socket objects. 
- [ ]  Decide when and how to trigger writing the objects back to disk. 
- [ ]  Implement it using clean code.

Safihre:
@puzzledsab this could maybe be interesting one? Seems you like these bigger core problems :)

puzzledsab:
I've been looking into it on and off for a few days already :D

I have some hunches about timestamps on the nzos and swapping out the articles on the nzos that were most recently added, then loading them all back if the nzo is accessed. Maybe you already have better ideas? Problem is I still don't know enough Python or understand SAB well enough to know if it's doable in practice or if there are major show stoppers. I'm digging, though.

I was trying to make the pause on full disk system use the actual complete directory of the nzo instead of the default dir, in order to only pause if the disk it will actually unpack to is full. It seemed simple enough and I even made a more generic caching OO replacement for diskspace(). However, digging through the code I realized that building the path was such a complex operation that it would use a bit much CPU to do it for every file. It can be cached, but then it needs to be cleared if it changes.

Safihre:
Yes, indeed a timestamp on the nzo is also what I was thinking. This can then be checked during a new check in the periodic checks in the main loop in SABnzbd.py for example. Has to be locked with `NZO_LOCK` probably, to avoid competing access.
That's another thing that could be improved, how to use a NZO specific lock instead of 1 lock for all NZO's. 

Feel free to send in draft PR's or send me an email at safihre@sabnzbd.org if you have any questions. I can already recommend PyCharm as an IDE, it opened up a whole new world for me when I started using that, especially the Debugger is amazing for SABnzbd. The community edition is free and has 99% of what you'd need. I have extra PyCharm Professional licenses that we get for free as a bigger open source project, in case you'd need it! 

Interested to see what OO version you made for diskspace, this way I don't have to do it! 😄

puzzledsab:
I've experimented a bit, and it seems that the pickled files and file.articles can be compressed by 60-80%. Maybe we could read the nzb, compress it immediately, attach it to the parent object, and decompress it on the fly when or just before it's needed?

[...]

PR #1758, opened 2021-01-20
Repickle loaded articles if the first server couldn't download them and the next is far behind

puzzledsab:
I'm sure there are problems but I'm unable to see them so I could use some feedback. It significantly reduces the memory usage if you test by adding a server with 1 day retention and a bunch of nzbs that are older than that. The marking of articles as seen is slowed down so that the other servers can start downloading and the pickler can keep up.

For normal usage I think it's the same as before.

Regarding previous comments:
I don't think it's possible for pickling and unpickling to happen simultaneously. At least that is the intention. The pickler won't touch files with the imported_finished flag = False and the unpickler won't touch them if it's True. Those flags are only reversed when the process is done. The unpickler also sets the last_used flag before it starts so the pickler won't touch it for 20 seconds. 

I have set the buffer of unpickled articles to 3 GB as the memory cost is negligible. I have also set a limit for how many files can be pickled before nzbqueue.pickle returns and the NzbQueueLocker is released. If it wasn't finished it will continue after 0.01s.

Issue #973


Safihre:
While I like this change a lot, I would like to postpone it for SABnzbd 3.3.0.
Since 3.2.0 is already filled with a lot of new features :)
So I will look at it in more depth and suggest some changes, but will wait a bit before merging it.

Safihre:
Let me know when you think it's ready and I should really take an in-depth look at it!

puzzledsab:
@Safihre I don't think there's much use in more nit picking before you take a look at it.

The search for articles that should be pickled or unpickled now runs every 3 seconds. Under normal circumstances it uses very little time. The biggest exception is if the queue is big, a server which too short retention unlocks all the nzos by setting `last_used`, and a servers is removed. That causes all the try_lists to be reset. Because `last_used` is set, the lower level servers will unpickle the entire list and add themselves to the try_list again. 

I added `pickle_lock_time` to the nzf in addition to last_used. It makes sure the articles aren't pickled again before the servers has had time to fetch an article and set last_used.

The unpickler has its own thread for several reasons.
- In case of the try_list reset described above, it lets all servers run ahead and unpickle all files. This would take forever if one server at a time does it, the whole list behind it is pickled while it does it, and then next server does the same thing. In the mean time the download thread is locked.
- It lets the downloader thread run even while unpickling. On my PC it can loop at least 4 times in the time it takes to unpickle one file.
- To avoid double unpicking with the pre-unpickler in the pickle function below.

nzbqueue.py:pickle will also unpickle files that will soon be needed. It was easy to add, and in most circumstances it will let the download thread run without having to wait for unpickling.

On very slow connections, less than one article may be fetched every 20 seconds. I added a check to see if the first article left in a file has fetcher set.

The `last_used` and `pickle_lock_time` time stamps are what is supposed to stop the pickler for pickling active articles. It has to wait at least 10 seconds before pickling a file that has just been unpickled. This is enough time to fetch an article and set last_used, or add itself to the try_list.

There is still a theoretical chance that get_article can somehow arrive at an unpickled file just when those 10 seconds are up and it is pickled before the check is complete. It would have to be a combination of very special circumstances because normally it would also be protected by the 2.5 GB buffer of unpickled data ahead of the file currently being downloaded. In the case where lower priority servers are running ahead they would add themselves to the try_list within a fraction of a second. Maybe it could happen if one server was delayed for just the right time before it could start catching up with the others.

Safihre:
Finally had time to take a look at this 🙂
I see the problem we have here: if you make it into a thread, there are these checks needed to prevent concurrent problems. Simpler would be to just add it to the main thread, but that could result in unwanted blocking. Also undesired.
The current code in this PR adds a whole layer of complexity that will be very hard to maintain in the coming years...

puzzledsab:
Yes, it is a bit big for solving a problem most users don't have. Fixing the XML leak is more important. I'll keep the branch for later, maybe some of it will be needed. I did notice that it can take 0.02s to load the articles for one file, even on SSD, so unpickling in a separate thread could help maximize the download speed as well.

Issue #1703, opened 2020-12-13
Test the adding of NZB's more extensively 


Safihre:
One of the things that isn't covered, but has given me troubles over the past few years is the adding of NZB's and all the checks performed there.
In particular selecting the right priority if duplicate detection and/or pre-queue scripts are active.
And the different situations based on what is set by the category and manual input. 
@jcfp you did some extensive api testing recently, can we easily add this as a functional set of tests? Or should it be unit tests?

jcfp:
Functional tests I'd say, as this is mainly about interaction between various choices/settings/input. Looks doable, afaict all relevant actions and settings (adding nzbs, scripts, duplicate-related stuff, categories) are available via the api. Could give it a go myself if this isn't too urgent.

Any scenarios or combinations of settings that are particularly troublesome?

Safihre:
I was hoping you'd have time, indeed no rush!
The combination of supplying either `-100` (default) and having a Pre queue script that sets a priority.
Then a category set from Pre queue that has a deviating priority from Default. And maybe what happens if on top of that a non-Default priority was supplied when adding the NZB, I think it should override the others, but maybe it doesn't. 😅

jcfp:
![images](https://user-images.githubusercontent.com/6666432/103505778-5c274500-4e5b-11eb-811f-918815fdbd37.png)

jcfp:
Got some work done on this one, which resulted in a couple of questions:
- Should a priority set upon adding the job ever get precendence over a later priority assignment?
- When setting a category from pre-q without explicitly setting a priority, should the job inherit the prio from that category?
- When setting a priority intended to modify the job state (such as Paused); what prio is supposed to be used after the job state gets set? Category-based, Default, always Normal, whatever is was before it was set to Paused, something else...?
- What is -100 (default) supposed to refer to; should it take into account category and/or any previous priority assignments?

Safihre:
- I think a priority-set-when-adding should always have precedence.
- Yes.
- What it was before it was paused, so most likely category based or something that came out of the pre-queue.
- -100 should just take the category one or the pre-queue'd one.

At least that's what I think, what do you think?

jcfp:
I agree for the most part, with the exception of priority-on-adding for which I don't see a good reason to let that override later (attempted) changes to the priority. But since you specifically mentioned that one in a comment and I didn't see it getting special treatment in an early test setup, I figured I better ask anyway.

[...]

PR #1760, opened 2021-01-21
Test adding nzbs

jcfp:
Fixed:
* category-based values from pre-queue didn't work at all, even if no explicit settings were given at any stage;
* explicit settings upon adding the job weren't given precedence over implicit values for pp, script, prio based on the category set by pre-queue;
* state-setting priorities (paused, dup) always used Normal priority as their fallback, instead of whatever priority (except Force which could override the state immediately after it was set) the job would have had based on its category;

For the tests I went with two separate test functions. The first is a reasonably-sized random sample, since a full run over multiple layers of parametrisation ends up too big for CI; the second deals with known issues using fixed sets of parameters.

To prevent configuration changes in functional tests from interfering with unrelated tests elsewhere, the `run_sabnzbd_sabnews_and_selenium` fixture was split and the scope of the new `run_sabnzbd` set to 'module'; the _sabnews_and_selenium_ part remains session-scoped. In the process, bugs surfaced in `tests/test_utils/{test_cert_gen,test_diskspeed,test_internetspeed}.py`, all of which were assuming the cache dir always exists.

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -2350,11 +2350,21 @@ def pre_queue(nzo: NzbObject, pp, cat):
         ret = p.wait()
         logging.info("Pre-queue script returns %s and output=\n%s", ret, output)
         if ret == 0:
+            try:
+                # Extract category line from pre-queue output
+                pre_queue_category = output.splitlines()[3].strip("\r\n '\"")
+            except IndexError:
+                pre_queue_category = None
             n = 0
             for line in output.split("\n"):
                 line = line.strip("\r\n '\"")
-                if n < len(values) and line:
-                    values[n] = line
+                if n < len(values):
+                    if line:
+                        values[n] = line
+                    elif pre_queue_category and n in (2, 4, 5):
+                        # Preserve empty pp, script, and priority lines to prevent
+                        # pre-existing values from overriding cat.-based settings
+                        values[n] = ""
                 n += 1
         accept = int_conv(values[0])
         if accept < 1:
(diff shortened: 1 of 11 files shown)


thezoggy:
can you also add adding nzb that should result in various failure states: empty nzb / malform nzb / valid nzb but file gets removed a second later (simulate flakey hdd)

jcfp:
> can you also add adding nzb that should result in various failure states: empty nzb / malform nzb / valid nzb but file gets removed a second later (simulate flakey hdd)

Testing the handling of nzb metadata was already on the (embedded) todo list; adding empty and malformed nzbs shouldn't be a problem. I guess the semi-broken hdd would result in only part of the file getting read, probably best simulated by cutting a valid nzb in two and feeding one of the pieces to sab?

thezoggy:
The flakey hdd is somewhat of a unknown, since usually sab only really checks at the start and if the file looks like it's fine.. so that first half might yield a partial download? The back half I would think it would fail to be added?

jcfp:
I expect sab won't accept either part, because these aren't correctly formed xml files in their own right.

jcfp:
Also added tests with dupe detection set to pause, which seems to work pretty much as expected: always labels, and pauses the job unless priority is set to Force. The changes should cover everything mentioned in #1703 now.

Safihre:
Great :)

I will wait to merge this after we split off the 3.2.x branch somewhere next week. There's already so much in the 3.2.0 release that I would like to put this in 3.3.0!

Safihre:
Thank you!

Issue #1759, opened 2021-01-20
Make "Use global interface settings" save all the options


puzzledsab:
I saw #1475, but it's a bit unclear if it's like that because that's what you want or because you don't think it's important. Is it ok if I try to change it so that everything on that page is persistent if "Use global interface settings" is chosen?

Safihre:
As mentioned it's like that to be able to have different settings on your tablet than on your desktop for example. But technically they could just all be remembered of course, although I would like maybe to store the other settings as one json object instead of seperate settings for each.

puzzledsab:
Done.

Issue #1475, opened 2020-05-29
[FR] Make web interface settings persistent


reloxx13:
Currently after cookie deleting or using another pc/browser, all my web interface settings are resseted to default, which is really annoying (items shown, tab view, compact view etc. ) .

I see no reason why someone would like diffrent settings per browser?!

Suggestion: Add global persistent web interface settings.

thezoggy:
we did this because people would access sab via their phone/desktop/etc and each place people wanted to customize what they saw because one layout or size doesnt fit all. (gotta realize that sab has been around for decade+.. well before the era of everything has an app)

reloxx13:
Your decision.

you r right, i use nzb360 app 😆 
Todays mobile displays also have a higher resolution and showing more is possible.

Safihre:
Indeed after testing on desktop, tablet and phone we decided not to make settings persistant. Only some settings are, like number of rows and refresh rate.

reloxx13:
What about a possibility to change the default settings?

What is "die globalen Interface-Einstellungen verwenden" (checkbox use the global settings).
I dont see that it does anything.

Safihre:
The top 3 settings (number of items and refresh rate) are stored on the server side, so persistant. If that box is checked.
Not very clear, I agree 🤣

PR #1761, opened 2021-01-21
Save all interface values if useGlobalOptions is true

puzzledsab:
This seemed a bit to easy so it feels like I may have missed something. I have tested it back and forth in Edge and Firefox and I can't find any problems.

Issue #1759 and #1475


sabnzbd/cfg.py
```diff
@@ -85,6 +85,7 @@ def validate_server(value):
 queue_complete_pers = OptionBool("misc", "queue_complete_pers", False)
 bandwidth_perc = OptionNumber("misc", "bandwidth_perc", 100, 0, 100)
 refresh_rate = OptionNumber("misc", "refresh_rate", 0)
+interface_settings = OptionStr("misc", "interface_settings", "'{}'")
```
Safihre, line 88: Is the extra `'` needed?

Safihre (changes requested):
You'll need to update the tavern API tests since the output changed 🥇

sabnzbd/cfg.py
puzzledsab: I don't know. They're there when it saves because there are two layers of stringifying so I added them just in case.

puzzledsab:
"'{}'" isn't '!anystr', apparently. I haven't figured out how to make this particular test run locally so I can't test it myself. Do you think !re_match "'\{.*\}'" or !re_fullmatch "'\{.*\}'" will work?

sabnzbd/cfg.py
Safihre: When it saves to `sabnzbd.ini`? Because the ConfigParser adds those on its own.
Does it work without the `'`?

puzzledsab: The page loads just fine with no default but when new values are stored it is in the format `'{"dateFormat":"fromNow",`, so for normal usage it is there.

jcfp:
> "'{}'" isn't '!anystr', apparently. I haven't figured out how to make this particular test run locally so I can't test it myself. Do you think `!re_match "'\{.*\}'"` or `!re_fullmatch "'\{.*\}'"` will work?

Tavern uses { and } as variable delimiters that can be escaped by doubling. Unfortunately, it keeps trying to interprete them as vars after finding a regex match, so it looks like that's a tavern bug. As a workaround, you could set strict to off/False (separate setting for json and xml, around line 5 resp. 145 in data/tavern/api_queue_format.yaml) and remove the new var from the expected output altogether for the time being.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -629,6 +629,37 @@ function ViewModel() {
         }
     })
 
+    // Save the rest in config if global-settings
+    var saveInterfaceSettings = function(newValue) {
+        if(self.useGlobalOptions()) {
+            var interfaceSettings = {
+                "dateFormat": self.dateFormat,
+                "extraQueueColumns": self.extraQueueColumns,
+                "extraHistoryColumns": self.extraHistoryColumns,
```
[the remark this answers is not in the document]
puzzledsab: Yes, the ko stringifier figures it all out :)

Safihre (approved):

Issue #1455, opened 2020-05-20
Feature Request: add notification when provider accounts are due to expire?


OneCDOnly:
Recently noticed a new addition to NZBHydra which allows the user to enter a VIP account expiry date for each indexer. NZBHydra then notifies you before this date so you'll remember to renew your indexer account.

It would be great if SABnzbd could do this with Usenet providers.

For example, I tell SABnzbd the date my account with XS News is due to expire. Then 1 week before this date, SABnzbd notifies me via a warning in its log. I can then decide whether to renew with XS News or not. This also gives me time to cancel an auto-renewing subscription.

Thank you.

thezoggy:
Xsnews already sends out email reminders that it's going to expire.

OneCDOnly:
Further activity has just occurred. :+1:

reloxx13:
> 
> 
> Xsnews already sends out email reminders that it's going to expire.

there are still other providers and it would be also nice to have a overview/see all at once in one central place of all expire dates. sab could just to this on the server config page.

Safihre:
Anyone is free to add this and open a pull request, I can give pointers where it should be added in the code!

sanderjo:
I don't like secondary / shadow administrations. Better to rely on one administration; I would prefer the provider tells it on NNTP login, so instead of a plain 281 like

```
      [C] AUTHINFO USER fred
      [S] 381 Enter passphrase
      [C] AUTHINFO PASS flintstone
      [S] 281 Authentication accepted
```

a 281 with additional info:

```
      [C] AUTHINFO USER fred
      [S] 381 Enter passphrase
      [C] AUTHINFO PASS flintstone
      [S] 281 Authentication accepted - Valid until 2020-07-07
```

or in case of a block account:

```
      [C] AUTHINFO USER fred
      [S] 381 Enter passphrase
      [C] AUTHINFO PASS flintstone
      [S] 281 Authentication accepted - 2543 GB left
```

If NNTP providers would do that, SAB could provide that info

OneCDOnly:
@sanderjo that looks like a great idea. :+1: 

Any thoughts on how to request this from the providers?

OneCDOnly:
Go-away stale bot.

reloxx13:
> 
> 
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

nuuuuu

puzzledsab:
[...]

PR #1762, opened 2021-01-22
Gives a warning when a server has downloaded a certain amount of bytes or a given date is reached

puzzledsab:
I tried to read the download data and old server data directly in ConfigServer/set_dict instead of using those hidden input fields but I couldn't make it work. It seemed like it tried to call that function before the data were available or something like that. If it's possible to do it in some way then please tell me how.

For instance,  when I call `sabnzbd.BPSMeter.grand_total[self.__name]` it would say 
`AttributeError: module 'sabnzbd' has no attribute 'BPSMeter'`

When I try to add `import sabnzbd.bpsmeter` somewhere I get
`ImportError: cannot import name 'OptionBool' from partially initialized module 'sabnzbd.config' (most likely due to a circular import) (sabnzbd/sabnzbd/config.py)`

Sorry, I'm still a Python newbie.

Issue #1455


interfaces/Config/templates/config_server.tmpl
```diff
@@ -223,6 +234,19 @@
                             <label class="config" for="displayname$cur">$T('srv-displayname')</label>
                             <input type="text" name="displayname" id="displayname$cur" value="$server['displayname']" />
                         </div>
+                        <div class="field-pair advanced-settings">
+                            <label class="config" for="expire_date">$T('srv-expire_date')</label>
+                            <input type="text" name="expire_date" id="expire_date$cur"  value="$server['expire_date']" />
```
Safihre, line 239: Let's use `type=date`.

sabnzbd/config.py
```diff
@@ -435,10 +438,17 @@ def set_dict(self, values: Dict[str, Any]):
...
+            "expire_date",
+            "quota_left",
+            "usage_at_start",
             "priority",
             "notes",
         ):
             try:
+                if kw == "usage_at_start":
+                    usage_at_start_new = values.get("usage_at_start_new", "")
+                    if usage_at_start_new and values["quota_left"] != values["quota_left_old"]:
```
Safihre, line 450: I'm not sure I understand what is happening here?

SABnzbd.py
```diff
@@ -1539,6 +1539,23 @@ def main():
 
         # 30 sec polling tasks
         if not timer % 10:
+            servers = config.get_servers()
+            for srv in config.get_servers():
+                server = servers[srv]
+                if server.expire_date():
+                    if time.time() > time.mktime(time.strptime(server.expire_date(), "%Y-%m-%d")):
```
Safihre, line 1546: I doubt there is a need to check this every 30 seconds. The date check should be done at program start and using the scheduler at midnight. We already do that for the general quoata check.

Although I wonder what would be the point of such a notification? Once your server expires you'll get notified anyway by the errors/warnings that the server is ignored due to errors. Usually the newsserver reports that the quota is exceeded and SABnzbd will report that.

SABnzbd.py
```diff
@@ -1539,6 +1539,23 @@ def main():
...
+                        logging.warning("Server %s is expiring", server.displayname())
+                        server.expire_date.set("")
+                        config.save_config()
+                if server.quota_left():
+                    if (
+                        server.quota_left.get_float() + server.usage_at_start.get_float()
+                        < sabnzbd.BPSMeter.grand_total[srv]
+                    ):
+                        logging.warning("Server %s has used the spcified quota", server.displayname())
```
Safihre, line 1555: I would have expected the code to be at the same spot as the regular quota check in the BPSMeter?

interfaces/Config/templates/config_server.tmpl
puzzledsab: Well, blush, I wasn't even aware of that type.

sabnzbd/config.py
puzzledsab: When the user changes the value (quota_left), the starting point for the usage calculation is changed to the current number of used bytes for that server and stored in usage_at_start. That way it's always relative to when the user added the quota and the quota doesn't have to be updated every time new bytes are downloaded.

SABnzbd.py
puzzledsab: That was just for testing, I forgot to change it. I intended it to be on the 5 minute mark, or create an even less regular one. It seemed simpler to keep the two tests together and only at one place in the code, and because it's done rarely the performance impact should be minimal.

Both are intended as soft early warnings, so you can renew or get a different subscription before they expire. If the main, usually time based, subscription expires you can end up using a lot of the data from the block based backup servers. On a backup server where you have 500 GB, you can add a warning when has used 400 GB so you can buy another block.

SABnzbd.py
puzzledsab: I wanted the performance impact to be minimal because all that happens is a warning in the program. Unless the user is monitoring it at the moment it happens, sending the warning at exactly the right moment or 5 minutes later makes no difference.

SABnzbd.py
Safihre: This spot just in `SABnzbd.py ` isn't intended for these checks. It should only be used for program-essential stuff.
I think we need to move this to the scheduler.

interfaces/Config/templates/config_server.tmpl
```diff
@@ -223,6 +234,19 @@
...
+                            <label class="config" for="expire_date">$T('srv-expire_date')</label>
+                            <input type="date" name="expire_date" id="expire_date$cur"  value="$server['expire_date']" />
+                            <span class="desc">$T('srv-explain-expire_date')</span>
+                        </div>
+                        <div class="field-pair advanced-settings">
+                            <label class="config" for="quota_left">$T('srv-quota_left')</label>
+                            <input type="text" name="quota_left" id="quota_left$cur" value="$server['quota_left']" />
+                            <input type="hidden" name="quota_left_old" id="quota_left_old$cur" value="$server['quota_left']" />
+                            <input type="hidden" name="usage_at_start" id="usage_at_start$cur" value="$server['usage_at_start']" />
+                            <input type="hidden" name="usage_at_start_new" id="usage_at_start_new$cur"  value="<!--#if 'amounts' in $server#-->$(server['amounts'][0])<!--#end if#-->" />
```
Safihre, line 247: I checked out the code and still don't understand why there need to be these 3 fields?

Safihre: Or should these be hidden?

sabnzbd/skintext.py
```diff
@@ -624,6 +624,14 @@
     "srv-password": TT("Password"),  #: Server password
     "srv-timeout": TT("Timeout"),  #: Server timeout
     "srv-connections": TT("Connections"),  #: Server: amount of connections
+    "srv-expire_date": TT("Account expire date"),
+    "srv-explain-expire_date": TT(
+        "Give warning on this date. Set a day or two before expiry to get an early warning. Format: YYYY-MM-DD"
```
Safihre, line 629: Since we use the date-input, it is no longer needed to specify the format.

sabnzbd/skintext.py
```diff
@@ -624,6 +624,14 @@
...
+    ),
+    "srv-quota_left": TT("Quota left on account"),
```
Safihre, line 631: If you use `Quota left` then we already have all the translations.

sabnzbd/config.py
Safihre: Aah now I understand 👍

sabnzbd/config.py
```diff
@@ -435,10 +438,17 @@ def set_dict(self, values: Dict[str, Any]):
...
+                        values["usage_at_start"] = usage_at_start_new
```
Safihre, line 451: I suggest we move this out of the `for` loop and not rely on information stored inside the HTML but just do the logic here if we need to update. It can be an if-statement below the `if not self.displayname():` one.

Safihre:
@sanderjo and @puzzledsab I see sometimes extra Black commits, I created a `pre-commit` git hook to just fix this for me.
All you have to do is create a file `pre-commit` in the folder `your-sab-git-folder\.git\hooks`.
And inside you put for Windows:
```
C:\\full_path_to\\python.exe -m black SABnzbd.py sabnzbd scripts tools builder builder\\SABnzbd.spec tests --line-length=120 --target-version=py36
```
And I guess for Linux:
```
python -m black SABnzbd.py sabnzbd scripts tools builder builder\\SABnzbd.spec tests --line-length=120 --target-version=py36
```
This will automatically run black after you commit something. So then you can append the black changes to the commit itself.

interfaces/Config/templates/config_server.tmpl
puzzledsab: The warning should be sent if `quota_left` (user input) + `usage_at_start` (what the BPSMeter grand_total is when the user enters quota_left) > `grand_total` (whenever the check is run).

If the user does a change to the server but the `quota_left` is not changed, the old `usage_at_start` should be kept. To know that the user has not changed quota_left, we need to know what it was before the save (unless there's another way to detect it). Ideally that could be done by loading the old server data before the new ones are saved, but I couldn't figure out how to do that. That's why I store it in the form as `quota_left_old`.

If the user has changed `quota_left`, we need replace usage_at_start with new data from BPSMeter grand_total. I couldn't figure out how to get that variable either in set_dict, so it's sent from the form as usage_at_start_new.

Yes, all three fields should be hidden.

puzzledsab:
Thanks, we'll soon find out if I added it correctly.

sabnzbd/config.py
puzzledsab: I first tried this:
```
servers = get_servers()
server = servers[self.__name]
logging.debug("quota_left_old: %s, from server: %s", values.get("quota_left_old", ""), server.quota_left())
```
When I do that, it crashes immediately:
```
$ python SABnzbd.py -f sabnzbd.ini -l 2 -b 0
  File "SABnzbd.py", line 1742, in <module>
    main()
  File "SABnzbd.py", line 995, in main
    res, msg = config.read_config(inifile)
  File "sabnzbd\sabnzbd\decorators.py", line 36, in call_func
    return f(*args, **kw)
  File "sabnzbd\sabnzbd\config.py", line 767, in read_config
    return _read_config(path)
  File "sabnzbd\sabnzbd\config.py", line 839, in _read_config
    ConfigServer(server.replace("{", "[").replace("}", "]"), CFG["servers"][server])
  File "sabnzbd\sabnzbd\config.py", line 421, in __init__
    self.set_dict(values)
  File "sabnzbd\sabnzbd\config.py", line 429, in set_dict
    logging.debug("quota_left_old: %s, from server: %s", values.get("quota_left_old", ""), server.quota_left())
AttributeError: 'dict' object has no attribute 'quota_left'
```
I tried `server.get("quota_left", "")` instead, but then I get `AttributeError: 'ConfigServer' object has no attribute 'get'` when I try to save. If you know how to solve this I can remove `quota_left_old` and `usage_at_start` from the form.

If you know how to get `sabnzbd.BPSMeter.grand_total` from the same place I can remove `usage_at_start_new` as well.

puzzledsab: Never mind, I figured it out. I assume it's too late to do it after the loop because then the old `quota_left` and `usage_at_start` have been replaced by the new values in the server object.

sanderjo:
> @sanderjo and @puzzledsab I see sometimes extra Black commits, I created a `pre-commit` git hook to just fix this for me.
> All you have to do is create a file `pre-commit` in the folder `your-sab-git-folder\.git\hooks`.

Ah, cool, I was looking for something like that.

```
sander@witte2004:~/git/sjo-sab$ cat .git/hooks/pre-commit
python3 -m black SABnzbd.py sabnzbd scripts tools builder builder/SABnzbd.spec tests --line-length=120 --target-version=py36
```
So: `python3` and a slash not a blackslash. And the script must have x bit set.

... and it works, also with commit from pycharm! Cool. Thanks.

Safihre:
@puzzledsab I pushed some changes, because it was easier to fix than to explain!
I moved some stuff around and also changed the timing of the warning to be 5 days in advance, daily.
I was thinking we could do something similar for the quota, for example warn also when 90% of quota is used. But then it became a whole mess to keep track if we already warned or not, since it's checked every 10 min.
Let me know what you think of the changes!

sabnzbd/downloader.py
```diff
@@ -1014,3 +1018,29 @@ def clues_pay(text: str) -> bool:
...
+
+def check_server_expiration():
+    """Check if user should get warning about server date expiration"""
+    for server in config.get_servers().values():
+        if server.expire_date():
+            days_to_expire = ceil(
+                (time.mktime(time.strptime(server.expire_date(), "%Y-%m-%d")) - time.time()) / (60 * 60 * 24)
+            )
+            # Notify from 5 days in advance
+            if 0 <= days_to_expire < 6:
```
puzzledsab, line 1031: I would skip the 0 <= so there is at least one warning, even if SAB is only started after it has expired. I think this feature is particularly useful for free xsusenet, which most users probably don't use as a primary server. There won't necessarily be a warning from the downloader for a while.

puzzledsab:
Except for that I think it's all good.

sabnzbd/downloader.py
Safihre: Done

Safihre:
I thought it would also be quite usefull to see the amount of quota left, so added that 🥇

PR #1764, opened 2021-01-24
Badssl: show ip address of bad NNTPS server

sanderjo:
Triggered by https://forums.sabnzbd.org/viewtopic.php?p=123401#p123401 (so: one of the N newsserver had a bad certificate), this PR prints the IP address of the server that was tried.

It works ... for English, but not in translations, because I changed the `"Server %s uses an untrusted certificate [%s]"` format to `"Server %s (%s) uses an untrusted certificate [%s]"`

image

@Safihre is it ok the change it in the .po files? It's just an extra `(%s)`, so AFAIK no translation work needed.

Or add it to the following line "%s - %s: %s"  ... less beautiful, but no impact on language files

Or even something else?

```
$ grep -A1  -irn "uses an untrusted certificate"  po/main/nl.po
1063:msgid "Server %s uses an untrusted certificate [%s]"
1064-msgstr "Server %s gebruikt een niet betrouwbaar certificaat [%s]"
```

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -380,7 +380,11 @@ class NNTP:
                 raw_error_str = T("Certificate not valid. This is most probably a server issue.")
 
             # Reformat error and overwrite str-representation
-            error_str = T("Server %s (%s) uses an untrusted certificate [%s]") % (self.nw.server.host, self.sock.getpeername()[0], raw_error_str)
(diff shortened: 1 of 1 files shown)


Safihre:
Solved this by always logging the exact IP we are connecting to in the `DEBUG` logging. This wasn't the case before for servers that only had 1 IP-address.

PR #1765, opened 2021-01-24
Validate input value for scripts

### LINKED ISSUE #1650: change_complete_action api call should verify input

jcfp:
Ensure the input points to a valid script, present in the script dir; refuse to set/run it if doesn't. The check should apply whenever script values are set or changed, and again when the script is eventually run.

Fixes #1650

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -472,7 +472,7 @@ def fix_unix_encoding(folder: str):
                         logging.info("Cannot correct name of %s", os.path.join(root, name))
 
 
-def is_valid_script(basename):
+def is_valid_script(basename: str) -> bool:
     """ Determine if 'basename' is a valid script """
     return basename in sabnzbd.api.list_scripts(default=False, none=False)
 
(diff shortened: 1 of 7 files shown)


sabnzbd/filesystem.py
```diff
@@ -472,12 +472,17 @@ def fix_unix_encoding(folder: str):
                         logging.info("Cannot correct name of %s", os.path.join(root, name))
 
 
+def is_valid_script(basename: str) -> bool:
+    """ Determine if 'basename' is a valid script """
+    return basename in sabnzbd.api.list_scripts(default=False, none=False)
```
Safihre, line 477: Could you maybe move `list_scripts` to `filesystem.py` since it's now not an api-specific function anymore?

Safihre (changes requested):
Thank you! 🎖️ 
Just a small request, looks good otherwise!

Issue #1430, opened 2020-05-08
Right-to-Left Support in SABnzbd


IONILz:
Hi,

Please consider to add support of right-to-left display in SABnzbd. Hebrew is written from right to left.

I can help with testing it.

puzzledsab:
Are you still there, @IONILz ? I did some small tests, and I realize that I have no idea how it should look in rtl-format. I have uploaded an image here: https://imgur.com/a/YmEeNdk

The first is with dir="rtl" attribute and all div right/left floats reversed. The one in the middle is standard ltr, and the one on the bottom is just dir="rtl".

If you could use a paint program to pick out every elements (images, text and numbers) and put them in the order they should be in proper rtl it would be helpful. Remember that it's the order that is important, not that they line up perfectly.

Safihre:
I just checked the statistics, and in the past 1 year only 18 people (0.02%) selected Hebrew during the setup wizard.
So code changes to add RTL-support should not be too invasive, since it's an extremely small userbase.

IONILz:
> 
> 
> Are you still there, @IONILz ? I did some small tests, and I realize that I have no idea how it should look in rtl-format. I have uploaded an image here: https://imgur.com/a/YmEeNdk
> 
> The first is with dir="rtl" attribute and all div right/left floats reversed. The one in the middle is standard ltr, and the one on the bottom is just dir="rtl".
> 
> If you could use a paint program to pick out every elements (images, text and numbers) and put them in the order they should be in proper rtl it would be helpful. Remember that it's the order that is important, not that they line up perfectly.

The third section would be just fine. In case it is possible without too much effort, I made a screenshot for a better result. Again, the third option is good enough and much better than the current state of SABnzbd.

https://imgur.com/a/PzamYhJ

puzzledsab:
@IONILz : You can test it now by going here: https://github.com/sabnzbd/sabnzbd/pull/1776/checks?check_run_id=1808945630
Then select a download from the "Artifacts (5)" drop down located at the right side of the screen.

Also take a look at the popups. That's the file lists in the queue, the history queue, add NZB and the wrench symbol.

IONILz:
I started testing it. On second thought, the "Queue", "Remaining size" and "Free Space" (and the list of working jobs) should be like in the first example. I made a new screenshot and wrote some comment on it. Also if it is possible to add RTL support in the quick wizard and in the settings pages - it would be great :P
[...]

PR #1776, opened 2021-02-01
Right-to-Left support for Glitter and Config

puzzledsab:
#1430

Can be tested by @IONILz if it builds correctly.


sabnzbd/lang.py
```diff
@@ -92,99 +92,100 @@ def list_languages():
...
+    "en": ("English", "English", 0, 0),
+    "eo": ("Esperanto", "Esperanto", 0, 0),
+    "et": ("Estonian", "Eesti", 0, 0),
+    "fo": ("Faroese", "Føroyskt", 0, 0),
+    "fj": ("Fijian", "Vosa Vakaviti", 0, 0),
+    "fi": ("Finnish", "Suomi", 0, 0),
+    "fr": ("French", "Français", 0, 0),
+    "gl": ("Galician", "Galego", 0, 0),
+    "de": ("German", "Deutsch", 0, 0),
+    "he": ("Hebrew", "עִבְרִית‎", 1255, 1),
```
Safihre, line 126: I'd prefer `True`/`False`.

sabnzbd/interface.py
```diff
@@ -384,6 +384,7 @@ def index(self, **kwargs):
             info["categories"] = list_cats(True)
             info["have_rss_defined"] = bool(config.get_rss())
             info["have_watched_dir"] = bool(cfg.dirscan_dir())
+            info["rtl"] = LanguageTable.get(cfg.language(), "en")[3]
```
Safihre, line 387: Let's introduce a function `is_rtl(lang)` in the `sabnzbd.lang` file. Nicer :)

Safihre (changes requested):
If the changes are this simple it would indeed be a good addition!

puzzledsab:
@Safihre : Do you know why the `<!---->` remains from template tags like `<!--#end if#-->` in `include_menu.tmpl`? It doesn't happen in `main.tmpl`. Is it possible to avoid?

thezoggy:
We used it as a tactic to safeguard from bad cheetah template leaking stuff and breaking the webpage.

Safihre:
@puzzledsab @thezoggy I never saw that before.. strange, is that really supposed to be like that?

Safihre:
Oh right, in the `include_` files you can just use `#` without need for `<!--` and `-->`.

Safihre:
I don't want to go quite this far. The number of RTL users is less than 0.5% of users! 
Only basic support should be added, because it adds a maintenance burden for years to come.. That's just not worth it.
Unless there's some major things wrong, this is as far as we can go.

puzzledsab:
@Safihre : Do you mean that any particular changes should be reversed or that I shouldn't add more?

puzzledsab:
@Safihre : I'm not going to do more changes now. Let me know if you think anything should be reverted.

Safihre:
I pushed some changes, shifting most of the stuff to the CSS. 
We can add some more of the needed changes there, if needed.
We can also do the config in the same way, to shift the checkboxes using CSS.
If it's really just the checkboxes, it's fine by me. If it really requires HTML changes I don't want to do it..

puzzledsab:
According to @IONILz it's a clear improvement to add dir=rtl to the config with no other changes, so any CSS changes would be a bonus. It doesn't look so good but the text is probably easier to understand if you read Hebrew.

Safihre:
Alright, I'll try to fix that asap :)

PR #1778, opened 2021-02-01
adding requests for sickGear support

farmerbean:
Hi folks,

I had a bit of a battle with sickGear/sabNZBD compatibility which I finally solved as an issue with requests missing from python venv within sabNZBD. Once I sourced /env and pip3 installed requests, this resolved the issue.

I've included the link to the job runner in the comment inside requirements.txt

Adding this into a PR for future.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,9 @@ portend
 chardet
 notify2
 
+# Support for SickGear https://github.com/SickGear/SickGear/blob/master/autoProcessTV/autoProcessTV.py
+requests>=2.25.0
+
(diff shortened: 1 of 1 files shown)


Safihre:
Sabnzbd only lists the the requirements to make it run, not other applications.
If you need requests, you need to add it to your python environment and not to ours :)

farmerbean:
Cheers @Safihre - digging round in the dark trying to fix Python issues :)

thezoggy:
Sounds like you should add requests to sickgears requirements :)

PR #1782, opened 2021-02-04
fix bonjour with localhost, retire LOCALHOSTS constant

### LINKED ISSUE #1777: disable - bonjour/zeroconf/SSDP
There are plenty of scenarios that people would NOT want to broadcast the fact they are running sabnzbd across their network.
We should offer option to disable 'bonjour/zeroconf/SSDP' broadcast

also upnp in general can be abused if left reachable to the world:
https://blog.cloudflare.com/ssdp-100gbps/

jcfp:
Changes:
* Fixed bonjour starting even though sab is only listening on localhost, see this excerpt from 3.2.0rc1:
  > INFO::[_cplogging:219] [...] ENGINE Serving on http://127.0.0.1:8080
  > DEBUG::[sabnzbdplus:1501] bonjour/zeroconf/SSDP using host: 10.111.246.57
  > DEBUG::[zconfig:81] Try to publish in Bonjour as "myhostname" (10.111.246.57:8080)
* Got rid of the `LOCALHOSTS` constant because it cannot deal with all valid loopback addresses (127.0.0.0/8), and replaced it with a tiny function in misc.py that can;
* Renamed `enable_bonjour` to `enable_broadcast` and made it apply to both bonjour and ssdp;
* Merged test_probablyip.py into test_misc.py, parametrized and expanded those tests and added some for the new functions too.

Maybe the `probablyipv[46]` functions should be renamed to `is_ipv[46]_addr`? Their current name suggests their code involves chance or otherwise delivers uncertain results...

Fixes #1777.


Safihre:
I don't fully understand the changes, what cases were going wrong before?

sanderjo:
> Maybe the `probablyipv[46]` functions should be renamed to `is_ipv[46]_addr`? Their current name suggests their code involves chance or otherwise delivers uncertain results...

The code was indeed a probably thing; it did not check if it was a formally correct IPv4 or IPv6 address.
Assuming `ipaddress.ip_address(ip).version == 4` does formal checks, we could indeed rename like you propose.

sanderjo:
if sabnzbd is only running on 127.0.0.1 port xyz, could it still be useful to broadcast it via bonjour / SSDP, and thus only for the local system? Does Bonjour accept 127.0.0.1:xyz ... ?

thezoggy:
note that we should NOT try to do SSDP to wan. as it can be used for amplification attacks and a well known attack vector

Safihre:
@thezoggy isn't that exactly what @jcfp made sure in this PR?

thezoggy:
> 
> 
> @thezoggy isn't that exactly what @jcfp made sure in this PR?

yep, took a sec to validate .is_loopback is infact rfc of local loopbacks only. so all good from my pov

jcfp:
> I don't fully understand the changes, what cases were going wrong before?

> Does Bonjour accept 127.0.0.1:xyz ... ?

See log excerpt: with sab listening on localhost, bonjour would still announce the service on the lan using an url with the system's lan ip address. There's a check at [zconfig.py#L74](https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/zconfig.py#L74) that notes bonjour+localhost never works and therefore tries to prevent announcing a localhost address, but fails at that because the call from [SABnzbd.py#L1501](https://github.com/sabnzbd/sabnzbd/blob/develop/SABnzbd.py#L1501) incorrectly sets the host to a lan ip (where one cannot actually reach sab).

I suspect (but haven't tested) that the current code also misbehaves on systems with direct external ipv4 connectivity (i.e. no local network, think consumer isp router in bridged mode, internet access via pppoe, hosting/vps, etc.) and sab configured to listen on the external ipv4 address.

The changes in this pr make having sab listening on something other than localhost a requirement for even bothering with any bonjour or ssdp related code, and verify the host used is actually a lan ip address.

sanderjo:
> See log excerpt: with sab listening on localhost, bonjour would still announce the service on the lan using an url with the system's lan ip address. There's a check at [zconfig.py#L74](https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/zconfig.py#L74) that notes bonjour+localhost never works and therefore tries to prevent announcing a localhost address, but fails at that because the call from [SABnzbd.py#L1501](https://github.com/sabnzbd/sabnzbd/blob/develop/SABnzbd.py#L1501) incorrectly sets the host to a lan ip (where one cannot actually reach sab).

Yes, the log shows wrong behaviour. I proposed something else.

sanderjo:
> I suspect (but haven't tested) that the current code also misbehaves on systems with direct external ipv4 connectivity (i.e. no local network, think consumer isp router in bridged mode, internet access via pppoe, hosting/vps, etc.) and sab configured to listen on the external ipv4 address.

Define "misbehave". I have a VPS with public IPv4, and indeed SAB announces itself. Misbehaviour, or correct?

sanderjo:
> The changes in this pr make having sab listening on something other than localhost a requirement for even bothering with any bonjour or ssdp related code, and verify the host used is actually a lan ip address.

Define "lan ip address"? Do you mean a RFC1918 address?

jcfp:
@sanderjo afaik both bonjour and ssdp are intended for announcing services to other system within a local network, as typically found in household or office settings. Is there ever any point of announcing sab to the rest of the internet/data center/all customers of the same isp?

sanderjo:
So ... we only broadcast unto a NAT-ed (RFC1918) LAN, because the other use cases are hypothecial / unwanted (not: wrong).

Yes, OK for me.

jcfp:
> note that we should NOT try to do SSDP to wan. as it can be used for amplification attacks and a well known attack vector

It did look into that as well, but afaict the socket handling in [sabnzbd/util/ssdp.py](https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/utils/ssdp.py#L120) already takes care of that by limiting the ttl on the broadcasts to just 2 hops (rather than binding to a specific address).

PR #1783, opened 2021-02-05
Quick and dirty support for auto night mode switching in Glitter

dotsam:
A quick and dirty css import that uses the prefers-color-scheme media query to only load the Night colour scheme when the UA is in dark mode.


Safihre:
Cooll

thezoggy:
pretty decent support for it also:
https://caniuse.com/prefers-color-scheme

Safihre:
I think this is very cool so I made this the default skin setting for new users. 🎉

PR #1785, opened 2021-02-05
Abort job if more than 50% of data is missing after trying x MB

puzzledsab:
https://forums.sabnzbd.org/viewtopic.php?f=4&t=25183

Sometimes you know the downloads will fail if there are a lot of missing articles early. After having tried `missing_threshold_mbytes` MB it will fail if the number of failed bytes exceeds the number of downloaded bytes.

I don't see any perfect solutions for this and it's a bit hackish so I didn't make a proper config. The threshold will have to be a guess, and if the user thinks it's a false positive they can retry. I'm very open for suggestions on how to improve it.

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -1343,6 +1343,7 @@ SPECIAL_VALUE_LIST = (
     "nomedia_marker",
     "max_url_retries",
     "req_completion_rate",
+    "missing_threshold_mbytes",
     "wait_ext_drive",
     "max_foldername_length",
     "show_sysload",
@@ -1352,7 +1353,6 @@ SPECIAL_VALUE_LIST = (
     "selftest_host",
     "rating_host",
     "ssdp_broadcast_interval",
-    "missing_threshold_mbytes",
 )
 SPECIAL_LIST_LIST = ("rss_odd_titles", "quick_check_ext_ignore", "host_whitelist")
 
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -1352,6 +1352,7 @@ SPECIAL_VALUE_LIST = (
     "selftest_host",
     "rating_host",
     "ssdp_broadcast_interval",
+    "missing_threshold_mbytes",
 )
 SPECIAL_LIST_LIST = ("rss_odd_titles", "quick_check_ext_ignore", "host_whitelist")
 
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -1173,6 +1173,8 @@ class NzbObject(TryList):
                 # Check the availability of these first articles
                 if cfg.fail_hopeless_jobs() and cfg.fast_fail():
                     job_can_succeed = self.check_first_article_availability()
+                    if not job_can_succeed:
+                        abort_reason = "https://sabnzbd.org/not-complete (check_first_article_availability)"
 
         # Remove from file-tracking
         articles_left = nzf.remove_article(article, success)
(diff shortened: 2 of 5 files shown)


thezoggy:
Isn't this already implemented as the % required completion value?

puzzledsab:
I don't think so, but I realize now that I don't understand how req_completion_rate is supposed to be used. Why would you require more or less than 100%?

Either way, I think this can let SAB mark the download as failed much earlier when there are a lot of par2-files.

sabnzbd/nzbstuff.py
```diff
@@ -1172,6 +1173,8 @@ def remove_article(self, article: Article, success: bool):
                 # Check the availability of these first articles
                 if cfg.fail_hopeless_jobs() and cfg.fast_fail():
                     job_can_succeed = self.check_first_article_availability()
+                    if not job_can_succeed:
+                        abort_reason = "https://sabnzbd.org/not-complete (check_first_article_availability)"
```
Safihre, line 1177: I understand you want to specify, but this is not usefull for regular users, let's not modify the text.
No need for `abort_reason`.

sabnzbd/nzbstuff.py
```diff
@@ -1187,14 +1190,22 @@ def remove_article(self, article: Article, success: bool):
 
         # Check if we can succeed when we have missing articles
         # Skip check if retry or first articles already deemed it hopeless
-        if not success and job_can_succeed and not self.reuse and cfg.fail_hopeless_jobs():
-            job_can_succeed, _ = self.check_availability_ratio()
+        if not success and job_can_succeed and not self.reuse:
+            # Abort if more than 50% is missing after reaching missing_threshold_mbytes
+            job_can_succeed = self.check_missing_threshold_mbytes()
```
Safihre, line 1195: This should only be called if `cfg.fail_hopeless_jobs() == True`, like the other checks.

sabnzbd/nzbstuff.py
```diff
@@ -1511,6 +1522,14 @@ def check_first_article_availability(self):
                     return False
         return True
 
+    def check_missing_threshold_mbytes(self):
+        """ Return false if more than 50% is missing after downloading missing_threshold_mbytes MB """
+        if self.bytes_missing > self.bytes_downloaded:
+            missing_threshold = cfg.missing_threshold_mbytes() * MEBI
+            if missing_threshold and self.bytes_tried > missing_threshold:
+                return False
```
Safihre, line 1530: Maybe add `INFO` logging here to notify why the job was aborted.

sabnzbd/interface.py
```diff
@@ -1352,6 +1352,7 @@ def saveSwitches(self, **kwargs):
     "selftest_host",
     "rating_host",
     "ssdp_broadcast_interval",
+    "missing_threshold_mbytes",
```
Safihre, line 1355: Could you move this after the `req_completion_rate`?
Both here and in `cfg.py`.

Safihre (changes requested):

sabnzbd/nzbstuff.py
puzzledsab: The other tests will only do a debug log. How about either a debug or info log for each test, like this?
`logging.debug('Abort job "%s", due to impossibility to complete it (test: check_first_article_availability)', self.final_name)`

puzzledsab:
I kept the old logging but added abort_reason so that all three checks behave the same way.

thezoggy:
this change looks to download 50M of a nzb and basing if its complete based on a % of that 50M? or if more than 50M is missing altogether?  either of these is prone to false failures...

puzzledsab:
More than half of the tried data must be available after reaching the configured limit. Yes, it can cause false positives, but then so can the first articles search.

Safihre:
@puzzledsab I did a small refactor.
You mentioned something about 50% but I couldn't find that anywhere?
I changed it now to trigger if "missing > succeeded and missing > threshold".

Safihre (approved):

puzzledsab:
`self.bytes_missing > self.bytes_downloaded` translates into failing if more than 50% is missing when the threshold is met. I agree that checking missing is better because it can fail faster.

Not sure how useful this is in the end but at least it's a small change with minimal performance impact.

Safihre:
@puzzledsab So should we still add it? I am also not certain if it is usefull..

puzzledsab:
I was hoping to hear back from nismozcar in the thread. I remember having had this problem myself a year ago, but I no longer see it because I haven't downloaded the kind of old postings that have that problem anymore. I think some files have enough of the first articles left not to trigger that check. In those cases it can take a while before it gives up because some servers take a long time to respond when articles are gone.

Issue #1006, opened 2017-08-22
"Maximum retries" is not actually number of retries per server, but per priority level


Safihre:
After digging deep to find the source of stalling issues, I have found that `article.tries` actually counts the number of tries per priority-level and not per-server.
This causes a problem when you have this set to 3 and you have 4 servers on the same priority. 
Additionally there is a problem with resetting the `TryList` when a server times out.

Still searching for a memory and CPU efficient way to keep the correct count.

puzzledsab:
I'm trying to understand this issue and code, but I'm not sure if I do. It seems to me that article.tries will always be increased, no matter if it fails or not? Also, the only place where I can find that it is being used is in downloader.py:
`if article.tries > cfg.max_art_tries() and (article.fetcher.optional or not cfg.max_art_opt())`

Doesn't this mean that there won't be any retries if three other servers have already been tried and __reset_nw is called? If so, I think that may explain why I once had a problem where servers a bit down the server list didn't seem to get checked when the first ones didn't have the article.

thezoggy:
its supposed to try all servers in order of priority, if you have multiple servers with the same priority then the retries are not reset for each server but shared, so you end up in a bad situation when retires are exhausted and it thinks it should be checking more.  

ideally the article.tries just needs to be independently scoped to the server its for rather than global

Safihre:
Damn, this is ages ago.. From looking at the code again it always does `+=1`, even if it is not the same server. 
So say it tries an article on server A, `tries=1`, but when it fails it can go to server B (with same priority as server A) and `tries=2`. 
You would expect this option to "try X amount for **each** server", but it doesn't do that.

puzzledsab:
I assume that if the server is set inactive it should not wait until it becomes available, but instead go on to the next server?

Safihre:
Indeed.

puzzledsab:
Wasn't this fixed when we added the article queue?

Safihre:
Indeed.

PR #1792, opened 2021-02-09
Change article.tries system and read multiple articles from queue

puzzledsab:
Issue #1006. Instead of resetting the `try_list` I have added an `article_queue` list to the server object, which the article is added to. I'm not sure how well this works because I don't know how to test the various retry scenarios properly. One issue I have seen is that if the nzb is deleted while it is being downloaded, the articles currently being downloaded will be retried.

`article_queue` can hopefully also be used to read a few articles in advance from `nzf.get_article()`, which would require a bit less resources than doing the full `NzbQueue.get_article()` every time.


puzzledsab:
I've found a problem that can possibly cause problems with the current version as well. When it's paused it will put the articles being downloaded back in the `article_queue`. I thought `__reset_nw` would stop the downloading so I tried to set `fetcher = None` for the articles in article_queue. However, the downloading can still continue, causing
```
Exception in thread Thread-5:
  File "C:\Users\t\AppData\Local\Programs\Python\Python38\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\gittestbranch\sabnzbd\sabnzbd\downloader.py", line 822, in run
    self.decode(article, nw.data)
  File "C:\gittestbranch\sabnzbd\sabnzbd\downloader.py", line 447, in decode
    sabnzbd.BPSMeter.register_server_article_tried(article.fetcher.id)
AttributeError: 'NoneType' object has no attribute 'id'
```
I haven't found a way to fix this yet, so maybe this PR no closer to a solution.

Anyway, I think this may be the reason for a problem I have seen once in a while when pausing and resuming. Some files will not finish downloading until I restart SAB. I suspect it's because the nw is put back into idle_threads with an unfinished article. It doesn't finish downloading so it's not sent to the decoder and article.fetcher is not reset so it's not picked up again by `get_article`.

puzzledsab:
Ignore the exception, I messed up in my first attempt.

sabnzbd/downloader.py
```diff
@@ -691,7 +710,8 @@ def run(self):
                                     # Don't count this for the tries (max_art_tries) on this server
                                     self.__reset_nw(nw, count_article_try=False, send_quit=True)
                                     self.plan_server(server, _PENALTY_TOOMANY)
-                                    server.threads -= 1
+                                    if server.threads > 1:
+                                        server.threads -= 1
```
Safihre, line 714: Why is this needed? Can it go below 0? That sounds like a problem somewhere else.

Safihre:
Looking at this, I don't want to merge this just for #1006. 
However, it seems this would really be a good base to extend it to fetch `server.threads` number of articles at once.

sabnzbd/downloader.py
puzzledsab: It's to stop it from going below 1, not below 0. It seems strange if 1 connection is too many.

thezoggy:
Seems like we are forcing it everywhere now? Maybe we should have just done force is default true unless explicitly passed not to

puzzledsab:
There doesn't seem to be any point in making the prefetching more complicated than the last update. It reduces the CPU usage of NzbQueue.get_article and NzbObject.get_article to a tiny fraction. It wasn't very high to start with, though. In total this update reduces the total CPU usage by a bit more than 1% in my tests.

I have just realized that the reason why I thought the get_article functions used so much CPU was that I had configured a top priority server with 1 day retention (for testing the pickling). That made it do 33 times as many article.get_article calls as without that server if the nzb is more than 1 day old. This seems a bit excessive, but it's a more limited problem which probably doesn't affect very many users.

Safihre:
>  In total this update reduces the total CPU usage by a bit more than 1% in my tests.

Interesting. I am surprised a bit. I always thought the `get_article` function was slow due to all the checks and called _a lot_. 
What I had in mind was kind of like what you made a long time ago: to not just pre-fetch 1 article, but maybe 5 at once. So also modify the `get_article` function.
But if you say that the time spent in those functions is negligible, it might indeed not be worth it to complicate things.

puzzledsab:
Rewrote to fetch multiple articles from nzf.get_article(). It doesn't matter when they are fetched for the top server because then it will be the next article in line in the nzf, but it's a huge improvement when it's not. When I have a 1 day retention server in front which will be marked as tried immediately it reduces the number of calls to article.get_article from 3.4M to 200K on my test NZB.

puzzledsab:
I realized that fetching a lot of articles if there are many connections may be counter productive so I set it to a fixed limit of 4. Under normal circumstances it will only be marginally faster to fetch 4 new articles in 1 call than to fetch 1 new article 4 times using nzf.get_articles. Fetching too many articles in one call may take enough time that the delay slows down the download speed on fast connections.

puzzledsab:
This version needs too many changes to work with my new pickler and also has some bugs I have fixed in the other branch. https://github.com/puzzledsab/sabnzbd/tree/feature/optimized

PR #1797, opened 2021-02-12
Another take on the pickling

puzzledsab:
New and significantly simpler version. This will only use the download thread and does nothing fancy to improve performance. It will slow down the speed on high speed connections when there is lots of pickling and unpickling going on. On the other hand I changed the server.retention loop to unpickle only one file for each server check. This will let the downloading continue at a reduced speed so it may be experienced as an improvement. All files that are being downloaded or where at least one article have been downloaded are skipped to avoid problems.

#973

There are also a few micro optimizations that have nothing to do with this (is_paused, self.valid and highest_server).


Safihre:
This looks much clearer 👍 
I wonder if we can't simplify things by always using pickles of the articles instead of saving the `raw_article_database` and loading it later.

puzzledsab:
It no longer saves the `raw_article_database`. It's used as a temporary list in RAM while the nzb is parsed and sent as a parameter to `finish_import`. `finish_import` is kept to create the article objects and parse old save files.

puzzledsab:
Not stale, I hope. It conflicts with the article.tries PR, but I'll fix that if the other one is merged.

sabnzbd/nzbstuff.py
```diff
@@ -371,15 +373,72 @@ def __init__(self, date, subject, raw_article_db, file_bytes, nzo):
...
-        """ Load the article objects from disk """
+    def pickle_articles(self, initial_pickle=False):
+        """ Pickle articles to file """
+        if not initial_pickle:
+            if (
+                not self.import_finished
+                # Don't pickle tiny files
+                or len(self.articles) < 3
+                # Don't pickle active files
+                or self.pickle_lock_time > time.time()
```
Safihre, line 390: You will need to set `self.pickle_lock_time = 0` in `__getstate__`, otherwise jobs from before this change will fail as non-existing attributes are set to `None`.
```python
>>> None > 1
  File "<input>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

sabnzbd/nzbstuff.py
```diff
@@ -371,15 +373,72 @@ def __init__(self, date, subject, raw_article_db, file_bytes, nzo):
...
+                or len(self.articles) != len(self.decodetable)
+            ):
+                return False
+            # Don't pickle active files
+            for article in self.articles:
+                if article.fetcher:
+                    return False
+
+        logging.debug("pickle %s", self.filename)
+        first_article = self.decodetable.pop(0)
```
Safihre, line 400: Why do we need to separate the first article again? I know there was a reason for it.

puzzledsab: The completion status check only checks the first article, and it has to always be imported because the import status is not checked before trying to read it.

sabnzbd/nzbstuff.py
puzzledsab: Thanks. this lead me to find a couple of other bugs. The `TypeError` bug would actually have been triggered very rarely because I didn't initialize `nzo.unpickled_files` either, so it would only have tried to pickle them in very special circumstances. Now an existing unpickled queue should be pickled after an upgrade if pickling is enabled.

Safihre:
I changed the setup a bit, where it is now triggered through the `check_all_tasks` interval, so not for every `get_article`. This should be plenty. Maybe we can even move it to the 300sec interval.

Only thing I would like to improve is that it now only works for jobs that haven't been touched due to the `len(self.articles) != len(self.decodetable)` check. It would be great if it could work also for jobs that for example were moved from the top spot all the way down.

puzzledsab:
I think doing it in the existing loop causes minimal overhead when there isn't anything that needs pickling. Combined with the speed limit on marking expired articles read will usually result in only one file having to be pickled or unpickled in each loop because they are spread out evenly in time. This lets SAB continue downloading.

No speed limit and pickling every 30 seconds means it will first unpickle NZBs continuously for up to 30 seconds to mark them tried if they are expired on a server, and then pickle everything in one go. Both will cause the downloading to stop completely, and in the mean time it may use as much RAM as without this patch.

This should work even better when combined with the the other patch (https://github.com/sabnzbd/sabnzbd/pull/1792), which significantly reduces the time it takes to read the remaining articles from one file to mark them tried.

I couldn't find a way to pickle files that have been touched in a way I was sure would be safe. Pickling stuff creates a copy of an object. If a reference to an article is also stored in the cache or elsewhere it will not refer to the same article after pickling and unpickling. Fixing this probably isn't trivial. I think there may also be some issues when it comes to doing it in a thread safe way. Maybe there is a way around it, but I left that for later.

sabnzbd/downloader.py
```diff
@@ -95,7 +95,7 @@ def __init__(
 
         self.busy_threads: List[NewsWrapper] = []
         self.idle_threads: List[NewsWrapper] = []
-        self.next_article_search: float = 0
+        self.next_article_search: float = 0.0
```
puzzledsab, line 98: There is no performance difference between 0 and 0.0.

```$ python -m timeit 'next_article_search: float = 0'
20000000 loops, best of 5: 11.1 nsec per loop

$ python -m timeit 'next_article_search: float = 0.0'
20000000 loops, best of 5: 11.1 nsec per loop```

Safihre: I know, it just looked not right to me to define it float but assign an integer.

Safihre:
> No speed limit and pickling every 30 seconds means it will first unpickle NZBs continuously for up to 30 seconds to mark them tried if they are expired on a server, and then pickle everything in one go. Both will cause the downloading to stop completely, and in the mean time it may use as much RAM as without this patch.

Yes I see your point. Maybe indeed better to have it back in `get_article`. Although, I don't really mind the pause when a job is added that exceeds the retention-time for a server. It's not a standard case for the regular user.

puzzledsab:
Another issue is that once the NZB has been unpickled the first time, all servers are allowed to unpickle it. Lower level servers will do this in order to mark articles temporarily as tried. If you add a big nzb that takes too long to unpickle, it will be pickled again immediately. Then in the next downloader loop it will be unpickled again if there is a low level server that checked the nzb prior to the top level server in the previous loop, and then it will be pickled again after the new time out.

There may be other such problems, which is why I think using it should be optional. Those with enough RAM and fast connections may prefer to keep everything in RAM in order to maximize download speed.

puzzledsab:
I'm now using a different version myself which also pickles partially downloaded files. https://github.com/puzzledsab/sabnzbd/tree/feature/optimized

Safihre:
We made some good progress in these optimizations, but in the end I think the biggest gain could be in the core download process. If only we could move more logic to the non-GIL code in the C-extension.
This doesn't have to be the current C-extension, we can create a new one. And it can also be a Rust-extension, maybe you have experience with that? I don't have experience with either of them 😃 
I always wished to implement SIMD for the decoding, but I just don't know enough of C: https://github.com/animetosho/node-yencode/issues/4

puzzledsab:
I don't know C very well and I've never done any Rust. I have read about the method NZBGet uses but I don't understand it well enough to implement it. Also I'm not sure what else could be written in those languages to make a significant improvement. It might help to make it possible to increase the number of decoder threads from 3. Both cheap ARM CPUs and x86-based CPUs can have more than 4 cores.

PR #1798, opened 2021-02-14
add unwanted extensions whitelist mode

jcfp:
Re: https://forums.sabnzbd.org/viewtopic.php?f=4&p=123658

This changeset adds a mode setting to the unwanted extensions feature. Blacklist mode is the equivalent of the existing unwanted extensions feature and the default, ensuring full compatibility with existing settings. In whitelist mode, only listed extensions are allowed with everything else considered unwanted; useful for those only downloading specific types of content.

Bug fixes:
- ~~always lowercase the extension before comparison to the unwanted extensions list~~ ditch unnecessary calls to lower();
- update the explanatory text to reflect that the unwanted extensions feature isn't limited to the content of rar files.


sanderjo:
> only listed extensions are allowed with everything else considered unwanted

... will (default) pause the download? I look forward to the results. FWIW: I designed Unwanted Extension to detect and halt malicious posts (with .exe and .com) early on. It was not my intention to halt a download if there is something unexpected but harmless (.url, .txt) ... for that, you can use the extension cleanup list. 
Anyway, as said, I look forward to the results.

jcfp:
> will (default) pause the download?

It will use whatever action is configured; "pause" is certainly a good choice to combine with whitelist mode.

sabnzbd/filesystem.py
```diff
@@ -58,6 +58,23 @@ def get_ext(filename: str) -> str:
         return ""
 
 
+def has_unwanted_extension(filename: str) -> bool:
+    """ Determine if a filename has an unwanted extension, given the configured mode """
+    extension = get_ext(filename).replace(".", "").lower()
```
Safihre, line 63: `get_ext` is always lower-cased, I did think of that 😉

sabnzbd/filesystem.py
```diff
@@ -58,6 +58,23 @@ def get_ext(filename: str) -> str:
...
+            # Blacklisted
+            sabnzbd.cfg.unwanted_extensions_mode() == 0
+            and extension in sabnzbd.cfg.unwanted_extensions()
+        ) or (
+            # Not whitelisted
+            sabnzbd.cfg.unwanted_extensions_mode() == 1
+            and extension not in sabnzbd.cfg.unwanted_extensions()
+        )
+    else:
+        return bool(sabnzbd.cfg.unwanted_extensions_mode())
```
Safihre, line 75: I am not sure I get this `else`? What case is it triggered?

sabnzbd/filesystem.py
jcfp: I should have notice, I put the new function right nextdoor...

sabnzbd/filesystem.py
jcfp: No extension in the filename and/or no extensions configured in the unwanted extension setting. In those cases the mode is basically all we have to go with.

sabnzbd/skintext.py
```diff
@@ -495,9 +495,13 @@
     "nodupes-tag": TT("Tag job"),  #: Four way switch for duplicates
     "abort": TT("Abort"),  #: Three way switch for encrypted posts
     "opt-action_on_unwanted_extensions": TT("Action when unwanted extension detected"),
-    "explain-action_on_unwanted_extensions": TT("Action when an unwanted extension is detected in RAR files"),
+    "explain-action_on_unwanted_extensions": TT("Action when an unwanted extension is detected"),
```
Safihre, line 498: Do we need to change the translation? Because it is detected in RAR files, either during downloading or after extraction. So the text isn't wrong.

jcfp: It applies to filenames that aren't rar or in rar files too, even before this change. See [nzbstuff.py:891](https://github.com/sabnzbd/sabnzbd/blob/10c56e08d4a7333a26f4a435025c314b4ce343f9/sabnzbd/nzbstuff.py#L891)

PR #1801, opened 2021-02-19
Reduce the number of full BPSMeter.update calls by caching the amounts

puzzledsab:
The sums of the amounts are stored in BPSMeter.temp_amount[server] and used when a full update is required.

### DIFF
--- a/sabnzbd/bpsmeter.py
+++ b/sabnzbd/bpsmeter.py
@@ -25,7 +25,8 @@ import re
 from typing import List, Dict, Optional
 
 import sabnzbd
-from sabnzbd.constants import BYTES_FILE_NAME, KIBI, MEBI
+from sabnzbd.constants import BYTES_FILE_NAME, KIBI
+from sabnzbd.misc import to_units
 import sabnzbd.cfg as cfg
 
 DAY = float(24 * 60 * 60)
@@ -297,7 +298,7 @@ class BPSMeter:
             self.reset()
 
         elif self.log_time < check_time:
-            logging.debug("MB/s: %.2f", self.bps / MEBI)
+            logging.debug("Speed: %sB/s", to_units(self.bps))
             self.log_time = t
 
         if self.speed_log_time < (t - 1.0):
--- a/sabnzbd/bpsmeter.py
+++ b/sabnzbd/bpsmeter.py
@@ -102,6 +102,7 @@ class BPSMeter:
 
         self.server_bps: Dict[str, float] = {}
         self.cached_amount: Dict[str, int] = {}
+        self.sum_cached_amount: int = 0
         self.day_total: Dict[str, int] = {}
         self.week_total: Dict[str, int] = {}
         self.month_total: Dict[str, int] = {}
@@ -216,6 +217,7 @@ class BPSMeter:
                 self.cached_amount[server] = 0
                 self.server_bps[server] = 0.0
             self.cached_amount[server] += amount
+            self.sum_cached_amount += amount
 
         # Wait at least 0.05 seconds between each full update
         if not force_full_update and t - self.last_update < 0.05:
@@ -236,12 +238,10 @@ class BPSMeter:
                 self.end_of_month = next_month(t) - 1.0
 
         # Add amounts that have been stored temporarily to statistics
-        total_amount = 0
         for srv in self.cached_amount:
             cached_amount = self.cached_amount[srv]
             if cached_amount:
                 self.cached_amount[srv] = 0
-                total_amount += cached_amount
                 if srv not in self.day_total:
                     self.day_total[srv] = 0
                 self.day_total[srv] += cached_amount
@@ -274,7 +274,7 @@ class BPSMeter:
 
         # Quota check
         if self.have_quota and self.quota_enabled:
-            self.left -= total_amount
+            self.left -= self.sum_cached_amount
             if self.left <= 0.0:
                 if not sabnzbd.Downloader.paused:
                     sabnzbd.Downloader.pause()
@@ -282,7 +282,9 @@ class BPSMeter:
 
         # Speedometer
         try:
-            self.bps = (self.bps * (self.last_update - self.start_time) + total_amount) / (t - self.start_time)
+            self.bps = (self.bps * (self.last_update - self.start_time) + self.sum_cached_amount) / (
+                t - self.start_time
+            )
         except:
             self.bps = 0.0
             self.server_bps = {}
@@ -290,6 +292,7 @@ class BPSMeter:
         self.last_update = t
 
         check_time = t - 5.0
+        self.sum_cached_amount = 0
 
         if self.start_time < check_time:
             self.start_time = check_time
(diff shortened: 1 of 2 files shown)


sanderjo:
Maybe a good moment to change logging to Mbps, with one decimal?

So not
2021-02-18 23:09:56,811::DEBUG::[bpsmeter:286] bps: 4453218.5451710895

biut
2021-02-18 23:09:56,811::DEBUG::[bpsmeter:286] Mbps: 4.4

... at a glance you can see what's going on: on DSL, coax, Fiber, and 10+ Gbps connections.

sabnzbd/bpsmeter.py
```diff
@@ -283,7 +297,7 @@ def update(self, server: Optional[str] = None, amount: int = 0):
             self.reset()
 
         elif self.log_time < check_time:
-            logging.debug("bps: %s", self.bps)
+            logging.debug("MB/s: %.2f", self.bps / MEBI)
```
Safihre, line 300: Let's use `to_units()`.

sabnzbd/bpsmeter.py
```diff
@@ -204,9 +205,21 @@ def read(self):
...
+
+        # Add amount to temporary storage
+        if server:
+            if server not in self.temp_amount:
+                self.temp_amount[server] = 0
+                self.server_bps[server] = 0.0
+            self.temp_amount[server] += amount
+
+        # Wait at least 0.021 seconds between each full update
+        if not force and t - self.last_update < 0.021:
```
Safihre, line 220: I think I need a bit more details why this _very_ specific number of seconds? 😋

Safihre:
Does this change benefit performance a lot? You probably made it for a reason (lot's time spent in calls to `bps.update()`?), so I would like to know 👨‍🎓

sabnzbd/bpsmeter.py
puzzledsab: Yes, that may not have been necessary. I wanted it to be more than 0.02 and the Windows internal clock is a bit kinky and has a tendency to be a few nanoseconds off a round number.

puzzledsab:
It was mostly because I really enjoy trying to make things faster, and a little bit because it annoyed me that I had made it slower with the individual server speed measurements.

The improvement is about 35% reduction in total time spent in the update function on my setup, according to yappi. Less than I had hoped, but it would probably be more if I had a faster connection and the sleep and force update didn't trigger quite as often.

sabnzbd/bpsmeter.py
Safihre: What happens if you use something like 0.5 or 0.1?
Probably also reduces CPU usage.

sabnzbd/downloader.py
```diff
@@ -659,7 +659,7 @@ def run(self):
                         if bytes_received + sabnzbd.BPSMeter.bps > limit:
                             while sabnzbd.BPSMeter.bps > limit:
                                 time.sleep(0.01)
-                                sabnzbd.BPSMeter.update()
+                                sabnzbd.BPSMeter.update(force=True)
```
Safihre, line 662: Probably somewhere in the shutdown routine we should also call this with force=True, so all bytes are counted.

sabnzbd/bpsmeter.py
Safihre: I tested setting it to `1` and it still works fine, visually. At most it's 1 second delayed, seems fine by me,
Maybe that increases the reduction of time spent in this function?

puzzledsab: I'm worried that a too long delay could make the transfer rate erratic for those who use the bps limiter. If it's not updated for 1 second then the transfer rate could speed up too much. Then it would sleep for too long to even it out, which would slow it down too much. 0.021 seconds seems safe, and reduces the number of full updates from potentially tens of thousands per second to less than 100.

sabnzbd/downloader.py
thezoggy: Seems like we are forcing it everywhere now? Maybe we should have just done force is default true unless explicitly passed not to

puzzledsab:
After I figured out my get_article problem, this PR reduces the total CPU usage of the processes running under GIL by about 2%.

sabnzbd/downloader.py
Safihre: There's only 1 call to `update` that includes `server.id, bytes_received`, so there's no need for `force`. 
We can keep the old behavior and only skip in case there's a `server` supplied.
What do you think @puzzledsab ?

puzzledsab: @Safihre : Sorry, I don't understand what you are suggesting?

@thezoggy : As far as I can tell there are currently 2 calls with force and 4 without in downloader, and a few without force in bpsmeter.

Safihre:
Made a few changes (mostly stylistic), and as suggested by @thezoggy I made `force=True` the default. So we only cache in case it's safe.
Do you agree with these changes @puzzledsab ?

puzzledsab:
The comment says "Wait at least 0.05 seconds" but the code says 0.25s. I think 0.05 is safer. Other than that I agree.

Safihre:
Turned out that when I used `0.25` the speedlimit was broken, it would overshoot the limit.

Safihre:
Even with `0.025` it overshoots the limit. If I set 20MB/s it would go to 21MB/s..
When I set it to `0.01` it still goes to 20.2MB/s instead of sticking to 20MB/s.
Without these changes it always sticks to exactly 20MB/s.

Safihre:
Ohhh that was the problem 😅 makes sense now.
Let mee see if we then can increase the cache time again!

sabnzbd/downloader.py
```diff
@@ -654,13 +657,13 @@ def run(self):
                         # In case nzf has disappeared because the file was deleted before the update could happen
                         pass
 
+                    sabnzbd.BPSMeter.update(server.id, bytes_received, force_full_update=False)
```
Safihre, line 660: Why move it up? This way it will still display 20.1MB/s for example even when user has set a limit of 20MB/s.
I know it's a bit fake, but has always worked like that.

sabnzbd/bpsmeter.py
```diff
@@ -204,9 +207,22 @@ def read(self):
...
+        # Add amount to temporary storage
+        if server:
+            if server not in self.cached_amount:
+                self.cached_amount[server] = 0
+                self.server_bps[server] = 0.0
+            self.cached_amount[server] += amount
+            self.sum_cached_amount += amount
+
+        # Wait at least 0.05 seconds between each full update
+        if not force_full_update and t - self.last_update < 0.05:
```
Safihre, line 223: I think we can easily bump this to `0.25`, I found no problems in my testing.

sabnzbd/downloader.py
puzzledsab: Mostly to make sure the amount is included for the next test, but if you think that gives the wrong number then it's probably ok to move it below.

puzzledsab: I just tested with 7 000 000 bytes as limit and a 50% speed limit. This results in a speed of 3492500 - 3495500 B/s when I debug log the bps as bytes from the BPS meter. It doesn't matter if the update call is above or below. Is it still too high if you measure it in bytes?

PR #1802, opened 2021-02-20
pre-create subdir if needed (POSIX, par2)

sanderjo:
This is based on the code from @puzzledsab from https://forums.sabnzbd.org/viewtopic.php?f=2&t=25155&p=123748#p123748

It solves the problem with Sample/ and other subdirs in a post with par2 files

Reference NZB with subdir (plus a bonus subdir in that): https://raw.githubusercontent.com/sanderjo/NZBs/master/unique_files_with_subdir_in_subdir_73698be2fe48.nzb

```
$ cat .sabnzbd/logs/sabnzbd.log | grep -i -e mkdir -e "quick-check will rename" -e "no path found"


2021-02-20 07:58:26,778::DEBUG::[newsunpack:2079] Quick-check will rename mysubdir+insub1 to mysubdir/insub1
2021-02-20 07:58:26,779::DEBUG::[newsunpack:2084] mkdir /home/sander/Downloads/incomplete/unique_files_with_subdir_ee0fba9604e5/mysubdir
2021-02-20 07:58:26,781::DEBUG::[newsunpack:2079] Quick-check will rename mysubdir+insub2 to mysubdir/insub2
2021-02-20 07:58:26,782::DEBUG::[newsunpack:2084] mkdir /home/sander/Downloads/incomplete/unique_files_with_subdir_ee0fba9604e5/mysubdir

2021-02-20 08:05:13,466::DEBUG::[newsunpack:2079] Quick-check will rename 63ccfd15f5bce5e6d7c782cde0a000a6 to xxx.s10e13.hog.heaven.720p.web.h264-caffeine.nfo
2021-02-20 08:05:13,466::DEBUG::[newsunpack:2090] No path found
2021-02-20 08:05:13,467::DEBUG::[newsunpack:2079] Quick-check will rename Sample+sample-xxx.s10e13.hog.heaven.720p.web.h264-caffeine.mkv to Sample/sample-xxx.s10e13.hog.heaven.720p.web.h264-caffeine.mkv
2021-02-20 08:05:13,467::DEBUG::[newsunpack:2084] mkdir /home/sander/Downloads/incomplete/moon-wth-pre-create-dir/Sample
```

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -2078,16 +2078,16 @@ def quick_check_set(set, nzo):
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
 
-                    # Let's check if there is a subdir in 'file'
-                    relative_new_path = re.search(r"(.*)(\\|/)", file)  # find directory separator \ or /
+                    # Let's check if there is a subdir in 'file', which SABs needs to create
+                    relative_new_path = re.search(r"(.*)[\\|/]", file)  # find directory before separator \ or /
                     # Proceed if subdir specified, and no malicous '..' in it
                     if relative_new_path and not os.pardir in relative_new_path.group(1):
                         full_new_path = os.path.join(nzo.download_path, relative_new_path.group(1))
-                        # to be sure, check again that subdir within path
+                        # to be sure, check again that subdir is within path
                         within_path = (
                             os.path.commonpath([nzo.download_path, os.path.abspath(full_new_path)]) == nzo.download_path
                         )
-                        if within_path and not os.path.exists(full_new_path):
+                        if (not os.path.exists(full_new_path)) and within_path:
                             logging.debug("Creating subdir %s", full_new_path)
                             try:
                                 os.makedirs(full_new_path)
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -2077,6 +2077,20 @@ def quick_check_set(set, nzo):
             if nzf.md5sum == md5pack[file]:
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
+
+                    # Let's check if there is a subdir in 'file' (only needed on POSIX so '/')
+                    new_path = re.search("(.*)/", file)
+                    if new_path:
+                        new_path = os.path.join(nzo.download_path, new_path.group(1))
+                        logging.debug("mkdir %s", new_path)
+                        try:
+                            os.mkdir(new_path)
+                        except:
+                            logging.debug("Failed to make path")
+                    else:
+                        logging.debug("No path found")
+
+
                     renamer(os.path.join(nzo.download_path, nzf.filename), os.path.join(nzo.download_path, file))
                     renames[file] = nzf.filename
                     nzf.filename = file
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -2082,10 +2082,6 @@ def quick_check_set(set, nzo):
                     # Note: file can and is allowed to have a subdir.
                     # This will only happen on POSIX (with par2), and as
                     # subdirs in par2 always contain "/" (not "\"), we're fine
-                    # Just for logging:
-                    relative_new_path = re.search(r"(.*)/", file)  # find directory before separator "/"
-                    if relative_new_path:
-                        logging.debug("Subdir found: %s", relative_new_path.group(1))
 
                     renamer(
                         os.path.join(nzo.download_path, nzf.filename),
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -2079,27 +2079,19 @@ def quick_check_set(set, nzo):
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
 
-                    # Let's check if there is a subdir in 'file', which SABs needs to create
-                    # This will only happen on POSIX (with par2)
-                    # subdirs in par2 always contain "/" (not "\")
+                    # Note: file can and is allowed to have a subdir.
+                    # This will only happen on POSIX (with par2), and as
+                    # subdirs in par2 always contain "/" (not "\"), we're fine
+                    # Just for logging:
                     relative_new_path = re.search(r"(.*)/", file)  # find directory before separator "/"
-                    # Proceed if subdir specified
                     if relative_new_path:
                         logging.debug("Subdir found: %s", relative_new_path.group(1))
-                        full_new_path = os.path.join(nzo.download_path, relative_new_path.group(1))
-                        # Only create subdir if it does not yet exist, and it's a valid subdir
-                        if not os.path.exists(full_new_path):
-                            if same_file(nzo.download_path, full_new_path) == 2:
-                                # Yes, a subdirectory
-                                logging.debug("Creating subdir %s", full_new_path)
-                                try:
-                                    os.makedirs(full_new_path)
-                                except:
-                                    logging.warning("Failed to create subdir path %s", full_new_path)
-                            else:
-                                logging.warning("par2 wanted to create a directory outside of main directory")
 
-                    renamer(os.path.join(nzo.download_path, nzf.filename), os.path.join(nzo.download_path, file))
+                    renamer(
+                        os.path.join(nzo.download_path, nzf.filename),
+                        os.path.join(nzo.download_path, file),
+                        create_local_directories=True,
+                    )
                     renames[file] = nzf.filename
                     nzf.filename = file
                     result &= True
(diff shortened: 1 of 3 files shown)


sanderjo:
Question: why not put the pre-create-sub-dir in filesystem.renamer() itself?

sabnzbd/newsunpack.py
```diff
@@ -2077,6 +2077,20 @@ def quick_check_set(set, nzo):
             if nzf.md5sum == md5pack[file]:
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
+
+                    # Let's check if there is a subdir in 'file' (only needed on POSIX so '/')
+                    new_path = re.search("(.*)/", file)
```
sanderjo, line 2082: what if the post is created on Windows? Will it contain a "/' or a Windows style "\"? 
Change to os.path.sep does not solve it, AFAIK

sabnzbd/newsunpack.py
```diff
@@ -2077,6 +2077,18 @@ def quick_check_set(set, nzo):
             if nzf.md5sum == md5pack[file]:
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
+
+                    # Let's check if there is a subdir in 'file', and if so, pre-create it
+                    new_path = re.search("(.*)/", file)
+                    if new_path:
+                        new_path = os.path.join(nzo.download_path, new_path.group(1))
```
Safihre, line 2084: And what if this is a malicious file and they add `../../` stuff?
There should be a check for that.

sabnzbd/newsunpack.py
puzzledsab: The move fails on Windows as well, but it uses "Parchive 2.0 client version 1.3.1.3 by Yutaka Sawada". It creates the directory and moves the the files. Yes, it's \ in Windows unless you use bash for Windows.

sanderjo:
@Safihre Do you understand why serverside black is complaining?

On my local system, black is fine:

```
sander@witte2004:~/git/sjo-sab$ python3 -m black -l120  sabnzbd/newsunpack.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
sander@witte2004:~/git/sjo-sab$ python3 -m black -l120 --target-version=py36  sabnzbd/newsunpack.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
```

sanderjo:
@Safihre I think the code is ready now.

sabnzbd/newsunpack.py
```diff
@@ -2077,6 +2077,23 @@ def quick_check_set(set, nzo):
             if nzf.md5sum == md5pack[file]:
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
+
+                    # Let's check if there is a subdir in 'file', which SABs needs to create
+                    relative_new_path = re.search(r"(.*)[\\|/]", file)  # find directory before separator \ or /
+                    # Proceed if subdir specified, and no malicous '..' in it
+                    if relative_new_path and not os.pardir in relative_new_path.group(1):
```
puzzledsab, line 2084: I don't think this is sufficient because .* will match everything up to the file. `../../file` would result in `../..`

sanderjo:
@puzzledsab I got a mail with you saying

```
+                    # Proceed if subdir specified, and no malicous '..' in it
+                    if relative_new_path and not os.pardir in relative_new_path.group(1):
I don't think this is sufficient because .* will match everything up to the file. ../../file would result in ../..
```

I don't see that remark here anymore? Did you revoke it?

To answer your question anyway: os.pardir is ".." on Windows/POSIX, so that would be caught by the `not os.pardir in relative_new_path.group(1):`

sabnzbd/newsunpack.py
```diff
@@ -2077,6 +2077,23 @@ def quick_check_set(set, nzo):
             if nzf.md5sum == md5pack[file]:
                 try:
                     logging.debug("Quick-check will rename %s to %s", nzf.filename, file)
+
+                    # Let's check if there is a subdir in 'file', which SABs needs to create
+                    relative_new_path = re.search(r"(.*)[\\|/]", file)  # find directory before separator \ or /
```
jcfp, line 2082: This regular expression doesn't work as intended: it hits on forward and backward slashes as well as `|`. Two of those three are legal in file and directory names on most unix filesystems, resulting in the creation of incorrect subdirs. Why not use os.path.split() or, if you want to go with a regex, os.sep?

sabnzbd/newsunpack.py
```diff
@@ -2077,6 +2077,23 @@ def quick_check_set(set, nzo):
...
+                    # Proceed if subdir specified, and no malicous '..' in it
+                    if relative_new_path and not os.pardir in relative_new_path.group(1):
+                        full_new_path = os.path.join(nzo.download_path, relative_new_path.group(1))
+                        # to be sure, check again that subdir is within path
+                        within_path = (
+                            os.path.commonpath([nzo.download_path, os.path.abspath(full_new_path)]) == nzo.download_path
```
jcfp, line 2088: Use of abspath might still allow escaping from nzo.downloadpath via symlinks (supported by many archive types), maybe realpath() is a safer option here?

sanderjo:
Maybe good to have an overview / summary:

myfile.txt
Sample/myfile.txt
Dir/SubDir/myfile.txt
and
Sample\myfile.txt (if that can happen at all?)


Also OK (despite the double dots):
My..Dir/myfile.txt
..blabla/myfile.txt

Not good because malicious:
../../dir/file.txt
/root/file.txt

Unsure (strange, mixed chars)
dir\subdir/file.txt
dir|more/file.txt


Furthermore:
I think we need unittests for this
Best to put it in a seperate function call, a bit like isvaliddir()

Safihre:
> Question: why not put the pre-create-sub-dir in filesystem.renamer() itself?

Yes, I think so. As an option `allow_directories` that defaults to `False`.
And maybe we can use the `same_file` detection? It has a sub-folder detection.

sanderjo:
> And maybe we can use the `same_file` detection? It has a sub-folder detection.

Interesting:

```
>>> from sabnzbd import filesystem

>>> filesystem.same_file("/my/big/", "/my/big/aaaa")
2
```

sanderjo:
Short: even if created on Windows, the directory separator is "/" (and not "\"). So no need to consider "/".

```
C:\Temp\MultiPar130\mycontent>..\par2j64.exe create mypar hallo.txt mysubdir\somefile.txt
Parchive 2.0 client version 1.3.0.6 by Yutaka Sawada

```
with result in the par2 file:
```
$ strings -9 mypar.par2 
mysubdir/somefile.txt
hallo.txt
```


```
$ par2 v mypar.par2 
...

Target: "mysubdir/somefile.txt" - missing.
Target: "hallo.txt" - missing.

```

sanderjo:
@Safihre I would like to have this feature in 3.2.0. As I presume 3.2.0 is near (correct?), time is short, so how about this approach:

Keep it simple & better safe than sorry / fail-safe. So for now, only accept 100% clean cases (for example: only letters in directory), and default the rest to failure (with logging). Later on improve

puzzledsab:
par2 wouldn't create these kinds of directories. What you have to consider is what the path can contain if it's created manually with bad intent.

sanderjo:
> par2 wouldn't create these kinds of directories. What you have to consider is what the path can contain if it's created manually with bad intent.

Maybe I don't understand you:
What do you mean with "these kinds of directories"? We've seen subdirs like "Sample/" in the wild. And both par2 and multipar-par2 can create those subdirectories, as I have shown above.

sanderjo:
More notes: this will only happen on POSIX and with par2 (not on Windows, not with multipar). I'll change the subject.

sanderjo:
> > Question: why not put the pre-create-sub-dir in filesystem.renamer() itself?
> 
> Yes, I think so. As an option `allow_directories` that defaults to `False`.
> And maybe we can use the `same_file` detection? It has a sub-folder detection.

@Safihre 

I played with the code, and I think there are **two** booleans:
1. allow_creation_directories
2. must_stay_within_parent_directory

For the use case of this thread (par2 subdirectory), both must be True. 
In other cases, both should be False (and default)
Agree? 
If so, we could combine them into one boolean parameter ... with the Frankenstein name must_stay_within_parent_directory_and_allow_creation_directories ... or maybe better subdir_exanpansion_only ?

Safihre:
Maybe `create_local_directories`? Sounds good to me!

Safihre (changes requested):

Safihre:
Looks good to me, @jcfp or @puzzledsab any feedback?

Safihre (approved):

puzzledsab:
There's a spelling error, Reusing instead of Refusing. Also, I think it should be made clearer to the user that it's probably a malicious NZB. Other than that I think it fixes everything we've talked about.

sanderjo:
> There's a spelling error, Reusing instead of Refusing. 

Thanks!

> Also, I think it should be made clearer to the user that it's probably a malicious NZB.

You mean: in that message now saying "Refusing to go outside directory" ... change to "Refusing to go outside directory. Malicious NZB?" ... ?

PR #1813, opened 2021-03-07
fix config auto_sort setting, broken by #1666

jcfp:
Same problem as with the javascript in the interface templates: things were written to work with the broken sorting function. Wish I spotted this one earlier so breaking user settings could have been avoided. :disappointed:

### DIFF
--- a/interfaces/Config/templates/config_switches.tmpl
+++ b/interfaces/Config/templates/config_switches.tmpl
@@ -140,8 +140,8 @@
                         <label class="config" for="auto_sort">$T('opt-auto_sort')</label>
                         <select name="auto_sort" id="auto_sort">
                             <option value="">$T('default')</option>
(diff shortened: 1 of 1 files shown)


interfaces/Config/templates/config_switches.tmpl
```diff
@@ -140,8 +140,8 @@
                         <label class="config" for="auto_sort">$T('opt-auto_sort')</label>
                         <select name="auto_sort" id="auto_sort">
                             <option value="">$T('default')</option>
-                            <option value="avg_age asc" <!--#if $auto_sort == "avg_age asc" then 'selected="selected"' else ""#--> >$T('Glitter-sortAgeAsc')</option>
-                            <option value="avg_age desc" <!--#if $auto_sort == "avg_age desc" then 'selected="selected"' else ""#--> >$T('Glitter-sortAgeDesc')</option>
+                            <option value="avg_age desc" <!--#if $auto_sort == "avg_age asc" then 'selected="selected"' else ""#--> >$T('Glitter-sortAgeAsc')</option>
```
Safihre, line 143: Shouldn't the if statement also be changed here?

PR #1815, opened 2021-03-08
Use binary mode to make write test less inaccurate on Windows

puzzledsab:


sanderjo:
How much accurate? Did you measure on Windows without and with your patch?

puzzledsab:
It's a bit off compared to the method used in the real file writing, but not quite as much. Using binary mode was about 44% faster for os.write() than text mode. open() with b (for binary) and <filehandle>.write(data) is still 25-40% faster. I wrote more than 200 GB when I tested to make sure caching wasn't a bit issue.

Issue #1821, opened 2021-03-12
Improvements to Deobfuscate


sanderjo:
Based on user feedback:

**Improvements to Deobfuscate**

- [x] Do not rename set of bigger files with the same extension. Reason: probably a collection (season pack, music), and renaming those files would not improve it. Based on https://github.com/sabnzbd/sabnzbd/issues/1818 Something else than the already exclusion of certain extensions. Method: add that extension as excluded extension
- [x] If we rename a (big) file, also rename other small files that start with the same basename, to the new basename. So if we rename blabla.iso to MyGreatISO2021.iso, then also rename blabla.srt to MyGreatISO2021.srt, and rename blabla.eng.srt to MyGreatISO2021.eng.srt. See https://github.com/sabnzbd/sabnzbd/issues/1811

I will work on this, hopefully the coming days.

Logic must land in https://github.com/sabnzbd/sabnzbd/blob/e8eec8069698fbb4e4a0f68595d4b59d50298824/sabnzbd/deobfuscate_filenames.py#L134-L178

sanderjo:
Pseudo sample sets:

myspeech_part1.mp3
myspeech_part2.mp3
myspeech_part3.mp3



someiso.abc
someiso.iso
someiso.srt
someiso.eng.srt
someiso.dut.srt


someiso.iso
someiso_sample.iso

someiso.iso
sample.iso

sanderjo:
Done: rename smaller files (with same basename) too:

Post:
```
-rw-rw-r--  1 sander sander        44 mrt 14 14:30 myiso.abc
-rw-rw-r--  1 sander sander        27 mrt 14 14:30 myiso.fra.srt
-rw-rw-r--  1 sander sander 100000000 mrt 14 14:29 myiso.iso
-rw-rw-r--  1 sander sander        22 mrt 14 14:30 myiso.srt
-rw-rw-r--  1 sander sander        10 mrt 14 14:42 nothing.txt
```


Result (with Post name "My Very Important Post"):
```
-rwxr-xr-x  1 sander sander        44 mrt 14 14:44 'My Very Important Post.abc'*
-rwxr-xr-x  1 sander sander        27 mrt 14 14:44 'My Very Important Post.fra.srt'*
-rwxr-xr-x  1 sander sander 100000000 mrt 14 14:44 'My Very Important Post.iso'*
-rwxr-xr-x  1 sander sander        22 mrt 14 14:44 'My Very Important Post.srt'*
[...]

PR #1826, opened 2021-03-14
deobfuscate: rename accompanying (smaller) files with same basename, and no renaming of collections with same extension

sanderjo:
Closes https://github.com/sabnzbd/sabnzbd/issues/1821

### DIFF
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -162,20 +162,25 @@ def deobfuscate_list(filelist, usefulname):
     if run_renamer:
 
         excluded_file_exts = EXCLUDED_FILE_EXTS
-        # If there is a collection (with bigger files), we don't want to rename it
-        bigfileslist = [f for f in filelist if os.path.getsize(f) > MIN_FILE_SIZE]
-        extlist = [os.path.splitext(f)[1] for f in bigfileslist]
-        # Count the extensions:
-        extcounterdict = dict((x, extlist.count(x)) for x in set(extlist))
-        for ext in extcounterdict:
-            if extcounterdict[ext] >= 3:
-                # probably a collection, so exclude this extension from renaming
+        # If there is a collection with bigger files with the same extension, we don't want to rename it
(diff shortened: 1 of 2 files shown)


sabnzbd/deobfuscate_filenames.py
```diff
@@ -158,7 +161,12 @@ def deobfuscate_list(filelist, usefulname):
     # No par2 files? Then we try to rename qualifying (big, not-excluded, obfuscated) files to the job-name
     if run_renamer:
         logging.debug("Trying to see if there are qualifying files to be deobfuscated")
+        # sort filelist on size, so start with biggest file
+        filelist = sorted(filelist, key=os.path.getsize, reverse=True)
```
Safihre, line 165: Maybe we can remove the `sorted` a few lines above?

sabnzbd/deobfuscate_filenames.py
```diff
@@ -168,11 +176,19 @@ def deobfuscate_list(filelist, usefulname):
...
-                # OK, rename
+                # Rename and make sure the new filename is unique
                 path, file = os.path.split(filename)
                 new_name = get_unique_filename("%s%s" % (os.path.join(path, usefulname), get_ext(filename)))
                 logging.info("Deobfuscate renaming %s to %s", filename, new_name)
-                # Rename and make sure the new filename is unique
                 renamer(filename, new_name)
+                # find other files with the same basename (in the same directory), and rename them in the same way:
+                basedirfile, _ = os.path.splitext(filename)
+                for samebasenamefile in glob.glob(basedirfile + "*"):
```
Safihre, line 186: Please use `filesystem.globber()`.

Safihre:
If you could add some unittests, that would be great. This kind of added functionality is easily broken.

sanderjo:
Closes #1821

neofright:
Unfortunately @sanderjo the linuxserver.io 'latest' tag is still lagging behind master.

I'll reply back when the container is updated to include these commits (if the issue is still open!) but I can't say how long that would be.

Issue #1837, opened 2021-03-26
Solution for MacOS and >4TB drives (plus bonus check for other OSes)


sanderjo:
In https://forums.sabnzbd.org/viewtopic.php?f=3&t=25264 we've found a solution / workaround for MacOS with big (>4TB) disks where python's and even MacOS' statvfs() reports too small available disk size, due to a rollover at 4TB (= 42 bit = 32bit counter for 10bit block sizes?). Since 2017 a known bug, but still not solved in Python nor MacOS (so maybe a ... feature?). 
In the forum thread we found a solution: calling statfs() in the MacOS C-lib (so not python's os.statvfs() )

See WiP PoC https://github.com/sanderjo/diskfree-statvfs-df

I'll implement this in SAB. 

Summary: if MacOS and df tells disk > 4TB then use clib's statfs()

Implementation:

- [ ] in SABnzbd.py if MacOS and df tells disk > 4TB, set boolean `use_statfs`
- [ ] bonus: in SABnzbd.py, for all POSIX: check if df and os.statvfs() report the same. If not: warning. Possible suspects: Linux with 32-bit OS, and >4TB drives
- [ ] in filesystem: implement the MacOS clib->statfs() call in a separate function
- [ ] in filesystem: implement a function that does "df -m". Rationale: " df -m" is always correct, but a heavier process
- [ ] in filesystem.diskspace_base(): if `use_statfs`, then call the function for MacOS

@Safihre  @jcfp feedback?

Safihre:
As discussed, no special check at startup. Just add the loading code to `__init__.py` and use it in the disk-check functions.

sanderjo:
> As discussed, no special check at startup. Just add the loading code to `__init__.py` and use it in the disk-check functions.

Nice. Less inter-module dependencies. My first step is then an implementation in only filesystem.py : if darwin, then call the clib-statfs()

sanderjo:
OK, step 1 working:
```
git clone https://github.com/sanderjo/sabnzbd.git sjo-sab
cd sjo-sab/
git checkout diskspace_macos

```


On MacOS:
```
server:sjo-sab sander$ python3


>>> import sabnzbd.filesystem
>>> sabnzbd.filesystem.disk_free_macos_clib_statfs64("/")
(999345127424, 545738301440)
>>> sabnzbd.filesystem.diskspace_base("/")
(930.7126770019531, 508.2583541870117)
>>> 

```

[...]

PR #1838, opened 2021-03-26
Diskspace macos

sanderjo:
related to https://github.com/sabnzbd/sabnzbd/issues/1837

Correct drive space reporting on MacOS for disk >4TB

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -29,6 +29,7 @@ import time
 import fnmatch
 import stat
 import zipfile
+import ctypes
 from typing import Union, List, Tuple, Any, Dict, Optional
 
 try:
@@ -944,16 +945,13 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
 ##############################################################################
 # Diskfree
 ##############################################################################
-def disk_free_macos_clib_statfs64(directory):
+def disk_free_macos_clib_statfs64(directory: str) -> Tuple[int, int]:
     # MacOS only!
     # direct system call to c-lib's statfs(), not python's os.statvfs()
     # because statvfs() on MacOS has a rollover at 4TB (possibly a 32bit rollover with 10bit block size)
     # Based on code of pudquick and blackntan
     # Input: directory.
     # Output: disksize and available space, in bytes
-
-    import ctypes
-    import ctypes.util
 
     # format & parameters: on MacOS, see "man statfs", lines starting at
     # "struct statfs { /* when _DARWIN_FEATURE_64_BIT_INODE is defined */"
@@ -982,14 +980,12 @@ def disk_free_macos_clib_statfs64(directory):
         ctypes.create_string_buffer(utob(directory)), ctypes.byref(fs_info64)
     )  # fs_info64 gets filled out via the byref()
     if result == 0:
-        # Upon successful completion, a value of 0 is returned.
-        disk_size = fs_info64.f_blocks * fs_info64.f_bsize
-        avail_size = fs_info64.f_bavail * fs_info64.f_bsize
+        # result = 0: "Upon successful completion, a value of 0 is returned."
+        return fs_info64.f_blocks * fs_info64.f_bsize, fs_info64.f_bavail * fs_info64.f_bsize
     else:
-        # Otherwise, -1 is returned and the global variable errno is set to indicate the error.
+        # restul = -1: Otherwise, -1 is returned and the global variable errno is set to indicate the error.
         logging.debug("Call to MACOSLIBC.statfs64 not successful. Value of errno is %s", ctypes.get_errno())
-        disk_size = avail_size = 0
-    return disk_size, avail_size
+        return 0, 0
 
 
 def diskspace_base(dir_to_check: str) -> Tuple[float, float]:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -987,7 +987,7 @@ def disk_free_macos_clib_statfs64(directory):
         avail_size = fs_info64.f_bavail * fs_info64.f_bsize
     else:
         # Otherwise, -1 is returned and the global variable errno is set to indicate the error.
-        logging.debug("result from statfs64 not 0, but %s", result)
+        logging.debug("Call to MACOSLIBC.statfs64 not successful. Value of errno is %s", ctypes.get_errno())
         disk_size = avail_size = 0
     return disk_size, avail_size
 
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -977,9 +977,8 @@ def disk_free_macos_clib_statfs64(directory):
             ("f_reserved", ctypes.c_uint32 * 8),
         ]
 
-    clib = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)  # the C library
     fs_info64 = statfs64()  # set up the parameters to be filled out
-    result = clib.statfs64(
+    result = sabnzbd.MACOSLIBC.statfs64(
         ctypes.create_string_buffer(utob(directory)), ctypes.byref(fs_info64)
     )  # fs_info64 gets filled out via the byref()
     if result == 0:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -944,6 +944,45 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
 ##############################################################################
 # Diskfree
 ##############################################################################
+def disk_free_macos_clib_statfs64(directory):
+    # MacOS only!
+    # direct system call to c-lib's statfs(), not python's os.statvfs()
+    # Based on code of pudquick and blackntan
+    # Input: directory.
+    # Output: disksize and available space, in bytes
+
+    import ctypes
+    import ctypes.util
+
+    class statfs64(ctypes.Structure):
+        _fields_ = [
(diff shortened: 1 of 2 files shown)


sabnzbd/filesystem.py
```diff
@@ -944,6 +944,45 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
 ##############################################################################
 # Diskfree
 ##############################################################################
+def disk_free_macos_clib_statfs64(directory):
+    # MacOS only!
+    # direct system call to c-lib's statfs(), not python's os.statvfs()
+    # Based on code of pudquick and blackntan
```
Safihre, line 950: Let's also add _why_ we made this: the 32bit-bug on macOS.

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,45 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+            ("f_fssubtype", ctypes.c_uint32),
+            ("f_fstypename", ctypes.c_char * 16),
+            ("f_mntonname", ctypes.c_char * 1024),
+            ("f_mntfromname", ctypes.c_char * 1024),
+            ("f_reserved", ctypes.c_uint32 * 8),
+        ]
+
+    kern = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)
+    fs_info64 = statfs64()
+    root_volume = ctypes.create_string_buffer(str.encode(directory))
```
Safihre, line 979: Use `utob` function instead of `str.encode`.

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,45 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+    result64 = kern.statfs64(root_volume, ctypes.byref(fs_info64))
```
Safihre, line 980: Seems `result64` is unused, so no need to declare it.
It can just be:
```python
kern.statfs64(root_volume, ctypes.byref(fs_info64))
```

And we could even remove the line above and make it:
```python
kern.statfs64(ctypes.create_string_buffer(utob(directory)), ctypes.byref(fs_info64))
```

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,45 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+            ("f_ffree", ctypes.c_uint64),
+            ("f_fsid", ctypes.c_uint64),
+            ("f_owner", ctypes.c_uint32),
+            ("f_type", ctypes.c_uint32),
+            ("f_flags", ctypes.c_uint32),
+            ("f_fssubtype", ctypes.c_uint32),
+            ("f_fstypename", ctypes.c_char * 16),
+            ("f_mntonname", ctypes.c_char * 1024),
+            ("f_mntfromname", ctypes.c_char * 1024),
+            ("f_reserved", ctypes.c_uint32 * 8),
```
Safihre, line 974: Do we need to declare all the fields, even though we only use a few? From the forum-thread it seems you can also only delcare the field you want to use?

Safihre (changes requested):

sabnzbd/filesystem.py
sanderjo: AFAIK: No. You really need all the fields ... otherwise the c code is writing to 'unspecified' memory ... hence the coredump I got on Linux with too little parameters.

sanderjo: Update: I verified, and yes, with too little parameters, on MacOS I get " Segmentation fault: 11" and on Linux "Segmentation fault (core dumped)". So: all parameters are needed.

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,55 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+    result = clib.statfs64(
+        ctypes.create_string_buffer(utob(directory)), ctypes.byref(fs_info64)
+    )  # fs_info64 gets filled out via the byref()
+    if result == 0:
+        # Upon successful completion, a value of 0 is returned.
+        disk_size = fs_info64.f_blocks * fs_info64.f_bsize
+        avail_size = fs_info64.f_bavail * fs_info64.f_bsize
+    else:
+        # Otherwise, -1 is returned and the global variable errno is set to indicate the error.
+        logging.debug("result from statfs64 not 0, but %s", result)
```
jcfp, line 990: If I read the comment above correctly, the value for result is always going to be -1 here. Meanwhile, the interesting info is stored in _errno_ which isn't logged.

jcfp:
Is there an upstream bug somewhere (apple? python?) that we can track to know if/when this workaround can be removed? If so, it would be useful to reference it in a comment.

sabnzbd/filesystem.py
sanderjo: Correct. FYI: it's text from "man statfs" on MacOS.

But what do you suggest? I see no way to get "errno". Do you?

sanderjo:
> Is there an upstream bug somewhere (apple? python?) that we can track to know if/when this workaround can be removed? If so, it would be useful to reference it in a comment.

AFIAK, not on python. Exists since 2017. Problem at MacOS level: also exists at C level for statvfs() but not statfs().

I don't own a MacOS system. I do have non-root user access via SSH to a MacOS via a friend, but without >4TB drive. So I'm reluctant to report a bug.

sabnzbd/filesystem.py
jcfp: Errno if possible, otherwise at least clearly state in the log msg that statfs64 failed. There's https://docs.python.org/3/library/ctypes.html#ctypes.get_errno - maybe that will do?

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,54 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+def disk_free_macos_clib_statfs64(directory):
+    # MacOS only!
+    # direct system call to c-lib's statfs(), not python's os.statvfs()
+    # because statvfs() on MacOS has a rollover at 4TB (possibly a 32bit rollover with 10bit block size)
+    # Based on code of pudquick and blackntan
+    # Input: directory.
+    # Output: disksize and available space, in bytes
+
+    import ctypes
+    import ctypes.util
```
Safihre, line 956: Remove the imports.

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,54 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+        ctypes.create_string_buffer(utob(directory)), ctypes.byref(fs_info64)
+    )  # fs_info64 gets filled out via the byref()
+    if result == 0:
+        # Upon successful completion, a value of 0 is returned.
+        disk_size = fs_info64.f_blocks * fs_info64.f_bsize
+        avail_size = fs_info64.f_bavail * fs_info64.f_bsize
+    else:
+        # Otherwise, -1 is returned and the global variable errno is set to indicate the error.
+        logging.debug("Call to MACOSLIBC.statfs64 not successful. Value of errno is %s", ctypes.get_errno())
+        disk_size = avail_size = 0
```
Safihre, line 991: Just `return 0.0, 0.0`

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,54 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
...
+        ]
+
+    fs_info64 = statfs64()  # set up the parameters to be filled out
+    result = sabnzbd.MACOSLIBC.statfs64(
+        ctypes.create_string_buffer(utob(directory)), ctypes.byref(fs_info64)
+    )  # fs_info64 gets filled out via the byref()
+    if result == 0:
+        # Upon successful completion, a value of 0 is returned.
+        disk_size = fs_info64.f_blocks * fs_info64.f_bsize
+        avail_size = fs_info64.f_bavail * fs_info64.f_bsize
```
Safihre, line 987: No need to declare variables, let's just right away:
`return fs_info64.f_blocks * fs_info64.f_bsize, fs_info64.f_bavail * fs_info64.f_bsize`

sabnzbd/filesystem.py
```diff
@@ -944,6 +944,54 @@ def remove_all(path: str, pattern: str = "*", keep_folder: bool = False, recursi
 ##############################################################################
 # Diskfree
 ##############################################################################
+def disk_free_macos_clib_statfs64(directory):
```
Safihre, line 947: Please add typing of input and output (see `diskspace_base`).

Safihre (changes requested):

sabnzbd/filesystem.py
sanderjo: Really remove? I would say they are needed by line 960 and further ...

sabnzbd/filesystem.py
sanderjo: and to be compliant with diskspace_base(), also make them floats?

sabnzbd/filesystem.py
sanderjo: I'll make it `return 0, 0` as it's bytes (like win32api.GetDiskFreeSpaceEx() ), not GIGIbytes (and thus float)

sabnzbd/filesystem.py
Safihre: Move them to the top of the file. As discussed.

PR #1841, opened 2021-03-29
Show server expiration date in server summary

puzzledsab:
Having used this feature for a while on several servers I've found that it's very inconvenient to have to open each server to see it.


Safihre:
Hmmm, but then it will be in yyyy-mm-dd notation, not very pretty. But we don't really have a better way to fix it at that spot.
In Glitter the format is an option.

Issue #1852, opened 2021-04-10
FYI: rar, SABnzbd and unrar can handle Unicode passworded rar files


sanderjo:
As a FYI: 

While working on improving the ambiguous message "Failed to read the password file %s" when reading a password file, I verified rar, SABnzbd and unrar can handle Unicode passworded rar files. Tested on Linux, not on Windows.

Proof:

I created a rar set with a unicode password `你好` using:
```
$ rar a blabla -p你好  -v10m -m0 *
```
I posted it, and let SAB download it. That worked, with the (unicoded) passwords in a password file:

```
2021-04-10 18:42:23,708::DEBUG::[misc:794] Read these passwords from file: ['secret123', '你好', 'Heizölrückstoßabdämpfung', 'einde']
2021-04-10 18:42:23,708::INFO::[misc:796] Read 4 passwords from file /home/sander/lees_file_ascii_of_unicode/some_passwords_unicode.txt
2021-04-10 18:42:23,708::INFO::[assembler:340] Trying password "你好" on job "unicode-password-included-8068accb9f8a"
2021-04-10 18:42:23,917::INFO::[assembler:376] Password "你好" matches for job "unicode-password-included-8068accb9f8a"
2021-04-10 18:42:23,918::DEBUG::[assembler:387] File contains: some_binary.iso
```

PR #1853, opened 2021-04-10
rar password file: clearer fault messages

sanderjo:
I find the current message "Failed to read the password file %s" too ambiguous and not useful to users. Therefor this PR.

If specified rar password does not exist at all:

image

File does exist, but is not a plain text file (not ASCII, not Unicode, see https://github.com/sabnzbd/sabnzbd/issues/1852 ), for example user saved as a Word file.

image

I think the message is printed twice because there are two rar files. Not nice if there are 10 - 100 rar files. 

Feedback welcome

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -802,6 +802,10 @@ def get_all_passwords(nzo):
                         "Your password file contains more than 30 passwords, testing all these passwords takes a lot of time. Try to only list useful passwords."
                     )
                 )
+        except UnicodeDecodeError:
+            logging.warning("Specified password file %s is not a plain text file", pw_file)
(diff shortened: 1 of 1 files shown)


Safihre:
I don't really see the point, personally, for these extra levels.
We changed in the last release that it logs the traceback why it failed (before it didn't), isn't that enough?

sanderjo:
My thoughts:
* tracebacks are nasty for normal users. IMHO a traceback should not happen with a normal user's mistake (like selecting not-valid password file)
* just a few lines of code ... Although two extra text lines to translate

Safihre:
Due to the extra translations and this being a quite advanced use case, I think the current message is enough.

PR #1856, opened 2021-04-15
switch prepend to sys.executable in misc.build_and_run_command

jcfp:
Prepend with sys.executable rather than "python", because there may be no such a thing on the PATH (for example if an OS vendor decided to provide only explicit python2 and/or python3 commands). I did keep "python" around as a fallback, since the [docs](https://docs.python.org/3/library/sys.html#sys.executable) mention that _if Python is unable to retrieve the real path to its executable, sys.executable will be an empty string or None_ in which case it's still better than nothing.

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -1011,7 +1011,7 @@ def build_and_run_command(command: List[str], flatten_command=False, **kwargs):
                     raise IOError
                 elif script_file.read(2) != "#!":
                     # No shebang (#!) defined, add default python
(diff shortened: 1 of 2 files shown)


sanderjo:
Ah, with (from https://docs.python.org/3/library/sys.html ):

> sys.executable
> A string giving the absolute path of the executable binary for the Python interpreter, on systems where this makes sense. If Python is unable to retrieve the real path to its executable, sys.executable will be an empty string or None.

Nice!

PR #1859, opened 2021-04-16
Default to block non-LAN traffic for empty Local network ranges

Safihre:
The the recent [malware problem](https://www.reddit.com/r/usenet/comments/mr9qom/beware_of_malware_targeting_unprotected/) highlighted that setting SABnzbd security right is not obvious for the regular user. 
Users saw the `No access` for `External internet access` in the Config, so assumed all is fine, but it's not.
Therefore, I changed the code to rely on `ipaddress.ip_address(referrer).is_private` to see if we are in a local network instead of demanding the `List of local network ranges` to be filled.

This is quite a behavior change, because now `No access` will actually work out of the box and block internet connections. So all the badly configured setups will now not be so badly configured and might block the user from using external apps, until they fix things.
But I think it's quite worth it, agreed @jcfp / @sanderjo? 

I now moved `List of local network ranges`  to Specials, but can't we just remove it? In what case would a user want to use that?

TODO: Implement unittests for `secured_expose`.


Safihre:
This also includes a refactor of the access-validator, so all is handled by `secured_expose`.

jcfp:
Requiring one to explicitly grant some level of access for what in most cases amount to 'outside' connections is probably not a bad thing. That said, it was already necessary to change the listening host from its _localhost_ default and (in most cases) setup port forwarding too - something many of the affected users seem to have done in order to remotely access sab via an app on their mobile devices. Given the demonstrated lack of security awareness and networking insight, it seems rather unlikely they could have achieved that feat without following some tutorial, which then left them with the web interface exposed without authentication. That kind of root cause won't go away by introducing an extra step before sab is remotely accessible.

> I now moved `List of local network ranges` to Specials, but can't we just remove it? In what case would a user want to use that?

There's quite a lot of weird networking setups out there, including ones where addresses in these private ranges aren't actually that private at all (as in, not limited to one's own household). Systems may also be connected to multiple private networks, but access only desirable from one of those (think a home lan on one side and an isp nat network on the other; or a connection to a corporate vpn in addition to a home lan).

sabnzbd/interface.py
```diff
@@ -150,25 +167,43 @@ def internal_wrap(*args, **kwargs):
...
+    is_allowed = False
+    if not cfg.local_ranges():
+        try:
+            is_allowed = ipaddress.ip_address(referrer).is_private
+        except ValueError:
+            # Something malformed, reject
+            pass
+    else:
+        is_allowed = bool(
+            [1 for r in cfg.local_ranges() if (referrer.startswith(r) or referrer.replace("::ffff:", "").startswith(r))]
```
jcfp, line 200: The replace part could be problematic with certain ipv6 addresses (unless cherrypy guarantees a fully written-out version of the address for ipv6, eliminating any chance of a '::' in the address). For example, someone configures `22ab:cd:ef:` as the local range, and the remote ip is something like `22::ffff:ab:cd:ef:12:34`. Limiting the replacing to addresses that aren't ipv6 would avoid this issue.

Separate form the above: comparing address ranges this way (starts with...) is yuck. Too bad it isn't in proper network notation so we could use ipaddress to check whether some address is in a given range.
```
>>> i = ipaddress.ip_address("4.3.2.1")
>>> n = ipaddress.ip_network("10.6.48.0/22")
>>> i in n
False
```

sabnzbd/interface.py
```diff
@@ -150,25 +167,43 @@ def internal_wrap(*args, **kwargs):
...
-    )
-    allowed = referrer in ("127.0.0.1", "::ffff:127.0.0.1", "::1") or range_ok or access_type <= cfg.inet_exposure()
-    if not allowed:
-        logging.debug("Refused connection from %s", referrer)
-    return allowed
+    # It will always contain that ::ffff: prefix, the ipaddress module can handle that
+    referrer = cherrypy.request.remote.ip
+
+    # Check for localhost
+    if referrer in ("127.0.0.1", "::ffff:127.0.0.1", "::1"):
```
jcfp, line 187: Use misc.is_loopback_addr() instead?

sabnzbd/interface.py
```diff
@@ -150,25 +167,43 @@ def internal_wrap(*args, **kwargs):
...
-    # It will always contain that ::ffff: prefix
-    range_ok = not cfg.local_ranges() or bool(
-        [1 for r in cfg.local_ranges() if (referrer.startswith(r) or referrer.replace("::ffff:", "").startswith(r))]
-    )
-    allowed = referrer in ("127.0.0.1", "::ffff:127.0.0.1", "::1") or range_ok or access_type <= cfg.inet_exposure()
-    if not allowed:
-        logging.debug("Refused connection from %s", referrer)
-    return allowed
+    # It will always contain that ::ffff: prefix, the ipaddress module can handle that
+    referrer = cherrypy.request.remote.ip
```
jcfp, line 184: To avoid confusion, you might want to change the variable name to something other than "referrer" - as that usually means the (user-set) http header.

jcfp (commented):
Didn't look at the @secured_expose stuff yet, too tired for that stuff right now.

sabnzbd/interface.py
Safihre: This is the same as the code was before, so far it worked but I admit I have no clue why and how.. The ipaddress module does recognize this format with `::ffff:` at the start. So maybe we can use it? Although I'm not sure how, do you maybe know? :)

jcfp: If it starts with `::ffff:` and contains dots, converting using `ipaddress.ip_address(input).ipv4_mapped` seems to work.

For the local ranges, we could start accepting the standard notation (e.g. _192.168.0.0/24_) `ipaddress.ip_network` takes, and internally convert legacy settings (easy to detect, no netmask) to that format. More flexibility for power users, and ipaddress methods.

Safihre: I will leave it now as it has been for a few years, maybe when you have time you could make this change? :)

Safihre:
Added the unittests and with those found some problems in the API permissions checking.
It allowed sub-functions like `mode=config&name=set_apikey` when the access was set to Basic-API (level 2), while those should only be allowed on Full-API (level 3).

Safihre:
@puzzledsab maybe you also want to review?

sanderjo:
The SAB GUI says "External internet access". Isn't that double? What is the value of "External" in this sentence? Can't we leave it out? 
Oh, to avoid confusion: I would rename it to "Access from Internet to SAB API". It does not cover the GUI, right?

It does define it as non-private IP addresses, right? So on a VPS (with a public IP address), it means not anything with a public IP address can access it ... Yes, that's good.


Furthermore, should the SAB GUI tell "Note: port forwarding needed" if SAB detects SAB is on a private IP address?

Safihre:
> Oh, to avoid confusion: I would rename it to "Access from Internet to SAB API". It does not cover the GUI, right?

It is also about the GUI, that's the last option on the list. 

> Furthermore, should the SAB GUI tell "Note: port forwarding needed" if SAB detects SAB is on a private IP address?

As @jcfp notes, clearly people are following some kind of guides to do this. So let's not encourage them..

thezoggy:
might want to step back and figure out what we are trying to resolve and how do users get in that state.

i understand wanting to make sab more secure, but also dont want to try make it just not work out of the box where lots of people would have no clue and think its broken and give up on it. 

sab rarely gets used as solo. most people are using it with 3rd party apps + tools and because it works 'better' out of the box they run with it. I've heard many times of people trying to use nzbget but fail to configure it to get it to play nice and then give up. we would not want sab to fall into that trap (like setting up a default password on sab out of the box) as some people wouldnt ever change it and that would cause more harm than good in the long run.

Safihre:
> i understand wanting to make sab more secure, but also dont want to try make it just not work out of the box where lots of people would have no clue and think its broken and give up on it.

I think this way it actually gets simpeler, there is just 1 dropdown where you select "External internet access", no need for local network ranges anymore. If you change the option, it does exactly what the option describes, block/allow external internet access.
They then have to manually setup a username/password.

sanderjo:
Just to be sure: **SABnzbd Host** must be set to 0.0.0.0 (or ::) for these settings to have, right?

Safihre:
> Just to be sure: SABnzbd Host must be set to 0.0.0.0 (or ::) for these settings to have, right?

I think so, right? @jcfp

puzzledsab:
I don't know much about this stuff, but could it be possible to access the server through IP spoofing if the computer is accessible from the internet and no password is set? If so, maybe you should add an extra warning to always set a password in such circumstances.

jcfp:
> > Just to be sure: SABnzbd Host must be set to 0.0.0.0 (or ::) for these settings to have, right?
> 
> I think so, right? @jcfp

Only the remote ip address matters: for a sab install configured to listen on an explicit private lan address (say 192.168.12.34 rather than 0.0.0.0/::), there could be a port forward active just the same (and thus inbound connections from anywhere).
The only listening host that could be exempted from this check is localhost, but that is already taken care of by the `is_loopback_addr(remote_ip)` check.

For less common cases such as private ranges that aren't actually private or reverse proxy setups (remote_ip appears as loopback but could have been forwarded from anywhere) there isn't much that can be done without making things break by default in typical home and small office networks.

@puzzledsab spoofing the remote ip address cannot be used to avoid these checks. Sab only does tcp, which requires a three-way handshake before a connection is established. Fake source address = no connection at all.

sabnzbd/interface.py
jcfp: Could just get rid of that notation altogether when assigning remote_ip - in a way that cannot break actual ipv6 addresses:
```
if remote_ip.startswith("::ffff:") and not remote_ip.find(".") < 0:
    remote_ip = remote_ip.replace("::ffff:", "")
```

thezoggy:
A use case not mentioned here is those that connect to a remote sab (seedbox/friends/parents) and they use the ability to add static wan up/prefixes to grant access or not have to deal with auth when managing sab. So limiting to only local ip would take away that functionality

jcfp:
@thezoggy afaict that kind of setup would continue to work. This changeset only blocks non-lan addresses in case no `cfg.local_ranges()` have been set.

PR #1860, opened 2021-04-18
Make num_decoders configurable through special variables

puzzledsab:
Hopefully increasing it can increase the download speed for some users with more than 4 cores.

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -1371,6 +1371,7 @@ SPECIAL_VALUE_LIST = (
     "max_foldername_length",
     "show_sysload",
     "url_base",
+    "num_decoders",
     "direct_unpack_threads",
     "ipv6_servers",
     "selftest_host",


Safihre:
Do you actually see a benefit on your system from increasing it? Because on my i7 system, increasing it doesn't help.
All the extra thread-switching that Python has to do seems to remove the benefit, on my slow Intel Celeron it actually slows things down.
When introducing the multi-decoder system it actually had the option to set `num_decoders` in Config Specials, but I removed that after testing.

puzzledsab:
I don't have a good way to test it myself, that's why I wanted to make it easier for others. The overhead in context switching in Python is very low for normal threads, but maybe there is additional overhead related to calling C stuff. If so, it might be a good idea to send multiple articles to the decoder. A simple way get an hunch about how much overhead there is would be to modify the decoder to decode each article 10 times and see how that influences the performance. I don't know how to do that on Windows, but maybe you could test it?

Safihre:
I didn't test the actual numeric performance, just the experience. But we can add it and see if it helps someone!

thezoggy:
if we are going to keep this as exposed we need to add it to wiki

PR #1862, opened 2021-04-22
Set nzo.deleted before calling logging.info

### LINKED ISSUE #1861: Sabnzbd duplicate download job after repair failed
Hi Dev's of Sabnzbd,

I like to report an issue according to the strange behavior of sabnzbd.
Initially, the issue was reported on the GitHub of Sonarr (https://github.com/Sonarr/Sonarr/issues/4451).
After some Discord chatting with Markus from Sonarr, we figured out that sabnzbd duplicates (3 times) the download job when sab tries to repair a broken download.
the duplication creates confusion on the end of Sonarr because the unique ID returned 3 times with different status from an API call from Sonarr.
At the end Sonarr got confused and can't track the status correctly & import the file.
image

You can see the behavior visually in the web GUI of sabnzbd. The same entry appears 3 times in the history tab:
image

This issue is reproducible with an nzb file that I can provide, debug logs also available if I can share it in a safe way.

puzzledsab:
I think this fixes #1861. I tested it many times and I could never make it happen when nzo.delete was set before logging.info.

I frequently got the problem with multiple items as well without this change, but for me I would get one that finished properly and one or two hanging and never finishing. They could not be deleted manually, but disappeared after a reset. Maybe this can help the issue some have seen with downloads that never finishes as well. Curiously I got 1 item that never finished when I had num_decoders=2 and 2 when it was set to 3, but that may have been a coincidence.

### DIFF
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -773,8 +773,8 @@ class NzbQueue:
         """ Send NZO to the post-processing queue """
         # Notify assembler to call postprocessor
         if not nzo.deleted:
-            logging.info("[%s] Ending job %s", caller_name(), nzo.final_name)
             nzo.deleted = True
+            logging.info("[%s] Ending job %s", caller_name(), nzo.final_name)
             if nzo.precheck:
                 nzo.save_to_disk()
                 # Check result


puzzledsab:
That's technically correct, but going directly from the if to the setting of the variable takes just a fraction of the time it takes to call logging.info first. In practice it probably won't happen ever, given how rarely this already occurs. The performance penalty of a lock would probably be minimal, though.

puzzledsab:
Come to think about it, logging.info probably does I/O which releases the thread lock and makes the problem happen.

Safihre:
Setting number of decoders to 8 it was easier to reproduce, but the lock fixed it I think.

PR #1869, opened 2021-04-28
Minor optimizations

puzzledsab:
I'm trying to split the changes in my own fork into bite size PRs, starting with the smallest ones. Regarding the continue/break, I don't see how any of the following files could have been imported if the first it encounters hasn't.


Safihre:
👍

PR #1870, opened 2021-04-28
Add article queue and change article tries system

puzzledsab:
- Please help me check that reset_article_queue is called in all the right places.
- The retries change could be a different commit, but it's a fairly small part.
- I still think there should be a delay in the marking of expired articles as tried, but according to my tests this version does it more than 10 times faster than the current develop so it doesn't make much difference when there is no repickling.
- A large portion of the speedup comes from using nzf.get_articles directly instead of going through NzbQueue.get_article. That's why there's not much point in fetching more than 4.


Safihre:
Isn't it easier (and cleaner) to use `nzo.get_articles` instead of `nzf.get_articles`? This way you don't have to do special checks for `lowest_partnum`, since that's part of the `nzo.get_articles`.

sabnzbd/downloader.py
```diff
@@ -532,21 +548,35 @@ def run(self):
                             server.request_info()
                         break
 
-                    article = sabnzbd.NzbQueue.get_article(server, self.servers)
+                    try:
+                        if server.article_queue[0].nzf.nzo.status == Status.PAUSED:
```
Safihre, line 552: I don't think this logic should be here. It's part of the pause-event to clear the queue, so maybe this should be part of `nzo.pause`.

puzzledsab: I'm not sure how well that would work if I understand you correctly. `nzo.pause` would have to go through all the server queues and look for articles belonging to that nzo, and there's the issue of locking. It could find an article, then the pause is called which clears the queue, and then that article is used to add 4 more articles to the queue. All those articles would be downloaded before it notices the pause.

puzzledsab:
> Isn't it easier (and cleaner) to use `nzo.get_articles` instead of `nzf.get_articles`?

A good portion of the speed up comes from bypassing `nzo.get_articles`, more than from bypassing `NzbQueue.get_article`. It has to loop through all files until it finds one with articles for that server, and call `nzf.server_in_try_list` for each of them. All that takes much more time than a simple boolean check. `nzf.get_articles` would still have to be the part that fetches multiple articles, otherwise it would have to start the loop through the articles all over for each call.

Safihre:
> A good portion of the speed up

Did you do by any chance some tests to see what % improvements this PR has? If its a huge amount I would be more tempted to accept less clean code..

puzzledsab:
I've tried, but I don't know how trustworthy they are. These are from Yappi:
```
                                                                                   ncall       tsub      ttot        tavg
develop nzbqueue.py:705 NzbQueue.get_article                                       22273       0.1718  1.765625    0.000079
queued nzbqueue.py:706 NzbQueue.get_article                                        12954       0.0104  0.343750    0.000027

develop nzbstuff.py:1584 NzbObject.get_article                                     9177        0.1666  1.375000    0.000150
queued nzbstuff.py:1588 NzbObject.get_article                                      49          0.0000  0.281250    0.005740

develop nzbstuff.py:430 NzbFile.get_article                                        9162        0.3750  0.890625    0.000097
queued nzbstuff.py:430 NzbFile.get_articles                                        2324        0.1875  0.421875    0.000182

develop nzbstuff.py:180 Article.get_article                                        46303       0.2604  0.359375    0.000008
queued nzbstuff.py:180 Article.get_article                                         20973       0.1666  0.234375    0.000011

develop nzbstuff.py:116 NzbObject.server_in_try_list                               41236       0.2135  0.234375    0.000006
queued nzbstuff.py:116 NzbObject.server_in_try_list                                22752       0.0729  0.046875    0.000002
```
The first number is the number of calls to that functions. The second is the average of 3 runs with the same nzb using the tsub value from Yappi, which is supposed to measure how much time was used in that function, but not sub functions. The two last numbers are just the first result in my grep. I don't think they're quite as useful as they include all the sub calls, so there is overlapping.

This is using an nzb file what was downloaded using my top level server so it's a best case scenario for the develop version. The benefits of skipping NzbQueue and NzbObject can be significantly higher when there are multiple levels of servers in use at the same time, if they have to loop through more nzbs and files to find valid articles. As I mentioned, marking articles as tried because they are expired is more than 10 times faster.

The total time used by SABnzbd for all 3 runs according to Yappi is 54.76 for the queued version and 57.79 for develop, so just over 5% reduction.

puzzledsab:
Not really relevant for this commit, but I was looking into the SSL performance and stumbled over this: https://github.com/python/cpython/pull/25478

Looks like it could help if they figure out how to solve the issues.

sabnzbd/downloader.py
Safihre: Then I think we can just remove this check. It's just not the place for it. Worst case it downloads 4 articles, which is like 750kb * 4.
It also doesn't account for changing the order of the queue for example, or setting job priority to Force and then back to normal etc etc. That's all fine for 4 articles.

puzzledsab: Unfortunately it would keep reusing articles from that file to download more until it was fully downloaded. I've changed that so that now it will add (threads / 5) articles once instead. That should reduce the problem with downloading articles.

The main advantages of the patch are a hopefully better retries system and much faster marking of expired articles as tried.

Safihre:
I hope you don't mind, but I pushed some changes to make it a bit more general. Now `NzbQueue.get_articles` just returns the requested number of articles.
For now I fixed the `fetch_limit=5`, because with `int( nr_threads/5)` you can get:
```python
int(4/5)
0
```
So probably needs to be `max(1, server.threads // 5)` (the `//` operator is faster than `int()`, but that's a nano-optimization).
Also wonder if `reset_article_queue` shouldn't just be part of `__reset_nw`? Since all the cases so far seem to also trigger that function.

The tries system looks fine. ✔️

puzzledsab:
It's actually more similar to the version I made a couple of years ago now, and when it doesn't reuse the article to load more it's probably a little bit faster for normal usage. `nzf.get_articles` will return at least one article if there are any so it wasn't necessary to check the limit for 0 with my version. If my last commit didn't mess up anything I think it's ok now.

puzzledsab:
One more thing, I think the comment `# Move to the next idle thread` is incorrect. It will move to the next server, not the next idle thread. Did you intend to do what it does or what the comment says?

Safihre:
> One more thing, I think the comment `# Move to the next idle thread` is incorrect.

You are right! 👍 

Some more questions:
1. `max(1, server.threads // 5)` maybe it can be a bit more? Since the default number of connections is 8, this would not be beneficial for users that are not tweaking their number of connections. How about `max(1, server.threads // 4)`?
2. Still wonder if `reset_article_queue` shouldn't just be part of `__reset_nw`? Since all the cases so far seem to also trigger that function.

puzzledsab:
1. Sure, or you could do + 1 or 2 for everything. I don't think this scales linearly because at some point it's better to do two shorter calls than one longer.
2. Forgot about that part while I looked at the other stuff. I don't think they quite the same, and I think I may have added                `reset_article_queue()` to more places than I should. Maybe it should be called when `__reset_nw` is called with retry = False, but that's partly a coincidence. I'll try to wrap my head around it when I'm less tired.

puzzledsab:
I think `reset_article_queue()` should only be called if SAB gives up on connecting to a server for a while or the downloading is stopped. `__reset_nw()` fixes problems with a specific connection so I don't think it should be called there. Instead, I moved it to `plan_server()`.

Safihre:
Alright, looks good.
Might have missed something, but we will do a few weeks of RC releases to work out all the edge cases we might have missed!

puzzledsab:
@Safihre, I think you should fork at the Beta 4 release and only merge fixes. It will make more people upgrade than the 3.21 release did. The stuff that was merged since B4 changes a lot and there is always a 3.4.

Safihre:
Really? We've worked/debated (😁) these changes so much, would be nice to have people using them.

puzzledsab:
There have already been 4 betas and these changes won't be noticed by most people. Beta testing fatigue will cause fewer and fewer to test the new releases, resulting in a rising chance for undiscovered bugs in edge cases.

There is already more than enough new features to justify a release. Regarding the par2 change I think you should make it clearer just how much faster it is because it will help some cases a lot on slower CPUs.

Safihre:
Yes, that's why I wanted to make a release candidate that includes these latest changes. Most of our beta testers that provide feedback use the Linux PPA so they get auto updated.

When you have the time, maybe it's worth to do a speed test again with your Rpi3 and compare 3.2.1 to develop? Wonder how all these changes over the past weeks combined affect the results.

thezoggy:
Also for giggles py3.9.5 came out last Monday. So won't hurt to have newer build with that. 

Have either of you tested sab with pyston? 
> Pyston is a fork of CPython 3.8.8 with additional optimizations for performance. 
https://github.com/pyston/pyston

puzzledsab:
```
develop      :  756.28 (750.97/764.43)   user 252.16 (244.71/258.73) system 69.75% CPU 4 rounds
develop-nopar:  677.93 (673.24/681.09)   user 231.61 (227.71/233.91) system 68.00% CPU 3 rounds
master       : 1134.85 (1131.06/1138.64) user 206.13 (205.58/206.68) system 72.50% CPU 2 rounds
master-nopar :  694.60 (693.01/696.08)   user 178.64 (174.71/182.89) system 63.25% CPU 4 rounds
```
I tested with the same nzb, but also tried a version where I removed all par2 files before downloading. I don't know why system time has gone up and I'm not sure how to find out.

Let's hope the SSLSocket patch is included sooner rather than later. Maybe it would be possible to release it as an installable module? `NewsWrapper.recv_chunk` and SSLSocket functions use 60-70% of the GIL CPU time now, according to Yappi. I looked some more into asyncio but I'm unable to make it use less CPU than SAB currently does.

Safihre:
Nice!
Indeed if somehow we could distribute that patch, but I wouldn't know how..

PR #1871, opened 2021-04-28
support prefix and netmask for local_ranges

jcfp:
In issue [#1859](https://github.com/sabnzbd/sabnzbd/pull/1859#discussion_r615269361) I noticed how the local_ranges setting couldn't handle the common prefix and netmask notations for networks. Especially now that the setting was moved to specials, more flexibility for power users would be nice. The old style local_ranges settings continue to work; the changes should be backward compatible with the exception of some irrelevant corner cases (see the _Bad old-style setting_ example in the tests).

While working on the above, I noticed weird results from some of the `misc.is_*_addr` functions when used with IPv4-mapped IPv6 addresses (e.g. `::ffff:192.168.1.1`). The root cause turned out to be several issues ([33433](https://bugs.python.org/issue33433), [38655](https://bugs.python.org/issue38655)) in the ipaddress library, so I added a function to strip that notation from ip addresses and applied it where needed plus some extra params to the tests for the affected functions.

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -201,7 +201,6 @@ def check_access(access_type: int = 4, warn_user: bool = False) -> bool:
     else:
         is_allowed = any(ip_in_subnet(remote_ip, r) for r in cfg.local_ranges())
 
-    # Reject
     if not is_allowed and warn_user:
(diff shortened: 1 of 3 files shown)


Safihre:
Thanks! Could you also update the documentation? :)

PR #1876, opened 2021-04-30
Stop reading par2 when all files have been found

puzzledsab:
I figured that there can theoretically be shared files between par2 sets but if the same filename is found again in the same par2 file then it must have finished reading one set of packets and found a duplicate. In my test case it reduced the CPU time spent in `parse_par2_file_packet` from 364 to 84 seconds.

### DIFF
--- a/sabnzbd/par2file.py
+++ b/sabnzbd/par2file.py
@@ -93,6 +93,7 @@ def parse_par2_file(fname: str, md5of16k: Dict[bytes, str]) -> Dict[str, bytes]:
     """
     table = {}
     duplicates16k = []
+    seenfiles = []
 
     try:
         with open(fname, "rb") as f:
@@ -100,6 +101,9 @@ def parse_par2_file(fname: str, md5of16k: Dict[bytes, str]) -> Dict[str, bytes]:
             while header:
                 name, filehash, hash16k = parse_par2_file_packet(f, header)
                 if name:
+                    if name in seenfiles:
+                        break
+                    seenfiles.append(name)
                     table[name] = filehash
                     if hash16k not in md5of16k:
                         md5of16k[hash16k] = name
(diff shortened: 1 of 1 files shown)


sabnzbd/par2file.py
```diff
@@ -93,13 +93,17 @@ def parse_par2_file(fname: str, md5of16k: Dict[bytes, str]) -> Dict[str, bytes]:
...
     duplicates16k = []
+    seenfiles = []
 
     try:
         with open(fname, "rb") as f:
             header = f.read(8)
             while header:
                 name, filehash, hash16k = parse_par2_file_packet(f, header)
                 if name:
+                    if name in seenfiles:
```
Safihre, line 104: Good catch, indeed we can assume all `FileDesc` is unique.
But I don't think we need `seenfiles`, why not just use `name in table`?

puzzledsab:
Time spent in `parse_par2_file_packet` is reduced to 1.7 seconds with the latest patch, so that's decent improvement :)

I still kept the duplicate check (using `table`, of course, I missed the obvious) because then we won't have to read all of the big par2 files. There's a small drawback, though. The creator id package is at the end of the par2 file, so it won't be found when there are duplicate files. It's only used for the debug logging and it will still be found in the smaller packages so I figured it was ok.

Safihre:
I took the time this morning to really dive into the par2 spec and found that the real slowdown was in the part:
```python
for offset in range(0, pack_len, 8):
```
Which is not needed at all, because the `offset` is always `16`. 
So it would do wayyyy too many loops.
I pushed the change to develop.

I tested on a smaller par2 file, running `parse_par2_file` 100x, and went down from 15 seconds to 1.7.
On a large 366MB (from the 10GB-test-download) par2 file it went down from 12 seconds to 1 second.

But I do have a super fast SSD, so maybe it can still be beneficial to prevent it having to read the whole giant file?

PR #1878, opened 2021-05-01
Don't slice data

puzzledsab:
I noticed the 16 byte offset too, I wasn't 100% sure if it could be trusted to always read the data from the right spot. That's why I kept the loop. However, the slicing is quite costly too, so my version was actually still faster (1.76 CPU time in `parse_par2_file_packet` instead of 4.15 for the current develop). This version takes it down to 1.72 seconds.

### DIFF
--- a/sabnzbd/par2file.py
+++ b/sabnzbd/par2file.py
@@ -100,6 +100,8 @@ def parse_par2_file(fname: str, md5of16k: Dict[bytes, str]) -> Dict[str, bytes]:
             while header:
                 name, filehash, hash16k = parse_par2_file_packet(f, header)
                 if name:
+                    if name in table:
+                        break
                     table[name] = filehash
                     if hash16k not in md5of16k:
                         md5of16k[hash16k] = name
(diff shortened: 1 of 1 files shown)


Safihre:
Sounds good. I wonder if your fix to stop reading if the same filenane is detected shouldn't also be applied?
I assume this would stop it reading the whole par2 file, right?

puzzledsab:
Yes, it does. It doesn't make much difference on an SSD, but it does reduce the parsing time as much as you'd expect when it only needs to read a small portion of the file.

puzzledsab:
I also thought about using the Recovery Set ID. If a duplicate file or the par2 creator package is reached then it means that all files have been read, so it could be added to a list of RS IDs that can be skipped if they are seen again. What do you think?

Safihre:
According to the documentation, packages can be placed in any order. So technically the recovery can be also at the beginning.
Wonder if that could also mean that duplicate files can be found?

Safihre:
@animetosho Maybe you can help us: when parsing `.par2` files, can we assume that `FileDesc`-packets will not repeat before all of them are listed?
So can we stop parsing the file once we see the same `FileDesc` filename again?

thezoggy:
http://parchive.sourceforge.net/docs/specifications/parity-volume-spec/article-spec.html#i__134603784_644

Safihre:
@puzzledsab I decided to parse the `Main` packet to find the number of files in the set. This resulted in never parsing the creator-informatoon, so to still get creator-information I decided to scan all files <10MB fully.
https://github.com/sabnzbd/sabnzbd/commit/22f1d2f642eaddedc74e578549083d0816f2e0d4

animetosho:
> @animetosho Maybe you can help us: when parsing .par2 files, can we assume that FileDesc-packets will not repeat before all of them are listed?

Specification doesn't guarantee this, though most applications probably will adhere to that. As you've mentioned, you can use the number of files as a guide to decide when to quit parsing.

PR #1880, opened 2021-05-01
also test with (partially) exploded ipv4-mapped addresses

jcfp:
After all, that was the reason for not sticking with string comparison in `misc.strip_ipv4_mapped_notation()`.


thezoggy:
saw this and made me think of you @jcfp https://www.bleepingcomputer.com/news/security/python-also-impacted-by-critical-ip-address-validation-vulnerability/

PR #1884, opened 2021-05-03
Check busy threads less often if nothing is wrong

puzzledsab:
This part uses 6.25% of the CPU time with just 8 threads according to Yappi, possibly much more with more threads. I think this is a safe and simple way to reduce it significantly.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -94,6 +94,8 @@ class Server:
         self.password: Optional[str] = password
 
         self.busy_threads: List[NewsWrapper] = []
+        self.check_busy_threads: bool = True
+        self.next_busy_check: int = 0
         self.idle_threads: List[NewsWrapper] = []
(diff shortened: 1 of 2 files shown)


Safihre:
Sounds good.
Should we introduce a constant? Maybe it can be the same the same for:
```
server.next_article_search = now + 0.5
```

puzzledsab:
I'm not sure, the use cases are a bit different. `next_article_search` could probably wait several seconds without causing problems, but `next_busy_threads_check` should be delayed less than 1 second for the timeout check. I don't mind either way.

Safihre:
Since timeouts are defined in full seconds, checking every `0.5` should be enough?
Or am I forgetting something?

puzzledsab:
I was a bit worried that `nntp.error_msg` could be set in some magical way I'm not aware of, but if it's not it should be ok.

PR #1885, opened 2021-05-03
Check header before entering parse_par2_packet

puzzledsab:
Reduces CPU time spent in parse_par2_file + parse_par2_packet from 7.8 seconds to 2.1 in my test.


Safihre:
Aah yes, nice. 
Maybe also remove the `offset` variable?

PR #1887, opened 2021-05-03
Use __slots__ on Server object

Safihre:
@puzzledsab I think you like this.
It shows with my tiny test-script about ~15% improvement in accessing attributes of the `Server` object, which we do a lot of course.

```python
import timeit
import sabnzbd.downloader
from statistics import mean

total = []
a=sabnzbd.downloader.Server('a','test','test',80,30,1,1,1,1,1,1)
what="a.idle_threads"

for _ in range(10):
    total.append(timeit.timeit(what, number=10000000,
                        setup="import sabnzbd.downloader; a=sabnzbd.downloader.Server('a','test','test',80,30,1,1,1,1,1,1)"))


print(mean(total))
```

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -57,6 +57,40 @@ TIMER_LOCK = RLock()
 
 
 class Server:
+    # Pre-define attributes to save memory and improve get/set performance
+    __slots__ = (
+        "id",
+        "newid",
+        "restart",
+        "displayname",
+        "host",
+        "port",
+        "timeout",
+        "threads",
+        "priority",
+        "ssl",
+        "ssl_verify",
+        "ssl_ciphers",
+        "optional",
+        "retention",
+        "send_group",
+        "username",
+        "password",
+        "busy_threads",
+        "next_busy_threads_check",
+        "idle_threads",
+        "next_article_search",
+        "active",
+        "bad_cons",
+        "errormsg",
+        "warning",
+        "info",
+        "ssl_info",
+        "request",
+        "have_body",
+        "have_stat",
+    )
+
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


Safihre:
Similar 15-20% improvement when setting-attributes.

puzzledsab:
Wohoo :) Now I'll need to read more about stuff like that too, I only know the generic optimization methods.

thezoggy:
neat, never knew this existed myself
https://book.pythontips.com/en/latest/__slots__magic.html

Safihre:
I only applied it previously to save memory on the Try List objects such as the article objects. At that time I didn't see the point of doing it for the Server objects, since we only have very few. But then today I read it also results in performance increases for getting/setting, so definitely worth it!

Safihre:
@puzzledsab Since you like micro-optimizations, check this out:
```python
import timeit
from statistics import median,mean

total = []

for _ in range(30):
    total.append(timeit.timeit("sabnzbd.cfg.enable_https_verification()", number=100000,
                        setup="import sabnzbd.cfg"))
print(100*mean(total))

for _ in range(30):
    total.append(timeit.timeit("enable_https_verification()", number=100000,
                        setup="from sabnzbd.cfg import enable_https_verification"))
print(100*mean(total))


for _ in range(30):
    total.append(timeit.timeit("cfg.enable_https_verification()", number=100000,
                        setup="import sabnzbd.cfg as cfg"))
print(100*mean(total))
```
Results in:
```
2.9916163333333348
2.7158140000000013
2.7330454444444445
```
So using `sabnzbd.cfg.X` is between 5-10% slower.
Again has to do with the look-up Python has to do if we import it from the `sabnzbd.X` namespace.

PR #1888, opened 2021-05-04
BPSMeter optimizations

puzzledsab:
Reduces CPU usage from `BPSMeter.update()` from 22 to 2.5 seconds in my tests. `Downloader.run` does not seem to use more CPU when updating the caching variables directly.

You should probably wait with this until 3.3.0 is released. I may have missed some obscure call which leaves variables uninitialized.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -142,6 +142,7 @@ class Server:
         self.request: bool = False  # True if a getaddrinfo() request is pending
         self.have_body: bool = True  # Assume server has "BODY", until proven otherwise
         self.have_stat: bool = True  # Assume server has "STAT", until proven otherwise
+        sabnzbd.BPSMeter.init_server_stats(server_id)
 
         for i in range(threads):
             self.idle_threads.append(NewsWrapper(self, i + 1))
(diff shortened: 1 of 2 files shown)


sabnzbd/downloader.py
```diff
@@ -217,26 +218,6 @@ def __repr__(self):
...
-        "shutdown",
-        "server_restarts",
-        "force_disconnect",
-        "read_fds",
-        "servers",
-        "server_dict",
-        "server_nr",
-        "timers",
-    )
-
```
Safihre, line 239: Why remove these?

puzzledsab: No idea why that happened, probably my mediocre Git skills.

puzzledsab:
I'll test a less radical approach for the variable initialization to make it more compatible with the tests.

thezoggy:
You can fix locally, rebate and force push to rewrite to keep it clean when it's merged

Safihre:
@puzzledsab I don't think moving `BPSMeter` to `Downloader` is needed, that might be a bit too much of a micro-optimization that would deviate from all the other classes that we access from `sabnzbd.X`. So maybe let's skip that optimization.

PR #1889, opened 2021-05-04
BPSMeter optimalizations

puzzledsab:
Same as last time, but without the stats initialization. This version used 2.7 CPU seconds in my test, 0.2 up from the previous version but still a lot down from the 22 seconds in the current develop.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -657,23 +657,22 @@ class Downloader(Thread):
             if readkeys:
                 read, _, _ = select.select(readkeys, (), (), 1.0)
 
-                if self.can_be_slowed != 0:
-                    # Add a sleep if there are too few results compared to the number of active connections
-                    if self.can_be_slowed:
-                        if len(read) < 1 + len(readkeys) / 10:
-                            time.sleep(self.sleep_time)
-                    else:
-                        # Need to initialize the check during first 20 seconds
-                        # Wait for stable speed to start testing
-                        if not self.can_be_slowed_timer and BPSMeter.get_stable_speed(timespan=10):
-                            self.can_be_slowed_timer = now
+                # Add a sleep if there are too few results compared to the number of active connections
+                if self.can_be_slowed and len(read) < 1 + len(readkeys) / 10:
+                    time.sleep(self.sleep_time)
 
-                        # Check 10 seconds after enabling slowdown
-                        if self.can_be_slowed_timer and now > self.can_be_slowed_timer + 10:
-                            # Now let's check if it was stable in the last 10 seconds
-                            self.can_be_slowed = BPSMeter.get_stable_speed(timespan=10)
-                            self.can_be_slowed_timer = 0
-                            logging.debug("Downloader-slowdown: %r", self.can_be_slowed)
+                # Need to initialize the check during first 20 seconds
+                if self.can_be_slowed is None or self.can_be_slowed_timer:
+                    # Wait for stable speed to start testing
+                    if not self.can_be_slowed_timer and BPSMeter.get_stable_speed(timespan=10):
+                        self.can_be_slowed_timer = time.time()
+
+                    # Check 10 seconds after enabling slowdown
+                    if self.can_be_slowed_timer and time.time() > self.can_be_slowed_timer + 10:
+                        # Now let's check if it was stable in the last 10 seconds
+                        self.can_be_slowed = BPSMeter.get_stable_speed(timespan=10)
+                        self.can_be_slowed_timer = 0
+                        logging.debug("Downloader-slowdown: %r", self.can_be_slowed)
 
             else:
                 read = []
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -657,22 +657,23 @@ class Downloader(Thread):
             if readkeys:
                 read, _, _ = select.select(readkeys, (), (), 1.0)
 
-                # Add a sleep if there are too few results compared to the number of active connections
-                if self.can_be_slowed and len(read) < 1 + len(readkeys) / 10:
-                    time.sleep(self.sleep_time)
+                if self.can_be_slowed != 0:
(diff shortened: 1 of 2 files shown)


sabnzbd/downloader.py
```diff
@@ -711,22 +716,23 @@ def run(self):
...
                     try:
-                        article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, bytes_received)
+                        article.nzf.nzo.update_download_stats(BPSMeter.bps, server.id, bytes_received)
                     except AttributeError:
                         # In case nzf has disappeared because the file was deleted before the update could happen
                         pass
 
-                    sabnzbd.BPSMeter.update(server.id, bytes_received, force_full_update=False)
+                    BPSMeter.cached_amount[server.id] += bytes_received
+                    BPSMeter.sum_cached_amount += bytes_received
```
Safihre, line 725: Not sure if I can approve of this.. We shouldn't be accessing `BPSMeter` internals outside. 
If this is not moved here but kept in the `.update()`, how much is the improvement still?
I kind of assume most of the improvement is due to the initialization of the variables beforehand.

puzzledsab: Hang on, I'll do some tests with only the time command to see if it's yappi that messes it up.

JackDandy:
```
It seems like it's the call itself that takes time, not what it does
```
Correct. We found exactly the same overhead savings by switching to attributes from function properties. Yes, the call itself takes time.

Safihre: In this case I would really like to see if it makes noticeable % difference in actual download performance. Clean code is also valuable.
Calling functions indeed is overhead, but that could be extended too far where everything is just thrown in the main Downloader loop.

puzzledsab:
The results are finally in...
```
develop      : 765.01 (760.03/777.21) user 254.76 (246.68/260.79) system 68.85% CPU 13 rounds
update       : 758.55 (753.13/764.81) user 256.40 (248.17/260.67) system 69.33% CPU 12 rounds
cached_amount: 753.90 (745.69/763.60) user 253.03 (249.02/257.36) system 68.64% CPU 11 rounds
nosum        : 752.65 (746.66/757.30) user 252.17 (245.42/257.27) system 68.67% CPU 6 rounds
```
Numbers are averages from time (min/max) for downloading my test nzb on an underclocked Rpi 3. I removed one outlier from update and one from cached_amount because of unusually high system time. I realized that the `sum_cached_amount` probably wasn't a good idea so I removed it and added it to the test. I used the direct cached_amount version but the differences are probably the same with `update()`.

The differences are there but not as clearly as in Yappi. Even though I would have used `cached_amount` directly, I assume you won't.

Safihre:
> Even though I would have used `cached_amount` directly, I assume you won't.

😁😁 Indeed!

So in your tests, the current version in this PR is `update` or `nosum`?

puzzledsab:
Yes, it uses update and the sum_cached_amount is removed. My guess is that it would use 757.30 user seconds.

Safihre:
So about 1-2%. Not earth-shattering, but still nice.

puzzledsab:
I just realized that the `can_be_slowed` test doesn't actually try to sleep while it's measuring so it can't know if doing so will slow it down.

sabnzbd/downloader.py
```diff
@@ -654,27 +657,28 @@ def run(self):
...
-                    if not self.can_be_slowed_timer and sabnzbd.BPSMeter.get_stable_speed(timespan=10):
-                        self.can_be_slowed_timer = time.time()
-
-                    # Check 10 seconds after enabling slowdown
-                    if self.can_be_slowed_timer and time.time() > self.can_be_slowed_timer + 10:
-                        # Now let's check if it was stable in the last 10 seconds
-                        self.can_be_slowed = sabnzbd.BPSMeter.get_stable_speed(timespan=10)
-                        self.can_be_slowed_timer = 0
-                        logging.debug("Downloader-slowdown: %r", self.can_be_slowed)
+                if self.can_be_slowed != 0:
```
Safihre, line 660: I don't understand why this is needed? 
Below is `self.can_be_slowed = BPSMeter.get_stable_speed(timespan=10)`, so it triggers the slow-down after 10 seconds of stable speed. After that it sets the timer `self.can_be_slowed_timer` while the slowdown is active and measures again for 10 seconds.

Safihre: It has also worked like this since 2.x so I'm pretty sure it's working 😄

puzzledsab: In the develop code, the sleep block is only entered if `self.can_be_slowed` is true. It can only be set true by https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/downloader.py#L654, but that block also sets `self.can_be_slowed_timer = 0`. The initialize block is only entered if `self.can_be_slowed is None` or `self.can_be_slowed_timer` is true, so as I under stand it, that `if` can't be true at the same time as `self.can_be_slowed` is true.

puzzledsab: The point of my refactor was to make it easier to read and not do more checks than necessary for all outcomes. If it is disabled, my code only needs to do one test. The current code does 3. (`if self.can_be_slowed` and `if self.can_be_slowed is None or self.can_be_slowed_timer`)

puzzledsab:
I think the can_be_slowed system needs a rewrite, but not one that belongs in this PR.

Safihre:
> I think the can_be_slowed system needs a rewrite, but not one that belongs in this PR. Let me know if you're working on it, if not I'll try.

Yes indeed. 
I have looked at this stupid problem so many times that I doubt I can find a better way, if you know one, that would be great :)

thezoggy:
out of curiosity, if one was to disable bpsmeter in the gui (only keep it for metrics after the fact). wonder if that makes a noticeable impact in the download speed when your not having to continuously record/update things

PR #1891, opened 2021-05-08
Changes to determining can_be_slowed

puzzledsab:
Changes:
- Change `can_be_slowed` to float because that's what `get_stable_speed()` will return if the speed is stable, and make the `get_stable_speed()` result consistent with that.
- Add a quick check to determine if speed fell by at least 15% in `get_stable_speed()`.
- Make `can_be_slowed_timer` local to the run method, I don't think it's useful outside of it and I added another timer variable as well.
- Add `next_stable_speed_check` to slow down polling of `get_stable_speed()`. It was using so much CPU that it could slow down the speed by itself.
- Only do tests if `self.sleep_time` is non-zero. This makes it possible to turn off the whole thing for end users.
- `self.can_be_slowed` is now set true for the second part of the test.
- Time of the second part is reduced to 8 seconds so that the first `bps_list` result used by `get_stable_speed()` is certain to be the maximum speed from the first part.


puzzledsab:
I have dug into the problem with increased system time usage compared to master in my tests. It seems to be mostly related to decreasing the sleep time from 0.01 to 0.001 seconds. I made a system for checking previous commits and these are the three I found with the largest impact in the last months (newest commit on top, numbers are user and system time):
commit 8c2d7243cca446be93fa1f81818c7f61058acd18 745.37 223.98
commit 49e67a0bef134175162575efdd20e4afabafe129 723.93 209.75

commit 7f702b70251c3c67687bbe29771c412766b0c648 709.74 204.09
commit db255a8b7eb94220621719a20964f0c1de9001eb 667.37 185.79

commit b748b05fbd0503f79493f8f63809cb864ca6ea8b 668.16 184.20
commit 9f2a9c32c09227e730014a37b2b392f4b653c3d0 693.87 188.42

I tried various timings using this branch (sleep time for full decode queue and can_be_slowed system):
```
0.01-0.001: 677.07 (674.66/679.79) user 231.57 (229.35/233.29) system 1309.4 seconds 5 rounds
0.01-0.005: 648.97 (644.73/651.23) user 198.23 (193.55/200.87) system 1274.7 seconds 5 rounds
0.01-0.01 : 640.18 (635.76/643.62) user 195.29 (189.51/201.43) system 1297.0 seconds 4 rounds
0.05-0.005: 651.11 (647.96/653.44) user 199.13 (194.38/201.83) system 1300.1 seconds 5 rounds
0.05-0.01 : 644.14 (640.63/648.37) user 192.76 (189.60/196.49) system 1323.3 seconds 5 rounds
```
So higher sleep time generally results in lower CPU usage but it quickly leads to slower download speed as well. 0.005 seems like the sweet spot in my test. This is on a Raspberry Pi 3 that is underclocked by 30% though, so it's hardly the benchmark we should aim for. I assume it will be lower for faster CPUs but I can't test it myself.

Safihre:
@puzzledsab I want to make an RC1. Should I still change something based on your findings or are we good to go?

puzzledsab:
I'm not sure what would be a good number yet so don't change anything.

puzzledsab:
@sanderjo : I made it a bit more generic and I didn't have anywhere to upload it so I put it on Github: https://github.com/puzzledsab/sabtest. I hope you'll test it.

In addition to the instructions in the README you need to check out the SAB code from this PR, set the processing of the testsab category in the instructions to Download only, and make the incoming/sabtest directory.

For this test it's preferable to use an nzb with articles and files that are as large as possible. I think the ones that are used in the https://sabnzbd.org/tests/ are a bit small, which make them use more CPU on stuff that isn't relevant for this test.

puzzledsab:
I have done some more testing and the numbers are different this time. I now think it's safer to keep it at 0.001.
```
sleep-0.01-0.0005: 707.06 (703.25/711.04) user 249.66 (243.68/255.13) system 1349.0 (1300.39/1403.48) seconds 11 rounds
sleep-0.01-0.001: 676.93 (670.82/682.96) user 228.62 (222.26/235.34) system 1344.8 (1278.28/1382.2) seconds 11 rounds
sleep-0.01-0.002: 661.35 (654.83/664.41) user 209.35 (201.10/218.70) system 1357.2 (1319.09/1406.1) seconds 10 rounds
sleep-0.01-0.003: 654.95 (652.97/656.53) user 204.45 (196.69/211.83) system 1365.5 (1332.66/1418.15) seconds 10 rounds
sleep-0.01-0.005: 649.79 (645.63/654.32) user 197.21 (193.74/200.33) system 1373.5 (1314.23/1460.52) seconds 10 rounds
```

So don't let this stop the progress of the PR.

noodle-:
I'd be happy to test whatever you want me to test on a (slow) linux system.

image

Just let me know :)

sabnzbd/bpsmeter.py
```diff
@@ -425,14 +425,22 @@ def get_bps_list(self):
         # We record every second, but display at the user's refresh-rate
         return self.bps_list[::refresh_rate]
 
-    def get_stable_speed(self, timespan=10):
+    def get_stable_speed(self, timespan: int = 10):
```
Safihre, line 428: Let's add return type typing

sabnzbd/downloader.py
```diff
@@ -243,7 +243,6 @@ class Downloader(Thread):
         "bandwidth_limit",
         "bandwidth_perc",
         "can_be_slowed",
-        "can_be_slowed_timer",
```
Safihre, line 246: I think we can then also make can_be_slowed local?

puzzledsab:
@noodle-: Thanks, but I don't think there's a better general number for downloader_sleep_time. On your system you might be slightly better off setting it to 20 or 30, though.

Issue #206, opened 2015-04-01
SOCKS proxy support


sharkone:
I think it would be great to add SOCKS proxy support.

I saw this post which seems to suggest that it should be a simple change:
https://forums.sabnzbd.org/viewtopic.php?f=4&t=3717

Maybe the only thing missing is UI support?

I guess I could look into it if needed.

Crafter6432:
Any reason why this wouldn't be added ?

Safihre:
None of us use it so we can't really test it. And nobody else made a pull request to add it.. 
You're very welcome to do so :)

Safihre:
I will re-open and maybe take it as an optional feature for 2.4.0 / 3.0.0.

CoderKiwi:
Any gauge on how likely this is to be implemented for the upcoming release @Safihre? I think quite a few users would appreciate this feature (me included).

Safihre:
Hmm not sure. How's your python skills? We welcome pull requests if anyone wants to add it!

CoderKiwi:
I actually did edit the source code in `newswrapper.py` as outlined in the now-ancient forum posts that @sharkone referenced. It was pretty straightforward. I got SOCKS5 proxying working through HTTP, but couldn't figure out how to verify it working through SSL. 

But that was hacky. I hardcoded proxy URLs and credentials within the code. So the main issue is how to add the GUI options to provide a SOCKS5 proxy and make those config options be used by `newswrapper.py` instead of hardcoding values. I haven't looked into how to do that. 

Seems like having me (try to) implement it is the least efficient way to do it though. I haven't actually ever submitted a pull request. The SAB devs should be able to just follow the guide at https://forums.sabnzbd.org/viewtopic.php?f=4&t=3717 to program it. 

I read elsewhere that the main obstacle isn't programming it but testing it. If I donate some money specifically for SOCKS5 proxy subscriptions/trials (i.e. TorGuard) would that make this feature more feasible?

CoderKiwi:
I've read elsewhere that the main obstacle to implementing this feature is a proxy to use for testing. I have credentials to both an HTTP(S) and SOCKS5 proxy service which I can share to help implement this. Would you be interested in that @Safihre?

CoderKiwi:
[...]

PR #1894, opened 2021-05-14
SOCKS 5 support

puzzledsab:
Issue #206
It was a bit frustrating to make it work so I'm not going to spend any time making it pretty until I have some confirmation that it is usable. Hopefully this will build installable versions. I tried PySocks first but I couldn't make it work with SSL. It currently only works for NNTP but HTTP requests shouldn't be too hard.

The configuration is at the bottom of the General config page. It uses blocking calls for making connections, so that takes longer than usual. I don't have a proxy that requires username and password so I haven't tested that.

Be aware that the username and password for the proxy server will be included in the log file if debug logging is enabled.


puzzledsab:
Installable versions for Windows and Mac are available here if you are logged in to Github: https://github.com/sabnzbd/sabnzbd/actions/runs/843529579

sabnzbd/newswrapper.py
```diff
@@ -269,17 +270,31 @@ def __init__(self, nw: NewsWrapper, host):
...
+        proxy_enabled = sabnzbd.cfg.proxy_enabled()
 
         # there will be a connect to host (or self.host, so let's force set 'af' to the correct value
         if is_ipv4_addr(self.host):
             af = socket.AF_INET
         if is_ipv6_addr(self.host):
             af = socket.AF_INET6
 
+        if proxy_enabled:
+            proxy_url = "socks5://"
```
jcfp, line 282: Out of curiosity: why hardcode the proxy type?

puzzledsab: Does anyone use anything other than socks 5?

jcfp: No idea about the popularity of various proxy types, I just noticed the module supports multiple types and your code (constructing a proxy_url parsed by `Proxy.from_url()`) looks like it would work out of the box for all of them.

puzzledsab:
Maybe I'll reopen it later if there is any interest.

dimqua:
Please reopen. :-)

Lord-Malgus:
Would love to see this integrated.

ImportTaste:
I'm also quite interested.

puzzledsab:
It seems that it's not possible or at least difficult to communicate with the proxy using SSL in Python. This means that the username and password is transferred in plain text. The content will be hidden if the connection to the end server is made using SSL but I assume it's possible to see the IP of the server for anyone between the client and proxy. Not sure if this defeats the purpose. Anyone using it should at least be aware that the traffic is not completely hidden.

ImportTaste:
> It seems that it's not possible or at least difficult to communicate with the proxy using SSL in Python. This means that the username and password is transferred in plain text. The content will be hidden if the connection to the end server is made using SSL but I assume it's possible to see the IP of the server for anyone between the client and proxy. Not sure if this defeats the purpose. Anyone using it should at least be aware that the traffic is not completely hidden.

Proxy passwords being stored in plaintext is unfortunately the norm, I don't know why that is. The main purpose, really, is to hide traffic from the ISP, so as long as it accomplishes that, it's fine.

puzzledsab:
If anyone has any suggestions on how to improve this code I'm all ears. I can't find a way to set the proxy for urlopen in a proper way for a specific connection. When I set it using socks.set_default_proxy it also seems to be used by the already proxied nntp connections because my proxy server (dante) complains that it tries to connect to itself. However, if I remove all the nntp specific proxy stuff from python_socks then it doesn't connect to the proxy at all.

I assume this won't work with IPv6 so it should probably be disabled when using proxy.

sabnzbd/utils/internetspeed.py
```diff
@@ -10,6 +10,10 @@
 import logging
 import urllib.request
 
+import sabnzbd.cfg as cfg
```
puzzledsab, line 13: Just noticed that this is useless. I'll fix it in the next batch.

Safihre:
Is it really needed to have a user-friendly input?
Why not just a Special-setting (since this is _really_ a _special_ setting) that has the whole connection string?
E.g.: `socks5://username:password`
Although I now see this notation isn't used anymore.

puzzledsab:
It's normal URL format, socks5://username:password@host:port. Username and password is optional. The socks5 part is ignored after parsing, it is assumed to be socks5. I don't think the failed test is because of the proxy code but I'm not 100% certain.

puzzledsab:
Too many commits? Binaries for the commit that just completed successfully are available here: https://github.com/sabnzbd/sabnzbd/actions/runs/1463556064

I'm fairly happy with the current result. It's not ideal to manipulate the socket like this but at least it ended up adding very little new code. I don't think it should create any problems for those not using a proxy. Let me know if I have missed any requests that should be proxied.

sabnzbd/getipaddress.py
```diff
@@ -101,6 +101,7 @@ def publicipv4():
             # specify the Host, because we only provide the IPv4 address in the URL:
             req.add_header("Host", sabnzbd.cfg.selftest_host())
             # get the response, timeout 2 seconds, in case the website is not accessible
+            socket.socket = sabnzbd.ORIGINAL_SOCKET
```
Safihre, line 104: I don't understand why we would want to return the original socket? When using a proxy wouldn't you want to see exactly the result _with_ proxy?
Would also simplify even some more code so we don't have to switch sockets.
We can just execute it once at startup and done!

sabnzbd/interface.py
```diff
@@ -928,6 +928,7 @@ def saveSwitches(self, **kwargs):
     "selftest_host",
     "rating_host",
     "ssdp_broadcast_interval",
+    "socks5_proxy_url",
```
Safihre, line 931: Thanks :)

sabnzbd/misc.py
```diff
@@ -322,11 +323,29 @@ def set_serv_parms(service, args):
     return True
 
 
+def set_socks5_proxy():
+    if cfg.socks5_proxy_url():
+        proxy = urllib.parse.urlparse(cfg.socks5_proxy_url())
+        logging.debug("Using proxy %s:%s", proxy.hostname, proxy.port)
```
Safihre, line 329: I didn't think of using `urllib`, smart!
Doesn't the library also support other type of proxies, or is only socks5 a relevant one?

sabnzbd/newswrapper.py
```diff
@@ -276,11 +276,17 @@ def __init__(self, nw: NewsWrapper, host):
         if is_ipv6_addr(self.host):
             af = socket.AF_INET6
 
+        sabnzbd.misc.set_socks5_proxy()
+        self.sock = socket.socket(af, socktype, proto)
+        if sabnzbd.cfg.socks5_proxy_url():
+            try:
+                self.sock.connect((self.host, self.nw.server.port))
```
Safihre, line 283: Why does this one have to be seperate from the regular `socket.connect`?

sabnzbd/misc.py
```diff
@@ -322,11 +323,29 @@ def set_serv_parms(service, args):
...
+        socket.socket = socks.socksocket
+    else:
+        socket.socket = sabnzbd.ORIGINAL_SOCKET
+
+
 def get_from_url(url: str) -> Optional[str]:
     """Retrieve URL and return content"""
     try:
         req = urllib.request.Request(url)
         req.add_header("User-Agent", "SABnzbd/%s" % sabnzbd.__version__)
```
Safihre, line 347: See comment above, let's just do it at start-up.

sabnzbd/getipaddress.py
puzzledsab: I was assuming it's trying to find the public IP of the server so that other servers can contact it. That won't work if it uses the proxy. In general, removing the proxy server settings will require a restart if the socket is not stored. As far as I have been able to find out it's not possible to remove the proxy settings from the socket.

sabnzbd/misc.py
puzzledsab: SOCKS4 came four years before SOCKS5, which came 25 years ago. I really doubt anything supports 4 and not 5 these days.

sabnzbd/newswrapper.py
puzzledsab: It has to connect before adding the SSL layer, otherwise it doesn't work. I don't know why but there doesn't seem to be any way around it.

thezoggy:
what about the config > rss ?

puzzledsab:
You're right, that needs to be added as well. I thought it was using one of the other functions to fetch the data. At least feedparser uses urllib so the same method should work.

sabnzbd/getipaddress.py
puzzledsab: I could add a callback to `set_socks5_proxy()` so that it's updated after editing instead of setting it before every request.

Regarding `publicipv4()` I think it should either return the IP of the user or nothing if it's using a proxy, not the IP of the proxy. I can't think of any use cases where that would be useful. In this case I suggest setting the original socket before the request and calling `set_socks5_proxy()` after.

Safihre: I am not a fan of these special set and unset steps, especially for exotic features like these.

Safihre: Especially so we don't have to think about it and possibly forget some request.
So just calling it once at startup should be enough and we can forget about it :)

ImportTaste:
I figure I should share my thoughts in case the maintainers have any reservations about merging this.

ISPs are getting more aggressive about monitoring traffic for 'infringement', and VPNs are not as convenient a solution as they used to be (I have to click on so many fucking boats and busses whenever I visit any website with it turned on).

The VPN software I use does have an experimental feature for exempting applications from tunneling, at least, but it's a bit dodgy. There are also times where I might forget the VPN isn't turned on because I had to disable it (due to all the services that are very aggressive toward VPN use these days).

Having this SOCKS5 option means I don't ever have to worry about any accidents happening after it's set up. VPNs and proxies don't play nice together though, so I'll likely have to have the VPN client exempt the traffic from sabnzbd, hopefully that works.

ImportTaste:
@puzzledsab I notice that this is currently configured to resolve DNS through the proxy. While this is safer and more reliable than using an ISP's DNS servers, it wouldn't have much upside compared to using third-party DNS servers except for preventing accidents, like if a user's DNS server settings get reset back to default for some reason. Though, speaking personally, since I set up DNS servers on my router, the chances of any accident happening there is slim.

Could you make this feature optional (though perhaps enabled by default)?

As an aside, I'd love it if sabnzbd could have its own DNS server configuration, but that's probably going to be deemed outside the scope of this PR.

puzzledsab:
@ImportTaste: Sorry, I think configurable DNS is a bit too much. If you really want it then you can run it from source and change True in line 1093 in `__init__.py` (https://github.com/sabnzbd/sabnzbd/pull/1894/files#diff-5d24cbf8a94f134acee67360be86efe95caca6947d48dbb0c76987ea190fa178R1093) to False.

Safihre:
@puzzledsab Looking great!
Will see if I can find some time to get a socks5 proxy to test this myself, curious to see how it works with SABnzbd and the effect. 
No more comments from my side.

puzzledsab:
Example config file for testing with the dante proxy server (no login required):
```
#debug: 2
internal: eth0 port = 1088
external: eth0
clientmethod: none
user.privileged: root
user.notprivileged: nobody
# Authentication methods
socksmethod: none
#socksmethod: username

client pass {
        from: 0.0.0.0/0 to: 0.0.0.0/0
        log: error connect disconnect
}
client block {
        from: 0.0.0.0/0 to: 0.0.0.0/0
        log: connect error
}
socks pass {
        from: 0.0.0.0/0 to: 0.0.0.0/0
        log: error connect disconnect
        # socksmethod: username
}
socks block {
        from: 0.0.0.0/0 to: 0.0.0.0/0
        log: connect error
}
```
Use `socksmethod: username` for authenticating with a local user. Both works for me.

puzzledsab:
Bad news, I'm get various problems with the global socket configuration. I can't use the performance test on the wrench page, and it seems like PySocks may have problems with IPv6 addresses. So far my only working fix is to use the proxy for specific connections and reset it to the original socket afterwards. The other stuff I've tried has created all kinds of new errors with cherrypy and other parts.

Safihre:
@puzzledsab I just tested it and it works well for me. Nice! 
I pushed 1 change so indeed the IPv6 doesn't crash.
Another thing I noticed would be that if `DEBUG` is enabled that there are a few requests that are not proxied. But we can't move it any earlier as it would crash CherryPy initialization.

You also added support to dynamically turn on/off the proxy, I wonder if we could even remove that and just require a restart after changing the setting. So we don't have to store the `ORIGINAL_SOCKET`.

Safihre:
I also removed the special `sock.connect` bit in `newswrapper` and it still works for me?

puzzledsab:
Downloading works without the early connect but it doesn't use the proxy. I can see this from the output of the proxy I'm running locally.

I don't mind requiring a restart to enable/disable the proxy. It's one less thing that can break.

Safihre:
> Downloading works without the early connect but it doesn't use the proxy. I can see this from the output of the proxy I'm running locally.

Great that you checked!
It bothers me a bit why that would be the case.. Will see if I maybe I can figure out why.

ImportTaste:
> Downloading works without the early connect but it doesn't use the proxy. I can see this from the output of the proxy I'm running locally.
> 
> I don't mind requiring a restart to enable/disable the proxy. It's one less thing that can break.

That would be for the best. Can you make the restart occur automatically after changing the proxy settings? Or at least put something in the UI telling the user that a restart is required for the change to apply.

puzzledsab:
I agree that it would be better if the user was notified but unfortunately it's not possible to add a description for those variables. I will add a description in the wiki later, though. Hopefully most users will do as the text on the top of the page says and read it.

Safihre:
@puzzledsab I didn't know the order could be changed, so now I moved all code after the `connect`.
Can you check if it works with the proxy correctly?

puzzledsab:
@Safihre : The proxy part seems to work fine. No idea if the order can make a difference.

thezoggy:
shame we cant just have a global wrapper and update one place for all calls to funnel through. might be possible if we used request lib then we could benefit from repooling and its own built in proxy support?

Safihre:
@thezoggy Not sure what you mean, it's exactly what we do now. We set the socket once and then all the new connections will use the proxy.

thezoggy:
right but the fact that we have to hunt and find what calls dont use said wrapper to get it to use a proxy is the point

Safihre:
Since the implementation became so graceful, I decided to just add this to the general Config page.
This way we can also force a restart after a change of the setting.

I also want to add some indicator to the Status and Interface Settings Window, only if the proxy is active.

Safihre:
It looks good to me, what do you think @puzzledsab ?

The CI failures are due to some Selenium update.

puzzledsab:
Seems great, I haven't seen any problems while running it.

Safihre:
Thank you @puzzledsab !

Safihre:
Could you maybe update the 3.5 documentation to specify how the connection-string is supposed to be entered?

Lord-Malgus:
Works perfectly, thank you for your work here.

Syntax:
socks5://<username>:<password>@<server address>:<port>

jcfp:
> Syntax: `socks5://<username>:<password>@<server address>:<port>`

I think it would be useful to include the expected input format in the config too, not just on the wiki. Either directly adjacent or as a greyed-out background text in the field itself.

Safihre:
That's on purpose, so that they read the documentation for example about the encryption notice and how to input if you don't use a username and password.

Safihre:
But a `placeholder` wouldn't be bad.

instylegithub:
Can't wait to try out SOCKv5 when this makes it to the main branch!  FYI, if anyone wants a cheap SOCKv5 service for testing, privadoVPN offers it as part of their VPN services. (so you can wrap VPN and SOCKSv5)  So I think this is a great addition to SABnzbd!

PR #1895, opened 2021-05-16
Re-work the Server's reset article queue

Safihre:
@puzzledsab I tried to work on this a bit, because it seemed to me the `reset_article_queue` wasn't in the right spot.
It only needs to be triggered when `server.active` is set to `False`.
Does it look fine to you?

But something is still wrong with the `develop` version and with this one. If a download is active and the primary server gets disabled through Config > Servers, the queue just stops. The current `master` just continue's to the next servers.
Seems to me that the `try_list` is not reset when clearing the `article_queue`. But I added that and it still doesn't work..
Maybe you know why?
In any case, something we need to fix for 3.3.0.


puzzledsab:
I think I may have missed the fact that there is a `server.enable` variable which is not the same as the `server.active`. From what I can tell, nothing is called when a server is disabled. The `article_queue` won't be emptied and the try lists won't be reset.

puzzledsab:
Hm, never mind, it's in the `if server.restart` block.

Safihre:
Found the problem. The old version would "work" due to the `stop_idle_jobs` that would tirgger.
But due to a different refactor the `server_nr` wasn't updated anymore.

puzzledsab:
Does the `server_nr` stuff work if there are inactive servers?

puzzledsab:
Regarding your original question, yes, I see that only calling `reset_article_queue` in `plan_server` can miss some cases where the server is deactivated so that wasn't the best idea.

Safihre:
> Does the `server_nr` stuff work if there are inactive servers?

You are right, it does not..

Safihre:
@puzzledsab What do you think of the changes? Anything I might have missed?

puzzledsab:
I think it's an improvement and I have no additional suggestions.

Issue #1898, opened 2021-05-20
Passwords not tried in the same sequence as provided in the password list file


ErikBrown2:
It appears that SABnzbd is trying the passwords to decrypt jobs from the password list file in a random order. It would make more sense to process the passwords in the order as provided in the password list. That will allow users to put frequently used passwords at the top of the list.

I made a modification to my post processing script so that it re-orders the passwords automatically and removes old passwords. But without processing the passwords in the sequence as listed, that will not have any effect.

Would it be possible to modify this and process the passwords according to their sequence in the password list file?

Below is a sample of the debug log that shows the sequence of the passwords. The first line shows the passwords in the sequence as they appear in the password list file. The following lines show the sequence that SABnznbd is trying these passwords. This appears to be in a random order.

2021-05-20 10:59:31,536::DEBUG::[misc:760] Read these passwords from file: ['TTwsMz', 'bQvucfDDEZJTWuVs', '999', '444', 'Army', 'army', 'UGDV', 'ugdv', 'Bones', 'bones', 'Spikesplace', 'nzbforyou', 'racefan', 'poema', 'dtod', 'Tiger', 'tiger', 'Tigers', 'tigers', 'Tigersplace', 'ilovevintagemusicvideos', 'roccat', 'adrenaline', 'Henan', 'Tarantula', 'pirates4all', 'terak', 'Zon', '456', '789', '234', '555', '888', 'XXXbyTCforTPO', 'aan', 'bosje', 'dwaas', 'makkie', 'qwert']
2021-05-20 10:59:31,536::INFO::[misc:762] Read 39 passwords from file \\?\C:\SABnzbd\PasswordList\MyPasswordList.txt
2021-05-20 10:59:31,536::INFO::[SABnzbd:118] Your password file contains more than 30 passwords, testing all these passwords takes a lot of time. Try to only list useful passwords.
2021-05-20 10:59:31,537::INFO::[assembler:340] Trying password "adrenaline" on job "This_is_my_job_name"
2021-05-20 10:59:31,651::INFO::[assembler:340] Trying password "roccat" on job "This_is_my_job_name"
2021-05-20 10:59:31,733::INFO::[assembler:340] Trying password "terak" on job "This_is_my_job_name"
2021-05-20 10:59:31,816::INFO::[assembler:340] Trying password "555" on job "This_is_my_job_name"
[...]

PR #1900, opened 2021-05-21
Keep password order

puzzledsab:
Issue #1898

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -779,7 +779,12 @@ def get_all_passwords(nzo):
         # If we're not sure about encryption, start with empty password
         # and make sure we have at least the empty password
         passwords.insert(0, "")
-    return set(passwords)
+
+    unique_passwords = []
+    for password in passwords:
+        if password not in unique_passwords:
+            unique_passwords.append(password)
+    return unique_passwords
 
 
 def find_on_path(targets):


sabnzbd/misc.py
```diff
@@ -779,7 +779,12 @@ def get_all_passwords(nzo):
         # If we're not sure about encryption, start with empty password
         # and make sure we have at least the empty password
         passwords.insert(0, "")
-    return set(passwords)
+
+    unique_passwords = []
+    for password in passwords:
+        if password not in unique_passwords:
+            unique_passwords.append(password)
+    return unique_passwords
```
Safihre, line 787: Wouldn't a nice 1 liner list comprehension be better here?

puzzledsab: Is that possible when it's comparing to the list it is generating? I don't know how.

JackDandy: perhaps...
```
>>> passwords = ['b', 'c', 'a', 'a', 'b']
>>> passwords
['b', 'c', 'a', 'a', 'b']
>>> unique_passwords = list(dict((pw, None) for pw in passwords))
>>> unique_passwords
['b', 'c', 'a']
```
py3 dict type is ordered so using the keys of a dict is an ordered set, and fast. ;)

Safihre: Oh yes sorry, can't tdo this really in a oneliner. Fine this way.

PR #1907, opened 2021-05-29
Regex for unwanted extensions

jcfp:
Can be used to avoid long lists of similar extensions such as r00-r99.

The only annoyance I came across is that certain chars (comma, ...) cannot be used in the regexp because of how the setting is stored and handled. Similar limitations seem to exist with rss feed filters, so I assume there's no easy way to avoid this.

Yes, I'll update the docs too. :grin:

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -59,6 +59,29 @@ def get_ext(filename: str) -> str:
         return ""
 
 
+def ext_is_listed(ext: str, ext_list: list) -> bool:
+    """Check if the extension is listed. In case of a regexp the entire extension must be matched;
+    partial matches aren't accepted (e.g. 'r[0-9]{2}' will be treated the same as '^r[0-9]{2}$' and
+    thus return false for extentions such as 'r007' despite the substring match on 'r00').
+
+    Since the extension list is stored as a comma-separated list in the config, regular expressions
+    cannot use a comma or they'll be split and interpreted as multiple listed extensions.
+    """
+    for item in ext_list:
+        if item.startswith("re:"):
+            try:
+                # Case-insensitive match, extensions from both cfg and get_ext are always lowercase
+                if len(re.compile(item[3:], re.I).match(ext).group()) == len(ext):
+                    return True
+            except Exception:
+                pass
+        else:
+            if ext in ext_list:
+                return True
+    # No match found
+    return False
+
+
 def has_unwanted_extension(filename: str) -> bool:
     """Determine if a filename has an unwanted extension, given the configured mode"""
     extension = get_ext(filename).replace(".", "")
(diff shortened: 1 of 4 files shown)


sabnzbd/filesystem.py
```diff
@@ -59,18 +59,40 @@ def get_ext(filename: str) -> str:
...
+    thus return false for extentions such as 'r007' despite the substring match on 'r00').
+
+    Since the extension list is stored as a comma-separated list in the config, regular expressions
+    cannot use a comma or they'll be split and interpreted as multiple listed extensions.
+    """
+    for item in ext_list:
+        if item.startswith("re:"):
+            try:
+                # Case-insensitive match, extensions from both cfg and get_ext are always lowercase
+                if len(re.compile(item[3:], re.I).match(ext).group()) == len(ext):
```
Safihre, line 74: Maybe it's best to use the same logic as `convert_filter` from `rss.py`? This also allows simple regex like `.r*`?

Safihre:
I do wonder when you would want to use this? Sounds a bit exotic 🌴😉

jcfp:
> I do wonder when you would want to use this? Sounds a bit exotic palm_treewink

I've grown fond of the extension whitelist feature, but using that for a while it did result in a lengthy list of extensions (hundreds of numbered rar parts and the likes) that could easily be replaced with just a few regular expressions.

sanderjo:
> > I do wonder when you would want to use this? Sounds a bit exotic palm_treewink
> 
> I've grown fond of the extension whitelist feature, but using that for a while did result in a **lengthy list of extensions** (hundreds of numbered rar parts and the likes) that could easily be replaced with just a few regular expressions.

That's a consequence of **whitelisting**, I suppose ...

When I implemented Unwanted Extensions, it was to early-ignore fake posts with fake/virus files like codec.exe. So having exe, com and bat as unwanted extension (thus just three on the blocklist) has always been enough for me.

Allow-listing plus regexps seems overly strict / difficult to me.

sabnzbd/filesystem.py
```diff
@@ -59,18 +59,37 @@ def get_ext(filename: str) -> str:
         return ""
 
 
+def is_listed_ext(ext: str, ext_list: list) -> bool:
+    """Check if the extension is listed. In case of a regexp the entire extension must be matched;
+    partial matches aren't accepted (e.g. 'r[0-9]{2}' will be treated the same as '^r[0-9]{2}$' and
+    thus return false for extentions such as 'r007' despite the substring match on 'r00').
+    """
+    for item in ext_list:
+        RE_EXT = sabnzbd.rss.convert_filter(item)
```
Safihre, line 68: Let's move it to `misc`, since we use it in more spots now.
After that we can merge.

jcfp:
> That's a consequence of **whitelisting**, I suppose ...

It's mostly a consequence of the rar devs using `.rNN` (etc.) as part extensions instead of sticking with `.partN.rar`. If microsoft were to make `.cmdNN` the equivalent of `.cmd` tomorrow, one would run into exactly the same issue with an extension blacklist.

Anyway, the simple notation continues to work so no harm done.

Issue #1913, opened 2021-06-11
Use puremagic module to find obfuscated extesion


Safihre:
Some downloads contain fully obfuscated files in such a way that even the extensions are obfuscated.
Proposal: use the [`puremagic`](https://github.com/cdgriffith/puremagic/) module to find the most likely extension of an obfuscated file.
We do need to determine when we will run the extension-detection, how do we detect a nonsense extension from a regular one?
It would then also only be activated if Deobfuscate final filenames is turned _on_.

@sanderjo has already been testing this module a bit.

sanderjo:
Example:

In my download:

`7986c213700bb0.8edb8bca0a2fcf7d1237e2cdf082f6.44c8`

the Linux command `file` (which puremagic can probably do too) tells

`PNG image data, 1920 x 1080, 8-bit/color RGB, non-interlaced`

So useful for user to let SAB rename that file to `7986c213700bb0.8edb8bca0a2fcf7d1237e2cdf082f6.44c8.png`  (so with correct extension)

... and after that, deobfuscate() could rename to jobname


@Safihre  is that what you mean?

Safihre:
Yes indeed.

puzzledsab:
You should make sure that the user gets a warning if they are included in the unwanted extension list or not in the white list, and that they are deleted if they are listed in the cleanup list.

sanderjo:
I propose this:

```
def correct_file_extension(file):

    return extension_needs_correction, proposed_file
```

- correct_file_extension("mypicture.png") => False, None
- correct_file_extension("blabla") => True, "blabla.png"

or

```
def correct_file_extension(file):

    return proposed_file # None if no improved extension needed or possible
```

- correct_file_extension("mypicture.png") => None
- correct_file_extension("blabla") => "blabla.png"

Feedback?

sanderjo:
Thanks for your feedback.

> You should make sure that it works with the unwanted extension list 

Unwanted extension is during the download. Deobfuscation is later on. So there is a conflict. Oh ... that is already the case: deobfuscation's par2-based renaming can already now result in unwanted extensions. 
Ah ... nice way for bad guys to get their "clickthis.exe" onto someone's harddrive anyway ... 

> and that they are deleted if they are listed in the cleanup list.

I think cleanup is at the very end, so no problem.

puzzledsab:
[...]

PR #1914, opened 2021-06-12
Puremagic extension

sanderjo:
related to https://github.com/sabnzbd/sabnzbd/issues/1913


sanderjo:
@Safihre Black wants to reformat the file puremagic.py (which is a straight copy from puremagic on github).

What to do? If to ignore, can I instruct github/Black to not Black-check that file?

jcfp:
> What to do? If to ignore, can I instruct github/Black to not Black-check that file?

Why include the entire module in the first place rather than listing it in requirements.txt?

sanderjo:
> > What to do? If to ignore, can I instruct github/Black to not Black-check that file?
> 
> Why include the entire module in the first place rather than listing it in requirements.txt?

... especially for you / PPA packaging ... but apparently I'm wrong?

Ubuntu does not provide puremagic:

```
apt search puremagic
Sorting... Done
Full Text Search... Done
```


There is a pypy module:
```
# python3 -m pip install puremagic
Collecting puremagic
  Downloading puremagic-1.10-py3-none-any.whl (31 kB)
Collecting argparse
  Downloading argparse-1.4.0-py2.py3-none-any.whl (23 kB)
Installing collected packages: argparse, puremagic
Successfully installed argparse-1.4.0 puremagic-1.10
```

If I put it into requirements.txt, that is OK for you, @jcfp ?

tests/data/test_extension/sources.txt
```diff
@@ -0,0 +1,3 @@
+
+cruise-ship-horn-sound.mp3: https://www.freesoundslibrary.com/cruise-ship-horn-sound/
+
```
jcfp, line 3: This file is licensed under CC-BY-4.0. That's fine, but this changeset fails to meet the basic [terms](https://creativecommons.org/licenses/by/4.0/) of that license.

jcfp:
> ... especially for you / PPA packaging ... but apparently I'm wrong?

Thanks, but no need to include. I can always create a package for that, or alternatively patch the sab package to include a copy of puremagic if needed.

tests/data/test_extension/sources.txt
sanderjo: OK ... created my own mp3 and removed the cruise ship.

sanderjo:
Benchmarking the results:

`find /media/sander/6665-6464/ -type f -follow -print | grep -v Witte |  awk '{ print "python3 sabnzbd/utils/correct_extension.py \"" $0 "\"" }'  | /bin/sh`

which works, but there must be a beautiful find | xargs method, but I'm struggling with spaces in the filenames. I tried  -print0 and -0 ... : 

`find /media/sander/6665-6464/ -type f -follow -print0  | xargs -0  python3 sabnzbd/utils/correct_extension.py
`
but that only finds three files instead of thousands?

sanderjo:
Some notes:

tests/data/test_extension/hello_world_mpthree.blabla is an MP3
but puremagic says: ['.mpga', '.sys']
whereas Linux' file says MPEG ADTS, layer III, v1, 224 kbps, 44.1 kHz, JntStereo

So ... I'm considering a translation table (based on training), that says

['.mpga', '.sys'] -> '.mp3'

... and that for a lot of extensions suggested by puremagic.

Input for the training data is stuff on my harddisk, with a common well-known extension

sanderjo:
OK, now in deobfuscate(): if file has no commonly known extension, add it.

Proof it works, from sabnzbd.log:

```

2021-06-13 18:42:51,382::INFO::[deobfuscate_filenames:177] Deobfuscate renaming (add extension) /media/sander/6665-6464/nzb_mix_no_ext_38f38a34acc2.4/apeeengeee to /media/sander/6665-6464/nzb_mix_no_ext_38f38a34acc2.4/apeeengeee.png

2021-06-13 18:42:51,388::INFO::[deobfuscate_filenames:177] Deobfuscate renaming (add extension) /media/sander/6665-6464/nzb_mix_no_ext_38f38a34acc2.4/somepeedeef to /media/sander/6665-6464/nzb_mix_no_ext_38f38a34acc2.4/somepeedeef.pdf

2021-06-13 18:42:51,391::INFO::[deobfuscate_filenames:177] Deobfuscate renaming (add extension) /media/sander/6665-6464/nzb_mix_no_ext_38f38a34acc2.4/vuur_matro to /media/sander/6665-6464/nzb_mix_no_ext_38f38a34acc2.4/vuur_matro.mkv
```

requirements.txt
```diff
@@ -1,3 +1,5 @@
+# usage: python3 -m pip install -r requirements.txt -U
+
```
Safihre, line 2: No need to add this, that's basic information that anyone running from sources should know or otherwise it's on our wiki.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -134,6 +135,10 @@ def is_probably_obfuscated(myinputfilename):
 def deobfuscate_list(filelist, usefulname):
     """Check all files in filelist, and if wanted, deobfuscate: rename to filename based on usefulname"""
 
+    # Two methods
+    # 1. based on par2 (if any)
+    # 2. based on detecting obfuscated filenames
```
Safihre, line 140: Add the deobfuscation of extensions?

sabnzbd/deobfuscate_filenames.py
```diff
@@ -154,6 +159,25 @@ def deobfuscate_list(filelist, usefulname):
...
+        for file in filelist:
+            if correct_extension.has_common_extension(file):
+                # common extension, like .doc or .iso, so assume OK and change nothing
+                logging.debug("extension of %s looks common", file)
+                newlist.append(file)
+            else:
+                # uncommon (so: obfuscated) extension
+                new_extension_to_add = correct_extension.most_likely_extension(file)
+                new_name = get_unique_filename("%s%s" % (file, new_extension_to_add))
+                logging.info("Deobfuscate renaming (add extension) %s to %s", file, new_name)
```
Safihre, line 176: `adding`

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,342 @@
...
+    "r18",
+    "r19",
+    "r21",
+    "r22",
+    "r23",
+    "r24",
+    "r25",
+    "r26",
+    "r27",
+    "r28",
```
Safihre, line 238: Feels a bit arbitrary to stop at `r28`. If you want to include these `.rXX` then I propose a specific check for `.rXX` and not all these separate extensions,

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,342 @@
...
+    "xhtml",
+    "xls",
+    "xlsm",
+    "xlsx",
+    "xml",
+    "z",
+    "zip",
+]
+
+downloadextlist = [
```
Safihre, line 161: For performance, let's remove all items of `downloadextlist` that are already in `commonextlist`. No need to check anything twice.

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,342 @@
...
+
+""" function to check and find correct extension of a (deobfuscated) file
+"""
+
+
+import puremagic
+import os
+
+# common extension from https://www.computerhope.com/issues/ch001789.htm
+commonextlist = [
```
Safihre, line 11: For consistency with the rest of SAB's code-base, we capitalize constants: `COMMON_EXT` and `DOWNLOAD_EXT`.
And let's also use a tuple instead of a list, since they are constants (so `( )` instead of `[ ]`).

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,342 @@
...
+            return possible_extension
+    # no common extension found, so just return the first
+    try:
+        return all_possible_extensions(file_path)[0]
+    except:
+        return ""
+
+
+def extension_matches(file_path: str) -> int:
+    import os
```
Safihre, line 294: Put at the top of the file.

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,342 @@
...
+
+    # TBD use SAB's own extension finder
+    filename, file_extension = os.path.splitext(file_path)
+    file_extension = file_extension[1:].lower()
+    return file_extension in all_possible_extensions(file_path)
+
+
+if __name__ == "__main__":
+    import sys
```
Safihre, line 303: Put at the top of the file.

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,342 @@
...
+            print(
+                "IRL-ext",
+                file_extension,
+                "most_likely",
+                most_likely_extension(file_path),
+                "puremagic",
+                all_possible_extensions(file_path),
+            )
+
+        """
```
Safihre, line 337: Remove comment.

tests/data/test_extension/sources.txt
```diff
@@ -0,0 +1,2 @@
...
```
Safihre, line 1: Is this file still needed?

tests/test_correct_extension.py
```diff
@@ -0,0 +1,61 @@
...
+        assert not correct_extension.has_common_extension("98ads098f098fa.a0ds98f098asdf")
+
+    def test_most_likely_extension(self):
+        assert correct_extension.most_likely_extension("tests/data/test_extension/hello_world_mpthree.blabla") == "mp3"
+
+
+class TestCorrect_Extension:
+    def test_all_possible_extensions(self):
+
+        import os
```
Safihre, line 51: Already imported.

tests/test_correct_extension.py
```diff
@@ -0,0 +1,61 @@
...
+
+class Test_Extension:
+    def test_has_common_extension(self):
+        assert correct_extension.has_common_extension("blabla/blabla.mkv")
+        assert correct_extension.has_common_extension("blabla/blabla.srt")
+        assert correct_extension.has_common_extension("djjddj/aaaaa.epub")
+        assert not correct_extension.has_common_extension("98ads098f098fa.a0ds98f098asdf")
+
+    def test_most_likely_extension(self):
+        assert correct_extension.most_likely_extension("tests/data/test_extension/hello_world_mpthree.blabla") == "mp3"
```
Safihre, line 45: Are you adding tests later that will use the `tests/data/test_extension/somempthree` file? Seems so far it is unused.

Safihre (changes requested):

requirements.txt
sanderjo: I have to google it each time. So I would like to keep it, if OK

sabnzbd/utils/correct_extension.py
sanderjo: Thanks. Please review my new implementation.

sabnzbd/utils/correct_extension.py
sanderjo: Please check my implementation to make it unique

sabnzbd/utils/correct_extension.py
sanderjo: Yes, please check my improved implementation

tests/test_correct_extension.py
sanderjo: Yes. And cleaning up unnecessary code / functions. So leaving this comment open until then.

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,316 @@
...
+    "xpi",
+)
+
+RAR_EXT = tuple(["r" + str(i).zfill(2) for i in range(0, 100)])  # "r00", "r01", ...
+
+# combine to one tuple, with unique entries:
+ALL_EXT = tuple(set(COMMON_EXT + DOWNLOAD_EXT + RAR_EXT))
+
+
+def has_common_extension(file_path: str) -> int:
```
jcfp, line 241: this function returns a boolean, not an integer

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,316 @@
...
+            # Yes, looks likely
+            return possible_extension
+    # no common extension found, so just return the first
+    try:
+        return all_possible_extensions(file_path)[0]
+    except:
+        return ""
+
+
+def extension_matches(file_path: str) -> int:
```
jcfp, line 271: this function too returns a boolean. Also consider making the function name reflect that (e.g. _is_this()_, _has_that()_ or similar).

jcfp (changes requested):

sabnzbd/utils/correct_extension.py
sanderjo: Ah, note to self: I tried `-> boolean `but it's `-> bool`

sabnzbd/utils/correct_extension.py
sanderjo: Yes, I have to review the naming of the functions and the module itself.

sabnzbd/utils/correct_extension.py
```diff
@@ -0,0 +1,316 @@
...
+
+def has_common_extension(file_path: str) -> int:
+    """returns boolean if the extension of file_path is a common, well-known extension"""
+    # TBD use SAB's own extension finder
+    filename, file_extension = os.path.splitext(file_path)
+    file_extension = file_extension[1:].lower()
+    return file_extension in ALL_EXT
+
+
+def all_possible_extensions(file_path: str) -> list:
```
Safihre, line 249: Use `List[puremagic.PureMagicWithConfidence]` 
See here for diference between `list` and `List`:
https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc

sanderjo:
Current module name is `correct_extension`. I find that confusing. So ... options:

1. - file_extension (I like this one as it makes clear it's about files and their extension)
2. - extension_check
3. - extension_info
4. - extension
5. - ext_checker
6. - ext_info (second best?)


Current function names:

1. - .has_common_extension() -> bool. Name seems OK to me
2. - .most_likely_extension() ->str,  which I want to rename to .what_is_most_likely_extension()

Feedback?

Safihre:
They all sound fine to me 👍 option 2 I would suggest.

sabnzbd/utils/correct_extension.py
sanderjo: typing.List works, but

```
>>> typing.List[puremagic.PureMagicWithConfidence]
/home/sander/.local/lib/python3.8/site-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.25.9) or chardet (4.0.0) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
AttributeError: module 'puremagic' has no attribute 'PureMagicWithConfidence'
```
FWIW

```
>>> puremagic.
puremagic.PureError(          puremagic.from_file(          puremagic.from_string(        puremagic.magic_footer_array  puremagic.magic_stream(       puremagic.main                
puremagic.ext_from_filename(  puremagic.from_stream(        puremagic.magic_file(         puremagic.magic_header_array  puremagic.magic_string(    
```

puzzledsab:
Sorry if I missed something, but won't this always add a new extension if it is determined to be "uncommon"? Is it possible that the file could have an "uncommon" but correct extension? If so you should check if the new extension is the same as the old one before renaming.

sanderjo:
> Sorry if I missed something, but won't this always add a new extension if it is determined to be "uncommon"? 

Yes.

> Is it possible that the file could have an "uncommon" but correct extension? 

Theoretically: Yes. Can you name one? If so, then I can add it the list of POPULAR_EXTENSIONS

Background: I got the popular extensions from:
* mentioned website (about 150 extensions)
* inspecting my own downloads
* adding r00 ... r99 to not interfer with strange rar files

> If so you should check if the new extension is the same as the old one before renaming.

That's logically impossible: the code checks if the file has a popular extension. If so, it won't do anything. If is not popular, it will add the most likely, popular extension ... which was not there in the first place (otherwise the code would not kick in).

Example:
filename is dfauasdf0sdf09asd0f89. No extension, so no popular extension. Or dfauasdf0sdf09asd0f89.asd0f89asd0f89, so no popular extension either.
puremagic will say, by inspecting the contents: ".koz, or less likely .mp3"
My code does not consider .koz as popular, so it will add .mp3 as extension

And the other way around: if filename is dfauasdf0sdf09asd0f89.mp3, then the extension is popular, and my code won't touch it.

sanderjo:
reftestnzb https://raw.githubusercontent.com/sanderjo/NZBs/master/reftestnzb__No_Extensions__062d817b1907.nzb pointing to files without any extension. My latest commit nicely adds extensions based on puremagic content guessing / inferring:

```
apeeengeee.png
my_matroska.mkv
some_nzb_file.nzb
somepeedeef.pdf
sometxtfile.txt
vuur_matro.mkv
```

jcfp:
> * adding r00 ... r99 to not interfer with strange rar files

I regularly come across the 7zip equivalent of these in my downloads, i.e. 7z.000 ... 7z.999. How good is puremagic's detection for those?

sanderjo:
> > * adding r00 ... r99 to not interfer with strange rar files
> 
> I regularly come across the 7zip equivalent of these in my downloads, i.e. 7z.000 ... 7z.999. How good is puremagic's detection for those?

What would like to be the result? Rename 7z.123 to 7z, or ... ?

jcfp:
Keep the name. Just wondered how it would handle those when I saw the r00 style explicitly added as a common extension.

sanderjo:
> Keep the name. Just wondered how it would handle those when I saw the r00 style explicitly added as a common extension.


head -c 100m </dev/urandom > random.bin
7z -v10m a blocks random.bin
resulting in
```blocks.7z.001
blocks.7z.002
blocks.7z.003
blocks.7z.004
blocks.7z.005
blocks.7z.006
blocks.7z.007
blocks.7z.008
blocks.7z.009
blocks.7z.010
blocks.7z.011
```

and puremagic
```
>>> import puremagic
>>> puremagic.magic_file("blocks.7z.005")
[]
```
... has no idea. Pity.

I can introduce 'if ends in ".7z.NNN" then do nothing'

Safihre:
Indeed, Sab understands these files and unpacks them. Same for split files, they can also have this extension but are picked up by Sab during unpacking.

While writing that I realize that we also don't need to exclude all the `rXX` because Sab already extracts those. By default Sab will have recurivse unpack enabled, and if there's any Rar files (no matter the extension) left, it will try to unpack them.

requirements.txt
Safihre: No, sorry. Please remove.
If anywhere, it should be in the `README.md`.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -131,16 +132,21 @@ def is_probably_obfuscated(myinputfilename):
     return True  # default not obfuscated
 
 
-def deobfuscate_list(filelist, usefulname):
+def deobfuscate_list(filelist, usefulname, extension_too=True):
```
Safihre, line 135: You added `extension_too` but except for tests not using it anywhere, so why add it?

sabnzbd/utils/correct_extension.py
sanderjo: @Safihre ... can you help?

jcfp: @sanderjo looks like you don't need that `PureMagicWithConfidence` type: the returned list contains only the extensions - and those are just regular `str`.

Safihre: True!
But it should still be `List[str]` and not `list`

sabnzbd/utils/file_extension.py
```diff
@@ -0,0 +1,313 @@
...
+ALL_EXT = tuple(set(POPULAR_EXT + DOWNLOAD_EXT))
+
+
+def has_popular_extension(file_path: str) -> bool:
+    """returns boolean if the extension of file_path is a popular, well-known extension"""
+    file_extension = get_ext(file_path)[1:]  # without the dot
+    return file_extension in ALL_EXT
+
+
+def all_possible_extensions(file_path: str) -> typing.List[str]:
```
jcfp, line 248: If you only import the things from typing that you actually need (i.e. `from typing import List`) and the `typing.` part can be dropped. Function definitions tend to be long enough already without extra baggage.

sanderjo:
@Safihre I think this PR is ready for pulling into SAB

sabnzbd/deobfuscate_filenames.py
```diff
@@ -150,65 +156,88 @@ def deobfuscate_list(filelist, usefulname):
...
-                for otherfile in filelist:
-                    if otherfile.startswith(basedirfile + ".") and os.path.isfile(otherfile):
-                        # yes, same basedirfile, only different extension
-                        remainingextension = otherfile.replace(basedirfile, "")  # might be long ext, like ".dut.srt"
-                        new_name = get_unique_filename("%s%s" % (os.path.join(path, usefulname), remainingextension))
-                        logging.info("Deobfuscate renaming %s to %s", otherfile, new_name)
-                        # Rename and make sure the new filename is unique
-                        renamer(otherfile, new_name)
+    if par2_renaming_done:
+        # TODO really needed to quit here? We could also proceed with the other actions. Anyway:
```
Safihre, line 164: Agree with your TODO, why stop here? Maybe the par2 file only renamed 1 file out of 100.

sabnzbd/utils/file_extension.py
```diff
@@ -0,0 +1,314 @@
...
+    # Check if text or NZB, as puremagic is not good at that.
+    try:
+        txt = Path(file_path).read_text()
+        # Yes, a text file ... so let's check if it's even an NZB:
+        if txt.lower().find("<nzb xmlns=") >= 0 or txt.lower().find("!doctype nzb public") >= 0:
+            # yes, contains NZB signals:
+            return ".nzb"
+        else:
+            return ".txt"
+    except:
```
Safihre, line 274: Can we make the `except` more specific?

sabnzbd/utils/file_extension.py
```diff
@@ -0,0 +1,314 @@
...
+        # not txt (and not nzb)
+        pass
+
+    # no popular extension found, so just trust puremagic and return the first extension (if any)
+    try:
+        return all_possible_extensions(file_path)[0]
+    except:
```
Safihre, line 281: Can we make the `except` more specific?

tests/test_file_extension.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+import os
+from tests.testhelper import *
+import sabnzbd.utils.file_extension as file_extension
+
+
+class TestPuremagic:
+    def test_puremagic_magic_file(self):
+        import puremagic  # direct access only in this test function
```
Safihre, line 29: I don't see the point of this test? Since it's testing basically `puremagic` itself.
If you want to make sure `par2` works, just make a copy of this file (but strip the extension) and then run our own functions on it.

Safihre (changes requested):
Only small comments, looks good otherwise.

But I think we do need to change the description in Config > Switches of "Deobfuscate final filenames" to include this extension renaming.

sabnzbd/deobfuscate_filenames.py
sanderjo: I tried this, but it has a lot of effects on the code, including the test code.
My proposal: I do this in a separate PR

Safihre: Agreed!

PR #1916, opened 2021-06-19
Use guessit for sorting and sample detection

jcfp:
* Made _year_, _decade_, and the _title-cased show/movie name_ sorting options universal (each previously only available for selected sorting types).
* Added `%GI<property>` that allows for using any valid guessit property in sorting. Available properties for the installed version of guessit  are listed under the _pattern key_ in the web interface.
* Added `episode_rename_limit` as a special option, and made it apply to both tv and date sorting. A similar option already exists for movies (with a 100M default), but for series is was hardcoded (to 20M; now the default value for the new option) and only applied to the _tv_ category.
* Removed the unused `tv_sort_countries` config variable.
* Got rid of alot of code duplication by introducing a common class for the sorters to inherit from.
* Added tests.

Sorting should work as expected, although the lack of test coverage for the old code at times made it hard enough to figure out what it was supposed to do, and nearly impossible to catch regressions so **please do test**.

The sample detection is almost inevitably going to yield different results: the old code was good at picking up keywords but equally fast at triggering false positives. Guessit's sample/proof detection does a pretty good job as long as releases are properly named, and is much harder to fool in that case than the previous simple regex-based approach.


sabnzbd/sorting.py
```diff
@@ -797,218 +536,181 @@ def filter_files(_file, current_path):
...
+        guess.setdefault("season", 1)
+
+    # Try to avoid setting the type to movie on arbitrary jobs (e.g. 'Setup.exe') just because guessit defaults to that
+    table = str.maketrans({char: "" for char in whitespace + "_.-()[]{}"})
+    if (
+        guess.get("type") == "movie"
+        and not sort_type == "movie"  # No movie hint
+        and not any([key in guess for key in ("year", "screen_size", "video_codec")])  # No typical movie properties set
+    ):
+        # Check for full name used as title, or weird chars in group
```
Safihre, line 671: It fails on weird chars? Just trying to understand why we need the `translate` parts.

sabnzbd/sorting.py
```diff
@@ -797,218 +536,181 @@ def filter_files(_file, current_path):
...
+        if guess.get("title").translate(table) == name.translate(table) or any(
+            c in guess.get("release_group", "") for c in (whitespace + punctuation)
+        ):
+            guess["type"] = "unknown"
+
+    # Remove sample indicators from groupnames, e.g. 'sample-groupname' or 'groupname-proof'
+    group = guess.get("release_group", "")
+    if group.lower().startswith(("sample-", "proof-")) or group.lower().endswith(("-sample", "-proof")):
```
Safihre, line 679: Sounds like a bug in `guessit`? Maybe we should report that?

Safihre (changes requested):
Nice with all the unittests, finally!

interfaces/Config/templates/config_sorting.tmpl
```diff
@@ -156,6 +171,43 @@
...
+                                <tr>
+                                    <td class="align-right"><b>$T('Example')</b></td>
+                                    <td>%GI&lt;audio_codec&gt;</td>
+                                    <td>DTS</td>
+                                </tr>
+                                <tr>
+                                    <td class="align-right" style="vertical-align:top;"><b>$T('Valid properties')</b></td>
+                                    <td colspan=2 style="word-spacing: 0.7em; max-width:275px; white-space: wrap;">
+                                        <!--#for $prop in $guessitProperties#-->
+                                        $prop
```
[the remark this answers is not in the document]
jcfp: I initially did just that, but that was quite long (vertically) while leaving lots of empty space. Idealy, this would have multiple columns, but my cheetah skills fell short.

requirements.txt
```diff
@@ -8,6 +8,7 @@ cherrypy
 portend
 chardet
 notify2
+guessit>=3.1.0  # https://github.com/guessit-io/guessit/pull/536 needed for sample detection
```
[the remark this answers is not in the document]
jcfp: I included that to track the specific reason for that version requirement while working on this and not knowing what features would make the finish line. I'll get rid of the comment now.

sabnzbd/sorting.py
```diff
@@ -1021,62 +723,48 @@ def path_subst(path, mapping):
...
+        # Try guessit next
+        if guess:
+            title = guess.get("title", "")
+            if title and "country" in guess:
+                title += " (" + guess.get("country").alpha2 + ")"  # Append ' (CC)'
+
+        # Fallback to the jobname if neither of the better options yielded a title
+        if not title:
+            # Format country codes in jobname
             for rep in COUNTRY_REP:
```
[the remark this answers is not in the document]
jcfp: Wouldn't that still leave the issue with title case, causing the country codes to end up Us, Nl, etc.?

sabnzbd/sorting.py
jcfp: Not really fail, but the title it returns is "too clean", so interpunction can throw the 'full name used as title' check off too easily. Think stuff like this:
```
guessit.guessit("[Setup]")
MatchesDict([('title', 'Setup'), ('type', 'movie')])
guessit.guessit("Setup.exe")
MatchesDict([('title', 'Setup exe'), ('mimetype', 'application/x-msdos-program'), ('type', 'movie')])
```
The groupname is another check that can stop badly named stuff from being considered a movie, as that sometimes seems to be used as a dumping ground by guessit:
```
guessit.guessit("Something [1.2.3 four five]")
MatchesDict([('title', 'Something'), ('release_group', '1.2.3 four five'), ('type', 'movie')])
```

sabnzbd/sorting.py
jcfp: Guessit thrives on "proper" scene naming. I'm not sure they'd consider things like this a bug if the input isn't properly formatted, but that said, there doesn't seem to be much of a downside for them standing in the way of fixing this either.

sabnzbd/sorting.py
```diff
@@ -444,350 +297,236 @@ def construct_path(self):
         # Lowercase all characters wrapped in {}
         path = to_lowercase(path)
 
-        # Strip any extra ' ' '.' or '_' around foldernames
-        path = strip_folders(path)
+        # Strip any extra spaces, dots, and underscores around directory names
+        path = strip_path_elements(path)
```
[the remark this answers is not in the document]
jcfp: A comment in the code said windows cannot handle path elements ending in a dot; I assumed that's (still) accurate. But even ignoring that, stripping the path elements is useful in case certain %patterns are replaced with nothing, resulting in ugly dangling separators. The function was broken before, in the sense that the success of the cleanup routine depended on the order of appearance of the chars it should strip.

sabnzbd/sorting.py
```diff
@@ -117,323 +119,174 @@
...
+            try:
+                self.info["decade"] = self.info["year"][2:3] + "0"
+                self.info["decade_two"] = self.info["year"][:3] + "0"
+            except TypeError:
+                pass
 
-        sorter = self.sort_string.replace("\\", "/")
+    def construct_path(self) -> str:
+        """Map all markers and replace the sort string with real values"""
+        sorter = self.sort_string
```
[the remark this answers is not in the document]
jcfp: Now set in super().__init__, and values passed there as an argument from each sorter's init.

sabnzbd/sorting.py
```diff
@@ -117,323 +119,174 @@
...
-        self.show_info = {}
+        self.do_rename = False
+        self.info = {}
+        self.type = None
+        self.guess = guess
+        self.force = force
 
-        # Check if it is a TV show on init()
+        # Check categories and do the guessing work, if necessary
         self.match()
```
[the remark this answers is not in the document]
jcfp: Obviously that class isn't intended to be used directly, and the other sorters always define this. Any standard python way to  signal that, or making pycharm understand? Adding "empty" methods just to get rid of this warning seems weird.

sabnzbd/sorting.py
```diff
@@ -797,218 +536,181 @@ def filter_files(_file, current_path):
...
-        self.date_info["resolution"] = get_resolution(self.original_job_name)
 
-        # - Get Title
-        self.date_info["ttitle"], self.date_info["ttitle_two"], self.date_info["ttitle_three"] = get_titles(
-            self.nzo, self.match_obj, self.original_job_name, True
-        )
-        self.date_info["title"], self.date_info["title_two"], self.date_info["title_three"] = get_titles(
-            self.nzo, self.match_obj, self.original_job_name
-        )
+def move_to_parent_directory(workdir: str) -> Tuple[str, bool]:
```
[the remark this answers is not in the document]
jcfp: Close, but that one only does the collapsing "when there's just a single subfolder".

sabnzbd/sorting.py
```diff
@@ -444,350 +297,236 @@ def construct_path(self):
...
-            self.rename_or_not = True
-        else:
-            head = path
+    def get_episodes(self) -> None:
+        """Fetch the guessed episode number(s)"""
+        self.format_series_numbers(self.guess.get("episode", 1), "episode_num")
 
-        if head:
-            return os.path.normpath(head)
+    def rename(self, files: List[str], current_path: str) -> Tuple[str, bool]:
```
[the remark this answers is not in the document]
jcfp: I'll modify the sorters' rename method to take a min_size argument as well, with their configured minimum size as the default value.

sabnzbd/sorting.py
```diff
@@ -797,218 +536,181 @@ def filter_files(_file, current_path):
...
+    # Avoid trouble with names starting with a digit (esp. with no year in the name)
+    digit_fix = "FIX" if name[0].isdigit() else ""
+
+    guessit_options = {
+        # "no-user-config": True,
+        "expected_title": [],  # This isn't empty by default?
+        # "allowed_countries": [],
+        # "allowed_languages": [],
+        "excludes": [
+            "part",  # Handled by more powerful check_for_sequence()
```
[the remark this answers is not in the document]
jcfp: It takes the 'part x' stuff out of titles, which is nice but not very useful for sab. For example:
```
guessit.guessit('The Dogfather: Part II (2024)')
MatchesDict([('title', 'The Dogfather'), ('part', 2), ('year', 2024), ('type', 'movie')])
```
All we can do with this is basically add it right back in.

thezoggy:
to note, this should also be expanded to fix the shoddy 'detect duplicate episodes in series' logic as well

jcfp:
> to note, this should also be expanded to fix the shoddy 'detect duplicate episodes in series' logic as well

The proper detection is currently broken by this patchset, because NzbObject's `has_duplicates` tries to find proper/repack/real tags in the episode name returned from `newsunpack.analyse_show` but with that now using guessit those aren't part of the episode name any longer. Should be easy to fix.

Safihre:
I also wonder if we maybe need to do a "conversion" of the history database to reanalyse all the jobs with the new detection? Just to make sure that the Series duplicate gets the same result.
Would at least me be a good test to run it on history and see if the guessit results differ from the stored result.

Safihre:
Got a crash while running this:
```
  File "C:\Users\saf\Documents\GitHub\sabnzbd3\sabnzbd\postproc.py", line 386, in process_job
    par_error, re_add = parring(nzo, workdir)
  File "C:\Users\saf\Documents\GitHub\sabnzbd3\sabnzbd\postproc.py", line 745, in parring
    if cfg.ignore_samples() and is_sample(setname.lower()):
  File "C:\Users\saf\Documents\GitHub\sabnzbd3\sabnzbd\sorting.py", line 702, in is_sample
    guess = guess_what(filename).get("other", "")
  File "C:\Users\saf\Documents\GitHub\sabnzbd3\sabnzbd\sorting.py", line 655, in guess_what
    guess["title"] = guess["title"][len(digit_fix) :]
KeyError: 'title'
```

jcfp:
> Would at least me be a good test to run it on history and see if the guessit results differ from the stored result.

Have a go at it:
```
from tests.testhelper import FakeHistoryDB
from sabnzbd.newsunpack import analyse_show

### FULL PATH TO HISTORY DB ###
db_path = "/tmp/history1.db"

good = 0
bad = 0
with FakeHistoryDB(db_path) as db:
    hist = db.fetch_history()
    names = [hist[0][i]["name"] for i in range(0, hist[1]) if hist[0][i]["series"]]

    for name in names:
        series, season, episode = analyse_show(name)[0:3]
        db_result = db.have_episode(series, season, episode)

        if db_result != True:
            print("MISSING: history name = %s but cannot find %s %s %s" % (name, series, season, episode))
            bad += 1
        else:
            good += 1

print("Found: %s, Missing: %s  [%s out of %s entries in %s have series info]" % (good, bad, len(names), hist[1], db_path))
```

jcfp:
fyi: I'm tracking a bug that causes the complete dir to be off. No eta.

interfaces/Config/templates/config_sorting.tmpl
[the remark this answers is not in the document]
jcfp: Switched back to a linebreak for now.

Safihre:
> > Would at least me be a good test to run it on history and see if the guessit results differ from the stored result.
> 
> Have a go at it:


I have only a small history, but still found some using your script:
```
MISSING: history name = Whose.Line.Is.It.Anyway.US.S17E02.1080p.WEB.h264-KOGi-2 (1) but cannot find Whose Line Is It Anyway (US) 17 2
MISSING: history name = Hells.Kitchen.US.S19E04.1080p.AMZN.WEB-DL.DDP2.0.H.264 but cannot find Hells Kitchen (US) 19 4
MISSING: history name = Hells.Kitchen.US.S19E03.1080p.AMZN.WEB-DL.DDP2.0.H.264-DR but cannot find Hells Kitchen (US) 19 3
MISSING: history name = Hells.Kitchen.US.S19E01.1080p.AMZN.WEB-DL.DDP2.0.H.264-DR but cannot find Hells Kitchen (US) 19 1
```

sabnzbd/sorting.py
```diff
@@ -1021,62 +732,48 @@ def path_subst(path, mapping):
...
+def get_titles(
+    nzo: Optional[NzbObject], guess: MatchesDict, jobname: str, titleing: bool = False
+) -> Tuple[str, str, str]:
+    """Get the title from NZB metadata or jobname, and return it in various formats. Formatting
+    mostly deals with working around quirks of Python's str.title(). NZB metadata is used as-is,
+    further processing done only for info obtained from guessit or the jobname."""
+    title = ""
     if nzo:
+        # Fetch NZB metadata
         title = nzo.nzo_info.get("propername")
```
Safihre, line 750: I think we should get rid of this, or at least not rely on this as the only input, what do you think>

sabnzbd/sorting.py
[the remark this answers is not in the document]
Safihre: If we move:
```
if title and "country" in guess:
                title += " (" + str(guess.get("country")) + ")"  # Append ' (CC)'
```
and replace this part:
```
            # Replace titled country names, (Us) with (US) and so on
            for rep in COUNTRY_REP:
                title = title.replace(rep.title(), rep)
``` 
Then we don't need `COUNTRY_REP`, right? We just rely on `guessit` to work correctly..

sabnzbd/sorting.py
Safihre: Yes, let's report it :)

sabnzbd/sorting.py
```diff
@@ -797,218 +546,180 @@ def filter_files(_file, current_path):
...
+    if guess.get("type") == "episode" and not "date" in guess:
+        guess.setdefault("season", 1)
+
+    # Try to avoid setting the type to movie on arbitrary jobs (e.g. 'Setup.exe') just because guessit defaults to that
+    table = str.maketrans({char: "" for char in whitespace + "_.-()[]{}"})
+    if guess.get("type") == "movie" and not sort_type == "movie":  # No movie hint
+        if (
+            guess.get("title", "").translate(table) == name.translate(table)  # Check for full name used as title
+            or any(
+                c in guess.get("release_group", "") for c in (whitespace + punctuation)
```
Safihre, line 678: I don't get this part? Why would that make it `unkown`?
In general (also due to the black-formatting) this `if` statement is a bit hard to read.

sabnzbd/sorting.py
```diff
@@ -797,218 +546,180 @@ def filter_files(_file, current_path):
...
+    guess = guessit.api.guessit(digit_fix + name, options=guessit_options)
+    logging.debug("Initial guess for %s is %s", digit_fix + name, guess)
+
+    if digit_fix:
+        # Unfix the title
+        guess["title"] = guess.get("title", "")[len(digit_fix) :]
+
+    # Force season to 1 for seasonless episodes with no date
+    if guess.get("type") == "episode" and not "date" in guess:
+        guess.setdefault("season", 1)
```
Safihre, line 670: Why do we need that? Is the season required to be set later on?

sabnzbd/sorting.py
```diff
@@ -797,218 +546,180 @@ def filter_files(_file, current_path):
...
-            mapping.append(("%_desc", self.date_info["ep_name_three"]))
+def guess_what(name: str, sort_type: Optional[str] = None) -> MatchesDict:
+    """Guess metadata for movies or episodes from their name. The sort_type ('movie' or 'episode')
+    is passed as a hint to guessit, if given."""
+
+    if not name:
+        raise ValueError("Need a name for guessing")
+
+    # Remove any passwords from the name
+    name = scan_password(name)[0]
```
Safihre, line 645: When we get to here, the name can still include the password? That should not be the case..

sabnzbd/sorting.py
jcfp: Because it's an indication the guessing is (way) off, see the earlier example of guessit using the release group as a dumping ground. Unless there's groupnames that include whitespace or punctuation?

The formatting is entirely black's doing: the original had one condition per line, with comments directly adjacent.

sabnzbd/sorting.py
jcfp: Why? This would be set by the uploader or indexer, both of whom sab has no choice but to trust when it comes to the content of the nzb. No reason not to extend that to the proper naming.

sabnzbd/sorting.py
jcfp: This is intended for mini series, documentaries and the likes, that have episodes but may come without an explicit season number. For practical purposes, the entire thing is often considered 'season one' by users and third party apps alike. 

Although the season number isn't required, defaulting to `1` can be benificial with sorting settings such as season folders, with little downside risk.

sabnzbd/sorting.py
jcfp: The sorting is called in more places than just postproc, and in at least some of those the password is still in the name. For example [`newsunpack.pre_queue`](https://github.com/sabnzbd/sabnzbd/blob/399de15792c878d74fedb0808d6435c620c5ac79/sabnzbd/newsunpack.py#L2337) that feeds the jobname - password included - to the `SeriesSorter` via `newsunpack.analyse_show` in order to supply the script with show info.

sabnzbd/sorting.py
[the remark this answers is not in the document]
jcfp: That could work, the only trade-off being that country codes get screwed up in the (unlikely?) case the fallback to the jobname as the title is used. Seems worth it.

jcfp:
> I have only a small history, but still found some using your script

Seems the parenthesis around the cc triggered those, should be fixed now.

Safihre:
I ran some basic tests, and it looks good to me. Only last suggestion I would have is to add this to `sorting.py` so the logs don't get fully spammed with all the `rebulk` output. 
```python
logging.getLogger("rebulk").setLevel(logging.WARNING)
```

jcfp:
> so the logs don't get fully spammed with all the `rebulk` output.

Thanks for reminding me, that one was plenty annoying when digging through the log but I still kept forgetting to shut it up.

Safihre:
Ready to merge?
Got one last warning:
```
DeprecationWarning: invalid escape sequence \.
    "\.{2,}", ".", title.replace(" - ", "-").replace(" ", ".").replace("_", ".").replace("(", ".").replace(")", ".")
```

Safihre:
Will merge now and maybe go over the file line by line again later, adding some comments/changes to make it easier to read many years from now :)

Safihre:
I just realized that with guessit producing the exact same results (since we compared it to the database), our old regexes were doing a pretty good job already.. Did we even need guessit? 😂

PR #1918, opened 2021-06-26
puremagic ... more

sanderjo:


Safihre:
@sanderjo I think you should checkout sabnzbd/develop again and re-apply your changes, because this pull request includes all your changes also of the previous one.

Safihre:
Or you want this one instead of the previous one?

PR #1919, opened 2021-07-01
create a record for the successful password

### LINKED ISSUE #1917: SABnzbd reports incorrect password in environment variable SAB_PASSWORD
When a NZB is added to the queue using the watched folder and a wrong RAR password is supplied but the correct password is in the password file, then the download will complete correctly. 
But SABnzbd reports in the environment variable 'SAB_PASSWORD' the wrongly provided password instead of the password from the password file that was actually used to decrypt the RAR file.

The following are some relevant parts from the logfile:

2021-06-20 19:52:55,839::INFO::[dirscanner:163] Trying to import \\?\C:\SABnzbd\WatchedFolder\books\Name_Of_My_Download_Job{{My_Wrong_Password}}.nzb

2021-06-20 19:53:02,933::INFO::[misc:739] Found a password that was set by the user: My_Wrong_Password
2021-06-20 19:53:02,934::DEBUG::[misc:760] Read these passwords from file: ['F28APT', 'daar', 'ster', '38HTS', ]
2021-06-20 19:53:02,934::INFO::[misc:762] Read 36 passwords from file \\?\C:\SABnzbd\PasswordList\MyPasswordList.txt
2021-06-20 19:53:02,934::INFO::[SABnzbd:118] Your password file contains more than 30 passwords, testing all these passwords takes a lot of time. Try to only list useful passwords.
2021-06-20 19:53:02,935::INFO::[assembler:339] Trying password "My_Wrong_Password" on job "Name_Of_My_Download_Job"
2021-06-20 19:53:03,131::INFO::[assembler:339] Trying password "F28APT" on job "Name_Of_My_Download_Job"
2021-06-20 19:53:03,282::INFO::[assembler:339] Trying password "daar" on job "Name_Of_My_Download_Job"
2021-06-20 19:53:03,383::INFO::[assembler:375] Password "daar" matches for job "Name_Of_My_Download_Job"

 'SAB_PASSWORD': 'My_Wrong_Password', 'SAB_PP': '3',

jcfp:
Keep track of the correct password, store that password in the history db, and make it available to the post-processing script via the `SAB_CORRECT_PASSWORD` environment variable. If the job doesn't require a password or the correct password remains unknown, the variable is empty.

The existing `SAB_PASSWORD` variable still provides the password supplied by the NZB or the user to both the pre-queue and post-processing scripts (as documented), but is never changed to the correct password anymore (previously: only if no password had been supplied by user or nzb).

Fixes #1917


Safihre:
Feels a bit strange to store the correct password in the database, but not just overwrite the `SAB_PASSWORD`.
Why not just replace the value in `SAB_PASSWORD`?
It's such an exotic variable, nobody would be very upset if we change this behavior in a 3.x.0 release.

jcfp:
Because by overwriting the value, the post-processing cannot know whether the user/nzb supplied password differs from the successful one - as was the case in #1917 - and use that info as part of automated password tracking or feedback to indexers.

Safihre:
If we introduce this, shouldn't we also use it a bit more?
For example in Direct Unpack, or in `get_all_passwords` to only return 1 value so it is used in `newsunpack` if there's no Direct Unpack?

jcfp:
In order to bypass the password collection logic once a correct password is known? That could work (unless there's multi-password jobs out there).

Does the directunpacker have any separate logic from the assembler thread when it comes to detecting a correct password?

Safihre:
In Direct Unpacker it just used `nzo.password`, to keep things simple. 
But in the Assembler the Direct Unpacker is only called after the password check, so in theory we should know the correct password by the time we get to the Direct Unpacker.
In practice that's not always the case, since sometimes `RarFile` crashes for some reason.
So we can't rely on the `correct_password` to always be available, but if filled, it should be used.

PR #1921, opened 2021-07-02
Continue after par2 renaming

sanderjo:
after par2 renaming, continue with usual named-based deobfuscating


Safihre:
So this code will run deobfuscate, even on files already processed by par2 right?
Shouldn't it actually ignore the files touched by par2, since par2 already renamed them?

Safihre:
Or, if par2 produced reasonable names, the Deobfuscate check just ignores them anyway?

sanderjo:
> Or, if par2 produced reasonable names, the Deobfuscate check just ignores them anyway?

Indeed.

sanderjo:
> So this code will run deobfuscate, even on files already processed by par2 right?
> Shouldn't it actually ignore the files touched by par2, since par2 already renamed them?

We could. It would make the code easier ... no need to see which new files were generated. Just a check which files in file_list still exist

sanderjo:
> Or, if par2 produced reasonable names, the Deobfuscate check just ignores them anyway?

Indeed.

tests/test_deobfuscate_filenames.py
```diff
@@ -324,3 +325,28 @@ def test_deobfuscate_par2(self):
...
+        )  # the 20MB file ... needed as deobfuscate only renames >10MB files
+
+        list_of_files = []
+        for (dirpath, dirnames, filenames) in os.walk(work_dir):
+            list_of_files += [os.path.join(dirpath, file) for file in filenames]
+
+        # deobfuscate: first par2 based renaming, then deobfuscate obfuscated names
+        deobfuscate_list(list_of_files, "My Great Download")
+
+        assert os.path.isfile(os.path.join(work_dir, "My Great Download.bin"))  # the twentymb.bin should be renamed
```
Safihre, line 349: Is this a good test? Because I feel even without the par2 this would be the resulting filename due to the regular deobfuscate?

sanderjo:
> > So this code will run deobfuscate, even on files already processed by par2 right?
> > Shouldn't it actually ignore the files touched by par2, since par2 already renamed them?
> 
> We could. It would make the code easier ... no need to see which new files were generated. Just a check which files in file_list still exist

@Safihre  if you want that, the most of the PR is unneeded: just one line with re-assessing file_list is enough.

Safihre:
> > > So this code will run deobfuscate, even on files already processed by par2 right?
> > > Shouldn't it actually ignore the files touched by par2, since par2 already renamed them?
> > 
> > 
> > We could. It would make the code easier ... no need to see which new files were generated. Just a check which files in file_list still exist
> 
> @Safihre if you want that, the most of the PR is unneeded: just one line with re-assessing file_list is enough.

Your current version is indeed the best one, let's keep it this way.
Looks good!

PR #1923, opened 2021-07-07
Add repickling

puzzledsab:
I'm trying this again. I've tested it for a while and found at least one bug related to partially downloaded files since last time. I think I also got rid of the problem where it would repeatedly pickle and unpickle the same files so I feel more confident in this version. I've put the optional setting back in because it adds overhead, there might still be bugs and it probably won't be very helpful in most use cases.


puzzledsab:
Anyone know why the Git Black stopped on this code? Neither my VS Code check nor `python -m black --line-length 120 --target-version=py36 sabnzbd/*.py` complains about it.

sanderjo:
> Anyone know why the Git Black stopped on this code? Neither my VS Code check nor `python -m black --line-length 120 --target-version=py36 sabnzbd/*.py` complains about it.

The same happens to me. Safihre suggested me to update the black python module, which I did, but I still have git-black errors. To me they seem intermittent ... so after an error, I give it a few extra tries.

thezoggy:
dunno if it matters but the github action is doing: `-line-length=120 --target-version=py36 --check --diff`

if you use vs.code here is what I have in my settings.json
```
    "python.formatting.blackArgs": [
        "--line-lenth=120",
        "--target-version=py36"
    ],
    "python.formatting.provider": "black",
    "files.trimTrailingWhitespace": true,
```

puzzledsab:
Should we just drop this? I think more RAM could be saved in a safer way by doing something about the XML leak.

Safihre:
Yeah probably..

Issue #1931, opened 2021-07-22
Always deobfuscate names from par2?


puzzledsab:
I just noticed that the names from par2 files may not be deobfuscated if they require prior processing, like unrar. Wouldn't it be more consistent to always rename those, even if deobfuscate is disabled? Using the names from the par2 files isn't guessing, like the normal deobfuscation process. If they are included in par2 files then they are intended to have those particular names. They would be renamed if they weren't compressed.

Safihre:
Not completely sure I understand. Are you talking about the Deobfuscate final filenames option?

puzzledsab:
Yes, this part: https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/deobfuscate_filenames.py#L155

thezoggy:
the option should determine if one wants to or not to maintain flexibility. as example i actually keep the option off since i dont need it and want less things that can go wrong

puzzledsab:
But if the same files aren't compressed then they will be deobfuscated during the recovery phase even if deobfuscate final filenames is disabled. I think this makes more sense and it's required for the files to be useful if it contains for instance a season set.

Safihre:
So to always execute the par2 renaming if we detect a par2 file in the unpacked files?
Indeed I have to agree with @thezoggy, by default we only unpack and don't try to be smarter than that. If users want us to be smarter (aka Deobfuscate finale filenames), they can enable it.

puzzledsab:
Problem is that it can't be enabled without enabling the part that is pure guessing and has a much higher chance of breaking things instead of helping. Using names from par2 files is in all likelihood what the poster intended, otherwise why would they be there? Using ones from the package is not actually different from using the ones from the nzb.

Safihre:
@sanderjo what do you think? Splitting the functionality?

sanderjo:
> @sanderjo what do you think? Splitting the functionality?

By **always** running par2-based-renaming? Yes, that would be OK for me. (Not yet another option to check / unckeck, unless under specials)
Remark: unwanted_extension does not work beyond par2-based-renaming, so a malicious poster can circumvent that unwanted-feature that way: let the par2-rename popup a nasty exe

Related: IMHO deobfuscate should be on by default. But that is another discussion.

puzzledsab:
[...]

PR #1935, opened 2021-07-29
Always deobfuscate names from par2

puzzledsab:
As discussed in #1931


sabnzbd/postproc.py
```diff
@@ -517,6 +517,9 @@ def process_job(nzo: NzbObject):
                         nzo.set_unpack_info("Unpack", T("Failed to move files"))
                         all_ok = False
 
+            # Use par2 files to deobfuscate unpacked file names
+            newfiles = deobfuscate.recover_par2_names(newfiles)
```
Safihre, line 521: Let's add a Special setting to disable this, in case somehow it causes problems.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -74,6 +74,30 @@ def decode_par2(parfile):
     return new_files
 
 
+def recover_par2_names(filelist):
```
Safihre, line 77: If you could add simple Typing, that would be great. `List[str] `

sabnzbd/cfg.py
```diff
@@ -258,8 +260,8 @@ def validate_script(value):
 ampm = OptionBool("misc", "ampm", False)
 replace_illegal = OptionBool("misc", "replace_illegal", True)
 start_paused = OptionBool("misc", "start_paused", False)
-enable_all_par = OptionBool("misc", "enable_all_par", False)
 enable_par_cleanup = OptionBool("misc", "enable_par_cleanup", True)
+par2_filenames = OptionBool("misc", "par2_filenames", True)
```
Safihre, line 264: That's a bit vague, how about `process_unpacked_par2`?

PR #1946, opened 2021-09-08
Implement regex to match the filename in the content-disposition header.

p0psicles:
With the chain of pymedusa -> prowlarr -> sabnzbd. Medusa sends prowlarr's nzb download url to sabnzbd.
Sabnzbd downloads the nzb and get's a `Content-Disposition` header.

An example header is: `filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb`

Sabnzbd fails to properly parse this string. This PR fixes that, while also supporting the existing formats using a regular expression.

Forum issue: https://forums.sabnzbd.org/viewtopic.php?f=2&t=25496

The following strings will match:
filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb
filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb;
filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb

### DIFF
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -192,7 +192,7 @@ class URLGrabber(Thread):
 
                         # Get filename from Content-Disposition header
                         if not filename and "filename=" in value:
-                            filename_match = re.match(r".*filename[*=]+(?:UTF-8''?)*([^=;]*);?", value")
+                            filename_match = re.match(r".*filename[*=]+(?:UTF-8''?)*([^=;]*);?", value)
                             if filename_match:
                                 filename = filename_match.group(1).strip('"')
 
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -22,6 +22,7 @@ sabnzbd.urlgrabber - Queue for grabbing NZB files from websites
 import os
 import sys
 import time
+import re
 import logging
 import queue
 import urllib.request
@@ -191,7 +192,9 @@ class URLGrabber(Thread):
 
                         # Get filename from Content-Disposition header
                         if not filename and "filename=" in value:
-                            filename = value[value.index("filename=") + 9 :].strip(";").strip('"')
+                            filename_match = re.match(r".*filename[*=]+(?:UTF-8''?)*([^=;]*);?", value")
+                            if filename_match:
+                                filename = filename_match.group(1).strip('"')
 
                 if wait:
                     # For sites that have a rate-limiting attribute
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -22,7 +22,6 @@ sabnzbd.urlgrabber - Queue for grabbing NZB files from websites
 import os
 import sys
 import time
-import re
 import logging
 import queue
 import urllib.request
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -212,7 +212,7 @@ class URLGrabber(Thread):
                             nzo_info[item] = value
 
                         # Get filename from Content-Disposition header
-                        if not filename and 'filename' in value:
+                        if not filename and "filename" in value:
                             filename = filename_from_content_disposition(value)
 
                 if wait:
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -212,7 +212,7 @@ class URLGrabber(Thread):
                             nzo_info[item] = value
 
                         # Get filename from Content-Disposition header
-                        if not filename:
+                        if not filename and 'filename' in value:
                             filename = filename_from_content_disposition(value)
 
                 if wait:
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -29,6 +29,7 @@ import urllib.request
 import urllib.error
 import urllib.parse
 from http.client import IncompleteRead, HTTPResponse
+from mailbox import Message
 from threading import Thread
 import base64
 from typing import Tuple, Optional
@@ -58,6 +59,27 @@ _RARTING_FIELDS = (
     "x-rating-passworded",
     "x-rating-confirmed-passworded",
 )
+
+
+def filename_from_content_disposition(content_disposition):
+    """
+    Extract and validate filename from a Content-Disposition header.
+
+    Origin: https://github.com/httpie/httpie/blob/4c8633c6e51f388523ab4fa649040934402a4fc9/httpie/downloads.py#L98
+    :param content_disposition: Content-Disposition value
+    :type content_disposition: str
+    :return: the filename if present and valid, otherwise `None`
+    :example:
+        filename_from_content_disposition('attachment; filename=jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz')
+        should return: 'jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz'
+    """
+    msg = Message(f'Content-Disposition: {content_disposition}')
+    filename = msg.get_filename()
+    if filename:
+        # Basic sanitation.
+        filename = os.path.basename(filename).lstrip('.').strip()
+        if filename:
+            return filename
 
 
 class URLGrabber(Thread):
@@ -191,10 +213,8 @@ class URLGrabber(Thread):
                             nzo_info[item] = value
 
                         # Get filename from Content-Disposition header
-                        if not filename and "filename=" in value:
-                            filename_match = re.match(r".*filename[*=]+(?:UTF-8''?)*([^=;]*);?", value)
-                            if filename_match:
-                                filename = filename_match.group(1).strip('"')
+                        if not filename:
+                            filename = filename_from_content_disposition(value)
 
                 if wait:
                     # For sites that have a rate-limiting attribute
--- a/sabnzbd/urlgrabber.py
+++ b/sabnzbd/urlgrabber.py
@@ -73,7 +73,7 @@ def filename_from_content_disposition(content_disposition):
         filename_from_content_disposition('attachment; filename=jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz')
         should return: 'jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz'
     """
-    msg = Message(f'Content-Disposition: {content_disposition}')
+    msg = Message(f'Content-Disposition: attachment; {content_disposition}')
     filename = msg.get_filename()
     if filename:
         # Basic sanitation.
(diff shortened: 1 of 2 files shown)


thezoggy:
https://stackoverflow.com/questions/8035900/how-to-get-filename-from-content-disposition-in-headers/27840405

jcfp:
Another interesting option could be this: https://github.com/httpie/httpie/blob/4c8633c6e51f388523ab4fa649040934402a4fc9/httpie/downloads.py#L98 (mailbox is part of stdlib).

RFC 6266 allows for quite some variation in those content-disposition headers (spaces, caps), so using a regex is tricky. And in any case, the code does need to guard against obvious bs such as an empty filename.

Safihre:
@p0psicles I like @jcfp suggestion, could you try it and see if it could work for us?

p0psicles:
> Another interesting option could be this: https://github.com/httpie/httpie/blob/4c8633c6e51f388523ab4fa649040934402a4fc9/httpie/downloads.py#L98 (mailbox is part of stdlib)

I'll check it out. But what do you mean with `mailbox is part of stdlib`?

jcfp:
> But what do you mean with `mailbox is part of stdlib`

The standard library is a collection of stuff that comes bundled with Python releases, meaning this solution doesn't require any third-party modules.

p0psicles:
But it's part of httpie, not mailbox.

p0psicles:
This could also be an option. But I don't know what your policy is on importing exernal libs. 
https://github.com/g2p/rfc6266/blob/master/rfc6266.py#L209

Safihre:
@p0psicles the `mailbox` module used by httpie is part of Python base installation, so we could write a similar function as httpie does using `mailbox`. That's what @jcfp is suggesting. So indeed, no extra external modules.

p0psicles:
I'll implement it. 
I'll also try to add some tests for this

p0psicles:
@Safihre @jcfp i've added some basic pytests. And as you can see this method does not catch the non-rfc examples.
I could as a fallback, also put in my regex solution. But for example for this string:
`"filename=my.filename.without-UTF-8.nzb; filename*=UTF-8''my.filename.with-UTF-8.nzb"` it will match to `my.filename.without-UTF-8.nzb` so I don't know if we'd think thats acceptable?

Safihre:
The code from `httpie` seems wrong, it should include `attatchment; `:
```python
>>> content_disposition="filename*=UTF-8''my.filename.with-UTF-8.nzb"
>>> Message(f'Content-Disposition: {content_disposition}').get_filename()
None
>>> Message(f'Content-Disposition: attachment; {content_disposition}').get_filename()
'my.filename.with-UTF-8.nzb'

```

Safihre:
For some reason Github requires me to manually run the workflows after each commit because you are a first time contributor, seems some new security feature.

sabnzbd/urlgrabber.py
```diff
@@ -22,12 +22,14 @@
 import os
 import sys
 import time
+import re
```
jcfp, line 25: Leftover?

tests/test_urlgrabber.py
```diff
@@ -27,6 +27,7 @@
 import sabnzbd.urlgrabber as urlgrabber
 import sabnzbd.version
 from sabnzbd.cfg import selftest_host
+from sabnzbd.urlgrabber import filename_from_content_disposition
```
jcfp, line 30: No need for this import, already imported a few lines earlier. Simply call the function as `urlgrabber.filename_from_content_disposition()`.

p0psicles:
okay cool. So now only "the issue" with `filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD-utf.nzb` remains.

tests/test_urlgrabber.py
```diff
@@ -158,3 +159,26 @@ def test_http_not_found(self):
...
+            (
+                "filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb;",
+                "Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb"
+            ),
+            (
+                "filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb",
+                "Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb"
+            ),
+        ]
+    )
```
jcfp, line 185: might want to add tests for corner cases too, e.g. input that causes the filename to come up empty and verify the function return correctly for those (i.e., `None` rather than an empty string).

jcfp: Other corner cases would be filenames that try to include path elements (/what/ever/file.name, ../../dir/escape.txt).

p0psicles: I agree that we could check for some corner cases. But we should asume at some point that where parsing RFC compliant strings imo

p0psicles: Never mind. I"ve added the "/what/ever/file.name" test. And it passes. It extracts the basename.

tests/test_urlgrabber.py
```diff
@@ -158,3 +158,43 @@ def test_http_not_found(self):
...
+            (
+                "filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb;",
+                "Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb"
+            ),
+            (
+                "filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb",
+                "Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb"
+            ),
+            (
+                "attachment; filename=jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz",
```
Safihre, line 181: Supplying additionally `attachment` here, it's added by the function later.

sabnzbd/urlgrabber.py
```diff
@@ -59,6 +60,27 @@
...
+    :type content_disposition: str
+    :return: the filename if present and valid, otherwise `None`
+    :example:
+        filename_from_content_disposition('attachment; filename=jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz')
+        should return: 'jakubroztocil-httpie-0.4.1-20-g40bd8f6.tar.gz'
+    """
+    msg = Message(f'Content-Disposition: attachment; {content_disposition}')
+    filename = msg.get_filename()
+    if filename:
+        # Basic sanitation.
```
Safihre, line 78: Just to note (also for @jcfp) is that later this is passed through `sanitize_filename` so we are safe enough.

tests/test_urlgrabber.py
p0psicles: A real example of what's being passed to the function is: `attachment; filename=Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb`. So that also has the `attachment` portion in it?

Or maybe i'm misunderstanding you?

sabnzbd/urlgrabber.py
```diff
@@ -190,8 +212,8 @@ def run(self):
                             nzo_info[item] = value
 
                         # Get filename from Content-Disposition header
-                        if not filename and "filename=" in value:
-                            filename = value[value.index("filename=") + 9 :].strip(";").strip('"')
+                        if not filename and 'filename' in value:
```
Safihre, line 215: Black-formatting requires the use of `"` instead of `'`.

tests/test_urlgrabber.py
Safihre: Check this line: https://github.com/sabnzbd/sabnzbd/pull/1946/files#diff-33d9efe9612e1d7ed61b9b8cdff238927ced12175411df050f92d3492912be34R75
In the function itself you add the `attachment; ` part, so no need to add to the input in the tests here.
You also don't add it a few tests-up: https://github.com/sabnzbd/sabnzbd/pull/1946/files#diff-3aa82f6816c2807e5b3fbd07c7c10c7affb12f4dcbb81988decaa492dcbf7dfbR178

Safihre (changes requested):
Just a few changes and we can merge!

tests/test_urlgrabber.py
p0psicles: I just wanted to have a test resemble  a real situation where `attachment` is also passed. 
Like with the prowlarr example. Attachment is included a the string.

p0psicles: @Safihre you've asked me to make this change. And i've added `attachment; ` here. 
```python
Message(f'Content-Disposition: attachment; {content_disposition}')
```

But when debugging i notice that it will effectively parse: 
```
Message(f'Content-Disposition: attachment; attachment; filename=Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb')
```

Now this isn't that much of an issue. As also proven by my tests, it will get the filename from it. So as prowlarr is sending a content-disposition header conform RFC like: `attachment; filename=Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb` I'd also like to have something like this in one of the tests.

tests/test_urlgrabber.py
```diff
@@ -158,3 +158,43 @@ def test_http_not_found(self):
...
             self._runner(self.httpbin.url + "/no/such/file", 404)
+
+
+class TestFilenameFromDispositionHeader:
+    @pytest.mark.parametrize(
+        "header, result",
+        [
+            (
+                # In this case the first filename (not the UTF-8 encoded) is parsed.
+                "filename=Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.480p.x264-mSD-utf.nzb",
```
p0psicles, line 169: So if we really want to test a string like send by prowlarr. I need change this to: `attachment; filename=Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb; filename*=UTF-8''Zombie.Land.Saga.Revenge.S02E12.1080p.WEB.H264-SENPAI.nzb`

Safihre: Indeed you are right and I am wrong!

Safihre (approved):

p0psicles:
should be good now

thezoggy:
Probably shouldn't put actual release stuff as test name.. generalize it so some dmca twat doesn't freak out

p0psicles:
sorry for all the commits. Had some issues with your linter. Hope all is fine now.

Safihre:
Thanks! Will release a new release candidate soon.

PR #1948, opened 2021-09-09
Add required server option

### LINKED ISSUE #1939: Add a "Required" flag for a server
This is a request to add a "Required" flag to a server and appropriate logic, so that if this server stops responding, say due to a billing or login issue, sabnzbd does not blindly proceed to empty a configured block account.

It should instead stall until the "Required" server starts responding again.

puzzledsab:
Fixes #1939
I've tried to test it by using a wrong password or wrong server name. I couldn't figure out how to do a proper test where one server becomes unavailable while it's running. I set the "Will resume..." message as debug because otherwise there would be two lines for every retry. Not sure what is preferable.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -844,15 +844,17 @@ class Downloader(Thread):
                                 penalty = _PENALTY_UNKNOWN
                                 block = True
                             if block or (penalty and server.optional):
+                                retry_article = False
                                 if server.active:
                                     if server.required:
                                         sabnzbd.Scheduler.plan_required_server_resume()
+                                        retry_article = True
                                     else:
                                         server.deactivate()
                                         if penalty and (block or server.optional):
                                             self.plan_server(server, penalty)
                                 # Note that the article is discard for this server
-                                self.__reset_nw(nw, retry_article=False, send_quit=True)
+                                self.__reset_nw(nw, retry_article=retry_article, send_quit=True)
                             continue
                         except:
                             logging.error(
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -853,7 +853,7 @@ class Downloader(Thread):
                                         server.deactivate()
                                         if penalty and (block or server.optional):
                                             self.plan_server(server, penalty)
-                                # Note that the article is discard for this server
+                                # Note that the article is discard for this server if the server is not required
                                 self.__reset_nw(nw, retry_article=retry_article, send_quit=True)
                             continue
                         except:
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -73,6 +73,7 @@ class Server:
         "ssl",
         "ssl_verify",
         "ssl_ciphers",
+        "required",
         "optional",
         "retention",
         "send_group",
@@ -109,6 +110,7 @@ class Server:
         send_group,
         username=None,
         password=None,
+        required=False,
         optional=False,
         retention=0,
     ):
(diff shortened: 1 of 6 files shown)


jcfp:
> I couldn't figure out how to do a proper test where one server becomes unavailable while it's running.

Adding a nonexistent server (e.g. some unused port on localhost, with _required_ set) combined with a working server (sabnews test instance at a lower prio) might work? With both the client and server side controlled by the test environment, checking the lower-prio server for activity could be used to verify the _required_ flag.

sabnzbd/skintext.py
```diff
@@ -644,6 +644,10 @@
     "ssl_verify-strict": TT("Strict"),
     "srv-priority": TT("Priority"),  #: Server priority
     "explain-svrprio": TT("0 is highest priority, 100 is the lowest priority"),  #: Explain server priority
+    "srv-required": TT("Required"),  #: Server required tickbox
+    "explain-required": TT(
+        "Download will be paused while connection or login fails for this server"
```
Safihre, line 649: I don't think this explains the actual way it works:
`In case of connection failures, the download queue will be paused for a few minutes instead of skipping this server`. Or something like that.

sabnzbd/scheduler.py
```diff
@@ -383,6 +383,12 @@ def __check_diskspace(self, full_dir: str, required_space: float):
         if not sabnzbd.Downloader.paused:
             self.cancel_resume_task()
 
+    def __resume_downloading(self):
+        """Unpause Downloader if paused"""
+        if sabnzbd.Downloader.paused:
```
Safihre, line 388: Seems the check for it being paused is redundant? The `Downloader.resume()` also performs this check.

sabnzbd/scheduler.py
```diff
@@ -391,11 +397,24 @@ def plan_diskspace_resume(self, full_dir: str, required_space: float):
             self.__check_diskspace, "check_diskspace", 5 * 60, 9 * 60, "threaded", args=[full_dir, required_space]
         )
 
+    def plan_required_server_resume(self, interval: int = 5):
```
Safihre, line 400: I wonder if we should use `plan_resume()` in this case? This way the user gets a visual indication how long the pause will last until it tries again.
This is different from the diskspace pause, because for the diskspace pause, it gives a warning before pausing. But here, the queue would just switch to paused mode and magically unpause again after a while.

puzzledsab:
It took a while before I figured out that using the paused flag was the easiest way to avoid doing duplicate checking for each connection. The __resume_downloading stuff was legacy from that which I didn't realize was redundant. I mean to look into what jcfp tipped about but I won't have time for it in the next few days.

sabnzbd/downloader.py
```diff
@@ -835,12 +844,17 @@ def run(self):
...
+                                retry_article = False
                                 if server.active:
-                                    server.deactivate()
-                                    if penalty and (block or server.optional):
-                                        self.plan_server(server, penalty)
-                                # Note that the article is discard for this server
-                                self.__reset_nw(nw, retry_article=False, send_quit=True)
+                                    if server.required:
+                                        sabnzbd.Scheduler.plan_required_server_resume()
+                                        retry_article = True
```
Safihre, line 851: Good catch!

sabnzbd/scheduler.py
```diff
@@ -391,11 +391,19 @@ def plan_diskspace_resume(self, full_dir: str, required_space: float):
...
+
     def cancel_resume_task(self):
         """Cancel the current auto resume task"""
         if self.resume_task:
             logging.debug("Cancelling existing resume_task '%s'", self.resume_task.name)
-            self.scheduler.cancel(self.resume_task)
+            try:
+                self.scheduler.cancel(self.resume_task)
+            except ValueError:
+                pass
```
Safihre, line 406: Is this change still needed now `plan_resume` is used?

Safihre (approved):

Safihre:
Will you also update the wiki?
I see we also didn't add `process_unpacked_par2` yet.

PR #1961, opened 2021-10-06
Add testable socks support

ImportTaste:
Originally posted by @puzzledsab in #1894 on May 14:
> Issue #206
> It was a bit frustrating to make it work so I'm not going to spend any time making it pretty until I have some confirmation that it is usable. Hopefully this will build installable versions. I tried PySocks first but I couldn't make it work with SSL. It currently only works for NNTP but HTTP requests shouldn't be too hard.
> 
> The configuration is at the bottom of the General config page. It uses blocking calls for making connections, so that takes longer than usual. I don't have a proxy that requires username and password so I haven't tested that.
> 
> Be aware that the username and password for the proxy server will be included in the log file if debug logging is enabled.
---

@puzzledsab closed this PR a couple weeks after opening it without any specific reason given, but it might've been due to a perceived lack of interest based on the comment given here: https://github.com/sabnzbd/sabnzbd/issues/1894#issuecomment-855385692

Given there seems to be a few people interested, myself included, I'm resubmitting it. If @puzzledsab comes back and would prefer to reopen their PR, that's probably better.

@jcfp reviewed the code (https://github.com/sabnzbd/sabnzbd/pull/1894#r633123349) and asked why the proxy type was hard-coded, given the module supports more than just SOCKS 5. @puzzledsab never answered, but, as for my reason, since I only have a SOCKS 5 proxy to test with, I'm leaving the code as-is. It can always be revised later.

### DIFF
--- a/interfaces/Config/templates/config_general.tmpl
+++ b/interfaces/Config/templates/config_general.tmpl
@@ -217,6 +217,43 @@
             </fieldset>
         </div><!-- /col1 -->
     </div><!-- /section -->
+    <div class="section">
+        <div class="col2">
+            <h3>$T('proxy') <a href="$helpuri$help_uri" target="_blank"><span class="glyphicon glyphicon-question-sign"></span></a></h3>
+        </div><!-- /col2 -->
+        <div class="col1">
+            <fieldset>
+                <div class="field-pair">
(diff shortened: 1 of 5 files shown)


thezoggy:
I know one guy that wanted to use it to get around china firewall...

Safihre:
Still nobody testing this? 🙃

PR #1966, opened 2021-10-19
remove unused imports

jcfp:


thezoggy:
be careful with this, some of these are done by design for building app/exe where they arent included unless we add them explicitly because the codepath wouldnt hit them during compile

jcfp:
> Did you check other imports, for example with importchecker?

The only imports I removed are the ones I was reasonably sure were actually unused leftovers made redundant by code changes. Linters may give plenty of hints, but have only limited understanding of anything more complex or non-standard.

PR #1969, opened 2021-10-28
reintroduce uu support

### LINKED ISSUE #1660: [Request] Restore UUEDECODE Support
I've run into a few binary posts in various groups that still use UUENCODING so it would be nice if there was _optional_ uudecode support in SABNZBd.

(Ran into one as recently as Oct 20 2020).

jcfp:
Seems to work for some example jobs, both single and multipart. Not sure about the performance, as the only somewhat larger uu-encoded job would only complete on the slow freebie server of xsusenet for me.

I considered adding a "type hint" at the nzo level too (indicating what decoding to try first for the next nzf in the same job), but didn't for now in order not to complicated things.

Closes #1660

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -301,7 +301,7 @@ def decode_uu(article: Article, raw_data: List[bytes]) -> bytes:
             raise BadUu
 
     def is_uu_junk(line: bytes) -> bool:
-        """Determine if the lilne is empty or contains known junk data"""
+        """Determine if the line is empty or contains known junk data"""
         return (not line) or line == b"-- " or line.startswith(b"Posted via ")
 
     # Check the uu 'begin' line
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -22,12 +22,15 @@ sabnzbd.decoder - article decoder
 import logging
 import hashlib
 import queue
+from binascii import a2b_uu
+from io import BytesIO
 from threading import Thread
 from typing import Tuple, List, Optional
 
 import sabnzbd
 import sabnzbd.cfg as cfg
 from sabnzbd.constants import SABYENC_VERSION_REQUIRED
+from sabnzbd.encoding import ubtou
 from sabnzbd.nzbstuff import Article
 from sabnzbd.misc import match_str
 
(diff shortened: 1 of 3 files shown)


sabnzbd/decoder.py
```diff
@@ -154,41 +162,50 @@ def run(self):
...
-                    # Examine headers (for precheck) or body (for download)
-                    # Look for DMCA clues (while skipping "X-" headers)
-                    # Detect potential UUencode
-                    for line in raw_data:
-                        lline = line.lower()
-                        if b"message-id:" in lline:
+                    # Try uu-decoding
+                    if (not nzo.precheck) and article.nzf.type != "yenc":
+                        try:
+                            decoded_data = decode_uu(article, raw_data)
```
Safihre, line 189: It feels a bit wrong to have this here.
Shouldn't we set the `article.nzf.type` and then `sabnzbd.Decoder.process()`?

jcfp: I don't think so: we cannot be certain an article is uu, just because we get to this point. The yenc decoding attempt could have failed for some other reason, e.g. corrupt data or even a complete lack of an encoded payload (like in one of the examples you forwarded). Note that setting the nzf type to uu here - without having successfully decoded the article as such - would causes any remaining articles in the same nzf to never be tried with yenc.

Safihre: I actually meant to keep the old detection, so if we find `\nbegin` that would be the trigger.

jcfp: That approach only really works for singlepart uu, where every article comes with a begin line. With multipart, the entire nzf would be written off if the lowest_partnum article is missing/damaged/slow/just plain weird (e.g. that aislin nzb) even if all other articles were fine.

You did just give me ideas for some nasty functional tests though. 😀

Safihre: Well, in that case I would propose to try yenc always first. Even if we know the file is uu.
Yes it's a performance impact, but who cares in case of uu?

jcfp: > Even if we know the file is uu.
 
Why?? Continuing to try the other encoding after the `nzf.type` was successfully determined is pointless, unless uploaders were to start mixing encodings **within a single nzf**.

> Yes it's a performance impact, but who cares in case of uu?

I can't think of any user that wants to see a program intentionally waste resources for no gain whatsoever, as a punishment for downloading an nzb that just happens to use a less common encoding type. An encoding type they have no way to know in advance no less.

And I sure cared when writing the code: by avoiding futile decoding attempts (for either decoding type; notice how the fallback to uu is skipped once the type is set to yenc), and preventing the addition of uu support from having a measurable performance impact on a typical yenc-encoded job.

PR #1981, opened 2021-11-20
Compare email_endjob with int, not string

puzzledsab:


puzzledsab:
I searched for the same mistake elsewhere as well but I didn't find any more cases.

Issue #1982, opened 2021-11-20
3.5.0b1 - SSLEOFError


thezoggy:
sab 3.5.0b1 windows binary

was updating my servers with notes about how much blocks i have left (in preparation of BF deals) and afterwards saw a bunch of warnings.. unable to reproduce on demand (adding note/testing) but figured I'd note:

```
2021-11-20 10:41:44,975::DEBUG::[interface:130] Request GET /config/server/ from 192.168.0.2 [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:94.0) Gecko/20100101 Firefox/94.0] {}
2021-11-20 10:41:45,477::INFO::[notifier:123] Sending notification: Warning - [20/Nov/2021:10:41:45] ENGINE socket.error 8
Traceback (most recent call last):
  File "cheroot\server.py", line 1287, in communicate
  File "cheroot\server.py", line 1077, in respond
  File "cheroot\wsgi.py", line 145, in respond
  File "cheroot\wsgi.py", line 231, in write
  File "cheroot\server.py", line 1133, in write
  File "cheroot\makefile.py", line 438, in write
  File "cheroot\makefile.py", line 36, in write
  File "cheroot\makefile.py", line 45, in _flush_unlocked
  File "socket.py", line 723, in write
  File "ssl.py", line 1205, in send
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:2384)
 (type=warning, job_cat=None)
2021-11-20 10:41:45,477::WARNING::[_cplogging:213] [20/Nov/2021:10:41:45] ENGINE socket.error 8
Traceback (most recent call last):
  File "cheroot\server.py", line 1287, in communicate
  File "cheroot\server.py", line 1077, in respond
  File "cheroot\wsgi.py", line 145, in respond
  File "cheroot\wsgi.py", line 231, in write
  File "cheroot\server.py", line 1133, in write
  File "cheroot\makefile.py", line 438, in write
  File "cheroot\makefile.py", line 36, in write
  File "cheroot\makefile.py", line 45, in _flush_unlocked
  File "socket.py", line 723, in write
  File "ssl.py", line 1205, in send
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:2384)

2021-11-20 10:41:45,477::WARNING::[_cplogging:213] [20/Nov/2021:10:41:45] ENGINE socket.error 8
Traceback (most recent call last):
[...]

PR #1983, opened 2021-11-23
Bugfix/ciphers

Safihre:
As described in #1982.

Because of Python 3.10 we need to force the SSL-socket to also allow TLSv1.2 ciphers if no specific one is set.
The `HIGH` one seems to cover all the TLSv1.3 ones.
I can also connect to TLSv1 servers with this setting and `allow_old_tls` (like `news.premium-news.de`), so seems it covers enough ciphers to allow even old stuff.

@jcfp @puzzledsab @sanderjo

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -189,7 +189,7 @@ class Server:
                 logging.debug("%s: Connecting to address %s", self.host, ip)
             elif cfg.load_balancing() == 2:
                 # RFC6555 / Happy Eyeballs:
-                ip = happyeyeballs(self.host, port=self.port, use_ssl=self.ssl)
+                ip = happyeyeballs(self.host, port=self.port)
                 if ip:
                     logging.debug("%s: Connecting to address %s", self.host, ip)
                 else:
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -279,7 +279,7 @@ class NNTP:
         # Create SSL-context if it is needed and not created yet
         if self.nw.server.ssl and not self.nw.server.ssl_context:
             # Setup the SSL socket
-            self.nw.server.ssl_context = ssl.create_default_context()
+            self.nw.server.ssl_context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
 
             if sabnzbd.cfg.allow_old_ssl_tls():
                 # Allow anything that the system has
@@ -299,6 +299,9 @@ class NNTP:
             if self.nw.server.ssl_ciphers:
                 # At their own risk, socket will error out in case it was invalid
                 self.nw.server.ssl_context.set_ciphers(self.nw.server.ssl_ciphers)
+            else:
+                # Support at least TLSv1.2+ ciphers, as some essential ones are removed by default in Python 3.10
+                self.nw.server.ssl_context.set_ciphers("HIGH:TLSv1.2")
 
             # Disable any verification if the setup is bad
             if not sabnzbd.CERTIFICATE_VALIDATION:
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -281,13 +281,6 @@ class NNTP:
             # Setup the SSL socket
             self.nw.server.ssl_context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
 
-            if sabnzbd.cfg.allow_old_ssl_tls():
-                # Allow anything that the system has
-                self.nw.server.ssl_context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
-            else:
-                # We want a modern TLS (1.2 or higher), so we disallow older protocol versions (<= TLS 1.1)
-                self.nw.server.ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
-
             # Only verify hostname when we're strict
             if self.nw.server.ssl_verify < 2:
                 self.nw.server.ssl_context.check_hostname = False
@@ -299,9 +292,18 @@ class NNTP:
             if self.nw.server.ssl_ciphers:
                 # At their own risk, socket will error out in case it was invalid
                 self.nw.server.ssl_context.set_ciphers(self.nw.server.ssl_ciphers)
+                # Python does not allow setting ciphers on TLSv1.3, so have to force TLSv1.2 as the maximum
+                self.nw.server.ssl_context.maximum_version = ssl.TLSVersion.TLSv1_2
             else:
                 # Support at least TLSv1.2+ ciphers, as some essential ones are removed by default in Python 3.10
-                self.nw.server.ssl_context.set_ciphers("HIGH:TLSv1.2")
+                self.nw.server.ssl_context.set_ciphers("HIGH")
+
+            if sabnzbd.cfg.allow_old_ssl_tls():
+                # Allow anything that the system has
+                self.nw.server.ssl_context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
+            else:
+                # We want a modern TLS (1.2 or higher), so we disallow older protocol versions (<= TLS 1.1)
+                self.nw.server.ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
 
             # Disable any verification if the setup is bad
             if not sabnzbd.CERTIFICATE_VALIDATION:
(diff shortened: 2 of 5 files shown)


jcfp:
Looks sane at first glance.

Don't we face a similar issue elsewhere, e.g. with rss feeds, the urlgrabber (uses urllib.request.urlopen() that accepts an SSLContext as an argument) or in sabnzbd.test_cert_checking() (esp. when a user-set host offers a more limited cipher set than the default test host)?

On the docs side there probably should be a link to the `allow_old_ssl_tls` setting on wiki/advanced/ssl-ciphers, as the (typically outdated) user-set ciphers and those supported by the default context are increasingly likely to be mutually exclusive. The python docs for SSLContext explicitly mention it doesn't guard against invalid combinations.

puzzledsab:
It's a bit confusing to me what's overrides what. Why does it connect to premium-news when HIGH is set? Do they support at least one of the required ciphers, or does it get overridden somewhere? Also, it seems like it's not necessary to set the TLS version (neither in `ssl_context.minimum_version` nor `set_ciphers`) to connect to `bulknews.eu` if `set_ciphers("HIGH")` is there.

Safihre:
> Why does it connect to premium-news when HIGH is set? Do they support at least one of the required ciphers, or does it get overridden somewhere?

Seems they support at least one of the `HIGH` ones:
```
@news.premium-news.de: Connected using TLSv1 (AES256-SHA)
```
> Also, it seems like it's not necessary to set the TLS version
That is the case on Python 3.10 which is TLSv1.2+. But older Python versions we still want to enforce TLSv1.2.

thezoggy:
curious to see if people complain about slower speeds once upgrading due to higher tls being used...

jcfp:
Hmm wait, looks like that TLSv1.2 thing does something else than I expected.

For `openssl ciphers -v 'HIGH'` there's 140 entries and the docs say: _"High" encryption cipher suites. This currently means those with key lengths larger than 128 bits, and some cipher suites with 128-bit keys_. Included in those 140 entries is a bunch of TLSv1/SSLv3 and all but one of the TLSv1.2 entries (only NULL-SHA256 isn't part of HIGH). Which means that using HIGH:TLSv1.2 over HIGH only adds that NULL entry, i.e. a cipher offering no encryption at all. So that's not good.

'DEFAULT' has 60 entries and is entirely a subset of HIGH, which explains why some servers will only work with HIGH. And TLSv1.3 stuff seems impossible to disabled even if we wanted to. Looks like just HIGH is the way to go, if the DEFAULT is too restrictive as #1982 suggests?

Safihre:
> curious to see if people complain about slower speeds once upgrading due to higher tls being used...

SABnzbd 3.4.2 already used the new ciphers and TLSv1.3, if offered by the server. The highest one is always picked. So won't be any impact.

> also do we need to go through users server ciphers strings and flag ones that wont play nice with py3.10 (if they dont have it enabled to do the older ciphers)?

If a lower one is set, Python still accepts it and will connect. The lower ones by default are disabled if no `set_ciphers` is called but they are allowed to be used.

puzzledsab:
Unless there's a better way to make it work for servers that support v1.3, I suggest adding `self.nw.server.ssl_context.maximum_version = ssl.TLSVersion.TLSv1_2` if the user has set a fixed cipher.

thezoggy:
Why max? Seems like that would break 1.3 then?

puzzledsab:
Setting a cipher doesn't work if it's using v1.3. It will use the best one.

Safihre:
> Setting a cipher doesn't work if it's using v1.3. It will use the best one.

This is already the case for probably a year or 2 that we use an Python build with OpenSSL supporting TLSv1.3.
I don't see a reason to change the behavior.

puzzledsab:
Why not? It seems like a bug to me. It breaks the feature of setting a cipher on most servers and I can't think of a use case where it's helpful if the user has so clearly decided that they want a lower cipher.

thezoggy:
Let's make sure we are all on the same page for the user experience. If a user specified a cipher for their Usenet server it should be used. Now what if that cipher is not available (because of python change.. openssl version.. because of the require modern tls option)?

puzzledsab:
If the user enters or uses a cipher that doesn't work then most likely they'll get an SSL connection error, and they'll understand that they'll have to remove it or try another one. To be extra helpful, a hint about removing it could be added to the warning if SSL Ciphers is defined and there is an error,

Currently it doesn't work with most servers but the user won't know unless they check the log. It will say "Connection Successful!" either way.

thezoggy:
Maybe during test, show if cipher request was good or if it was discarded?

Safihre:
> Currently it doesn't work with most servers but the user won't know unless they check the log. It will say "Connection Successful!" either way.

The Cipher used is also shown on the Status and Interface settings window for each server.

But indeed.. it seems the setting you own cipher thing has been broken for a long time now. As almost everyone uses TLSv1.3 now: https://www.appelboor.com/newsservers/newsservers-with-SSL.html

So let's say we force TLSv1.2 if a cipher is set, but the cipher specified by the user is one from TLSv1.3?
Or should we say we only support TLSv1.2 ciphers..

And no: I'm not putting in nice error messages.. These are advanced settings that people should only touch if they know what they are doing.

sanderjo:
> Or should we say we only support TLSv1.2 ciphers..

If you mean: only support cipher with TLS1.2, then I would say Yes. Deprecate the option. It's something from 10 years ago, when SSL put a big load on systems without OS/HW supported encryption.

puzzledsab:
> So let's say we force TLSv1.2 if a cipher is set, but the cipher specified by the user is one from TLSv1.3?

That will fail either way. Only three options from 1.3 are available on most servers; `TLS_AES_128_GCM_SHA256`, `TLS_AES_256_GCM_SHA384` and `TLS_CHACHA20_POLY1305_SHA256`. If you try to force `TLS_AES_128_GCM_SHA256` it will say "No cipher can be selected" because setting it is not supported in 1.3. https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_ciphers

Safihre:
Right.
So, indeed, for user-configured ciphers to work we need to set the maximum TLS to 1.2 indeed.
@jcfp @sanderjo agreed?
We should also update the documentation to reflect this.

sanderjo:
So:
* if a user specifies a cipher, SAB will always use TLS.1.2, with that cipher.
* f a user does not specify a cipher, SAB will leave TLS version and cipher to the SSL negotiation?

If so: nice!

jcfp:
> Right. So, indeed, for user-configured ciphers to work we need to set the maximum TLS to 1.2 indeed.

So it seems (assuming they don't override that as well).

> We should also update the documentation to reflect this.

This needs tests too, not in the least because it relies quite heavily on implementation details (tlsv1.3 ciphers cannot be directly set, tlsv1.3 is always added to the list, setting the max to tlsv1.2 is actually honoured).

jcfp:
> * if a user specifies a cipher, SAB will always use TLS.1.2, with that cipher.

Not quite. The code will try to set the user-specified cipher, but the (python? openssl?) implementation still adds all of tlsv1.3 which is counteracted by setting the maximum to tlsv1.2 making the user-set value the only usable client-side cipher.

> * f a user does not specify a cipher, SAB will leave TLS version and cipher to the SSL negotiation

Without a user-set cipher, sab accepts only ciphers that are part of "HIGH" - unless the special option to permit old ssl stuff is active and every cipher available on the system is allowed. AFAIK ssl implementations will try to negotiate the strongest possible cipher shared between client and server, so this would typically use TLSv1.3 or something else on the upper end in terms of security.

Safihre:
> This needs tests too, not in the least because it relies quite heavily on implementation details (tlsv1.3 ciphers cannot be directly set, tlsv1.3 is always added to the list, setting the max to tlsv1.2 is actually honoured).

😢 Oh no.. tests.. This will be a pain.

Safihre:
> This needs tests too, not in the least because it relies quite heavily on implementation details (tlsv1.3 ciphers cannot be directly set, tlsv1.3 is always added to the list, setting the max to tlsv1.2 is actually honoured).

😢 Oh no.. tests.. This will be a pain.
But of course, you are right!

Safihre:
@jcfp Could you take a look at the unittests that I added?
What more can/should I test?
I did ignore `allow_old_ssl_tls` testing, but I also don't realllyy care about it. 
It also took me hours to build the current tests, figuring out all the right mocking and how to properly make the fake server.

jcfp:
Looks like you got the important stuff covered.

PR #1985, opened 2021-11-25
Add all passwords to list every time

### LINKED ISSUE #1979: Handle multiple passwords in the same nzb
As discussed in https://forums.sabnzbd.org/viewtopic.php?f=2&t=25621

puzzledsab:
Fixes #1979

The support was already pretty much there. I considered @jcfp's suggestion about moving the last working password to the end of the list, but then it gets really slow if that password is used for several or all the archives. I figure this will be used so rarely that it doesn't matter.

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -752,16 +752,15 @@ def create_https_certificates(ssl_cert, ssl_key):
 
 
 def get_all_passwords(nzo) -> List[str]:
-    """Get all passwords, from the NZB, meta and password file. In case the correct password is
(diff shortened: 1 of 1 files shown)


Safihre:
Looking at this, shouldn't we just remove getting the password from `nzo_info`?
User-set password is stored in `nzo.password`. Right?

puzzledsab:
The nzo_info password can also be set here: https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/urlgrabber.py#L183

PR #1986, opened 2021-11-28
Print low level Windows status error on assembler IOError

puzzledsab:
As discussed in https://forums.sabnzbd.org/viewtopic.php?f=2&t=25261

There are some other places where it could be added but this seems like the spot that usually gives Errno 13. It would be nice to get some feedback from those who have seen the problem.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -25,6 +25,7 @@ import logging
 import re
 from threading import Thread
 import hashlib
+import ctypes
 from typing import Tuple, Optional, List
 
 import sabnzbd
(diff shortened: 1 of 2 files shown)


Safihre:
Maybe you can also add it to `renamer`, since it's very often used.

puzzledsab:
Problem with the renamer is that it's used a lot of places but it doesn't catch it's own exceptions. I'm not sure how to add it in a way that doesn't break anything, except for adding it everywhere it's used.

Safihre:
Why not log it in this section?
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/filesystem.py#L945
That way at least it's logged when users report things.

Safihre:
Great!

PR #1987, opened 2021-11-30
Use regex parser to save RAM

puzzledsab:
I know regexes aren't popular but I've tried to make it check every line and fall back to the XML parser if there is any real content it doesn't understand. It reduces the RAM usage slightly more than lxml in my tests and doesn't require additional modules.

Rewriting it to read the file from disk instead of loading it into memory first would save a lot more RAM for very big nzbs.


puzzledsab:
I assume there's something about the nzb file it made that made it hang but haven't figured out how to replicate it. I tried various binary files, broken nzbs and files without trailing newline but they always fail the way they are supposed to. Maybe it's OS specific?

How do I run the tests on Windows?  `python -m unittest` does not seem to be it.

Safihre:
> How do I run the tests on Windows? `python -m unittest` does not seem to be it.
https://github.com/sabnzbd/sabnzbd/blob/develop/.github/workflows/integration_testing.yml#L38

Or if you want to run specific tests: `pytest -k test_functional`

puzzledsab:
It's the new certification tests in `tests/test_newswrapper.py`. They always work for me, though.

Another thing, maybe you could reduce the timeout for killing tests on Github? 6 hours is a bit much.

Safihre:
Pushed improvements: 207c5430c1dfaf641d15372334551b7985a15268

puzzledsab:
@Safihre: What's your opinion about this? The reduction in memory consumption is 70-80% for large queues and you know how I feel about optimization. lxml is a good alternative if you accept the increase in file size, though.

puzzledsab:
I think I found a better way.

Safihre:
Looking forward to that :)
I'd rather report this ElementTree memory leak to Python than to include the regex parser.
Already made some progress in creating a basic reproduction, just have to finish and report.

PR #1990, opened 2021-12-02
Preserve Downloader.paused state on restart

### LINKED ISSUE #1062: Feature request: preserve pause state on shutdown/restart
Just update to 2.3.1~rc2 on my Ubuntu 17.04 server. Before the update SABnzbd was paused. After the update SABnzbd automatically restarted and began downloading again.

I have a monthly data cap, if I didn't have a telegram bot I wouldn't have known that I was about to blast through it. 

Not sure if this is a bug or a feature request.

puzzledsab:
Fixes #1062.

A drop down box with three options; Resume, Paused and Preserve state would be prettier but it would still need two variables.


sabnzbd/cfg.py
```diff
@@ -260,6 +260,7 @@ def validate_script(value):
 ampm = OptionBool("misc", "ampm", False)
 replace_illegal = OptionBool("misc", "replace_illegal", True)
 start_paused = OptionBool("misc", "start_paused", False)
+preserve_state = OptionBool("misc", "preserve_state", False)
```
Safihre, line 263: Let's make it a bit more clear what it means: `preserve_paused_state`

sabnzbd/downloader.py
```diff
@@ -372,12 +372,17 @@ def remove_socket(self, nw: NewsWrapper):
     @NzbQueueLocker
     def set_paused_state(self, state: bool):
         """Set downloader to specified paused state"""
-        self.paused = state
+        if self.paused != state:
```
Safihre, line 375: What happens here? Why is this needed?
This code needs come comments.

sabnzbd/downloader.py
```diff
@@ -372,12 +372,17 @@ def remove_socket(self, nw: NewsWrapper):
...
+            if cfg.preserve_state():
+                cfg.start_paused.set(int(state))
+            self.paused = state
 
     @NzbQueueLocker
     def resume(self):
         # Do not notify when SABnzbd is still starting
         if self.paused and sabnzbd.WEB_DIR:
+            if cfg.preserve_state():
+                cfg.start_paused.set(0)
```
Safihre, line 385: `start_paused` is an `OptionBool` so use `.set(False)` and `.set(True)`.
And let's put it below the `sabnzbd.notifier.send_notification`

sabnzbd/downloader.py
```diff
@@ -386,6 +391,8 @@ def resume(self):
     def pause(self):
         """Pause the downloader, optionally saving admin"""
         if not self.paused:
+            if cfg.preserve_state():
+                cfg.start_paused.set(1)
```
Safihre, line 395: Same comments as the `resume`.

Safihre (changes requested):
Just so small changes requested.

sabnzbd/downloader.py
puzzledsab: It's not strictly needed but it avoids doing the check and update if there is no change. It's also not necessary to set paused if it's unchanged so I moved it to the same level.

Safihre (approved):

Safihre:
Could you also add it to the documentation for 3.5?

PR #1992, opened 2021-12-05
Read XML from file using iterparse

puzzledsab:
This reads the nzb file in the admin dir in increments instead of keeping the full text in RAM until it's parsed. This works much better because it doesn't have to keep as much stuff in RAM at the same time and it seems to leak a lot less.

I'm not sure why the line `re.sub(r"""\s(xmlns="[^"]+"|xmlns='[^']+')""", "", raw_data, count=1)` is there (to ignore the charset?) but it is now done before the nzb file is saved to the admin dir, but after the backup nzb is saved. The backup is deleted if the import fails. Hopefully it's not too important that the one in the admin dir is modified.

At least twice as big nzbs can be imported in the same amount of RAM and large queues use about 70% less after import. It would be an advantage if we didn't need to read the full nzb file into RAM but currently that's necessary to run `correct_unknown_encoding`.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -600,15 +600,15 @@ def backup_exists(filename: str) -> bool:
     return path and os.path.exists(os.path.join(path, filename + ".gz"))
 
 
-def backup_nzb(filename: str, data: AnyStr):
-    """Backup NZB file"""
+def backup_nzb(filename: str, data: AnyStr) -> Optional[str]:
+    """Backup NZB file, return path to nzb if it was saved"""
     path = cfg.nzb_backup_dir.get_path()
     if path:
         return save_compressed(path, filename, data)
 
 
-def save_compressed(folder: str, filename: str, data: AnyStr):
-    """Save compressed NZB file in folder"""
+def save_compressed(folder: str, filename: str, data: AnyStr) -> str:
+    """Save compressed NZB file in folder, return path to saved nzb file"""
     if filename.endswith(".nzb"):
         filename += ".gz"
     else:
@@ -623,7 +623,7 @@ def save_compressed(folder: str, filename: str, data: AnyStr):
             f.flush()
             f.close()
     except:
-        logging.error(T("Saving %s failed"), os.path.join(folder, filename))
+        logging.error(T("Saving %s failed"), full_nzb_path)
         logging.info("Traceback: ", exc_info=True)
 
     return full_nzb_path
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -604,7 +604,7 @@ def backup_nzb(filename: str, data: AnyStr):
     """Backup NZB file"""
     path = cfg.nzb_backup_dir.get_path()
     if path:
-        save_compressed(path, filename, data)
+        return save_compressed(path, filename, data)
 
 
 def save_compressed(folder: str, filename: str, data: AnyStr):
@@ -613,10 +613,11 @@ def save_compressed(folder: str, filename: str, data: AnyStr):
         filename += ".gz"
     else:
         filename += ".nzb.gz"
-    logging.info("Backing up %s", os.path.join(folder, filename))
+    full_nzb_path = os.path.join(folder, filename)
+    logging.info("Backing up %s", full_nzb_path)
     try:
         # Have to get around the path being put inside the tgz
-        with open(os.path.join(folder, filename), "wb") as tgz_file:
+        with open(full_nzb_path, "wb") as tgz_file:
             f = gzip.GzipFile(filename, fileobj=tgz_file, mode="wb")
             f.write(encoding.utob(data))
             f.flush()
@@ -624,6 +625,8 @@ def save_compressed(folder: str, filename: str, data: AnyStr):
     except:
         logging.error(T("Saving %s failed"), os.path.join(folder, filename))
         logging.info("Traceback: ", exc_info=True)
+
+    return full_nzb_path
 
 
 ##############################################################################
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -18,6 +18,7 @@
 import os
 import logging
 import datetime
+import shutil
 import tempfile
 import pickle
 import ctypes.util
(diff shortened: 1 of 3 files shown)


thezoggy:
Was to strip the namespace, since it wasn't always included we got rid of it before the rest of the check

The xmlns attribute specifies the xml namespace for a document.

Safihre:
> Was to strip the namespace, since it wasn't always included we got rid of it before the rest of the check
> 
> The xmlns attribute specifies the xml namespace for a document.

Indeed, @puzzledsab if the iterparse can do without the replacement, but before it was required to work reliable. Otherwise for every ElementTree command you needed to specify the namespace.

Safihre:
I have this massive 750GB download some user send me, and on the `develop` parser it bumps the memory usage of SABnzbd to a whopping 1.5GB and then get's stuck at 500MB after the import.
However, on this branch the memory gets up to 180MB max and then drops to 80MB (almost same as before) after the import.
Nice work 👍

Safihre:
I always run SABnzbd with ` python -X dev SABnzbd.py` and it shows me always this:
```
C:\Users\safihre\Documents\GitHub\sabnzbd3\venv\lib\site-packages\cherrypy\_cpwsgi.py:282: ResourceWarning: unclosed file <_io.BufferedRandom name=4>
  self.cpapp.release_serving()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
Maybe that's where our remaining lost memory is coming from?

sabnzbd/nzbstuff.py
```diff
@@ -763,8 +763,11 @@ def __init__(
             remove_all(admin_dir, "SABnzbd_article_*", keep_folder=True)
 
         if nzb_data and "<nzb" in nzb_data:
+            backup_nzb = sabnzbd.backup_nzb(filename, nzb_data)
+            full_nzb_path = sabnzbd.save_compressed(admin_dir, filename, nzb_data)
+            nzb_data = None
```
Safihre, line 768: Just a small note:
This doesn't really do anything because the reference to the string-data is still alive in the calling function.

sabnzbd/nzbparser.py
```diff
@@ -54,98 +49,123 @@ def nzbfile_parser(raw_data: str, nzo):
...
+        except:
+            # Remove all data added to the nzo
+            for nzf in nzo.files:
+                nzf.remove_admin()
+            nzo.first_articles = []
+            nzo.first_articles_count = 0
+            nzo.bytes_par2 = 0
+            nzo.files = []
+            nzo.files_table = {}
+            nzo.bytes = 0
```
Safihre, line 167: I saw you added this, but why?
It breaks the `allow_incomplete_nzb` functionality. I know that's a weird one, but some users wanted it.
Additionally, if `allow_incomplete_nzb` is disabled, the `nzo.purge_data()` in `NzoObject.__init__` will clear all the stuff out, right?

sabnzbd/nzbparser.py
```diff
@@ -54,98 +49,123 @@ def nzbfile_parser(raw_data: str, nzo):
...
-                        else:
-                            logging.info("Skipping duplicate article (%s)", article_id)
-                    elif segment_size <= 0 or segment_size >= 2 ** 23:
-                        # Perform sanity check (not negative, 0 or larger than 8MB) on article size
-                        # We use this value later to allocate memory in cache and sabyenc
-                        logging.info("Skipping article %s due to strange size (%s)", article_id, segment_size)
-                        nzo.increase_bad_articles_counter("bad_articles")
+            for _, element in xml.etree.ElementTree.iterparse(nzb_fh):
+                # Ignore namespace
+                _, has_namespace, postfix = element.tag.partition("}")
```
Safihre, line 57: Nice 👍

sabnzbd/nzbstuff.py
puzzledsab: That's true. I initially added it because removing the namespace from the data string was done there so it would be a new copy, which is why I moved `backup_nzb()` above `nzbfile_parser()`.

sabnzbd/nzbparser.py
puzzledsab: I'm not entirely sure if it should be there. This part of the parsing doesn't work quite like before. In the old version, if the XML parser found a problem then it would throw an exception and the import loop would never start so the `nzo` would always be empty.

Now that I think of it, I think it shouldn't be there. I added it while working on the regex parser and then I figured I needed it here as well because both import the file incrementally. However, the reason why it was necessary in the regex version was that the XML parser would use the same `nzo`.

Safihre:
Perfect.
Now we would really need some unit tests for this.. 
In `testhelper` there is the `create_nzb` function that can create an NZB from all files in a folder.
We could use that to create an NZB and check the contents against the files from the sources folder. As in: number of files, filenames and filesizes.

Safihre:
@puzzledsab Looking back at the old 2.3.x code, I don't see anywhere that we do any encoding fixing. 
I suspect I just added it "just in case" when doing the Python 2 to 3 conversion, but it's not needed.
If that's the case (99% it is), we can just use file handlers everywhere instead of loading the data in memory.
Or maybe we should keep that for a next PR.

puzzledsab:
I looked into using a filehandle and I think it should be a different PR. The check for incomplete nzbs would have to be done differently and I haven't quite understood the whole zipfile/rarfile procedure yet.

sabnzbd/__init__.py
```diff
@@ -600,31 +601,34 @@ def backup_exists(filename: str) -> bool:
...
         # Have to get around the path being put inside the tgz
-        with open(os.path.join(folder, filename), "wb") as tgz_file:
-            f = gzip.GzipFile(filename, fileobj=tgz_file, mode="wb")
-            f.write(encoding.utob(data))
-            f.flush()
-            f.close()
+        with open(full_nzb_path, "wb") as tgz_file:
+            # We only need minimal compression to prevent huge files
+            with gzip.GzipFile(filename, mode="wb", compresslevel=1, fileobj=tgz_file) as gzip_file:
+                gzip_file.write(encoding.utob(data))
```
jcfp, line 625: Opening tgz_file in a separate context manager before calling GzipFile isn't necessary: in case no fileobj is passed, the given filename is automatically opened to provide a file object.

Safihre: @jcfp No, otherwise it puts the whole filepath in the `tgz` if we don't do the double context manager.

jcfp: In that case, never mind...

puzzledsab: Isn't try and then two levels of with a bit excessive, though?

Safihre: Why? Each `with` serves the purpose of closing their respective file descriptors (error or no error) and the try/except serves the purpose of catching the error.

puzzledsab: I figured the first `with` would close the file anyway, but looking closer it's probably safer to do it this way.

Safihre:
@puzzledsab do you want to make the unit tests? If not, I can merge it and add the later.

puzzledsab:
I'll look into it but I need to learn how it works so it may take some time.

Safihre:
Ok, will merge this for now!

Safihre:
@puzzledsab Tried with the filepointers and got it to work for `.nzb` and `.nzb.gz`, but it fails for `.zip`/`.rar`/`.7z`.
So probably we can keep it the current way.

PR #1994, opened 2021-12-09
Rework adding of NZB's to use filehandlers

Safihre:
@puzzledsab This way it works.
What do you think?

### DIFF
--- a/sabnzbd/nzbparser.py
+++ b/sabnzbd/nzbparser.py
@@ -21,6 +21,7 @@ sabnzbd.nzbparser - Parse and import NZB files
 import bz2
 import gzip
 import time
+import io
 import logging
 import hashlib
 import xml.etree.ElementTree
@@ -29,7 +30,7 @@ from typing import Optional, Dict, Any, Union
 
 import sabnzbd
 from sabnzbd import filesystem, nzbstuff
-from sabnzbd.encoding import utob, correct_unknown_encoding
+from sabnzbd.encoding import utob
 from sabnzbd.filesystem import is_archive, get_filename
 from sabnzbd.misc import name_to_cat
 
@@ -50,7 +51,7 @@ def nzbfile_parser(full_nzb_path: str, nzo):
     valid_files = 0
 
     # Use nzb.gz file from admin dir
-    with gzip.open(full_nzb_path) as nzb_fh:
+    with gzip.open(full_nzb_path, "r") as nzb_fh:
         for _, element in xml.etree.ElementTree.iterparse(nzb_fh):
             # For type-hinting
             element: xml.etree.ElementTree.Element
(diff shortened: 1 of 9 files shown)


Safihre:
I did notice that this way is a bit slower, probably due to the chunky read/write when writing the compressed file.

puzzledsab:
Doesn't `io.BytesIO(zf.read(name))` mean that the full text is read from the file and then made into an object which feeds the text through a filehandle? I don't think it saves any RAM.

I think the reasons why it's slow are that reading the data through the io object is slow (I noticed this when I used the same method for feeding the regex version) and `shutil.copyfileobj` only uses 16 kB buffer by default. I don't know if it reacquires the GIL between each read/write but if it does then a significantly larger buffer would be much better.

sabnzbd/nzbparser.py
```diff
@@ -349,13 +354,17 @@ def process_single_nzb(
         # Empty
         return 1, nzo_ids
     except:
-        if data.find("<nzb") >= 0 > data.find("</nzb"):
+        # Read all the data to check if it's incomplete
+        data = nzb_fp.read()
```
puzzledsab, line 358: I think you need to do seek(0) first.

Safihre:
The gzip, Rar and 7zip handlers do not support a full file interface (so no `seek()` etc). So indeed, for those we have to load them into memory and fake them as file pointers..

I'm not really convinced the changes in this PR are really needed, only for huge NZB files it makes a difference.

puzzledsab:
Yes, unfortunately I don't see how it can help if it's always loading the whole file into RAM anyway.

puzzledsab:
Replacing `datap = io.BytesIO(zf.read(name))` with `datap = zf.open(name)` seems to work for both zip and rar, though? I haven't tested 7zip but I assume that works too.

puzzledsab:
Didn't work with 7zip, but you could make an exception for those. Alternatively, the nzb files could always be extracted/copied to the root of the incomplete dir and moved when the subdir is created or deleted if it isn't imported correctly.

Safihre:
> Didn't work with 7zip, but you could make an exception for those. Alternatively, the nzb files could always be extracted/copied to the root of the incomplete dir and moved when the subdir is created or deleted if it isn't imported correctly.

We can fix that, since the `SevenZip` is our own implementation (see `newsunpack.py`).
For now I applied your change of `datap = zf.open(name)` and indeed it works!
Will work on `SevenZip` next.

Safihre:
Updated, all working now.
Any comments @sabnzbd/contributers ?

thezoggy:
If I recall correctly, if I had different nzb names the same release. Added them to sab it would keep the 2nd one with a .1 with your change does it now silently toss it aside because it's the same name and assumed the same contents?

puzzledsab:
I would probably have extracted it to a temporary file to save the RAM but this is much simpler. Most of the benefit comes from the xml import and I don't think 7zip is common for nzbs anyway. Going through the `nzb_reader_handler` is probably not necessary anymore, you could just open it directly with the correct handler.

Regarding the .1 extension for duplicates, that seems to work like before for me. The names in the GUI are the same but the second will get a .1 extension to the path in incomplete. nzbs with the same name in the nzb backup dir will get overwritten but they always would.

Issue #1989, opened 2021-12-01
Handling permissions in ACL-based environments like DSM7


ludovicchabant:
Hi,

I installed SABnzbd on a Synology running DSM7 and ran into some permission issues. These issues are detailed in [this SynoCommunity ticket][1], with [detailed steps in this message][2].

Based on what [BenjV says][3], DSM7 is using ACLs and, as such, SABnzbd setting permissions using umasks might not work very well. 

The problem may be fixed if SABnzbd was able to recognize that ACLs are in use, and just skip running `set_permissions` altogether in that case. Apparently, it would make the completed downloads naturally inherit their parent directory's permissions.

Worst case, if ACLs can't be auto-detected, maybe the Synology package can install SABnzbd with a default `permissions` config setting set to a special value (like, say, `__acl__` or whatever), which can be checked to ignore setting old-style permissions.

[1]: https://github.com/SynoCommunity/spksrc/issues/4940
[2]: https://github.com/SynoCommunity/spksrc/issues/4940#issuecomment-982919167
[3]: https://github.com/SynoCommunity/spksrc/issues/4940#issuecomment-984004910

Safihre:
@jcfp @sanderjo Help, Linux permissions.. Why again do we set permissions even though no `umask` is set? Aren't permissions fine by default if SABnzbd creates the folders?
Never understood that part, I just know it kinda works.

sanderjo:
> Why again do we set permissions even though no `umask` is set? 

Oh, do we? Where/how?

> Aren't permissions fine by default if SABnzbd creates the folders?

Yes, I would say so.

But the synology DSM7 problems might be different: it seems each process on Synology-DSM7 is now a different user, causing that other process (*arr) can't do everything on stuff delivered by SAB ...

puzzledsab:
Digging back in the commit history indicates that it was initially added to make sure Sab could read the files it had written. It seems to me like the default should be to not set permissions, and then have a switch to turn on the current functionality. Previously it could be set with a command line parameter so maybe it has to be on by default for Sab to start in some cases.

thezoggy:
The preferred solution for dsm7 is to use docker

Safihre:
> The preferred solution for dsm7 is to use docker

.... How does that change anything? SAB breaks the permission system of Synology because it sets low level unix permissions, whether in Docker or not.

ludovicchabant:
> Oh, do we? Where/how?

AFAICT this [happens here][1].

[...]

PR #1999, opened 2021-12-19
check default posix acl before setting permissions

jcfp:
Change `filesystem.set_permissions` to not act if the given path is subject to a default posix access control list. In that case, it leaves things alone and assumes whoever set the ACL knows best.

I'm aware the CI trips over the acl dev header, but E_NOTIME right now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,6 +18,9 @@ pywin32>=227; sys_platform == 'win32'
 # macOS system calls
 pyobjc; sys_platform == 'darwin'
 
+# Optional detection of default posix ACLs; requires libacl1 headers
+pylibacl; sys_platform != 'win32'
+
 # Optional support for *nix tray icon.
 # Note that pygobject depends on pycairo, which requires pkg-config and cairo headers.
 # See https://pycairo.readthedocs.io/en/latest/getting_started.html
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -32,6 +32,13 @@ import ctypes
 from typing import Union, List, Tuple, Dict, Optional
 
 try:
+    import posix1e
+
+    HAVE_POSIX_ACL = True
+except ImportError:
+    HAVE_POSIX_ACL = False
+
+try:
     import win32api
     import win32file
 except ImportError:
@@ -42,6 +49,7 @@ from sabnzbd.decorators import synchronized
 from sabnzbd.constants import FUTURE_Q_FOLDER, JOB_ADMIN, GIGI, DEF_FILE_MAX
 from sabnzbd.encoding import correct_unknown_encoding, utob, ubtou
 from sabnzbd.utils import rarfile
+
 
 # For Windows: determine executable extensions
 if os.name == "nt":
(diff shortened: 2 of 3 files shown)


sanderjo:
poor man's ACL checker if posix1e and getfacl are not available (as on a default DSM7 and Ubuntu):

```
def is_ACL_set_3(mydir):
	cmd = 'ls -al ' + mydir
	alllines = os.popen(cmd).readlines()
	# the second line contains the directory info we want
	return alllines[1].rstrip("\n").split()[0].endswith("+")
```

puzzledsab:
How about making more generic flag for turning off setting of file permissions? Either a variable that is on by default and turned off on start up if permissions should not be set or a special variable that can also be turned off by the user.

Safihre:
@jcfp Before we go down this road, I still have no answer why we do these forced-permssions in the first place?

What if we just disable applying any permissions if `umask = ""`? And then we wait for users to report problems...

jcfp:
> Before we go down this road, I still have no answer why we do these forced-permssions in the first place?

See the initial line of [1989](https://github.com/sabnzbd/sabnzbd/issues/1989#issuecomment-985332671).

What changed in [3705e1f](https://github.com/sabnzbd/sabnzbd/commit/3705e1f3ee626732816f9cffc678cef098256ee7) is taking the session umask setting into account (note: that is separate from the sab setting of the same name), not sab dealing with permissions. The session umask is a useful default in environments that rely on a combo of permissions and group membership to enable access to files (less advanced but easier to understand than posix ACLs).

> How about making more generic flag for turning off setting of file permissions? Either a variable that is on by default and turned off on start up if permissions should not be set or a special variable that can also be turned off by the user.

I considered that approach, but it would only work reliably in locked-down environments (e.g. some nas appliances) where someone knowledgable can tightly control startup arguments. Otherwise the ACL would be overruled as soon as sab is started without that flag. Most users will only read up on such things after they ran into trouble. The benefit of autodetecting posix acls is that it can avoid such issues and make sab work with them out of the box (without any real disadvantage).

Safihre:
I checked what NZBGet does, and it just doesn't do any permissions. It just allows you to configure the application umask.
So why do we need to do something special?
Shouldn't we just check at start-up if the default session umask allows SABnzbd to access it's own files if no Permissions are set? And if so, we just don't do any permissions?
And maybe check if we need to remove the extra execute bit if we find any set? NZBGet doesn't seem to do that either..

And maybe that's also something for the ACL-detection approach? Don't we still need to make sure there's no execute bits?

jcfp:
Whether or not to bother with setting permissions at all is a design choice (from way back, but still). In general, sab tends to err on the side of caution and noob-friendlyness.

Checking ACLs on startup isn't optimal as these apply to specific directories (rather than the entire session like the umask). Both the dirs used by sab as well as the ACLs are subject to change at any time while the program is running, with users reasonably expecting such changes to take effect immediately without the need for a restart, logout or system reboot.

Default posix acl causes files to be created without the execute bit set, i.e. the `user::rwx,group::---,other::r-x` used in the tests sets `rwx---r-x` for dirs but `rw----r--` for files. So no need for a check there afaict.

Safihre:
> Checking ACLs on startup isn't optimal as these apply to specific directories (rather than the entire session like the umask)

I meant to check the umask of the session at startup, not the ACL's.

Safihre:
Will work a bit on an alternative and if I am successful you can be the judge what's best :)

PR #2000, opened 2021-12-21
if too little disk space reported by 7z unzipping, report so in GUI

sanderjo:
Triggered by https://forums.sabnzbd.org/viewtopic.php?f=3&t=25652 

Result in case of too little disk space:

image

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -127,7 +127,7 @@ def find_programs(curdir):
         if not sabnzbd.newsunpack.ZIP_COMMAND:
             sabnzbd.newsunpack.ZIP_COMMAND = find_on_path("unzip")
         if not sabnzbd.newsunpack.SEVEN_COMMAND:
-            sabnzbd.newsunpack.SEVEN_COMMAND = find_on_path("7za") # 7za = 7z stand-alone executable
+            sabnzbd.newsunpack.SEVEN_COMMAND = find_on_path("7za")  # 7za = 7z stand-alone executable
         if not sabnzbd.newsunpack.SEVEN_COMMAND:
             sabnzbd.newsunpack.SEVEN_COMMAND = find_on_path("7z")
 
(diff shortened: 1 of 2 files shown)


sabnzbd/newsunpack.py
```diff
@@ -1053,6 +1053,10 @@ def seven_extract_core(sevenset, extensions, extraction_path, one_folder, delete
         # We can output a more general error
         ret = 1
         msg = T('ERROR: CRC failed in "%s"') % setname_from_path(sevenset)
+    if ret == 2 and (("Disk full." in output) or ("No space left on device" in output)):
+        # note: the above does not work with 7z version 16.02, and does from with 7z 19.00 and higher
+        ret = 1  # to avoid ret = 2, causing error about password
+        msg = T('ERROR: Not enough disk space for unzipping "%s"') % setname_from_path(sevenset)
```
Safihre, line 1059: No no, we don't do that `ERROR` stuff anymore. That's from ancient times.
Let's make it:
`Unpacking failed, no space left on device.

sabnzbd/newsunpack.py
```diff
@@ -1053,6 +1053,10 @@ def seven_extract_core(sevenset, extensions, extraction_path, one_folder, delete
         # We can output a more general error
         ret = 1
         msg = T('ERROR: CRC failed in "%s"') % setname_from_path(sevenset)
+    if ret == 2 and (("Disk full." in output) or ("No space left on device" in output)):
```
Safihre, line 1056: Let's make this a nice `if / elif`, so first check for the disk-full and then `elif` check for the CRC-failure.

Safihre (changes requested):

sabnzbd/newsunpack.py
sanderjo: plain "unpacking"? Not "unzipping"?

sanderjo:
note to self, and @Safihre : after the failed download, SAB says "complete". That is not good, I assume?

Safihre:
> note to self, and @Safihre : after the failed download, SAB says "complete". That is not good, I assume?

No that's wrong, it should set the right flags to make it show failed.

sabnzbd/newsunpack.py
sanderjo: > > note to self, and @Safihre : after the failed download, SAB says "complete". That is not good, I assume?
> 
> No that's wrong, it should set the right flags to make it show failed.

Found the cause:

- with just a .7z file, it results in nice failure message in the GUI, including the download itself
- however, with several .7z.001 etc files, there is a failure message in the GUI, but the GUI says "completed" for the download. Probable cause: SAB/7z does a two-pass method on .7z.001 files: first converts it to one big .7z (OK), and then unpacks that (not OK if too little disk space)

See below.

And proof no translation is needed: it works in Nederlands:

image

sanderjo:
image

PR #2004, opened 2021-12-26
Do not apply permissions if not requested by the user, but remove xbits

Safihre:
I propose to change the way we handle permissions and no longer force any if the user didn't request them.
With 1 exception: we remove any execution-related bits on files if we find them.
@jcfp What do you think?

If you approve, there is other things that still need to be done:

- [x] Check `os.umask` at start-up to see if it at least allows user read/write on created files.
- [x] Check the Incomplete Directory at startup to see if we have read/write permissions.
- [x] More tests, for example for `removexbits`.
- [ ] Documentation.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1205,6 +1205,13 @@ def main():
             sabnzbd.encoding.CODEPAGE,
         )
 
+    # Verify umask, we need at least 700
+    if not sabnzbd.WIN32 and sabnzbd.ORG_UMASK > int("077", 8):
+        sabnzbd.misc.helpful_warning(
+            T("Current umask (%o) might deny SABnzbd access to the files and folders it creates."),
+            sabnzbd.ORG_UMASK,
+        )
+
     # SSL Information
     logging.info("SSL version = %s", ssl.OPENSSL_VERSION)
 
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -248,9 +248,6 @@ def initialize(pause_downloader=False, clean_up=False, repair=0):
         cfg.download_dir.set(cfg.download_dir(), create=True)
     cfg.download_dir.set_create(True)
 
-    # Set access rights for "incomplete" base folder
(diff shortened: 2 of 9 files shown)


sanderjo:
> I propose to change the way we handle permissions and no longer force any if the user didn't request them.

... which will take care of ACL setups (on DSM7, ad other POSIX systems): if a user wants to use the ACL setup, she/he should not fill out permission settings?


> With 1 exception: we remove any execution-related bits if we find them.

Sounds OK, but why? Becasue of security ... ? 
Please note:
* directories have x-bits. AFAIK they are needed to access and read that directory. Google hit: "A directory with its x bit set allows the user to cd (change directory) into this directory, and access the files in it. The ability to read the names of files stored in this directory. The ability to rename files in the directory, create new files, or delete existing files, if you also have Execute permissions."
* all my files on my exFAT drive have -rwxrwxrwx . Nothing SAB can do about that. However, trying to remove the x bit does not result in an error or a problem

Safihre:
> Sounds OK, but why? Becasue of security ... ? Please note:
> 
> * directories have x-bits. AFAIK they are needed to access and read that directory. Google hit: "A directory with its x bit set allows the user to cd (change directory) into this directory, and access the files in it. The ability to read the names of files stored in this directory. The ability to rename files in the directory, create new files, or delete existing files, if you also have Execute permissions."
> * all my files on my exFAT drive have -rwxrwxrwx . Nothing SAB can do about that. However, trying to remove the x bit does not result in an error or a problem

Should have been more clear: only removes x-bits on _files_.

Safihre:
@jcfp what do you think about the changes?

jcfp:
> @jcfp what do you think about the changes?

I'll comment tomorrow, time permitting.

jcfp (commented):
I ignored the possibility that users would set something weird for the permissions (for which they'd receive a warning anyways), particularly anything that has a value of zero when converted to int. That kind of setting could throw off certain checks though, with int(zero) evaluating to False in statements checking if custom permissions have been set.

jcfp (approved):

Issue #2006, opened 2021-12-26
After update to latest commit, sabnzbd won't run


anonvoy:
I'm running sabnzbd from source, python 3.10 on OpenSUSE Leap 15.2. After updating to commit a848283d5, sabnzbd will start, but encounters a problem and won't run. Returning to commit b860741ab solves the problem instantly. I guess the actual problem is with commit 9d5c7d33172d435d5cd2ccf22897c574809c63e1 (I had the same problem yesterday, when this was the latest commit except for "translatable texts", but just went back to the previous commit because I didn't have time to open an issue here).

sabnzbd.error.log says:

```
Traceback (most recent call last):
  File "/home/abc/.sabnzbd/SABnzbd.py", line 1787, in <module>
    main()
  File "/home/abc/.sabnzbd/SABnzbd.py", line 1285, in main
    print_modules()
  File "/home/abc/.sabnzbd/SABnzbd.py", line 480, in print_modules
    seven_command_output = sabnzbd.misc.run_command(sabnzbd.newsunpack.SEVEN_COMMAND)
  File "/home/abc/.sabnzbd/sabnzbd/misc.py", line 1104, in run_command
    with build_and_run_command(cmd, **kwargs) as p:
  File "/home/abc/.sabnzbd/sabnzbd/misc.py", line 1065, in build_and_run_command
    command = ionice + command
TypeError: can only concatenate list (not "str") to list
```

sabnzbd.log says:
```
2021-12-26 17:55:05,991::INFO::[SABnzbd:1183] --------------------------------
2021-12-26 17:55:05,991::INFO::[SABnzbd:1184] SABnzbd.py-develop
2021-12-26 17:55:05,992::INFO::[misc:1097] [sabnzbd.misc.run_command] Running external command: ['git', 'rev-parse', '--short', 'HEAD']
2021-12-26 17:55:05,992::DEBUG::[misc:1098] Popen arguments: {'stdin': -1, 'stdout': -1, 'stderr': -2, 'startupinfo': None, 'creationflags': 0, 'cwd': '/home/abc/.sabnzbd'}
2021-12-26 17:55:05,995::INFO::[SABnzbd:1194] Commit = a848283d5
2021-12-26 17:55:05,995::INFO::[SABnzbd:1196] Full executable path = /home/abc/.sabnzbd/SABnzbd.py
[...]

PR #2007, opened 2021-12-26
bugfix with 7z if ionice is set: command must be a list

sanderjo:
solves https://github.com/sabnzbd/sabnzbd/issues/2006


Safihre:
Ah yes, the input is indeed specified as a `List[str]`. Too bad we don't use static type checking.

PR #2009, opened 2021-12-31
7z password detection relocation

sanderjo:
code more easier to understand. No functional changes

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -1048,6 +1048,7 @@ def seven_extract_core(sevenset, extensions, extraction_path, one_folder, delete
     sabnzbd.PostProcessor.external_process = p
     output = platform_btou(p.stdout.read())
     logging.debug("7za output: %s", output)
+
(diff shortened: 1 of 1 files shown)


sabnzbd/newsunpack.py
```diff
@@ -1081,6 +1070,28 @@ def seven_extract_core(sevenset, extensions, extraction_path, one_folder, delete
...
+            ret = 1
+            msg = T('ERROR: CRC failed in "%s"') % setname_from_path(sevenset)
+        else:
+            # Default message
+            msg = T("Could not unpack %s") % setname_from_path(sevenset)
+
+    # 7z cases we don't handle
+    if ret > 2:
+        # 7 = Bad command line parameters, 8 = Not enough memory for operation, 255 = User stopped the process
+        msg = "Unexpected 7z error. Check debug logging"
```
Safihre, line 1093: Let's include this with the code above to output 1 general message.
Maybe we already have "Check the logs" type message in the translations?

sanderjo: I now use two existing messages.

PR #2018, opened 2022-01-06
Show estimated time left on par2 repair and unrar unpack 

Safihre:
Based on the time spent repairing and unpacking, I figured we can make a rough prediction of the time it will take to finish.

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.main.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.main.js
@@ -267,11 +267,6 @@ function ViewModel() {
             Download timing and pausing
         ***/
         var timeString = response.queue.timeleft;
-        if(timeString === '') {
-            timeString = '0:00';
-        } else {
-            timeString = rewriteTime(response.queue.timeleft)
-        }
 
         // Paused main queue
         if(self.downloadsPaused()) {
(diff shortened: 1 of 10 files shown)


sanderjo:
image


Nice!!!

thezoggy:
is the format hh:mm? so in sander's image its 24 mins left?

Safihre:
All other times in the GUI are also mm:ss and only hh:mm:ss if mm>60. So it matches the rest and should not be confusing.
Additionally, you see it go down every second so that clears any confusion.. in a second 😅

thezoggy:
So time was h:mm:ss but you changed to mm:ss. This broke apps. So prob be good idea to revert it back

bakerboy448:
> So time was h:mm:ss but you changed to mm:ss. This broke apps (*adarr). So prob be good idea to revert it back

https://github.com/sabnzbd/sabnzbd/commit/1d3a922a7bb6e8643678dada3a6a595e371258bc#commitcomment-63089839

> Is this change really necessary? It means v3.5 is not compatible with Sonarr/Radarr/Readarr/Lidarr/Prowlarr

> ref https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Sabnzbd/JsonConverters/SabnzbdQueueTimeConverter.cs#L26

> This would require patching the aforementioned apps and a new master/main release and would break Sab for hundreds of thousands of users for many months if not years (...looking at Sonarr's delays between develop and main) until new masters/mains are released.

PR #2023, opened 2022-01-10
Backup and restore admin data

puzzledsab:
This is a very basic implementation that I could use some feedback for. It adds a backup and restore functionality to the bottom of the general config page. You can click the link to download a zip file and then upload it later to overwrite sabnzbd.ini and the admin dir.

Issues:
- Where should the code be put?
- How do I give feedback to the users? I haven't figured out how to show error messages from this page and I think the user should be sent to the main page after successfully uploading a backup.
- Security?
- Delete the old files in admin before overwriting?
- I'll add it to the wizard when the rest is figured out.


thezoggy:
Would be a very easy way for someone to get users sensitive data (like servers or email pw)..  I would think we might not want to allow this by default or add a way to lock down (config only option) so someone remote could not do something bad

puzzledsab:
It's only supposed to be possible to download the backup if you have access to the config pages and the configlock is not set, which I assume means that you must already have access to that data? I could strip the path before writing the files in the restore function just to be sure, but I think the zipfile library already makes sure it only returns safe relative paths.

sanderjo:
Nice feature.

Personally, I'm interested in copying newsserver (and maybe RSS) settings from one SAB instance to another SAB instance. And preferrably merging those settings into existing settings. Reason: I have a lot of _test_ SAB instances running
I guess that is not possible with this feature? 
Your feature is meant for all settings and configs, onto the same system?

image

image

thezoggy:
The backup should be like how *adarrs do it. Include the relevant .db so it's an actual snapshot not just the ini. Named with date and version so you have context

Safihre:
> The backup should be like how *adarrs do it. Include the relevant .db so it's an actual snapshot not just the ini. Named with date and version so you have context

See @sanderjo's screenshot, it includes the db.

jcfp:
> Would be a very easy way for someone to get users sensitive data (like servers or email pw)

I think this would be the main security consideration. Obviously any legit user knows his own passwords, but for anyone else out there (e.g. scriptkiddies using shoodan to locate sab installs) the export function provides a means of accessing credentials that thus far were out of reach even if one had access to the web interface and/or api.

puzzledsab:
If you have access to change the config then you can set up your own server, change the server info to point to it, do a test connect to log the username and password, and change it back. If the lock is set so that you can't change it then it can't be downloaded either (if it works like I intended).

@sanderjo: Yes, that's a bit out of scope. I would probably hack it with some Perl scripts.

jcfp:
> do a test connect to log the username and password

You're right, that works for anything with a test button.

Safihre:
Still not sure what to think about this.. The code is probably fine, but do we want this feature... Is it safe enough?
NZBGet and Sonarr indeed also support this..

interfaces/Config/templates/config_general.tmpl
```diff
@@ -372,6 +402,40 @@
...
 
+    // Upload config
+    \$('.uploadConfig').click(function(event){
+        removeObfuscation()
+        var theButton = \$(this)
+        var resultBox = theButton.parents('.col1').find('.result-box .alert');
+        theButton.attr("disabled", "disabled")
+        theButton.find('span').toggleClass('glyphicon-sort glyphicon-refresh spin-glyphicon')
+        formData = new FormData(document.querySelector('#uploadConfigForm'))
+        \$.ajax({
```
Safihre, line 413: Is this really needed? Why not use the regular `.fullform` triggers? Seems like a lot of code-duplication.
https://github.com/sabnzbd/sabnzbd/blob/develop/interfaces/Config/templates/staticcfg/js/script.js#L344

sabnzbd/interface.py
```diff
@@ -446,6 +450,28 @@ def shutdown(self, **kwargs):
         sabnzbd.shutdown_program()
         return T("SABnzbd shutdown finished")
 
+    @secured_expose(check_configlock=True)
+    def backup(self, **kwargs):
+        adminpath = cfg.admin_dir.get_path()
+        logging.debug("Backing up %s + %s", adminpath, config.CFG.filename)
+        adminfiles = [f for f in os.listdir(adminpath) if os.path.isfile(os.path.join(adminpath, f))]
```
Safihre, line 457: Why use a wildcard `os.listdir`? Can't we just create a new constant in `sabnzbd.constants` that includes all the files we want to backup? Since in `sabnzbd.constants` we have constants for each of the filenames.
So we don't include any random files.

sabnzbd/interface.py
```diff
@@ -446,6 +450,28 @@ def shutdown(self, **kwargs):
...
+
+        zip_buffer = io.BytesIO()
```
Safihre, line 459: Can we wrapped in a `with`.

sabnzbd/interface.py
```diff
@@ -1078,6 +1104,37 @@ def saveGeneral(self, **kwargs):
...
+        admin_backup = kwargs.get("admin_backup")
+
+        # Only accept the backup file if it can be opened as a zip archive and contains sabnzbd.ini
+        valid_backup = False
+        try:
+            # Read the zip file into RAM so that it can be closed and deleted before restart
+            backup_ref = io.BytesIO(admin_backup.file.read())
+            admin_backup.file.close()
+            admin_zip_ref = zipfile.ZipFile(backup_ref, "r")
+            for f in admin_zip_ref.infolist():
```
Safihre, line 1118: See other comment, if we use a constant with all the expected filenames, we can just loop over that.

sabnzbd/interface.py
```diff
@@ -1078,6 +1104,37 @@ def saveGeneral(self, **kwargs):
...
+                    valid_backup = True
+        except:
+            valid_backup = False
+
+        if valid_backup:
+            sabnzbd.RESTORE_DATA = admin_zip_ref
+            Thread(target=sabnzbd.trigger_restart, kwargs={"timeout": 1}).start()
+            return sabnzbd.api.report(data={"result": True, "message": "Success"})
+        else:
+            logging.warning("Invalid backup archive, bad file or missing sabnzbd.ini")
```
Safihre, line 1133: Why throw a `warning` if we return the error?

Safihre (changes requested):

interfaces/Config/templates/config_general.tmpl
puzzledsab: I hadn't noticed that script. I've never been able to follow JS and CSS. I started with the script for testing a server and changed it using cut and paste and a lot of fiddling to make it work with file upload.

I've changed it to use the existing system. However, now it won't be imported immediately if the user answers no to the restart request. It will be imported later if there is a restart but not if it's shut down. That might be unexpected to the user.

sabnzbd/interface.py
puzzledsab: I didn't know what it might contain. Is this the complete list?
```
ADMIN_FILES = (
    BYTES_FILE_NAME,
    QUEUE_FILE_NAME,
    POSTPROC_QUEUE_FILE_NAME,
    RSS_FILE_NAME,
    SCAN_FILE_NAME,
    RATING_FILE_NAME,
    DB_HISTORY_NAME,
)
```

puzzledsab:
I don't need this feature my self but there has been regular requests for it. I can see why it would be useful for people with less knowledge about the file system or if it's installed on a system which is only accessed through a web interface.

sabnzbd/interface.py
```diff
@@ -1107,33 +1111,34 @@ def saveGeneral(self, **kwargs):
...
+                        if f.filename == "sabnzbd.ini":
+                            valid_backup = True
         except:
             valid_backup = False
 
         if valid_backup:
-            sabnzbd.RESTORE_DATA = admin_zip_ref
-            Thread(target=sabnzbd.trigger_restart, kwargs={"timeout": 1}).start()
-            return sabnzbd.api.report(data={"result": True, "message": "Success"})
+            sabnzbd.RESTORE_DATA = admin_backup_data
```
puzzledsab, line 1138: I changed this to a copy of the file data because the backup_ref file was being closed, probably by the garbage collector.

Safihre:
Looks good to continue.
Indeed a button in the wizard that redirects to `config/general/#admin_backup` would be nice, they can upload their backup there. No need to build a whole upload form in the wizard.

puzzledsab:
Maybe we should make Sab store the password in encrypted format before adding this? It would make a copy of the backup less useful. On the other hand, the most valuable passwords are probably those for other sites.

Safihre:
> Maybe we should make Sab store the password in encrypted format before adding this? It would make a copy of the backup less useful. On the other hand, the most valuable passwords are probably those for other sites.

Other applications that offer a settings backup like Sonarr and Nzbget also have all authentication in plain text.. Not saying that's best, just that seems to be accepted.

Safihre:
I wonder if we really should backup all the files. Why for example back up the queue file? It's pointless without the actual incomplete file folders.
Don't we just need the Config, database and the totals file?

thezoggy:
Also you would need to reload once importing? Since the config/db may need migrations or to populate rss task

puzzledsab:
@Safihre: I kept the RSS and rating databases because I'm not entirely sure what they do but it seems like they may be useful.

 I tried backing up with items in the queue and it didn't work properly. I made a copy of the temp directory. paused, backed up, deleted the item from the temp directory, restored the temp directory and then imported the backup. The items would be in the queue but they would always start downloading from the beginning. I think it's better to not import the queue and instead use the orphaned jobs functionality.

@thezoggy: The files are restored during a restart, at the last moment after the server is shut down but before it's restarted. That way nothing should interfere and the files should be consistent when the server starts.

Safihre (approved):

Safihre:
I also pushed some code changes and noticed that if the recovered config has a different value for `admin_dir`, things would be recovered to the wrong place. That's fixed now.

This does need some unit-tests though, I know you are not a fan of that @puzzledsab.. But at least the creating and validating of zip's should be tested.

Safihre:
@puzzledsab do you want to work on the unit-tests or not at all? :)

sabnzbd/config.py
```diff
@@ -925,6 +927,66 @@ def save_config(force=False):
...
+                read_config(CFG.filename)
+
+                # Write the rest of the admin files that we want to recover
+                adminpath = sabnzbd.cfg.admin_dir.get_path()
+                for filename in CONFIG_BACKUP_FILES:
+                    try:
+                        zip_ref.getinfo(filename)
+                        destination_file = os.path.join(adminpath, filename)
+                        logging.debug("Writing backup of %s to %s", filename, destination_file)
+                        print(filename)
```
Safihre, line 977: Remove print.

tests/test_config.py
```diff
@@ -0,0 +1,61 @@
...
+from sabnzbd import config
+
+
+@pytest.mark.usefixtures("clean_cache_dir")
+class TestConfig:
+    @set_config({"admin_dir": os.path.join(SAB_CACHE_DIR, "test_config")})
+    def test_config(self):
+        FULL_INI_PATH = os.path.join(sabnzbd.cfg.admin_dir.get_path(), DEF_INI_FILE)
+        shutil.copyfile(os.path.join(SAB_DATA_DIR, "sabnzbd.basic.ini"), FULL_INI_PATH)
+        _, _ = config.read_config(FULL_INI_PATH)
```
Safihre, line 38: Don't need the _, _

Safihre (approved):
Looking good, just small cleanup and we can merge!

Safihre:
@puzzledsab Thanks!

thezoggy:
so when you go to restore:
- it prompt to pick any file. is it only meant to accept the .ini or the previous exported backup?
- can someone do bad things by just having it upload a random file?

when backing up the file:
- should we offer to add the same pw to the zip that the user used to login (if set)
- adopt a standard format like other apps do: `sabnzbd_backup_v<version of sab thats running>_<iso date>.zip`
- ex: `sabnzbd_backup_v3.5.3_2022-03-17.zip`

PR #2032, opened 2022-01-20
mod tests to exclude forced jobs from dupe check

jcfp:


Safihre:
Thanks for helping out!
I'm still confused: is my change working as expected? Why do you need to do the additional check for `prio_add` on top of the `expected`?

Safihre:
Thanks for helping me @jcfp, your unit-tests showed me that I just made the code badly.
It should skip the duplicate check if the job somehow ends up forced, whatever way it is.

Issue #2033, opened 2022-01-21
Better handle spontaneous / immediate "500 Access Denied" by newsservers like news.kpn.nl / newszilla.xs4all.nl


sanderjo:
When not on the KPN/Xs4all-network, trying to connect to news.kpn.nl / newszilla.xs4all.nl, SAB -> servertest now says `Could not determine connection result ()`

That's a bit unclear.

A manual NNTP/telnet connection shows that those newsservers immediatly say `500 Access Denied`

I think it would be nice if SAB would tell that, plus "before any login". I'll check if I can arrange that in https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/utils/servertests.py#L135-L144

```
sander@brixit:~$ telnet -4 newszilla.xs4all.nl nntp
Trying 94.232.116.172...
Connected to newszilla.xs4all.nl.
Escape character is '^]'.
500 Access Denied
Connection closed by foreign host.
```

```
sander@brixit:~$ telnet -6 newszilla.xs4all.nl nntp
Trying 2001:67c:174:101:0:65:ff02:172...
Connected to newszilla.xs4all.nl.
Escape character is '^]'.
500 Access Denied
Connection closed by foreign host.
```

Also: SAB keeps trying that server. 

```
2022-01-21 17:52:20,415::INFO::[downloader:662] 4@news.kpn.nl: Initiating connection
2022-01-21 17:52:20,415::DEBUG::[downloader:174] news.kpn.nl: Re-using address 94.232.116.171
2022-01-21 17:52:20,416::INFO::[downloader:662] 5@news.kpn.nl: Initiating connection
2022-01-21 17:52:20,416::DEBUG::[downloader:174] news.kpn.nl: Re-using address 94.232.116.171
2022-01-21 17:52:20,417::INFO::[downloader:662] 6@news.kpn.nl: Initiating connection
2022-01-21 17:52:20,418::DEBUG::[downloader:174] news.kpn.nl: Re-using address 94.232.116.171
2022-01-21 17:52:20,419::INFO::[downloader:662] 7@news.kpn.nl: Initiating connection
2022-01-21 17:52:20,419::DEBUG::[downloader:174] news.kpn.nl: Re-using address 94.232.116.171
2022-01-21 17:52:20,420::INFO::[downloader:662] 8@news.kpn.nl: Initiating connection
[...]

PR #2034, opened 2022-01-21
Handle 500 access denied

sanderjo:
WIP ... still hacky stuff: it works, but feedback needed

### DIFF
--- a/sabnzbd/utils/servertests.py
+++ b/sabnzbd/utils/servertests.py
@@ -94,17 +94,9 @@ def test_nntp_server(host, port, server=None, username=None, password=None, ssl=
         while not nw.connected:
             nw.clear_data()
             nw.recv_chunk(block=True)
-            #print("SJ nw.status_code: ", nw.status_code)
-            print("SJ nw.data: ", nw.data)
             if nw.status_code == 500:
                 # we got "500 Access Denied" instead of nice welcome message "200 Welcome to ..."
-                print("breakie breakie")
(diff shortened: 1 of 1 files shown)


jcfp:
I think any nntp status code indicating a hard no from the server at this point should trigger this, not just the `500` kpn decided to use for that purpose. The message displayed to the end user should make abunduntly clear it's server reply rather than an internal sab thing, e.g. `Server replied with "<woteva msg>" instead of a welcome message`.

Does sab handle this kind of server reply correctly in actual use? If not, that should be fixed too.

sanderjo:
> Does sab handle this kind of server reply correctly in actual use? If not, that should be fixed too.

No. It keeps on trying news.kpn.nl, with resulting download speed 0. I should have mentioned https://github.com/sabnzbd/sabnzbd/issues/2033 in this PR

puzzledsab:
How about if you instead of the current change do a break in the loop if this line returns an error code of >= 500
https://github.com/sabnzbd/sabnzbd/blob/56d0feab121a66fcd3c740f0f44cbcb6ae182b58/sabnzbd/utils/servertests.py#L96
and add a check for `nw.connected` at
https://github.com/sabnzbd/sabnzbd/blob/56d0feab121a66fcd3c740f0f44cbcb6ae182b58/sabnzbd/utils/servertests.py#L126

That way it should fall back to
https://github.com/sabnzbd/sabnzbd/blob/56d0feab121a66fcd3c740f0f44cbcb6ae182b58/sabnzbd/utils/servertests.py#L144

PR #2031, opened 2022-01-20
Add shortcuts

SKBotNL:
This pull request adds the following shortcuts:
A: add nzb
P: pause the queue
S: show the status popup
C: opens the config

A good idea may be to make a config entry to disable and edit the keybinds (which I can try to do later)

This also closes #1556

PR #2037, opened 2022-01-22
Close status popup when 's' is pressed for a second time

SKBotNL:
This is a small addition to #2031 that I forgot to push


Safihre:
That's not needed. All modals close when Esc key is pressed.

PR #2040, opened 2022-01-24
add appstream metadata, sample desktop and mimeinfo files

jcfp:
This change adds [appstream](https://www.freedesktop.org/software/appstream/docs/) metadata, desktop and mimeinfo files. Various app stores and package managers increasingly rely on such files, which means providing known-good versions with a number of optional fields filled in can benefit the project vs. someone else (distro packagers, etc.) writing a minimal/skeleton version from scratch (with possibly incorrect info).

Note:
- File locations match those in the existing systemd example.
- The appdata file can be verified using: `appstreamcli validate linux/org.sabnzbd.sabnzbd.appdata.xml` (from the `appstream` package on ubuntu).
- The metadata license in the appdata file is set to MIT in line with the considerations at [freedesktop.org](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-metadata_license).
- It's possible to list releases too, see [spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases).
- The sharedmimeinfo file defines a compressed nzb file format, extending the regular nzb filetype. Properly installed alongside the desktop file, this allows for setting sabnzbd as the default program in file managers and desktop environments for `nzb`, `nzb.gz`, and `nzb.bz2`. Regular archive formats such as zip, rar, 7z - while supported by the program - aren't listed to prevent sabnzbd from ending up as a suggested or default application for those.

### DIFF
--- a/linux/org.sabnzbd.sabnzbd.appdata.xml
+++ b/linux/org.sabnzbd.sabnzbd.appdata.xml
@@ -7,7 +7,7 @@
   <summary>Free and easy binary newsreader</summary>
   <description>
     <p>
-      SABnzbd+ is a free and Open Source web-based binary newsreader,
+      SABnzbd is a free and Open Source web-based binary newsreader,
       with support for the popular nzb file format. It greatly simplifies
       the process of downloading from Usenet, thanks to a friendly
       web-based user interface and advanced built-in post-processing
--- a/linux/org.sabnzbd.sabnzbd.appdata.xml
+++ b/linux/org.sabnzbd.sabnzbd.appdata.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2022 The SABnzbd-Team <team@sabnzbd.org> -->
+<component type="desktop-application">
+  <id>org.sabnzbd.sabnzbd</id>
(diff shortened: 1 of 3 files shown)


linux/org.sabnzbd.sabnzbd.appdata.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2022 The SABnzbd-Team <team@sabnzbd.org> -->
+<component type="desktop-application">
+  <id>org.sabnzbd.sabnzbd</id>
+  <metadata_license>MIT</metadata_license>
+  <name>SABnzbd</name>
+  <summary>Free and easy binary newsreader</summary>
+  <description>
+    <p>
+      SABnzbd+ is a free and Open Source web-based binary newsreader,
```
Safihre, line 10: Noo, not the `+`.

Safihre:
It would also be included in the `src` release, that's fine right?

jcfp:
> It would also be included in the `src` release, that's fine right?

Very much so. The files don't have any effect unless installed to specific locations where *nix systems expect them.

Issue #2038, opened 2022-01-23
keybinding: CTRL-C (Copy) jumps to Config. Plus: keybinding wishes


sanderjo:
Running https://github.com/sabnzbd/sabnzbd/commit/c1a8470

On the frontpage, when I select a text with my mouse ("History" for example), and type CTRL-C to copy ... SAB jumps to Config.

That is not what I want.

Wanted:
Plain C keypress: jump to Config
CTRL-C: do not catch that, do not jump to Config. Keep it a Copy command

Pinging @SKBotNL

SKBotNL:
Ill try to fix this right now

Safihre:
Fixed it. Leaving this open in case anyone finds more problems with the keyboard shortcuts.

SKBotNL:
> Fixed it. Leaving this open in case anyone finds more problems with the keyboard shortcuts.

Seems good

sanderjo:
> Leaving this open in case anyone finds more problems with the keyboard shortcuts.

May I also tell more wishes?

After pressing C (-> Config), I would to press a key to back to the main screen. "M"? "ESC"? "Backspace"?

Within Config, I would like to key-navigate to sub-options. Press G to go to General, F to Folders? Or press cursor-right (and cursor-left) to walk through the sub-options?

![image](https://user-images.githubusercontent.com/1273502/150682568-10947d84-4ae0-44ef-8b2f-ec70a48ff6f0.png)

SKBotNL:
@sanderjo 
Seems like an interesting thing to add, I can try to do it (unless Safihre does it).

Safihre:
@SKBotNL Go ahead :)

SKBotNL:
> > Leaving this open in case anyone finds more problems with the keyboard shortcuts.
> 
> May I also tell more wishes?
> 
> After pressing C (-> Config), I would to press a key to back to the main screen. "M"? "ESC"? "Backspace"?
> 
> Within Config, I would like to key-navigate to sub-options. Press G to go to General, F to Folders? Or press cursor-right (and cursor-left) to walk through the sub-options?
> 
> ![image](https://user-images.githubusercontent.com/1273502/150682568-10947d84-4ae0-44ef-8b2f-ec70a48ff6f0.png)

Or just using 'C' again or would that be inconvenient?

sanderjo:
> Or just using 'C' again or would that be inconvenient?

[...]

PR #2043, opened 2022-01-26
Add shortcuts to the config

SKBotNL:
What this adds:
- M: Go back to main menu
- S: Open the search popup
- Navigate through the tabs with the arrow keys

The code is probably not in the right place.


SKBotNL:
This adds the keybinding wishes from #2038

Safihre:
Now that I see what code is needed, I'm not sure if I'm still a fan.. Does it really add a lot of value to add these shortcuts @sanderjo?
Since so much in the Config is based on using your mouse..

sanderjo:
I use keyboard-shortcuts all the time in Gmail: c, tab, ctrl-enter, j k u.
I would love them in SABnzbd. Now I need a lot of mouse movement and mouse clicks to get to information. For example: Config -> General and Config -> Servers: tiny icons in upper right corner, then move mouse to the left to click in upper bar. Then move mouse to middle. I find that annoying.
My hover-over request, which solves the same problem via easier mouse movements, was rejected.

> Does it really add a lot of value 

For me: yes. 
For other users: I think keyboard-users are a minority

SKBotNL:
Although I also use shortcuts a lot I do agree with @Safihre that the config is mostly based on using your mouse

Safihre:
Will close this for now, as I am not sure I want to add it this way.

PR #2044, opened 2022-01-26
Full Apple M1 compatibility

Safihre:
@jcfp Maybe you can be of help?
@mreid-tt can you test the release? https://github.com/sabnzbd/sabnzbd/actions/runs/1751497555

After a whole day of hours feeling completly lost (again!) in the wonderous world of `clang`, `./configure` and `make` I finally managed to compile a multithreaded version of par2cmdline that works.
However, I feel I didn't do it properly because I had to include the `libomp.dylib` in the folder for it to work instead of statically linking it inside.. 
See: https://github.com/sabnzbd/sabnzbd/tree/feature/unrar-7zip-par2cmdline-update/osx/par2/arm64

I tried (a combination of) these things:
- `-lomp` in link-step
- `-static-openmp` in link-step
- Various combinations of these and `-L/xxx`
- Homebrew `libomp` package

I do have a cloud-M1 still running (it's only 3 euro/day), if you (@jcfp) maybe want to experiment? 
It can be accessed through VNC, so I can send you the details easily.


mreid-tt:
> @mreid-tt can you test the release? https://github.com/sabnzbd/sabnzbd/actions/runs/1751497555

Downloaded release and executed a test download successfully. File unpacked while downloading as usual. No errors encountered.

jcfp:
Macos and its compiler and other quircks aren't exactly familiar territory for me. That said, the filename of libomp.dylib suggests a shared library. What happens if you do a static build of openmp itself (setting LIBOMP_ENABLE_SHARED to FALSE), then linking the result of that (libomp.a or so?) into par2?

Safihre:
Yeah that seems to work, but feels like even more of a Frankenstein monster 👾😅 maybe the dylib isn't that bad.

jcfp:
> maybe the dylib isn't that bad.

As long as macos always looks for the dynamic lib in the location where sab provides it (and does so before looking elsewhere on the library path), supplying the dylib should work. Otherwise, the actual static binary may be preferable.

Either way, consider documenting how to build this thing somewhere on the repo while your memories are still fresh.

osx/7zip/License.txt
```diff
@@ -1,17 +1,21 @@
...
+  7-Zip
+  ~~~~~
   License for use and distribution
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  7-Zip Copyright (C) 1999-2016 Igor Pavlov.
+  7-Zip Copyright (C) 1999-2021 Igor Pavlov.
 
-  Licenses for files are:
+  The licenses for 7zz file are:
```
jcfp, line 8: typo? (7zz)

Safihre: No, for some reason the osx binary is called `7zz` now... No clue why. The license file is just copied from the official release.

Issue #2066, opened 2022-02-09
Slow devices: Wrench Status: no values on manual run (=refresh click)


sanderjo:
Strange thing, that seems to happen on slow devices (pystone 11.000 or lower)

After starting SAB, the Wrench -> Status does show IP addresses (but no values in lower screen):

![image](https://user-images.githubusercontent.com/1273502/153212041-61b42b82-c686-4a59-9559-da6e93d39d90.png)
That is expected: those values are already there in the startup sequence of SAB:
```
2022-02-09 14:34:12,448::DEBUG::[SABnzbd:1235] Local IPv4 address = 192.168.1.3
2022-02-09 14:34:12,725::DEBUG::[SABnzbd:1236] Public IPv4 address = x.x.162.65
2022-02-09 14:34:12,726::DEBUG::[SABnzbd:1237] IPv6 address = 2001:470:x.x::2
2022-02-09 14:34:12,984::DEBUG::[SABnzbd:1240] CPU Pystone available performance = 11802
```

But, then, clicking on Refresh results in ... nothing ... even after 30 minutes of waiting:
![image](https://user-images.githubusercontent.com/1273502/153212800-7d9ad960-6de8-44dd-b098-2b815869c0e0.png)

STDOUT shows
```
2022-02-09 14:38:22,175::DEBUG::[internetspeed:48] Basic measurement, with small download:
2022-02-09 14:38:22,177::DEBUG::[internetspeed:22] URL is https://sabnzbd.org/tests/internetspeed/5MB.bin
2022-02-09 14:38:23,408::DEBUG::[internetspeed:34] Downloaded bytes: 5242880
2022-02-09 14:38:23,415::DEBUG::[internetspeed:35] Duration in seconds: 1.229444
2022-02-09 14:38:23,419::DEBUG::[internetspeed:51] Speed in MB/s: 4.07
2022-02-09 14:38:23,420::DEBUG::[internetspeed:76] https://sabnzbd.org/tests/internetspeed/10MB.bin
2022-02-09 14:38:23,421::DEBUG::[internetspeed:22] URL is https://sabnzbd.org/tests/internetspeed/10MB.bin
2022-02-09 14:38:24,891::DEBUG::[internetspeed:34] Downloaded bytes: 10485760
2022-02-09 14:38:24,895::DEBUG::[internetspeed:35] Duration in seconds: 1.468383
[...]

PR #2071, opened 2022-02-11
Logging of performance measurements

sanderjo:
Related to https://github.com/sabnzbd/sabnzbd/issues/2066 ... more logging to see how long measurements take ...

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -193,7 +193,7 @@ class Assembler(Thread):
                             # Log traceback
                             logging.info("Traceback: ", exc_info=True)
                             if sabnzbd.WIN32:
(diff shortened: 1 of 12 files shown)


puzzledsab:
Haven't you already figured out that the reason why it fails is that it takes too long on slow machines? If so, is it really necessary to add this debugging for everyone? Sorry if I'm missing anything.

sanderjo:
> Haven't you already figured out that the reason why it fails is that it takes too long on slow machines? If so, is it really necessary to add this debugging for everyone? Sorry if I'm missing anything.

Yes, in an ideal world, no logging and no Wrench is needed at all.
In the real world, strange things happen, and logging helps in finding the cause of problems.
Indeed, thanks to logging, I found the cause of the Wrench not showing info. And I found the cause of another weird problem. And I expect that thanks to the logging, future problems are easier to analyze.

It's +DEBUG, so only if someone turns that on.

PR #2075, opened 2022-02-12
Do not fail, if attribute subject ist missing

wolfmanx:
There are some NZBs out there, where `file` elements do not have a `subject` attribute. However, they have a `poster` attribute, which can be used instead. The attached NZB was downloaded from nzbgeek.info. [sample-from-nzbgeek.info.nzb.gz](https://github.com/sabnzbd/sabnzbd/files/8053645/sample-from-nzbgeek.info.nzb.gz)

### DIFF
--- a/sabnzbd/nzbparser.py
+++ b/sabnzbd/nzbparser.py
@@ -400,12 +400,10 @@ def nzbfile_parser(full_nzb_path: str, nzo):
 
             # Parse the files
             if element.tag.lower() == "file":
-                # Get subject and date
-                file_name = ""
+                # Get subject and date, don't fail, if subject is missing
+                file_name = "No subject"
                 if element.attrib.get("subject"):
                     file_name = element.attrib.get("subject")
-                elif element.attrib.get("poster"):
-                    file_name = element.attrib.get("poster")
 
                 # Don't fail if no date present
                 try:
--- a/sabnzbd/nzbparser.py
+++ b/sabnzbd/nzbparser.py
@@ -404,6 +404,8 @@ def nzbfile_parser(full_nzb_path: str, nzo):
                 file_name = ""
                 if element.attrib.get("subject"):
                     file_name = element.attrib.get("subject")
+                elif element.attrib.get("poster"):
+                    file_name = element.attrib.get("poster")
 
                 # Don't fail if no date present
                 try:
--- a/sabnzbd/nzbparser.py
+++ b/sabnzbd/nzbparser.py
@@ -400,8 +400,9 @@ def nzbfile_parser(full_nzb_path: str, nzo):
 
             # Parse the files
             if element.tag.lower() == "file":
-                # Get subject and date, don't fail, if subject is missing
-                file_name = "No subject"
+                # Get subject and date
+                # Don't fail, if subject is missing
+                file_name = "unknown"
                 if element.attrib.get("subject"):
                     file_name = element.attrib.get("subject")
 


Safihre:
Why would they do that 😢

jcfp:
Unless I'm missing something, every `<file>` in the example nzb seems to have a subject?

thezoggy:
Also poster can be randomized per part.. so not sure how this well this would work for obfuscated posts as well

wolfmanx:
@jcfp You are right, that was the wrong example, here is a better one: [real-sample-from-nzbgeek.info.nzb.gz](https://github.com/sabnzbd/sabnzbd/files/8054378/real-sample-from-nzbgeek.info.nzb.gz)

wolfmanx:
nzbget segfaults on ubuntu 18.04 and 20.04. IMHO, that is not an appropriate reaction to faulty input.
When the `poster` attributes are renamed as `subject` attributes, both sabnzbd+ and nzbget download the files perfectly.

As I said before, it is really a hail mary stop gap measure, which either works with such NZBs or it simply fails the same way as if the `subject` attributes were faulty. If the hack works, it's a win. If it does not work, it's not worse than before.

If you do not plan to accept this hack, please, make the error mesage less misleading: "missing subject attribute for files" is a lot more helpful than "no files found", which is simply wrong. Even "nn invalid files skipped" gives a better hint.

jcfp:
@wolfmanx imho you should bring this to the attention of nzbgeek, and get them to stop handing out these broken nzb files.

puzzledsab:
How about replacing the line `file_name = ""` with `file_name = "No subject"`? The nzb is missing the date attribute as well, so it seems like it's just a coincidence that the poster is still there. Next time that one might be missing as well so depending on it being there would could also break.

Safihre:
Yeah indeed.. It would be extremely strange to just randomly shuffle data from one field to the other.
If anything, indeed it can be replaced by a random string. And then we can try to fix it with our anti-obfuscation techniques. If the post has par2, we can fix it.

wolfmanx:
@jcfp First, it's not just nzbgeek. But second I really do not want to notify them. Being a programmer myself, I imagine the solution will probably be to filter out such NZBs instead of actually fixing them. That means, I will no longer get them at all. While this "solves" the problem for nzbgeek and sabnzbd+, it does not solve my problem of getting the data. I still prefer getting faulty NZBs, which are easily repairable:
```sh
sed -i.faulty '/<file.*subject="/n;/<file/s,poster=",subject",' some.nzb
```
@puzzledsab Why do you assume the poster "is still there"? A poster attribute, which is a *correct* subject attribute, is for all intends and purposes an *incorrect* poster attribute. So the poster was never there to begin with. Neither was the subject. The subject information was just put in the wrong attribute.
Since the poster attribute is also the *only* attribute, there is nothing random about using it. A case could be made, that if a `file` element has no `subject` attribute but only a singular `foo` attribute, that this attribute should be examined whether it is suitable to be used as a `subject`.
When you look at the code, you will see, that nothing happens, when the poster attribute is not there, so nothing breaks, if there is no poster attribute. ALso, nothing breaks any differently than it does with any other mangled subject.

@Safihre When considering the obfuscation technique:
```
put the valid subject information into the poster attribute
do not add any other attributes
```
one possible anti-obfuscation technique is:
```
if subject is missing:
  if poster is present:
    if poster conforms to requirements for a valid subject:
      use poster with warning
    else:
      use poster as random string
  else:
      use random string
```
So, yes, I heartily agree to deploy any further anti-obfuscation techniques, with the contents of an existing `poster` attribute as "random string" ;-) If the "random" `poster` information happens to be a correct `subject`, the download will just magically succeed without the need for any other de-obfuscation.

wolfmanx:
So, I checked the case, where the `file` element only has a `subject` attribute with a fixed string (identical for all files).

Assuming, there is no cached information about previous downloads:
1. It works great, even downloading with the correct filenames, when there is par2 information.
2. I also removed the par2 information. The download still worked with the correct file names.

So it seems, that the file_name derived from the `subject` attribute is not really necessary at all.
It also means, that a non-random fixed string is good enough as @puzzledsab suggested.
It further means, that the current behaviour to fail, if there is no `subject` attribute, is definitely sub-optimal and should be fixed.
And last, but not least, it is not necessary to notify anybody.

I have modified the pull-request accordingly.

thezoggy:
Looking at that sample nzb you shared from nzbgeek and the actual scene nzb.. they are a bit drastically different (scene) 368 KB vs (nzbgeek) 652 KB

looks like this was someones re-post where nzbgeek assembled the crosspost, not certain if JBinDown (posting app) is doing this by default or its just nzbgeek doing it on processing. woulndt be to hard to tell by looking at other nzbs... but anyways this is prob done by design to avoid dmca and stop public raw sites from indexing..    

but anyways, i assume sab would handle the job just fine as its getting the nzb job name (release name) from the nzb + indexer.

Safihre:
Made a small textual change to match what we do in `santize_filename` if it's bad input.

PR #2135, opened 2022-03-25
User defined extensions 2022 03 25

sanderjo:
... that should not get an extension added by deobfuscate

Should solve https://forums.sabnzbd.org/viewtopic.php?t=25792

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -418,7 +418,6 @@ socks5_proxy_url = OptionStr("misc", "socks5_proxy_url")
 user_defined_well_known_extensions = OptionList("misc", "user_defined_well_known_extensions", ["abc", "ppp", "xxx"])
 
 
-
 ##############################################################################
 # Config - Notifications
 ##############################################################################
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -87,12 +87,14 @@ def all_lowercase(value):
         return None, [item.lower() for item in value]
     return None, value.lower()
 
+
 def lower_case_extensions_without_dot(value):
-    """ generate lower case extension(s), without dot"""
+    """generate lower case extension(s), without dot"""
     if isinstance(value, list):
         # If list, for each item
         return None, [item.lower().split(".")[-1] for item in value]
     return None, value.lower().split(".")[-1]
+
 
 def validate_no_unc(root, value, default):
     """Check if path isn't a UNC path"""
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -418,6 +418,7 @@ socks5_proxy_url = OptionStr("misc", "socks5_proxy_url")
 user_defined_well_known_extensions = OptionList("misc", "user_defined_well_known_extensions", ["abc", "ppp", "xxx"])
 
 
+
 ##############################################################################
 # Config - Notifications
 ##############################################################################
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -415,6 +415,7 @@ max_url_retries = OptionNumber("misc", "max_url_retries", 10, 1)
 downloader_sleep_time = OptionNumber("misc", "downloader_sleep_time", 10, 0)
 ssdp_broadcast_interval = OptionNumber("misc", "ssdp_broadcast_interval", 15, 1, 600)
 socks5_proxy_url = OptionStr("misc", "socks5_proxy_url")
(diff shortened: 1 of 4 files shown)


Safihre:
Sounds like something that's a good addition, but maybe more as a Special setting?

Safihre:
Oops. Yes 👍

sanderjo:
Reference NZB: https://raw.githubusercontent.com/sanderjo/NZBs/master/reftestnzb_4f00cdd7d42f_with_xxx_extension.nzb

text file lorem_ipsum.xxx ... to test user-defined-extensions

sanderjo:
@Safihre now with unit-test! And thus ready for review.

sabnzbd/cfg.py
```diff
@@ -415,6 +415,7 @@ def validate_notempty(root, value, default):
 downloader_sleep_time = OptionNumber("misc", "downloader_sleep_time", 10, 0)
 ssdp_broadcast_interval = OptionNumber("misc", "ssdp_broadcast_interval", 15, 1, 600)
 socks5_proxy_url = OptionStr("misc", "socks5_proxy_url")
+user_defined_well_known_extensions = OptionList("misc", "user_defined_well_known_extensions", ["myext", "specialext"])
```
Safihre, line 418: There must be a shorter version 😂
To keep it similar to the other option above, what about: `ext_rename_skip`?
And why these default values? Just leave it empty and update the Wiki to have examples.

sabnzbd/utils/file_extension.py
```diff
@@ -255,14 +256,20 @@
 
 # Combine to one tuple, with unique entries:
 ALL_EXT = tuple(set(POPULAR_EXT + DOWNLOAD_EXT))
-# Prepend a dot to each extension, because we work with a leading dot in extensions
+# ... and prepend a dot to each extension, because we work with a leading dot in extensions
 ALL_EXT = tuple(["." + i for i in ALL_EXT])
 
 
+def all_extensions() -> tuple:
```
Safihre, line 263: Should be `Tuple[str,...]`

sabnzbd/utils/file_extension.py
```diff
@@ -293,7 +300,7 @@ def what_is_most_likely_extension(file_path: str) -> str:
 
     for possible_extension in all_possible_extensions(file_path):
         # let's see if technically-suggested extension by puremagic is also likely IRL
-        if possible_extension in ALL_EXT:
+        if possible_extension in all_extensions():
```
Safihre, line 303: The new function is now called again and again for every iteration of the for loop.
Maybe assign it to a variable before the loop?

Safihre (changes requested):

jcfp:
Do things still work as expected in case a user sets an extension using caps or with the dot included? IIRC the unwanted extensions code handles that stuff automatically; if this special option is to forgo such niceties, do make sure to clearly define the expected input format.

sabnzbd/utils/file_extension.py
sanderjo: please check if format is now correct

sanderjo:
> Do things still work as expected in case a user sets an extension using caps or with the dot included?

... now with validation=lower_case_extensions_without_dot. Seems to work nice: if you type "BLABLA.XXX" in the GUI and click Save, it is nicely rewritten to "xxx" in the GUI and settings.

Safihre:
Nice. We can actually use your new validation also for the Cleanup list and the Unwanted extensions.

sabnzbd/utils/file_extension.py
```diff
@@ -255,14 +256,20 @@
 
 # Combine to one tuple, with unique entries:
 ALL_EXT = tuple(set(POPULAR_EXT + DOWNLOAD_EXT))
-# Prepend a dot to each extension, because we work with a leading dot in extensions
+# ... and prepend a dot to each extension, because we work with a leading dot in extensions
 ALL_EXT = tuple(["." + i for i in ALL_EXT])
 
 
+def all_extensions() -> Tuple[List[str]]:
```
Safihre, line 263: No, that's not what I said ;)
`-> Tuple[str, ...]`
See https://stackoverflow.com/a/54747327/5235502

Safihre: Your PyCharm must have also indicated this, it will highlight the `return` in slight yellow highlight to indicate it doesn't match what you specified.

sanderjo: Oh, "-> Tuple[str, ...]" litterally? OK.

Safihre (approved):

PR #2166, opened 2022-04-30
pre-sanitize filename inside rar file for usage FAT-like filesystems

sanderjo:
Solves https://github.com/sabnzbd/sabnzbd/issues/2165#issuecomment-1113250148

So: sanitize filename inside rar file before unrarring, so that that unrarring is safe on FAT-like systems.

Note: needs `rar` (not `unrar`), for the `rn` (rename) option

PR #2167, opened 2022-04-30
Check filesystem capabilty, especially for filenames with special characters

sanderjo:
to detect FAT and FAT-like filesystems 
solves https://github.com/sabnzbd/sabnzbd/pull/2166#issuecomment-1114006105

image

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -1247,7 +1247,7 @@ def save_compressed(folder: str, filename: str, data_fp: BinaryIO) -> str:
     return full_nzb_path
 
 
-def directory_is_writable(mydir, myfilename):
+def directory_is_writable_with_file(mydir, myfilename):
     filename = os.path.join(mydir, myfilename)
     if os.path.exists(filename):
         try:
@@ -1263,17 +1263,17 @@ def directory_is_writable(mydir, myfilename):
         return False
 
 
-def check_directory_writing_capability(dir):
-    """Checks if dir is writable at all, and writable with special chars. Returns Error, Message"""
-    SOME_WRITING_ERROR = True
-    if directory_is_writable(dir, "sab_test.txt"):
+def directory_is_writable(dir):
+    """Checks if dir is writable at all, and writable with special chars. Returns True if all OK, otherwise False"""
+    if directory_is_writable_with_file(dir, "sab_test.txt"):
         if not sabnzbd.WIN32 and not sabnzbd.cfg.sanitize_safe():
-            if not directory_is_writable(dir, "sab_test \\ bla :: , bla.txt"):
-                return (
-                    SOME_WRITING_ERROR,
-                    T("%s is not writeable with special character filenames. This can cause problems.") % dir,
+            if not directory_is_writable_with_file(dir, "sab_test \\ bla :: , bla.txt"):
+                sabnzbd.misc.helpful_warning(
+                    T("%s is not writeable with special character filenames. This can cause problems.") % dir
                 )
+                return False
     else:
-        return SOME_WRITING_ERROR, T("%s is not writable at all. This blocks downloads.") % dir
-    # no problems detected
-    return not SOME_WRITING_ERROR, None
+        sabnzbd.misc.helpful_warning(T("%s is not writable at all. This blocks downloads.") % dir)
+        return False
+    # no problems detected, no directory is writable:
+    return Return
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -1263,7 +1263,7 @@ def directory_is_writable_with_file(mydir, myfilename):
         return False
 
 
-def directory_is_writable(dir):
+def directory_is_writable(dir: str) -> bool:
     """Checks if dir is writable at all, and writable with special chars. Returns True if all OK, otherwise False"""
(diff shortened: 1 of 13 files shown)


sabnzbd/utils/checkdir.py
```diff
@@ -103,6 +103,30 @@ def isFAT(check_dir):
...
+        with open(filename, "w") as f:
+            f.write("Some random content")
+        os.remove(filename)
+        return True
+    except:
+        return False
+
+
+def directory_is_writable_basic(mydir):
+    return directory_is_writable(mydir, "sab_test.txt")
```
Safihre, line 123: Do we really need a basic check? Don't we just want to check the expected "regular" case?

sabnzbd/utils/checkdir.py
```diff
@@ -103,6 +103,30 @@ def isFAT(check_dir):
...
+
+
+def directory_is_writable_special_chars(mydir):
+    return directory_is_writable(mydir, "sab_test \\ bla :: , bla.txt")
```
Safihre, line 127: But, this will always fail on Windows... 
image
See also `sanitize_filename` and `CH_LEGAL_WIN`.

Safihre (changes requested):

sabnzbd/utils/checkdir.py
sanderjo: Ah, yes. I had tested from Linux (plain Ubuntu and WSL-on-Windows) onto FAT and onto NTFS. From both Linux and WSL to NTFS, I can write those chars (although resulting in strange characters on the filename when viewed from Windows, I now see).

Let me think about that. Maybe detect it, and force-turn on Sanitize_Safe. Or exclude that test if on Windows, because then we already know the outcome and because SAB will already take care of it.

sanderjo: PS: "interesting" that CI-test of this PR run OK ... also for the Windows-instance. So in that Windows-test, my PR could write "blabla \\ bla : " to disk?

Safihre: It only throws a warning and doesn't block anything, and as you can see, also on CI it's thrown:

https://github.com/sabnzbd/sabnzbd/runs/6245689918?check_suite_focus=true#step:7:145

sanderjo: Ah, yes, so `2022-05-01 08:32:42,389::INFO::[notifier:123] Sending notification: Warning - \\?\D:\a\sabnzbd\sabnzbd\tests\cache\Downloads\incomplete is not writeable with special character filenames. This can cause problems. (type=warning, job_cat=None)`

I'll think about a solution on Windows.

Safihre: Seems we can just skip it, or at least this check, on Windows.

sabnzbd/postproc.py
```diff
@@ -230,6 +230,18 @@ def run(self):
         else:
             logging.debug("Completed Download Folder %s is not on FAT", complete_dir)
 
+        def check_directory_writing_capability(dir):
```
Safihre, line 233: I don't like this function-inside-a-function (`run(self)`). Please put it with the the other functions, or in `sabnzbd.filesystem`.

sabnzbd/utils/checkdir.py
Safihre: Ok!

sabnzbd/utils/checkdir.py
sanderjo: now skipping if on windows, or safe_sanitize is on

sabnzbd/filesystem.py
```diff
@@ -1244,3 +1245,35 @@ def save_compressed(folder: str, filename: str, data_fp: BinaryIO) -> str:
...
+        return False
+
+
+def check_directory_writing_capability(dir):
+    """Checks if dir is writable at all, and writable with special chars. Returns Error, Message"""
+    SOME_WRITING_ERROR = True
+    if directory_is_writable(dir, "sab_test.txt"):
+        if not sabnzbd.WIN32 and not sabnzbd.cfg.sanitize_safe():
+            if not directory_is_writable(dir, "sab_test \\ bla :: , bla.txt"):
+                return (
```
Safihre, line 1272: Let's just use `True` and `False` as return variables.
You can call `helpful_warning` in `filesystem.py` by calling `sabnzbd.misc.helpful_warning(T('..'))`

Safihre (changes requested):

sabnzbd/filesystem.py
sanderjo: OK, but if we do the helpful_warning (and thus printing) in filesystem, is there still a return value needed? Maybe only for the unit_test?

sabnzbd/filesystem.py
```diff
@@ -1244,3 +1245,35 @@ def save_compressed(folder: str, filename: str, data_fp: BinaryIO) -> str:
...
+            f.write("Some random content")
+        os.remove(filename)
+        return True
+    except:
+        return False
+
+
+def check_directory_writing_capability(dir):
+    """Checks if dir is writable at all, and writable with special chars. Returns Error, Message"""
+    SOME_WRITING_ERROR = True
```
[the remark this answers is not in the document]
sanderjo: removed

Safihre (approved):

PR #2175, opened 2022-05-11
add bash completion

jcfp:
This change adds an autocomplete file for SABnzbd.py. It only actually does something if installed to a location where bash expects it (intended use), or when the user manually sources the file into bash; please do include in releases just like the desktop and service files.

To test, run a bash shell, source the autocomplete file, then use one of the recognized forms of the sabnzbd command and hit the TAB key one or more times:
```
source linux/sabnzbd.bash-autocomplete
SABnzbd.py <TAB>
SABnzbd.py -<TAB>
SABnzbd.py --<TAB>
```
Command-line options are automatically suggested and completed, argument values suggested, and directories and supported file types autocompleted for loading nzb files from the command line.

It deals with every option that takes arguments manually rather than parsing the program's --help output, as the latter is really slow for anything using interpreted languages (noticable lag when hitting TAB to start autocompletion).

For the --server option, only the possible argument formats are indicated rather than trying to hint actual values such as hostnames and port. The multitude of possible formats makes an autocomplete function for actual hostnames, ipv4 and ipv6 addresses and ports too complex to be worthwhile.

### DIFF
--- a/linux/sabnzbd.bash-completion
+++ b/linux/sabnzbd.bash-completion
@@ -47,6 +47,7 @@ _sabnzbd()
         --server|-!(-*)[s])
             # suggest possible formats
             COMPREPLY=( $(compgen -W 'hostname :port hostname:port ipv4 ipv4:port [ipv6] [ipv6]:port' -- "$cur") )
+            return
             ;;
     esac
 
--- a/linux/sabnzbd.bash-completion
+++ b/linux/sabnzbd.bash-completion
@@ -1,0 +1,62 @@
+# bash completion for sabnzbd
+
+_sabnzbd()
+{
+    local cur prev words cword split
+    # list all options here
+    local all_options="-f --config-file --pidfile -t --templates --pid -l --logging -w --weblogging -d --daemon -h
+                       --help -v --version -c --clean -p --pause --repair --repair-all --no-login --log-all --console
+                       --disable-file-log --new -b --browser --ipv6_hosting --inet_exposure --https -s --server"
+    _init_completion -s || return
+
+    # handle options that take arguments
+    case $prev in
+        # 0..1
+        --browser|--ipv6_hosting|-!(-*)[b])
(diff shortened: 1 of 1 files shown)


Safihre:
So this only works if the user manually runs the `source` command?

The file should be included automatically in the source distribution: https://github.com/sabnzbd/sabnzbd/blob/b42e600285b663d69487372163f364775dbaa775/builder/package.py#L437

But you can check that in the build output.

jcfp:
> * how/when are directories useful as input for SAB?

The program takes one or more (collections of) nzb files to load as command line arguments. The autocomplete for that is limited to directories and a list of accepted extensions.

> So this only works if the user manually runs the `source` command?

That's the simple way to load this for testing purposes. In typical use, the file would be installed in locations where bash automatically finds it, system-wide by a package in /usr/share/bash-completion/completions/ or in a local directory such as ~/.local/share/bash-completion/completetions by an individual user.

sanderjo:
> system-wide by a package in /usr/share/bash-completion/completions/

Ah, that is how it works! Cool.

On my Ubuntu, /usr/share/bash-completion/completions/ contains 900+ entries. For example for curl and wget.

And thus soon for sabnzbdplus and SABnzbd.py.

jcfp:
is it me or does the file in the rc1 -src.tar.gz have windows line endings all of a sudden?

Safihre:
It seems we only "correct" line endings for a specific set of file extensions:
https://github.com/sabnzbd/sabnzbd/blob/develop/builder/package.py#L448-L453

Safihre:
@jcfp I moved the source building to the macOS job, now the line endings are fine.
https://github.com/sabnzbd/sabnzbd/commit/1894ed72e38aae5d4d870e58f879fd8e831e6a9a
I could have made it's own job, but that seems like a waste of resources (installing deps etc).

Issue #2053, opened 2022-02-02
Downloads stall with unique characters in password


kage1414:
This is a random issue I came across and THINK I found the source of. I am not a python dev and haven't worked on the project at all, but I figure you all would appreciate knowing about this bug

I recently changed all the passwords of my self-hosted programs to random 48-character strings that included special characters. After doing this, downloads halted on sab and couldn't make a connection to the server. I notice 'if len(dest_pair) != 2 or dest_pair[0].startswith("[")' in the log file pointing to an open bracket somewhere (I'm not sure where this came from because the new password didn't include this character). After changing the password back and removing special characters from the new password, sabnzbd was acting normal and downloading again.

Snippet of sabnzbd.log:

`2022-02-02 10:42:04,902::INFO::[downloader:662] 1@news.newshosting.com: Initiating connection
2022-02-02 10:42:04,902::INFO::[downloader:662] 2@news.newshosting.com: Initiating connection
...
2022-02-02 10:42:04,968::INFO::[downloader:662] 100@news.newshosting.com: Initiating connection
2022-02-02 10:43:05,030::INFO::[downloader:942] Thread 1@news.newshosting.com: timed out
2022-02-02 10:43:05,030::INFO::[downloader:942] Thread 2@news.newshosting.com: timed out
2022-02-02 10:43:05,031::INFO::[downloader:942] Thread 3@news.newshosting.com: timed out
...
2022-02-02 10:43:05,049::INFO::[downloader:942] Thread 100@news.newshosting.com: timed out
2022-02-02 10:43:31,781::INFO::[happyeyeballs:97] Quickest IP address for news.newshosting.com (port 563, preferipv6 False) is None`

Snippet of sabnzbd.error.log:

`  File "/usr/lib/python3.10/site-packages/socks.py", line 738, in connect
AttributeError: 'NoneType' object has no attribute 'startswith'
  File "/usr/lib/python3.10/site-packages/socks.py", line 738, in connect
  File "/usr/lib/python3.10/site-packages/socks.py", line 47, in wrapper
  File "/usr/lib/sabnzbd/sabnzbd/newswrapper.py", line 332, in connect
  File "/usr/lib/python3.10/site-packages/socks.py", line 738, in connect
    if len(dest_pair) != 2 or dest_pair[0].startswith("["):
    if len(dest_pair) != 2 or dest_pair[0].startswith("["):
    return function(*args, **kwargs)
    self.sock.connect((self.host, self.nw.server.port))
[...]

PR #2177, opened 2022-05-12
Disable password manager autofill on socks5 field

20Koen02:
#2053 

Bitwarden autofills the socks5 field, even when advanced settings is unchecked.
I'm guessing a lot of us use bitwarden, so I'm proposing this small compromise.

If this is not the direction we want to go, maybe we should look into getting correct input sanitization and error handling for wrong/unreachable socks5 connection strings


thezoggy:
in addition, one could add `autocomplete="off"` to the input fields, but not all browsers respect that (and password mangers have options to ignore it as well)

Safihre:
If this works, sure!
We had to do a lot of weird things for the server username not to be auto filled, like hashing the field names. These browser extensions don't listen to anything.

PR #2186, opened 2022-05-22
Add sort by completedness

puzzledsab:
As discussed in #2176 

Using the existing sort functionality was simple, compatible with the existing sorting alternatives and seems to have minimal performance impact. I assume you don't want the 30 second resorting in SABnzbd.py but I couldn't remember or figure out where you put similar tasks.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1571,9 +1571,6 @@ def main():
             if not sabnzbd.check_all_tasks():
                 autorestarted = True
                 sabnzbd.TRIGGER_RESTART = True
-            # Sort by completed status
-            if sabnzbd.cfg.auto_sort() and sabnzbd.cfg.auto_sort().split()[0] == "completed":
-                sabnzbd.NzbQueue.sort_by_completed()
 
         # 3 sec polling tasks
         # Check for auto-restart request
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1571,6 +1571,9 @@ def main():
             if not sabnzbd.check_all_tasks():
                 autorestarted = True
                 sabnzbd.TRIGGER_RESTART = True
+            # Sort by completed status
+            if sabnzbd.cfg.auto_sort() and sabnzbd.cfg.auto_sort().split()[0] == "completed":
+                sabnzbd.NzbQueue.sort_by_completed()
 
         # 3 sec polling tasks
         # Check for auto-restart request
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -580,8 +580,13 @@ class NzbQueue:
         logging.info("Sorting by size... (reversed: %s)", reverse)
         self.__nzo_list = sort_queue_function(self.__nzo_list, _nzo_size_cmp, reverse)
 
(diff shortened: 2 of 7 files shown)


sabnzbd/skintext.py
```diff
@@ -862,6 +862,7 @@
     "Glitter-logText": TT(
         "All usernames, passwords and API-keys are automatically removed from the log and the included copy of your settings."
     ),
+    "Glitter-sortCompleted": TT("Sort by Completed <small>Most&rarr;Least</small>"),
```
Safihre, line 865: Don't you think `Sort by percentage remaining` or something like that is better?

SABnzbd.py
```diff
@@ -1571,6 +1571,9 @@ def main():
             if not sabnzbd.check_all_tasks():
                 autorestarted = True
                 sabnzbd.TRIGGER_RESTART = True
+            # Sort by completed status
+            if sabnzbd.cfg.auto_sort() and sabnzbd.cfg.auto_sort().split()[0] == "completed":
```
Safihre, line 1575: Why does this one get special treatment of being run often?
Shouldn't we then just always run this for _all_ auto-sorters?

Safihre (changes requested):

sabnzbd/skintext.py
thezoggy: Sort by Comp% ?

puzzledsab: Sort by % downloaded?

SABnzbd.py
Safihre: Let's move this to `NzbQueue` and call it from:
https://github.com/puzzledsab/sabnzbd/blob/00e1fb61aa8de9f9e0bbea7dd1de61d3bdc2bd1a/sabnzbd/__init__.py#L492

sabnzbd/nzbqueue.py
```diff
@@ -580,8 +580,13 @@ def sort_by_size(self, reverse: bool = False):
         logging.info("Sorting by size... (reversed: %s)", reverse)
         self.__nzo_list = sort_queue_function(self.__nzo_list, _nzo_size_cmp, reverse)
 
+    @NzbQueueLocker
+    def sort_by_completed(self, reverse: bool = True):
+        logging.debug("Sorting by completed...")
+        self.__nzo_list = sort_queue_function(self.__nzo_list, _nzo_completed_cmp, reverse)
```
Safihre, line 586: I also worry about it just randomly sorting jobs at the bottom that have 0% completion. Did you check that?
Otherwise we should only apply the sort on jobs with >0%.

puzzledsab: It shouldn't happen.

> The [sort()](https://docs.python.org/3/library/stdtypes.html#list.sort) method is guaranteed to be stable. A sort is stable if it guarantees not to change the relative order of elements that compare equal [...]

Safihre:
Great ⚡

thezoggy:
looks like this never got added to the test_functional_api.py / wiki to support sorting by 'remaining'
https://sabnzbd.org/wiki/configuration/3.7/api#sort

PR #2191, opened 2022-05-26
Remove redundant if in _nzo_completed_cmp

puzzledsab:
I'm not even sure why I added it anymore so it's probably bad considering that nzo.remaining is calculated. I don't think it's necessary to optimize it further. The queue will mostly be in the correct order which means each nzo.remaining is called only once or twice.


Safihre:
Maybe you can take a look at the actual sort function. It seems maybe it could also be smarter? Now it first creates seperate lists for each priority level. There must be a way without doing that using some smarter input to Python's sort.

PR #2192, opened 2022-05-26
Optimize sort_queue_function

puzzledsab:
Simpler and about twice as fast for all sort methods when I tested it by running sort 1000 times on a queue with about 100 items with various priorities.


Safihre:
Nice, very elegant. Just to be sure, you tested it?

puzzledsab:
I have tested the performance and tried to check that the sorting order is correct after sorting and doesn't randomly change in the case of sorting by % downloaded.  From what I understand the sort function is intentionally keeping the relative order so that this kind of sorting by multiple levels is possible. I'm not aware of anything else that could break.

PR #2193, opened 2022-05-27
Restart direct unpack after rename, category change or file removal

georgeto:


Safihre:
Thanks for the PR.
I always implemented Direct Unpack as a nice-to-have feature, but now for users it has become part of the standard.
So instead of allowing a restart, I think it's best if we implement something like:
`DirectUnpacker.unpack_dir_changed()` that re-evaluates `prepare_extraction_path` and checks if `tmp_workdir_complete` & `one_folder` (these are the only ones that we care about) have changed compared to `nzo.direct_unpacker.unpack_dir_info`.
So we can change `nzo.abort_direct_unpacker(self, only_if_dir_changed: bool = False)`, that in case of `only_if_dir_changed=True` checks if `unpack_dir_changed` and only aborts.
Or something like that.

albino1:
Commenting to remove the stale tag because I think it's a good idea even if the original PR implementation necessarily the best way to handle it.  One of the things that annoys me is when I edit a job name and direct unpack aborts and never starts up again.

Safihre:
@albino1 Agreed. Will implement if @georgeto doesn't respond anymore.

georgeto:
Sorry, I am hesitating because I think that the suggested alternative implementation does not work for the most prominent use case (at least for me), rename or category change that leads to a different extract location.
If I understand @albino1 post correctly, the annoyance he describes is "abort on rename", which would not be resolved by the alternative implementation.

The optimal solution would probably be to abort direct unpacking only if either the unpack base path or the final base path changes (might lie on a different disk/partition), but not in the case that only job name is changed (rename).
However, this could be complicated to implement, as the new name must be used for the final move.

Even with this optimal solution, an option to restart the direct unpacker automatically (or manually?) could be desirable.

albino1:
Since it's always direct unpacking to a temp directory anyway, I've never understood why it needs to abort on a rename or category change in the first place.  The name of the temp directory doesn't need to change, all that matters is the final output destination/folder/filename.

Safihre:
> Since it's always direct unpacking to a temp directory anyway, I've never understood why it needs to abort on a rename or category change in the first place. The name of the temp directory doesn't need to change, all that matters is the final output destination/folder/filename.

The reason is that currently in Post-Processing these directories are linked to each other, so if the job name is `ABC` it expects the temp unpack folder to be named `__UNPACK__ABC`. It would crash if for example the job was renamed to `EFG`, since it will look for `__UNPACK__EFG`.
That's something that could be changed of course.

Safihre:
Oke it turns out to be much more of a pain than I expected. Thought I could fix this in 1 hour, but requires quite a refactor of where and how we create the final directories.
It's a work in progress, for some time this summer.

albino1:
Commenting to keep alive.

PR #2194, opened 2022-05-28
Change variable names and values from completed to remaining for consistency and fix sort order

puzzledsab:


Safihre:
Now the variable names are right, but don't match the description (remaining vs downloaded). Maybe that's also confusing? 😅

puzzledsab:
Yes, I felt that the priority for the description should be to be user friendly, not programmer friendly. There are many more users than programmers. The GUI shows % downloaded, starting from 0, so the intuitive way to think about it for a user is that they want to have the most downloaded on top. If the text says "Sort by % remaining Least -> Most" then the problem is turned on its head which would probably be confusing to many users.

Safihre:
Ok! 👍

PR #2209, opened 2022-06-15
Update requirements sabyenc3

Joulinar:
Update requirements sabyenc3 to new version 5.4.3 as of following: 

- Issue: https://github.com/sabnzbd/sabyenc/issues/56
- Fixed with: https://github.com/sabnzbd/sabyenc/commit/a51b23f41d9a994fd7ce92aeb5e68ccd6d64d371


Safihre:
Also need to update this line: https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/constants.py#L53
That's why the CI tests fail.

Joulinar:
thx for the hint. Was actually already looking for it. But wasn't sure where it is located exactly.

PR #2235, opened 2022-07-03
tweak hotkeys, add pagination navigation, use jquery.hotkeys

### LINKED ISSUE #2229: keyboard shortcut - 3.6.0
shortcut for Azerty keyboard is "Q", not "A" for ADD

thezoggy:
Hopefully this fixes #2229 

vscode linter wanted to cleanup formatting for a few things on the .js, sorry for the whitespace

Rather than try and re-invent the wheel just used jquery.hotkeys, which allowed for simplification and better detection than what we had before. This fixes the fact that previously we were still firing hotkeys off on comboboxes, which I ran into during testing on various modals.

Added shortcuts to navigate queue/history pagination:
> shift+up    -- first page
> shift+down  -- last page
> shift+left  -- previous page
> shift+right -- next page

Limited hotkeys to shift+<x> to not try and mess with normal web browser navigation.

About the pagination bindings, they only apply to the active tab - which works nice on tabbed layout.
On non-tabbed layout, usually the 'active' tab is queue so the hotkeys only apply there rather than both tables at the same time.

I did notice while playing around with switching to tab/non-tab that sometimes the history would become the 'active' tab..
Could prob make things a bit more stable when switching to tab/non-tab by clearing active tab and default to queue?

Also, I found that we do not have any documentation of our hotkeys.. do we want to add it in app or link to wiki?
Just question because it means adding more clutter to the status modal, but also we normally dont do wiki pages for glitter specific stuff


thezoggy:
Both are same, just submitted to different branches (dev / 3.6.x). Didn't know which you'd want since dev had other changes like the category filtering

jcfp:
@thezoggy could you add a short comment with info on origin, license, and exact version to the minified hotkeys file?

thezoggy:
I used the min file from cdn which oddly doesn't have it. I'd have to grab from non min file but I'm traveling for next few days so it will be a few

thezoggy:
added license blurp from non min file

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -1186,71 +1186,77 @@ function ViewModel() {
...
+        $(document).bind('keydown', 'shift+p', function(e) {
+            self.pauseToggle();
+        });
+        $(document).bind('keydown', 'shift+a', function(e) {
+            $('#modal-add-nzb').modal('show');
+        });
+        $(document).bind('keydown', 'shift+c', function(e) {
+            window.location.href = './config/';
+        });
+        $(document).bind('keydown', 'shift+s', function(e) {
```
Safihre, line 1241: Why include the `shift`? The current version works without shift, so why change?

thezoggy: Current version for me does it on key or shift+key combo. So when adding the arrow keys for the pagination, it would break accessibility if we were using the standard arrow keys without using the shift modifier. So using shift for all keeps things consistent.

Safihre: I understand for the pagination that we should use the shift, but the current release uses non-shift for the P/A/S/C commands, so let's keep it that way?

thezoggy: The current way works with shift and without

Safihre: Only without shift for the existing shortcuts, with shift for the new pagination ones.

thezoggy:
ok updated, dropped shift for c/a/s/p and kept it for the pagination ones. also added logic to not only fire off a/s modals if a modal isnt shown. since you could previously make both show which then was a bit of a mess.

PR #2238, opened 2022-07-08
Update German translation

theRealBaccata:
Added missing percent character.


Safihre:
Thanks for the correction! But this is not how translations need to be changed, please see: https://sabnzbd.org/wiki/translate

Otherwise this change would just be reverted on the next update.

Issue #2225, opened 2022-06-26
Aggressive deobfuscation occasionally renames files that shouldn't be renamed


bambury:
I've noticed lately that in some cases, when downloading MP3 music scene releases, one or two (empirically it's either none, one or two, never more) MP3 files are renamed to the NZB name.

Today I took a look at the debug log and it looks like it's an aggressive (and unnecessary) deobfuscation attempt that causes the issue.

I've uploaded an extract of the debug log [here](https://0bin.net/paste/sk+GZxjF#CS6fs-iVQuqoN3JEjXh2iN7dGhhu7IpFBhqSAlHtiVS), the relevant bit starts at line 451.

Also, it's probably unrelated but there are two mentions of using the "tv sorter" for the download, even though it's categorized and music and eventually treated as such (lines 74 and 510).

The NZB file is [here](https://anonfiles.com/d8Gf9es0y6/Mark_Knopfler-Down_The_Road_Wherever-_Deluxe_Edition_-2018-RiBS_nzb).

Thanks for the attention, I'll be happy to do tests or provide more information if necessary.

sanderjo:
I see.
The code https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/deobfuscate_filenames.py#L220-L240 could use some improvement.

Are you running SABnzbd from source (on Linux)? Can you find the file sabnzbd/deobfuscate_filenames.py ?

bambury:
I am running SABnzbd 3.6.0 in a Docker container on LibreElec at the moment, but I can run a second instance from source on my Linux machine if it helps. Would you like me to make some changes to deobfuscate_filenames.py and see if the problem is fixed?

sanderjo:
I'm still thinking about a real solution, but in the meantime this should work for you:

https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/deobfuscate_filenames.py#L44

add ".mp3" somewhere in that line defining EXCLUDED_FILE_EXTS

bambury:
Got it, thanks!

bambury:
I just wanted to confirm that the workaround you suggested works. I took a quick look at the code you mentioned (lines 220-240) and I'm not sure I can see what's wrong there. Then again, I didn't sleep much and I'm pretty tired today :) Maybe it should iterate on the extcounter array after the for loop?

sanderjo:
New reference NZBs:

[...]

PR #2241, opened 2022-07-17
Only pick biggest file for deobfuscation

sanderjo:
to make Deobfuscation less aggressive, which was a problem for sets. 
Solves https://github.com/sabnzbd/sabnzbd/issues/2225

### DIFF
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -182,10 +182,10 @@ def big_small(filelist):
 
 
 def deobfuscate(nzo, filelist: List[str], usefulname: str):
-    """Check all files in filelist, and deobfuscate: rename to filename based on usefulname.
-    Methods
+    """For files in filelist:
     # 1. if no meaningful extension, add it
-    # 2. based on detecting obfuscated filenames"""
+    # 2. pick biggest file (and its lookalikes), and deobfuscate (if needed), to usefulname"""
+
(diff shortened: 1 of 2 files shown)


sabnzbd/deobfuscate_filenames.py
```diff
@@ -217,64 +230,47 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+        path, file = os.path.split(biggest_file)
+        # construct new_name: <path><usefulname><extension>
+        new_name = get_unique_filename("%s%s" % (os.path.join(path, usefulname), get_ext(biggest_file)))
+        logging.info("Deobfuscate renaming %s to %s", biggest_file, new_name)
+        renamer(biggest_file, new_name)
+        nr_files_renamed += 1
+        # find other files with the same basename in filelist, and rename them in the same way:
+        basedirfile, _ = os.path.splitext(biggest_file)  # something like "/home/this/myiso"
+        for otherfile in filelist:
+            if otherfile.startswith(basedirfile + ".") and os.path.isfile(otherfile):
```
puzzledsab, line 262: Wouldn't it be better to not require the dot? That way it could also rename for instance xxx-sample.ext.

sanderjo: Yes, great idea! I'll change it in the code, and I'll create a unit test for it.
Thanks.

PR #2243, opened 2022-07-19
set OSType in pyfakefs instead of separate properties

jcfp:
This change makes the tests use the recommended way of setting os-specific properties for the fake filesystem, and in doing so works around a regression in pyfakefs 4.6.2 (eliminating 10 out of the 30 related failures discussed in #2242). A fix for the rest of them is being implemented on the pyfakefs side as per https://github.com/jmcgeheeiv/pyfakefs/issues/692#issuecomment-1189388181.

Patch borrowed from https://github.com/mrbean-bremen/sabnzbd/commit/e10e924f21a0ad16c3364993c21f0b9ea4b32070

### DIFF
--- a/tests/test_filesystem.py
+++ b/tests/test_filesystem.py
@@ -27,6 +27,7 @@ from pathlib import Path
 import tempfile
 
 import pyfakefs.fake_filesystem_unittest as ffs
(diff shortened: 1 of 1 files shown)


Safihre:
Thanks! Also for the bug report at pyfakefs!

PR #2246, opened 2022-07-22
Deobfuscation: comments and logging

sanderjo:
Three examples in the comments

### DIFF
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -238,7 +238,7 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
     Case 1:
     
     bla.iso (1000MB, so much bigger than next biggest file)
-    bla-sample.iso (10MB)
+    bla-sample.iso (100MB)
     bla.txt (1MB)
     something.txt (1MB)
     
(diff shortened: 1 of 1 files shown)


sanderjo:
@Safihre @puzzledsab @jcfp Any feedback welcome!

* the comment block with three typical cases: does that make things clearer?
* I wanted better logging why a file was NOT to be deobfuscated (relative size, extension, or not obfuscated), which led to another code structure: more step-by-step, less indention. Better this way?

sabnzbd/deobfuscate_filenames.py
```diff
@@ -230,6 +230,41 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
         nzo.set_unpack_info("Deobfuscate", T("Deobfuscate corrected the extension of %d file(s)") % nr_ext_renamed)
         filelist = newlist
 
+    """
```
Safihre, line 233: Using triple quotes is usually placed directly under the `def` to explain the function. I would suggest to move this there to.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -240,40 +275,40 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
         biggest_file = filelist[0]
     else:
         biggest_file = None
-
+    if not (biggest_file and os.path.isfile(biggest_file)):
```
Safihre, line 278: Cleaner to use `if not biggest_file or not os.path.exists` instead of the single `not`.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -240,40 +275,40 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+        # no file found
```
Safihre, line 279: No logging for this case?

Safihre (changes requested):

sabnzbd/deobfuscate_filenames.py
sanderjo: It's a fault case. I should never happen. I'll put in an ERROR logging.

Safihre (approved):

PR #2255, opened 2022-07-30
servers - clear pw on host change

thezoggy:
reduce people using stale passwords on updating servers and make it harder to siphon pw. see discord dev for more info


Safihre:
With the discussion in Discord, I don't see the point in this. It won't stop any real attacker. Maybe we should apply #2261.

thezoggy:
Config backup is unrelated to what this is helping guard against.

PR #2263, opened 2022-08-08
Add option to replace underscores in folder name.

thezoggy:


thezoggy:
we prob should add a test set for the switches>naming options. 

reason this came about is that on some random sites there are some posters using `_` instead of the traditional `.` or ` ` in release names. Which then is a pita to do regex against when processing because of how `_` is treated..  So rather than force somone to use pre-queue script we can just add this option to go with the other folder rename options.

Safihre:
The wiki test requires the label to be the same as in the application. I think one uses "folder name" while the other one uses "foldername".

thezoggy:
> The wiki test requires the label to be the same as in the application. I think one uses "folder name" while the other one uses "foldername".

I noticed for the other options, the skintext would say "foldername" but in gui it shows "folder name"..  Do I need to re-gen the POT file or something after adding a skintext to make it actually reflect correctly?

Safihre:
It also took me a while to figure it out, but then I remembered that long ago you corrected the wrong spelling of "foldername"  to "folder name" through the `en.po` file 😃 That's why the GUI has a different value than the `skintext.py`.
So the label needs to be adjusted in `skintext.py` to match the label on the wiki.

thezoggy:
I see the change 7 years ago to en.po, did we just not update the skintext because we didnt want to invalidate the existing translations?

it makes a bit more sense now that ive stepped through it, the test_consistency checks the english translated text (en.po) to the wiki option and is case insensitive.

thezoggy:
tests failing are unrelated to this change. 
looks like all the tests that use `shutil.rmtree` fail on windows with python 3.10.x

in py 3.10 changelog:
> [bpo-42014](https://bugs.python.org/issue?@action=redirect&bpo=42014): The onerror callback from shutil.rmtree now receives correct function when os.open fails.

options:
- add code to handle OSError in our calls in 3.10+ since its normalized across os
- update our rmtree calls to just hide filesystem errors with `ignore_errors=True`
- adding a brief delay (1sec sleep) + retry loop (which only might fix some errors)

Safihre:
@thezoggy I think that had to do with the pyfakefs that needed a patch, on develop all tests succeeded.

PR #2268, opened 2022-08-14
Handle Unicoded filenames add via SAB GUI (cherrypy)

sanderjo:
in SABnzbd's GUI, if you upload a NZB file with unicode in it, the filename gets garbled in SABnzbd.
the cause is in cherrypy
this PR is a workaround so that you can upload unicoded nzb filenames via the SAB GUI.

image


sanderjo:
@Safihre ready for review

sabnzbd/encoding.py
```diff
@@ -77,6 +77,16 @@ def correct_unknown_encoding(str_or_bytes_in: AnyStr) -> str:
             return str_or_bytes_in.decode(chardet.detect(str_or_bytes_in)["encoding"])
 
 
+def correct_cherrypy_encoding(somestring: str) -> str:
+    """convert somestring with seperate, individual chars (1-255) to valid string (with UTF8 encoding)"""
+    try:
+        correctedstring = somestring.encode("raw_unicode_escape").decode("utf8")
```
Safihre, line 83: Let's just do `return` right here

sabnzbd/encoding.py
```diff
@@ -77,6 +77,16 @@ def correct_unknown_encoding(str_or_bytes_in: AnyStr) -> str:
...
+    except:
+        # not possible to convert to UTF8, so don't change anything:
+        correctedstring = somestring
```
Safihre, line 86: Same!

Safihre (changes requested):
Small comment. And maybe put a reference to the issue in the PR text.

PR #2282, opened 2022-08-27
Determine password from <meta> before running preprocessing script

### LINKED ISSUE #2281: pre-script: SAB_PASSWORD not set if password originates from inside the nzb (<meta>)
The pre-script documentation https://sabnzbd.org/wiki/scripts/pre-queue-scripts states that `SAB_PASSWORD` contains the password 
> [...] supplied **by the NZB** or the user

However, `SAB_PASSWORD` is only filled in a pre-script if the password originates from the file name, not if it originates from the `<meta type="password">` element inside the NZB.

m0vie:
This way the environment variable SAB_PASSWORD (whose documentation says is supplied by user OR the nzb) is filled properly and the password is available in a preprocessing script.

Fixes #2281.

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -841,6 +841,10 @@ class NzbObject(TryList):
         self.set_priority(priority)
         self.repair, self.unpack, self.delete = pp_to_opts(pp_tmp)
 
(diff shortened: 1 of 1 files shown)


Safihre:
Thanks for the fix!

PR #2285, opened 2022-09-03
remove some extensions that are non-mainstream

sanderjo:
Solves https://www.reddit.com/r/SABnzbd/comments/wu9o1y/downloaded_files_extensions_are_obfuscated/

### DIFF
--- a/sabnzbd/utils/file_extension.py
+++ b/sabnzbd/utils/file_extension.py
@@ -218,7 +218,6 @@ DOWNLOAD_EXT = (
     "mp3",
     "mp4",
     "mpls",
-    "mx",
     "nfo",
     "nib",
     "nzb",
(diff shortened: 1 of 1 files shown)


sabnzbd/utils/file_extension.py
```diff
@@ -227,7 +226,6 @@
     "part",
     "pdf",
     "pem",
-    "php",
```
Safihre, line 230: Php seems like one we shouldn't rename?

sanderjo: My reasoning: "PHP in a download is very unlikely. And if it's there, it's probably malevolent."

I checked my own drive, and there are 3 .php files. Two look certainly not php, one probably not. Maybe extension-renaming gave them the php extension. Reason more to remove .php from the list?

Issue #2320, opened 2022-10-29
Download hangs forever if server cannot be resolved


thespad:
If a server's hostname cannot be resolved then a download trying to check against it will hang forever. Disabling the server, pausing and restarting the download, restarting sab entirely, does not allow it to recover. Only removing the download entirely and re-importing (or the server in question coming back online) seems to allow it to resume.

![image](https://user-images.githubusercontent.com/8425502/198843471-8253308c-1b3a-47ef-85a4-1b0f9bd6d7cc.png)

I would have assumed that there was some kind of timeout in place for name resolution, similar to or using the timeout value attached to the server configuration, but it never seems to fail and just tries indefinitely.

Running 3.7.0 RC1 in docker.

Safihre:
There's indeed no timeout.
@sanderjo do you maybe know why here we don't use the functions from `getipaddress` that timeout after 3 seconds?
Seems pretty straightforward to use those also for getting newsserver info, wonder why we never did..

sanderjo:
AFAIK, and I might be completely wrong, but anyway:

> why here we don't use the functions from `getipaddress` that timeout after 3 seconds?

Would that solve it? 
My guess at the above problem is that SAB does not declare an unresolvable server as 'dirty' / 'ignore for 10 minutes', so that other newssrvers can take over.

A bit like filling out www.google.nl (resolvable, but no newsserver) as newsserver with best prio ... SAB will keep trying, and will bog down. See screenshot and logging.

SAB does have a mechanism "if failed login => ignore for 10 minutes", right? If so, use that for any misbehaving newsserver?

![image](https://user-images.githubusercontent.com/1273502/198847085-052d2e55-aac3-4596-aa9b-1b699a95c7e9.png)



```
2022-10-29 20:23:14,571::DEBUG::[downloader:173] www.google.nl: Re-using address None
2022-10-29 20:23:14,573::INFO::[downloader:670] 6@www.google.nl: Initiating connection
2022-10-29 20:23:14,575::DEBUG::[downloader:173] www.google.nl: Re-using address None
2022-10-29 20:23:14,575::INFO::[downloader:670] 7@www.google.nl: Initiating connection
2022-10-29 20:23:14,575::DEBUG::[downloader:173] www.google.nl: Re-using address None
2022-10-29 20:23:14,576::INFO::[downloader:670] 8@www.google.nl: Initiating connection
[...]

PR #2330, opened 2022-11-05
Do a more thorough check when a bad try_list is detected

puzzledsab:
First part of #2320.

If the last server in the priority list is deactivated and it was the only missing server in the article try_list then none of the active servers will check the article. This means that the article, file and nzb may never be registered as finished because that is only done after checking an article.

The bad try_list check will notice this because if there are 3 active servers and the last is deactivated then there will be 2 servers in the try_list and 2 active servers. If this is still the case after checking that all the servers in the try_list are currently active then the article will be marked as failed.

### DIFF
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -887,6 +887,10 @@ class NzbQueue:
         nr_servers = len(active_servers)
         empty = []
 
+        if nr_servers <= 0:
+            logging.debug("Skipping stop_idle_jobs because no servers are active")
+            return
+
         for nzo in self.__nzo_list:
             if not nzo.futuretype and not nzo.files and nzo.status not in (Status.PAUSED, Status.GRABBING):
                 logging.info("Found idle job %s", nzo.final_name)
@@ -902,12 +906,12 @@ class NzbQueue:
                     if len(nzf.try_list) >= nr_servers:
                         logging.info("Resetting bad trylist for file %s in job %s", nzf.filename, nzo.final_name)
                         nzf.reset_try_list()
-                        # Make sure there are no deactivated servers in the article try_list
+
+                        # Check for articles where all active servers have already been tried
                         for article in nzf.articles:
                             if nzf.deleted or nzo.removed_from_queue:
                                 break
-                            article.cleanup_try_list(active_servers)
-                            if len(article.try_list) >= nr_servers:
+                            if article.all_servers_in_try_list(active_servers):
                                 sabnzbd.NzbQueue.register_article(article, success=False)
                                 nzo.increase_bad_articles_counter("missing_articles")
                                 nzf.has_bad_articles = True
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -903,18 +903,16 @@ class NzbQueue:
                 for nzf in nzo.files:
                     if nzo.removed_from_queue:
                         break
-                    if len(nzf.try_list) >= nr_servers:
-                        logging.info("Resetting bad trylist for file %s in job %s", nzf.filename, nzo.final_name)
-                        nzf.reset_try_list()
 
+                    if len(nzf.try_list) >= nr_servers:
                         # Check for articles where all active servers have already been tried
-                        for article in nzf.articles:
-                            if nzf.deleted or nzo.removed_from_queue:
-                                break
+                        for article in nzf.articles[:]:
                             if article.all_servers_in_try_list(active_servers):
                                 sabnzbd.NzbQueue.register_article(article, success=False)
                                 nzo.increase_bad_articles_counter("missing_articles")
-                                nzf.has_bad_articles = True
+
+                        logging.info("Resetting bad trylist for file %s in job %s", nzf.filename, nzo.final_name)
+                        nzf.reset_try_list()
 
                 # Reset main trylist, minimal performance impact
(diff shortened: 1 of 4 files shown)


Safihre:
Looking good, I didn't think of this indeed!

Currently I'm less available due to (happy) family circumstances 🍼, will look at in depth in a few days.

puzzledsab:
Congratulations :)

I have changed the test so that it doesn't delete inactive servers. I was afraid that there might be cases where the same server would try to download the same article multiple times if it was disabled and then reenabled.

Regarding the DNS lookup issue, it turns out it was already supposed to disable the server for 10 minutes in `maybe_disable_server()`, but it didn't trigger because `get_server_addrinfo()` didn't return `False` like it expected. I fixed it by setting it to `False` explicitly because server.info can contain different false values (ie `None`) that should be treated differently. I included it in the same PR because it was such a small change.

sabnzbd/nzbqueue.py
```diff
@@ -896,11 +901,21 @@ def stop_idle_jobs(self):
             if len(nzo.try_list) >= nr_servers:
                 # Maybe the NZF's need a reset too?
                 for nzf in nzo.files:
+                    if nzo.removed_from_queue:
```
Safihre, line 904: Maybe we should wrap this whole section in `with NZBQUEUE_LOCK ` so we don't have to check for these things?
Because also below I wonder if we won't end up in a race condition with nzf.articles?
If one is removed during looping then it will result in a crash.

puzzledsab: I only added those checks for performance. Initially in my test it would go through all the files and articles in the nzo, even though the quick completion test would move the nzo out of the queue quickly because less than 100% including par2 was available.

My test consisted of an nzb where all the articles were missing on all the servers, and the lowest priority server had an invalid host name. I then disabled the invalid server when all the valid servers had tried all the articles. You have to comment out `self.info = False` in this PR to do the same test.

The locking problem I don't properly understand. Intuitively I would agree, but the more I dig into it, the stranger it seems. `sabnzbd.NzbQueue.register_article()` is called inside the article loop and did not cause problems in my tests. I assume it removed articles from the list while looping.

Also the downloader thread loops through the articles in an unlocked mode in `get_articles()` even though the decoder thread regularly deletes articles from `nzf.articles` through `NzbQueue.register_article()`. That doesn't seem to cause problems either. What am I missing?

I could change it so that the check is done before looping the articles in `nzf.get_articles()` in the downloader thread, using `nzf.articles[:]` to avoid deleting from the same list as it's looping. It would reduce some of the concurrency. It shouldn't cause a noticeable performance hit in normal circumstances if the servers parameter is changed so that it only includes active servers.

Safihre: Since this is only called in case an idle job is found, do we really need to worry about performance? I guess it doesn't hurt, the `if`'s just bothered me a tiny bit 😅

From what I can see, the only place `nzf.articles` has articles removed is using `nzf.remove_article`. Which is called from `nzo.remove_article` which is locked by `NZO_LOCK`. 
https://github.com/sabnzbd/sabnzbd/blob/be7ae3d151339fd379d1774f0a86b688887df9ff/sabnzbd/nzbstuff.py#L1190-L1191
So I think we either need to wrap this is `NZO_LOCK` or use `nzf.articles[:]`.

sabnzbd/nzbqueue.py
```diff
@@ -896,11 +901,21 @@ def stop_idle_jobs(self):
...
                         nzf.reset_try_list()
 
+                        # Check for articles where all active servers have already been tried
+                        for article in nzf.articles:
+                            if nzf.deleted or nzo.removed_from_queue:
+                                break
+                            if article.all_servers_in_try_list(active_servers):
+                                sabnzbd.NzbQueue.register_article(article, success=False)
+                                nzo.increase_bad_articles_counter("missing_articles")
+                                nzf.has_bad_articles = True
```
Safihre, line 917: I just removed this everywhere, was a leftover from the failed new QuickCheck.

sabnzbd/nzbqueue.py
puzzledsab: I removed one of the if blocks. I also realized that removing elements from the list I'm looping over creates problems because the element position is shifted. If element 0 is removed then element 1 becomes 0 and so on. The loop has already read element 0 so the original element 1 will be skipped. If trying to remove every element then this causes half of them to be skipped silently if [:] isn't used.

Regarding the lock, I was thinking about this part: https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/nzbstuff.py#L432

The remove_article methods are called from other threads than the downloader. They could theoretically remove an article from nzf.articles while the downloader thread is using the get_articles() code to loop over the article list. As far as I can tell no level of the get_articles system is locked so using the NZO_LOCK before removing doesn't fix it. I don't think it would print any errors or crash if it happens but it may skip one of the remaining articles in the list. Adding [:] would avoid it.

puzzledsab: Btw, `NzbQueue.register_article()` is also using `nzo.remove_article()`, so NZO_LOCK will be set while an article is removed.

Safihre:
It looks good to me. 
Do you think this fully resolves #2320?

puzzledsab:
Yes.

Safihre:
Thanks! 🎉

PR #2344, opened 2022-11-17
Refactor article.get_article

puzzledsab:
It's a bit shorter, probably faster and in my opinion easier to read. It should give the same results as the old one. The main difference in logic is that it takes advantage of the servers list being sorted by priority.

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -187,65 +187,45 @@ class Article(TryList):
     def get_article(self, server: Server, servers: List[Server]):
         """Return article when appropriate for specified server"""
         log = sabnzbd.LOG_ALL
-        if not self.fetcher and not self.server_in_try_list(server):
+
+        if self.fetcher or self.server_in_try_list(server):
             if log:
-                logging.debug("Article %s | Server: %s | in second if", self.article, server.host)
-                # Is the current selected server of the same priority as this article?
-                logging.debug(
-                    "Article %s | Server: %s | Article priority: %s", self.article, server.host, self.fetcher_priority
-                )
-                logging.debug(
-                    "Article %s | Server: %s | Server priority: %s", self.article, server.host, server.priority
-                )
-            if server.priority == self.fetcher_priority:
-                self.fetcher = server
-                self.tries += 1
+                logging.debug("Article %s | Server: %s | Returning None", self.article, server.host)
+            return None
+
+        if log:
+            logging.debug(
+                "Article %s | Server: %s | Article priority: %s", self.article, server.host, self.fetcher_priority
+            )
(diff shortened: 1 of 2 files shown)


Safihre:
Looks great!
Now what would be perfect is if we create unit-tests for this.
Should be easy since `Article` is an "easy" object and `Server` is easily `Mock`ed (I think).
Then we can also remove all those `if log:` since we have unittests to check ourselves.
Could you add the unittests or shall I do that?

puzzledsab:
Not sure if what I implemented is what you meant by mock, or if you meant the mock library. I couldn't figure out how to use it for this. Anyway, I think I'm out of ideas for tests to add.

sabnzbd/nzbstuff.py
```diff
@@ -187,65 +187,45 @@ def reset_try_list(self):
...
-                self.tries += 1
-                if log:
-                    logging.debug("Article %s | Server: %s | same priority, use it", self.article, server.host)
-                    logging.debug("Article %s | Server: %s | Article-try: %s", self.article, server.host, self.tries)
-                return self
-            else:
+                logging.debug("Article %s | Server: %s | Returning None", self.article, server.host)
+            return None
+
+        if log:
```
Safihre, line 196: Looks good. Let's get rid of all the `if log`'s. They make the code very hard to read.

tests/test_nzbstuff.py
```diff
@@ -50,6 +50,52 @@ def test_nzo_basic(self):
...
+        # Test fetching when there is a higher priority server available
+        server = servers[2]
+        assert article.fetcher_priority == 10
+        assert article.get_article(server, servers) == None
+        assert article.fetcher_priority == 20
+
+        # Server should be used even if article.fetcher_priority is a higher number than server.priority
+        article.fetcher_priority = 30
+        server = servers[1]
+        assert article.get_article(server, servers) == article
```
Safihre, line 88: I ran the unittest on the old code and then it fails at this line. So, what is changed? I don't fully understand to be honest.

Safihre: Ah I understand, but now I wonder why the old code didn't do this... Or is this a situation that shouldn't happen.

puzzledsab: I suspected that there might be a way to have the priority of all the remaining articles set such that after changing the priority of an untried server to a lower priority number than the articles, it would end up being skipped because the old version would skip it. 

I haven't been able to create it in a real world test so maybe there is something I've missed that avoids such situations. It's still a bug if it can happen because then the lower pri server would probably continue downloading the articles even though there would be a higher pri server available.

Safihre:
Thanks a lot!

Issue #2357, opened 2022-12-04
Download speed regression in 3.7.1RC1


Cpuroast:
Hi,

I noticed a pretty download big speed regression in 3.7.1RC1. 
This is probably related to the recent download optimizations, but I'm not sure which commit caused it.

Using 3.7.1RC1 using the 1 or 10GB download tests, speeds up goes slowly and tops out at ~34MB/sec, it will sometimes go up to ~40MB but come back down to ~32-34MB/sec

Using 3.7.0 using the 1 or 10GB download tests, speed goes up very quickly and tops out at ~53MB/sec which is the limit of my connection and stays there.

I can easily replicate it by switching back and forth between the 2 versions, so it's not random.

This is on macOS Big Sur, downloading to a SATA SSD.
CPU: I7-3720QM which is a 3rd gen mobile hyper-threaded quad-core CPU.
Maximum line speed: 53MB
Percentage of line speed: 0
Article Cache Limit: -1

Cpuroast:
I don't test individual git commits, only betas, RCs and stable. Probably could have caught it earlier had there been a beta instead of going straight to RC1.

Cpuroast:
anyways, somewhere between 3.7.0 and 3.7.1RC1 the speed regression occurred.

puzzledsab:
It's https://github.com/sabnzbd/sabnzbd/commit/557b9ef71d928857c42637774fe007ab7157e971 or the next one. I can't be sure because this one hangs.

Safihre:
@puzzledsab why? He doesn't have problems with failing to connecting, just with speed.
I expect it's one of the recent optimizations.. But which one..

sanderjo:
in SAB: Wrench symbol (upper right corner), what does the popup say at "Download speed limited by"

Mine:
Download speed limited by  CPU (95x) Disk speed (192x)

puzzledsab:
I don't know why for sure but it's when I check out https://github.com/sabnzbd/sabnzbd/commit/8597784bc64b3cbc935cbcfbbae4df1baea1fbf9 that the download speed first goes down. https://github.com/sabnzbd/sabnzbd/commit/c7791a478ae61d38703971f606e098d9ae788424 downloads at full speed. I think it may be downloading in blocking mode.

puzzledsab:
Replacing `self.sock.settimeout(None)` with `self.sock.setblocking(False)` in develop increases the speed to almost max speed but not completely so there may be other issues.

[...]

PR #2361, opened 2022-12-05
Bugfix/speed

puzzledsab:
@thezoggy : I have reverted the 5 commits I listed as possible in #2357 plus one that was a followup to one of those issues. Can you test this?

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -93,7 +93,7 @@ class Decoder:
     def stop(self):
         # Put multiple to stop all decoders
         for _ in self.decoder_workers:
-            self.decoder_queue.put((None, None, None))
+            self.decoder_queue.put((None, None))
 
     def join(self):
         # Wait for all decoders to finish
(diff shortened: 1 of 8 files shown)


puzzledsab:
I'm not sure if I understand what you mean. It looked like 3.7.0 was much faster than both this and the newest develop version from your screen shots in #2357? 81 MB/s for 3.7.0 and 62-65 for the ones that include the blocking fix?

thezoggy:
> I'm not sure if I understand what you mean. It looked like 3.7.0 was much faster than both this and the newest develop version from your screen shots in #2357? 81 MB/s for 3.7.0 and 62-65 for the ones that include the blocking fix?

3.7.0 - getting 80-90 MB/s
3.7.1rc1 - getting 50-70 MB/s (not stable)
3.8.0 dev (blocking fix) - start off spike to 80MB/s then hung in the 60-70MB/s
Bugfix/speed #2361 -- looks to be same as 3.8.0 dev, start off good then hover little lower. cpu seemed higher than 3.8.0 dev

all the builds have been the built windows installer binaries. just dirty install over existing instance.

PR #2362, opened 2022-12-05
Bugfix/speed python311

puzzledsab:
@thezoggy : Sorry, this may take some time. On this branch I intend to revert everything that is done in the opposite order. I now think it is either a dependency change or different Python versions that is the cause. The first try reverts everything back to the first time Python 3.11 was used. I'm not sure what's used in the good version.


thezoggy:
ok the built binary from this is "3.7.0Beta1" and uses py3.11

can see this one is good
sab

puzzledsab:
I've put back some stuff but I skipped the commits that broke blocking.

thezoggy:
ok updated to latest build:
firefox_2022-12-05_17-02-38

looks to be a little slower:
Taskmgr_2022-12-05_17-05-24

puzzledsab:
I think the cryptography update was actually included in 3.7.0. I've reverted the most recent dependency update, I'll do them one at a time now. It's unfortunate that you only got part of the speed back because then there must be at least one more problem in the other part of the code as well.

puzzledsab:
Was the version you were just testing and 3.7.0 using different ciphers?

thezoggy:
> Was the version you were just testing and 3.7.0 using different ciphers?

it was the previous revert that you did, which just had some translations/dependencies etc. when i looked at those 11 changes the only thing that looked like it could be anything was the cryptography change due to different openssl being built with..  You can see the same build with it using tls 1.3 vs manually setting the cert there was ~10MB/s difference and no cpu change. 

cleared server ssl cipher. shutdown, grabbed this newer build and installed. going to test and report back

thezoggy:
with this latest build:

firefox_2022-12-05_17-30-42

Taskmgr_2022-12-05_17-32-48

now setting main server to use `AES128-SHA`, slight slowdown
[sab-ssl2](http://zoggy.net/Taskmgr_2022-12-05_17-35-43.png)

so setting the ssl cipher prob doesnt really matter just variance showing up randomly ?

thezoggy:
> Curiouser and curiouser... So this can be even faster than 3.7.0, at almost 100 MB/s?

my max speed is 112-114MB/s, over the life cycle of sab ive gotten this speed off and on but never consistently.

puzzledsab:
What do you think, @Safihre ?

Safihre:
I'm not sure what this tells us? To remove all improvements we made?

Issue #2363, opened 2022-12-05
Inconsistent `stage_log` typing in history API calls


JagandeepBrar:
When fetching the SABnzbd history it is possible for `stage_log` to be a String instead of a list of stages. This happens specifically when an NZB can fail when being added, and presumably no stages are logged.

Ideally, an empty array/list would be returned instead of an empty string to allow for type-safe languages to consistently deserialize the object.

Relevant issue in LunaSea's repository: https://github.com/JagandeepBrar/LunaSea/issues/289

thezoggy:
How is the API call being made? 

Supported output types are json, xml and text and can be specified in the request:

http://host:port/sabnzbd/api?output=json

JagandeepBrar:
@thezoggy `http://host:port/sabnzbd/api?output=json&mode=history&limit=200`

In the vast majority of cases, including with most failed NZBs, this does not occur and works as expected. There are rare cases where this can occur, I am trying to find or create an example NZB that will replicate this issue.

thezoggy:
Just a note, changing our API response will prob break the other apps that handle this gracefully...  Not that we shouldn't be consistent but need to see if it causes even more collateral damage

JagandeepBrar:
A list of stages is what is [documented in the API](https://sabnzbd.org/wiki/configuration/3.7/api#history) and what is returned in the _VAST_ majority of cases. I would argue that fixing this would not break any configurations, since making it not return a different type would still be handled by anything that is integrating the API and only improve the experience for type-safe languages that need to add special logic to handle this case which shouldn't exist in the first place.

PR #2364, opened 2022-12-06
fix: set stage_log to empty array on missing database content

JagandeepBrar:
This should solve #2363 as [`stage_log` will be an empty string](https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/database.py#L467-L470) if `nzo.unpack_info` is an empty dictionary. This later will [fail the implicit booleanness check](https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/database.py#L519-L520) and an empty string will remain as the `stage_log`.

This causes `stage_log` to be an empty array instead on any false nullish value (including an empty string), keeping in-line with the expected data type for `stage_log` (an array of stage result dicts).

This keeps the response type true to the expected response type and should not impact any consuming applications or services of the API. The unpack information is only empty/null in very rare scenarios so this does not happen frequently but is problematic for type-safe languages.


Safihre:
Seems logical

PR #2366, opened 2022-12-06
Limited refactoring of recv_chunk

puzzledsab:
I've currently given up on trying to read more than one chunk for each call. It seems like it's more important to read each connection frequently. Anyway, I've looked at this code for so long now that I figured I'd make a PR of some possible improvements I noticed.

- `block` parameter is removed because setting blocking mode makes sure data is read and `ssl.SSLWantReadError` isn't triggered. Instead blocking mode is set explicitly before calling `recv_chunk`. It removes the need for the `while 1` and the `if` in the `except ssl.SSLWantReadError` part.
- I changed the order of `try` and `if self.nntp.nw.server.ssl` because try is only needed for ssl.
- I moved setting `self.timeout` below reading the data. This makes more sense to me because then it will only be updated if data has been read (at least for ssl).

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -178,28 +178,20 @@ class NewsWrapper:
         self.nntp.sock.sendall(command)
         self.clear_data()
 
-    def recv_chunk(self, block: bool = False) -> Tuple[int, bool, bool]:
+    def recv_chunk(self) -> Tuple[int, bool, bool]:
         """Receive data, return #bytes, done, skip"""
-        self.timeout = time.time() + self.server.timeout
-        while 1:
+        if self.nntp.nw.server.ssl:
             try:
-                if self.nntp.nw.server.ssl:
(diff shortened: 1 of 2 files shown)


sabnzbd/newswrapper.py
```diff
@@ -178,28 +178,20 @@ def send_group(self, group: str):
...
-                if block:
-                    # time.sleep(0.0001)
-                    continue
-                else:
-                    return 0, False, True
+                return 0, False, True
+        else:
+            chunk = self.nntp.sock.recv(262144)
+
+        self.timeout = time.time() + self.server.timeout
```
Safihre, line 194: The timeout reset is done at the start for all the other commands, not necessarily meaning it should be at the start but at least all should be the same. Either at start or finish.

thezoggy:
just to note, running the same test i've been doing with this pull:
sab

so more or less the same as before.

PR #2367, opened 2022-12-08
More recv_chunk refactoring

puzzledsab:
I'm not 100% sure it's faster but I think it's cleaner and easier to understand.


Safihre:
Interesting!

PR #2368, opened 2022-12-08
Put some of the nntp connection handling code in a separate method

puzzledsab:
This part of the code is a bit long and hard to follow so I've tried to isolate a chunk of it. It makes it clearer that this is the only way it influences the rest of the loop and it's easier to follow the indenting for the rest of the block. The name could probably be better. I'm still not sure if the handling of `done` is always correct or if it should be set true or false in some cases but at least it should work like before.


Safihre:
Trusting you that it works the same 😅

PR #2370, opened 2022-12-09
better logging with login from multiple IP

sanderjo:
Closes https://github.com/sabnzbd/sabnzbd/discussions/2354

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -856,10 +856,8 @@ class Downloader(Thread):
                 # Login from (too many) different IP addresses
                 if server.active:
                     errormsg = T("Login from too many different IP addresses to server") + display_msg
-                    logging.debug("SJ111: %s", display_msg)
                     if server.errormsg != errormsg:
                         server.errormsg = errormsg
-                        # servername = " (%s)" % server.host
                         logging.warning(
                             T("Login from too many different IP addresses to server")
                             + " "
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -855,13 +855,10 @@ class Downloader(Thread):
             elif error.code in (502, 481, 482) and clues_too_many_ip(error.msg):
                 # Login from (too many) different IP addresses
                 if server.active:
-                    errormsg = (
-                        T("Login from too many different IP addresses to server")
-                        + " "
-                        + server.host
-                        + display_msg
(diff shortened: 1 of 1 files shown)


Safihre:
Could you share a screenshot of the new 1 liner Warning? It seems to me we might need to change the wording maybe.

sanderjo:
image

puzzledsab:
I suggest removing "(too many)" and maybe use %s to place the address. Having it at the end may not be correct for all languages.

sanderjo:
So how about "Multiple IP login"? Not 100% correct, but short

sabnzbd/downloader.py
```diff
@@ -853,13 +853,19 @@ def __finish_connect_nw(self, nw: NewsWrapper) -> bool:
                     self.plan_server(server, _PENALTY_TOOMANY)
                     server.threads -= 1
             elif error.code in (502, 481, 482) and clues_too_many_ip(error.msg):
-                # Account sharing?
+                # Login from (too many) different IP addresses
                 if server.active:
-                    errormsg = T("Probable account sharing") + display_msg
+                    errormsg = T("Login from (too many) different IP addresses") + display_msg
+                    logging.debug("SJ100 %s", display_msg)
```
Safihre, line 859: Remove 😉

sabnzbd/downloader.py
```diff
@@ -853,13 +853,19 @@ def __finish_connect_nw(self, nw: NewsWrapper) -> bool:
                     self.plan_server(server, _PENALTY_TOOMANY)
                     server.threads -= 1
             elif error.code in (502, 481, 482) and clues_too_many_ip(error.msg):
-                # Account sharing?
+                # Login from (too many) different IP addresses
                 if server.active:
-                    errormsg = T("Probable account sharing") + display_msg
+                    errormsg = T("Login from (too many) different IP addresses") + display_msg
```
Safihre, line 858: Agree partially with @puzzledsab, let's make it `Login from too many different IP addresses to server %s`.

Safihre:
I made some small textual changes to the wiki text and 2 small comments, then we can merge it.
After that I will probably improve this code for all these errors, such ugly IF's!

sanderjo:
Message is now

image

sanderjo:
solves https://github.com/sabnzbd/sabnzbd/discussions/2354

PR #2372, opened 2022-12-09
Put */default category first in lists

### LINKED ISSUE #2371: Category order bug
Whenever I install and setup SAB, the category order mirrors the one in the settings but after a while the order always changes for some reason. I usually like `Default` at the top but looking at the screeenshot its found its way to the bottom of the list. I'm on 3.7 linux currently. Its also the same issue on another instance of SAB docker install on my synology.

Screenshot 2022-12-09 114333 - Copy
Screenshot 2022-12-09 114610 - Copy

puzzledsab:
Fixes #2371. I have a feeling there is a prettier way to do this.

### DIFF
--- a/sabnzbd/config.py
+++ b/sabnzbd/config.py
@@ -400,7 +400,7 @@ class ConfigServer:
 
         self.displayname = OptionStr(name, "displayname", add=False)
         self.host = OptionStr(name, "host", add=False)
-        self.port = OptionNumber(name, "port", 119, 0, 2**16 - 1, add=False)
+        self.port = OptionNumber(name, "port", 119, 0, 2 ** 16 - 1, add=False)
         self.timeout = OptionNumber(name, "timeout", 60, 20, 240, add=False)
         self.username = OptionStr(name, "username", add=False)
         self.password = OptionPassword(name, "password", add=False)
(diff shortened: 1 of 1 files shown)


thezoggy:
could we not just remove it, then insert it back as the first item?
```
blah.remote('*')
blah.insert(0, '*')
```

puzzledsab:
I tried that but I couldn't get it right. I figured it out now, I had to use the conf reference from `get_deconfig`.

Safihre:
I don't think this fixes it. Look at the example screenshot and you'll also see that the ordering of the other categories does not match.
I didn't have time to investigate yet, but I was thinking to make sure that the internal list of categories is sorted when loading or modifying the Config.
Currently we have a special sorted way of getting them, which clearly isn't called anymore due to the API refractor in 3.7.0 where categories were removed from the queue call.
https://github.com/sabnzbd/sabnzbd/blob/5cd5f00df787544811b3802d6960d969842f6a99/sabnzbd/config.py#L1034

puzzledsab:
I did notice get_ordered_categories but I got confused by all the slightly different ways the data structure was accessed so I didn't think it was usable. Also, in my sabnzbd.ini, order was 0 for all the elements so I thought it wouldn't work anyway.

Safihre:
Seems my idea of base-sorting things is too complicated.
This is much easier.

That code to get/set the config with the API with the is such a mess, I have never dared to refactor 😭

PR #2379, opened 2022-12-19
Make downloader sleep test optional

puzzledsab:
Sometimes my download rate is a bit unstable so SABnzbd disables can_be_slowed. If this new switch is disabled then can_be_slowed will be set true without doing the test first. I'll add a description to the wiki if you agree.

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -414,6 +414,7 @@ api_warnings = OptionBool("misc", "api_warnings", True, protect=True)
 no_penalties = OptionBool("misc", "no_penalties", False)
 x_frame_options = OptionBool("misc", "x_frame_options", True)
 allow_old_ssl_tls = OptionBool("misc", "allow_old_ssl_tls", False)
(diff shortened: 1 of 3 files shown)


puzzledsab:
In case you're waiting for me to fix the failing tests, I believe they only fail because the wiki does not have the new variable.

Safihre:
Just not completely sure if we need this.
You have a use case, but why would a regular user every enable it? Do we need it to help users on the forums?

puzzledsab:
I believe it can improve performance because sleeping leaves more time available for the other threads. Alternatively it reduces power consumption and heat so the fan doesn't have to spin as fast. I agree that it's not something a normal user would ask for but I think a fair share would enable it if they are made aware of what it does through the documentation.

thezoggy:
whats the purpose of having the downloader sleep.. to help reduce downloading when system is cpu bound?

puzzledsab:
Alternatively we could increase the required time of stable speed to 20 seconds before starting the test. That seems to give much more reliable results, at least for me.

puzzledsab:
Removing because the new recv will probably change the sleep logic.

PR #2391, opened 2022-12-30
Experimental changes to adapt to unlocked_ssl_recv

puzzledsab:
Hoping it will build executables.


Safihre:
@puzzledsab the macOS build fails because the new function doesn't yet work on macOS. The right library name isn't listed yet.
Will fix that soon but currently not near development system.

sabnzbd/downloader.py
```diff
@@ -823,9 +810,13 @@ def run(self):
 
                     # Reset connection for new activity
                     nw.soft_reset()
-                    server.busy_threads.remove(nw)
-                    server.idle_threads.append(nw)
-                    self.remove_socket(nw)
+                    if server.article_queue and nw.connected:
+                        nw.article = server.article_queue.pop(0)
```
Safihre, line 814: Why pop(0)? It's less efdecient and doesn't really matter for us which article we get next? Or does it? 
https://stackoverflow.com/a/195839

puzzledsab: The difference is minimal for small lists and it helps the assembler a bit if the data arrive more or less in sequence. Here are the timings of doing pop(0), pop() and reading the elements using counting and index on a list with 20 items 1000000 times.
```
pop0
1.8746365000260994
pop
1.426083300029859
count
1.4542528000893071
```

Safihre: 👍 👍

mnightingale:
I've not tried this PR since a couple of days ago but I found it reduced my speed with my in-memory test setup by a couple of hundred MB/s I think this is partly because assemble gets called much more frequently.

I also think this PR could also use a lot of memory; the assembler loads articles from the cache or disk, this creates its own queue of article data therefore on some devices if MD5 hashing speed was slow but it could write to storage quickly memory usage could bloon.

I'll put some thoughts on #2398 when I get a chance.

sabnzbd/decoder.py
```diff
@@ -239,7 +239,9 @@ def run(self):
                 # If the data needs to be written to disk due to full cache, this will be slow
                 # Causing the decoder-queue to fill up and delay the downloader
                 sabnzbd.ArticleCache.save_article(article, decoded_data)
-
+            elif not nzo.precheck:
+                # Nothing to save
+                article.on_disk = True
```
puzzledsab, line 244: @Safihre : Do you know if this can break anything? The intent is to let the assembler continue writing a file if an article is confirmed missing so that the cache doesn't fill up.

PR #2392, opened 2022-12-30
Use index for assembler to reduce looping and file opening

puzzledsab:
Inspired by animetosho's suggestion to use an iterator. Saving assembler_index is not really required, I can skip that if you like.


Safihre:
Is this really worth it to skip a few `if on_disk`?
It makes things much harder to read in my opinion. Some of the other optimizations discussed like removing md5 probably have much more impact?

puzzledsab:
Maybe you'll like this version better. Not quite as optimal as the first one but it will avoid opening the file and entering the loop if the next article in line isn't ready for saving.

puzzledsab:
Does not seem to improve speed noticeably.

PR #2401, opened 2023-01-03
fix: extracted file permissions on Synology

mreid-tt:
For archives extracted under non-Windows platforms the "-ai" switch was excluded with the logic that it was not needed. It is proposed that the switch is still required as it allows the extracted files to ignore the file attributes stored in the archive.

This is important for archives which may have no group attributes set as these extract with permissions only set for the user. This results in problems post-processing the files with other packages on Synology platforms running DSM 7 as demonstrated in the following example:

```
Sample archive contents:
Archive: ./rar_files/Rihanna - Loud.part1.rar
Details: RAR 4, volume, lock

 Attributes      Size    Packed Ratio    Date    Time   Checksum  Name
----------- ---------  -------- ----- ---------- -----  --------  ----
    ..A....      1907       686  35%  2014-03-16 18:10  C4966751  00._Rihanna_-_Loud.nfo
    ..A....   9800349   9308923  94%  2010-11-14 18:29  0F544CC5  01 S&M.mp3
[---snip---]
```

When this is extracted using the standard switches, the files look like this:

```
$ sudo -u sc-sabnzbd /var/packages/sabnzbd/target/bin/unrar x -vp -idp -scf -o+ -p- ./rar_files/Rihanna\ -\ Loud.part1.rar ./test4

UNRAR 6.12 freeware      Copyright (c) 1993-2022 Alexander Roshal

Extracting from ./rar_files/Rihanna - Loud.part1.rar
[---snip---]
Extracting  ./test4/00._Rihanna_-_Loud.nfo                            OK 
Extracting  ./test4/01 S&M.mp3                                        OK 
[---snip---]
Extracting  ./test4/Rihanna - Loud - Front.bmp                        OK 
All OK
```
```
$ ls -all test4
total 129220
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      750 Jan  3 12:38  .
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      122 Jan  3 12:37  ..
-rw-------  1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rw-------  1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

For the extracted files they have the attributes 'rw' for the owner and no attributes otherwise. This is different than the attributes which are set when we include the "-ai" switch in the following example:

```
$ sudo -u sc-sabnzbd /var/packages/sabnzbd/target/bin/unrar x -vp -idp -scf -o+ -ai -p- ./rar_files/Rihanna\ -\ Loud.part1.rar ./test3

UNRAR 6.12 freeware      Copyright (c) 1993-2022 Alexander Roshal

Extracting from ./rar_files/Rihanna - Loud.part1.rar
[---snip---]
Extracting  ./test3/00._Rihanna_-_Loud.nfo                            OK 
Extracting  ./test3/01 S&M.mp3                                        OK 
[---snip---]
Extracting  ./test3/Rihanna - Loud - Front.bmp                        OK 
All OK
```
```
$ ls -all test3/
total 129220
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      750 Jan  3 12:17  .
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      140 Jan  3 12:16  ..
-rwxrwxrwx+ 1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rwxrwxrwx+ 1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

In this case, the files have attributes assigned by the operating system for newly created files. Now this would include executable permissions but the built-in set_permissions script would clean this up where files contain 'xbits'.

This was proven by applying these modifications to a Virtual DSM unit running DSM 7.1.1. The debug logs are included below:
*	[sabnzbd.log (without "-ai")](https://pastebin.com/dbB346NT)
*	[sabnzbd.log (with "-ai")](https://pastebin.com/mUgxejzG)

This fix should be simple enough to address the permissions issue on Synology as noted in: https://github.com/SynoCommunity/spksrc/issues/4940.

### DIFF
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -414,7 +414,6 @@ class DirectUnpacker(threading.Thread):
                 "%s\\" % long_path(extraction_path),
             ]
         else:
-            # Don't use "-ai" (not needed for non-Windows)
             # The -scf forces the output to be UTF8
             command = [
                 sabnzbd.newsunpack.RAR_COMMAND,
@@ -423,6 +422,7 @@ class DirectUnpacker(threading.Thread):
                 "-idp",
                 "-scf",
                 "-o+",
+                "-ai",
                 password_command,
                 rarfile_path,
                 "%s/" % extraction_path,
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -624,13 +624,16 @@ def set_permissions(path: str, recursive: bool = True):
                 # If custom permissions are absent, set them when doing recursion
                 if not custom_permissions:
                     logging.debug("Setting permission bits of files based on parent folder")
-                    custom_permissions = int(os.stat(path).st_mode)
+                    root_permissions = int(os.stat(path).st_mode)
                 # Parse the dir/file tree and set permissions
                 for root, _, files in os.walk(path):
                     if custom_permissions:
                         set_chmod(root, custom_permissions)
                     for name in files:
-                        removexbits(os.path.join(root, name), custom_permissions)
+                        if custom_permissions:
+                            removexbits(os.path.join(root, name), custom_permissions)
+                        else:
+                            removexbits(os.path.join(root, name), root_permissions)
             elif custom_permissions:
                 set_chmod(path, custom_permissions)
         else:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -621,10 +621,6 @@ def set_permissions(path: str, recursive: bool = True):
 
         if os.path.isdir(path):
             if recursive:
-                # If custom permissions are absent, set them when doing recursion
-                if not custom_permissions:
-                    logging.debug("Setting permission bits of files based on parent folder")
-                    custom_permissions = int(os.stat(path).st_mode)
                 # Parse the dir/file tree and set permissions
                 for root, _, files in os.walk(path):
                     if custom_permissions:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -621,6 +621,10 @@ def set_permissions(path: str, recursive: bool = True):
 
         if os.path.isdir(path):
             if recursive:
+                # If custom permissions are absent, set them when doing recursion
+                if not custom_permissions:
+                    logging.debug("Setting permission bits of files based on parent folder")
+                    custom_permissions = int(os.stat(path).st_mode)
                 # Parse the dir/file tree and set permissions
                 for root, _, files in os.walk(path):
                     if custom_permissions:
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -624,16 +624,13 @@ def set_permissions(path: str, recursive: bool = True):
                 # If custom permissions are absent, set them when doing recursion
                 if not custom_permissions:
                     logging.debug("Setting permission bits of files based on parent folder")
-                    root_permissions = int(os.stat(path).st_mode)
+                    custom_permissions = int(os.stat(path).st_mode)
                 # Parse the dir/file tree and set permissions
                 for root, _, files in os.walk(path):
                     if custom_permissions:
                         set_chmod(root, custom_permissions)
                     for name in files:
-                        if custom_permissions:
-                            removexbits(os.path.join(root, name), custom_permissions)
-                        else:
-                            removexbits(os.path.join(root, name), root_permissions)
+                        removexbits(os.path.join(root, name), custom_permissions)
             elif custom_permissions:
                 set_chmod(path, custom_permissions)
         else:
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -701,7 +701,6 @@ def rar_extract_core(
             "%s/" % extraction_path,
         ]
     else:
-        # Don't use "-ai" (not needed for non-Windows)
         # The -scf forces the output to be UTF8
         command = [
             RAR_COMMAND,
@@ -710,6 +709,7 @@ def rar_extract_core(
             "-scf",
             overwrite,
             rename,
+            "-ai",
             password_command,
             rarfile_path,
             "%s/" % extraction_path,


Safihre:
@sanderjo and I were already discussing this yesterday. The `ai` switch is not left out because of `scf` but because of some historical reason 10 years ago. Seems back then unrar on Linux did not support it.

I contacted rarlabs to ask if this switch removes any other attributes, just to be sure. Usually they are very quick to respond.

Interesting this only pops up now, since we had this for years.

Another thing @sanderjo and I found was that 7zip also has the same "problem" but no switch to remove the permissions when unpacking.

mreid-tt:
> Another thing @sanderjo and I found was that 7zip also has the same "problem" but no switch to remove the permissions when unpacking.

Interesting. I guess I've not come across this because most files on newsgroups seem to be RAR compressed. A possible direction that can be looked into is to simply set the file permissions to '0666' for **all** files and not just the ones with the 'xbits'.

thezoggy:
> Interesting. I guess I've not come across this because most files on newsgroups seem to be RAR compressed. A possible direction that can be looked into is to simply set the file permissions to '0666' for **all** files and not just the ones containing 'xbits'.

its actually against the rules to use compression, why majority of stuff is only store not compressed with rar. rarset is just used to split files.

Safihre:
> Interesting. I guess I've not come across this because most files on newsgroups seem to be RAR compressed. A possible direction that can be looked into is to simply set the file permissions to '0666' for **all** files and not just the ones containing 'xbits'.

We approach it that if users want that, they can configure explicitly the Permissions setting. Otherwise we won't touch it. Since some users do no want 666 but have the permissions inherited from the parents and/or process's permissions.

mreid-tt:
That makes sense (regarding some users not wanting 666), since it may conflict with the "Permissions for completed downloads" setting. If we were to consider new logic it could perhaps be something like this:

1. Extract files using the default attributes (permissions included in the archive)
2. Update the extracted files to match the permissions of the parent folder
3. Examine files for the 'xbits' and remove the bit using something like `chmod a-x` (rather than setting 666)

What do you think?

Safihre:
No this will not work. 
If we use chmod, it will destroy ACL permissions. At least it did that in all my testing.
What was wrong with your previous code of just adding the extra switch to unrar?

mreid-tt:
@Safihre, I've changed the approach to this issue. Instead of modifying the `unrar` calls, I've instead added a check for the `set_permissions` function. The check will set the custom permissions based on the parent folder if we are doing recursion. This was done because you mentioned that this fix will not work for `7zip`.

Using the same original source RAR file these debug logs were generated:
* Original debug log - https://pastebin.com/YzbVqBnm
* Debug log using new code - https://pastebin.com/0zbY6sMQ

Original files extracted:
```
$ ls -all Rihanna\ Loud/
total 129224
drwx------+ 1 sc-sabnzbd synocommunity      786 Jan  3 20:24  .
drwx------+ 1 sc-sabnzbd synocommunity       24 Jan  3 20:24  ..
-rw-------  1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rw-------  1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

Files extracted using new code:
```
$ ls -all Rihanna\ Loud/
Password: 
total 129224
drwxrwxrwx  1 sc-sabnzbd synocommunity      786 Jan  3 20:17  .
drwx------+ 1 sc-sabnzbd synocommunity       24 Jan  3 20:17  ..
-rw-rw-rw-  1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rw-rw-rw-  1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

As you can see it has the same effect as including "-ai" in the `unrar` call.

Safihre:
No it does nog have the same effect, look at the permissions:
`-rw-rw-rw-` (last code)
`-rwxrwxrwx+` (adding the switch)
The important bit is the `+` at the end which indicates that ACL permissions are active, and it is missing in the lastest code.

thezoggy:
Why does syno need this fix compared to any other os?

Safihre:
> Why does syno need this fix compared to any other os?

Because the way the ACL permission system works. If you explicitly use chmod it destroys that and breaks DSM permissions. So it has to be correct right away, to prevent using chmod.

mreid-tt:
> No it does nog have the same effect, look at the permissions:
> `-rw-rw-rw-` (last code)
> `-rwxrwxrwx+` (adding the switch)
> The important bit is the `+` at the end which indicates that ACL permissions are active, and it is missing in the lastest code.

You are correct based on the example directories provided however for the code with the switch, this is when the `unrar` is triggered manually. If we apply it to SABnzbd and take a look at the directory after it runs the `set_permissions` script it looks like this:

```
ls -all Rihanna\ Loud/
total 129224
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      786 Jan  4 04:26  .
drwxrwxrwx+ 1 sc-sabnzbd synocommunity       24 Jan  4 04:26  ..
-rw-rw-rw-  1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rw-rw-rw-  1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

As you can see it gets stripped anyway; the only difference is the permission bits (octal) used. The particular evidence is seen in the debug logs I provided:

Debug log based on unrar switch:

```
2023-01-03 11:06:58,060::DEBUG::[filesystem:604] Applying permissions 0o100666 (octal) to /volume1/downloads/complete/_UNPACK_Rihanna Loud/00._Rihanna_-_Loud.nfo
2023-01-03 11:06:58,060::DEBUG::[filesystem:604] Applying permissions 0o100666 (octal) to /volume1/downloads/complete/_UNPACK_Rihanna Loud/01 S&M.mp3
[---snip---]
```

Debug log based on recursion using parent:

```
2023-01-03 20:17:45,783::DEBUG::[filesystem:604] Applying permissions 0o40777 (octal) to /volume1/downloads/complete/_UNPACK_Rihanna Loud
2023-01-03 20:17:45,783::DEBUG::[filesystem:604] Applying permissions 0o40666 (octal) to /volume1/downloads/complete/_UNPACK_Rihanna Loud/00._Rihanna_-_Loud.nfo
2023-01-03 20:17:45,784::DEBUG::[filesystem:604] Applying permissions 0o40666 (octal) to /volume1/downloads/complete/_UNPACK_Rihanna Loud/01 S&M.mp3
[---snip---]
```

Effectively, the results are the same.

mreid-tt:
> > Why does syno need this fix compared to any other os?
> 
> Because the way the ACL permission system works. If you explicitly use chmod it destroys that and breaks DSM permissions. So it has to be correct right away, to prevent using chmod.

I am not sure this is correct. I don't have a deployment in Linux but I suspect the mechanics would be the same. When we extract files from an archive, initially the files are written with the permission bits set based on what is in the archive. With the particular archive in the example, this results in files with 'rw' permissions only on the owner bits. As such, if there were any other apps which needed to access these files they would not have the permissions to do so.

Manually setting the file permissions to include permissions for the group (and/or others) would allow other apps to post-process the download. The difference is that with the "-ai" switch, it sets the permission bits of the files to the default for newly created files, whereas with the recursion code it sets the permission bits of the files to the parent folder. This is effectively the same because the parent folder would have the permission bits of a newly created file.

mreid-tt:
> I am confused. When we use `-ai` it will use the permissions for new files, which _includes_ the ACL. It will not include the execute permissions bits (because we used `-ai`), so `set_permissions` will not be triggered and the ACL will be preserved. Am I wrong?

From what I've observed, the `set_permissions` is triggered during post processing of the download. In the method, it will be skipped if running on Windows platforms. If on non-Windows, it will check if the argument is a file and if so remove the 'xbits'. If the argument is a directory and custom permissions are set, it will apply these to the directory. If recursion is true it will remove the 'xbits' for each file in the tree. As such, if there are no custom permissions set and recursion is on, it will attempt to remove the 'xbits' for each file in the path. Because the default O/S permissions for newly created files includes the execute permissions, `set_permissions` is triggered.

Further, within the logic of the `removexbits` method, it will first check if custom permissions are set and if not it will extract the existing permissions of the object. If the existing permissions contain execute permissions a `set_chmod` is called which does strip the ACL. As such, when we are using the "-ai" switch, it will always strip the ACL because the O/S has execute bits enabled for newly created files.

Safihre:
Will check when I get back to my NAS.

I got response from Rarlabs, to confirm the impact of `-ai`:
```
Hello,

> The -ai switch says it ignores file attributes, I was wondering: which ones?

In Windows it affects archive, system, hidden and read-only attributes.
in Unix - user, group and other file permissions.

Eugene
```

mreid-tt:
> Will check when I get back to my NAS.

Synology ACLs seem to be very weird. Referring to their KB for [Assign Shared Folder Permissions](https://kb.synology.com/en-id/DSM/help/DSM/AdminCenter/file_share_privilege), it lists the following:

> * Read:
>   - Traverse folders/Execute files: This controls whether a user can run a program file.
> * Write:
>   - Create files/Write data: This controls whether a user can change the contents of a file.

The problem is if you set the share permissions to enable "Create files/Write data" but disable "Traverse folders/Execute files" it gives you a permission denied when trying to create a file with `touch` as well as throws errors in SABnzbd. As such, it seems that you cannot get away from the O/S enabling execution bits when creating a new file in Synology unlike in other `*nix` systems.

mreid-tt:
> @mreid-tt Could uou share another test NZB?

Sure, here's the file again: https://file.io/wCSMsRi9DuM3 or https://www.filemail.com/d/bibuxlosszomidp

mreid-tt:
> Maybe they're using only a limited subset of ACL capabilities to keep things simple?

From their KB for [Assign Shared Folder Permissions](https://kb.synology.com/en-id/DSM/help/DSM/AdminCenter/file_share_privilege), it notes that:

> In DSM 5.0 or later version, the access permissions of shared folders are based on Windows ACL by default. Newly created shared folders implement the permissions settings of Windows ACL, which also allows for customizing the permissions of individual files and subfolders.

So I guess this is the reason why it behaves differently than standard `*nix`?

> IIRC from when they introduced this in DSM, the key part for sab was not getting in the way of the ownership and permissions inherited from directories. The removexbits function only targets files, which have no descendants to worry about. Running chmod on directories - as proposed in this PR - may have unintended side effects if the affected directory sees further use (e.g. in case of multiple archives in a job, a no-job-dirs flat directory structure configured in sab, or a postproc script creating new files/dirs).

Hmm, while you make sense that the `removexbits` function should only target files, I've tested it with a modification to the code (https://github.com/sabnzbd/sabnzbd/pull/2401/commits/0ef72f603810d31cd5e97d6bda1138482cec17d7) and but having it can cause traversal issues for other Synology packages. Take for example this archive:

```
Archive: hYDECdeL.rar
Details: RAR 4, recovery record

 Attributes      Size    Packed Ratio    Date    Time   Checksum  Name
----------- ---------  -------- ----- ---------- -----  --------  ----
    ..A....       107       107 100%  2017-10-13 21:00  2206A86B  hYDECdeL/lees dit even  door/nieuwe site, voor al uw films .url
    ..A....       201       201 100%  2017-11-17 09:37  2D2B1357  hYDECdeL/lees dit even  door/wij zoeken staff leden.txt
    ..A....   8717877   8717877 100%  2017-12-01 16:22  4FEB3D39  hYDECdeL/Sam Smith - The Thrill Of It All (2017) 320/1. Too Good at Goodbyes.mp3
[---snip---]
```

If we do an extraction and only run the `removexbits` on files, we get the following folder structure:

```
$ ls -all /volume1/downloads/complete/Sam_Smith_-_The_Thrill_of_It_All_2017_MP3/
total 48
drwxrwxrwx+ 1 sc-sabnzbd synocommunity    40 Jan  5 07:11 .
drwxrwxrwx+ 1 sc-sabnzbd synocommunity    82 Jan  5 07:11 ..
drwx------  1 sc-sabnzbd synocommunity   124 Dec  1  2017 hYDECdeL
-rw-rw-rw-  1 sc-sabnzbd synocommunity 45780 Jan  5 07:11 hYDECdeL.nzb

$ ls -all /volume1/downloads/complete/Sam_Smith_-_The_Thrill_of_It_All_2017_MP3/hYDECdeL
ls: cannot open directory '/volume1/downloads/complete/Sam_Smith_-_The_Thrill_of_It_All_2017_MP3/hYDECdeL': Permission denied
```

As noted in the above, because the extracted files are in a sub-folder where only `sc-sabnzbd` has access, other packages (like lidarr) would not be able to post-process these files. So from what I can tell, not applying the permissions to the entire tree is not ideal either.

Sharing the details from my tests in case there is need to reproduce these results:
* https://pastebin.com/BZhCacfx (debug log with directories modified)
* https://pastebin.com/SUkUDj0j (debug log with directories skipped)
* https://file.io/uasUoVkwpTUl or https://www.filemail.com/d/czisagirdaxfmat (original NZB file)

Safihre:
@mreid-tt I can confirm your findings that the files have xbits set and get removed by SABnzbd.
So does this change now "fix" that Lidarr can import it? Despite that we called chmod on the files?

Safihre:
Ah sorry, I see you updated your message that this fixes it! 👍

mreid-tt:
> Ah sorry, I see you updated your message that this fixes it! 👍

Yes, based on the comment from @jcfp, the concern regarding 7zip archives should be relatively uncommon and thus this fix should all that would be required. Should I squash the commits to prepare for merging?

jpdsc:
@mreid-tt , I have tried this fix and using the develop branch on the hotio image (efd7d1a4a0cae6eb36064e62805b7c1705c35a08)
The unpacked download still removes the Synology permissions. Basically ignoring the -ai flag as it was doing before and applying 777.

I checked the option in settings and there is the field to apply permissions is empty.

Any clue?

Safihre:
We remove executable bits on files if we see them.
You will now most likely ask for an option to disable that, however, that would create an easy attack vector for all people that accidentally have their SABnzbd exposed to the internet without password.

jpdsc:
> We remove executable bits on files if we see them. You will now most likely ask for an option to disable that, however, that would create an easy attack vector for all people that accidentally have their SABnzbd exposed to the internet without password.

I might have my SABnzbd set-up different then most here.
My SABnzbd isn't running on my Synology NAS. Instead, it's running on a VM on other HW in Docker. It moves downloads after or during unpacking to the Synology NAS.
When the download is unpacked or during unpack; it is changing the permissions instead of ignoring them using -ai which causes the permissions in the Synology NAS to be removed and 777 applied.

Does this also apply to your comment?

mreid-tt:
My understanding of the way SABnzbd is setup is that if you have files with the executable bit enabled it will disable it. If it has to disable it, it will use the `chmod` command to do so. In using this command the ACL permissions from Synology will be stripped. This typically is not a concern especially for files (as compared to folders) so all in all you should be good.

The only difference with the `-ai` switch is that it will override the permissions in the archive to the default for newly created files. In the Synology platform from what I can tell newly created files have the equivalent of a '777' permissions set with ACL. Because of this SABnzbd will strip these down to `666` for files with no ACL. Folders should not be touched. Are you experiencing something different?

jpdsc:
> My understanding of the way SABnzbd is setup is that if you have files with the executable bit enabled it will disable it. If it has to disable it, it will use the `chmod` command to do so. In using this command the ACL permissions from Synology will be stripped. This typically is not a concern especially for files (as compared to folders) so all in all you should be good.
> 
> The only difference with the `-ai` switch is that it will override the permissions in the archive to the default for newly created files. In the Synology platform from what I can tell newly created files have the equivalent of a '777' permissions set with ACL. Because of this SABnzbd will strip these down to '666' for files with no ACL. Folders should not be touched. Are you experiencing something different?

Thank you for the explanation.
The files are actually moved to a NFS share upon unpack or during direct unpack. 
Now, if I do this with NZBget with the -ai flag, it skips any permissions and just moves the file using the container UID. With SABnzbd the file permissions get changed to 777/666.

The folders remain intact, this is because I used UMASK 007 in the compose. If I don't add UMASK 007, it also changes the permissions on the Completed download folder in the Synology NAS.
So in an overview:

SABnzbd downloads file and unpacks/direct unpacks
The file is moved to the NFS share during this process
However, during this process, the permissions are changed by SABnzbd to use 777/666 and therefore the permissions from Synology are stripped.
The folder itself is created by Radarr/Sonarr so there is no change with the folder regarding the Synology permissions.

Is there anything else I can do to ensure the Synology permissions stay intact for files? It shouldn't be an issue in Synology, however, it is creating a mess with permissions.
I'm using Plex in container on the NAS, if I would use the native app, the Video/PlexMediaServer permissions would also be stripped and therefore no access to Plex.

mreid-tt:
> SABnzbd downloads file and unpacks/direct unpacks The file is moved to the NFS share during this process However, during this process, the permissions are changed by SABnzbd to use 777/666 and therefore the permissions from Synology are stripped. The folder itself is created by Radarr/Sonarr so there is no change with the folder regarding the Synology permissions.
> 
> Is there anything else I can do to ensure the Synology permissions stay intact for files? It shouldn't be an issue in Synology, however, it is creating a mess with permissions. I'm using Plex in container on the NAS, if I would use the native app, the Video/PlexMediaServer permissions would also be stripped and therefore no access to Plex.

This sounds like a new issue. The reason I say this is because before the `-ai` switch, if the archive contained executable bits, the `chmod` would fire anyway. Now it just happens more consistently. If you have already setup the file share permissions as recommended by the SynoCommunity [Permission Management](https://github.com/SynoCommunity/spksrc/wiki/Permission-Management) page and you are still getting issues I would suggest you open a new issue to explore this more fully (do include your logs and other relevant data for examination).

jpdsc:
> > SABnzbd downloads file and unpacks/direct unpacks The file is moved to the NFS share during this process However, during this process, the permissions are changed by SABnzbd to use 777/666 and therefore the permissions from Synology are stripped. The folder itself is created by Radarr/Sonarr so there is no change with the folder regarding the Synology permissions.
> > Is there anything else I can do to ensure the Synology permissions stay intact for files? It shouldn't be an issue in Synology, however, it is creating a mess with permissions. I'm using Plex in container on the NAS, if I would use the native app, the Video/PlexMediaServer permissions would also be stripped and therefore no access to Plex.
> 
> This sounds like a new issue. The reason I say this is because before the `-ai` switch, if the archive contained executable bits, the `chmod` would fire anyway. Now it just happens more consistently. If you have already setup the file share permissions as recommended by the SynoCommunity [Permission Management](https://github.com/SynoCommunity/spksrc/wiki/Permission-Management) page and you are still getting issues I would suggest you open a new issue to explore this more fully (do include your logs and other relevant data for examination).

I think this is just how SABnzbd deals with completed downloads. It performs chmod even if the option is blank/not filled in.
It would be great if there was an option to disable any permissions being set (apart from the x). This will ensure that any file or folder created uses the UID set in the docker container.

@Safihre, is it possible to get such an option in Special?

Safihre:
For the **last** time: SABnzbd does not apply permissions if the option is empty, it only performs a chmod if the files have X bit set so those are removed. Again, this is for security reasons.

PR #2402, opened 2023-01-04
Set on_disk = True on unavailable articles

puzzledsab:
To avoid them stopping the assembler and the cache filling up. Do you know if this can break anything?


puzzledsab:
The article data is saved to cache if there are any data but the assembler will only save it if `article.decoded = True`. Shouldn't decoded be set true if there are any data at all? Otherwise, why save it?

puzzledsab:
I need to think some more about this.

PR #2403, opened 2023-01-04
Use broken article data if no valid data exists and let the assembler continue if an article is missing

puzzledsab:
I tested it by making a single article nzb and putting a wrong CRC in mnightingale's server, with some real servers on a lower priority. It seems to work and saved the 800 000 byte file. The dev version saves a 0 byte file. The bad_article count stayed at 0, though. I don't understand why.

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -167,9 +167,6 @@ class DecoderWorker(Thread):
             except BadData as error:
                 # Continue to the next one if we found new server
                 if search_new_server(article):
-                    if error.data and not article.decoded:
-                        article.decoded = True
-                        sabnzbd.ArticleCache.save_article(article, error.data)
                     continue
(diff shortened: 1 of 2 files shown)


Safihre:
At this point I don't think I want to add that code complexity for an edge case where only 1 server has the article with an error, but others have them all missing.
I do like the change to set `on_disk` for missing articles.

puzzledsab:
I realized that it was a bit broken anyway. By setting article.decoded immediately on articles with CRC errors there's a good chance that the assembler will find it and try to save it while the downloader checks the other servers. This is true for the current version as well so I moved `article.decoded = True` until after it's saved. Hypothetically it could even do it with a valid article in the time it takes from decoded is set to it's saved.

Safihre:
That's a good catch! Interesting.

Issue #2396, opened 2023-01-02
Consider replacing MD5 with CRC32 for quick check


animetosho:
To not stray too far off topic in a [previous conversation](https://github.com/sabnzbd/sabnzbd/discussions/2352#discussioncomment-4549416), I've decided to move the discussion here.

Summary so far:

* `assemble` computes the MD5 of the file it's assembling. MD5 computation may become a performance limiter in the future (and already consumes a fair bit of CPU)
* @puzzledsab [pointed out](https://github.com/sabnzbd/sabnzbd/discussions/2352#discussioncomment-4557885) that MD5 could probably be replaced with CRC32
* CRC32 is [already being computed](https://github.com/sabnzbd/sabyenc/issues/29#issuecomment-1126122276) during yEnc decoding, so adopting CRC32 would effectively eliminate the need to separately compute a hash
* only checking yEnc CRC32 against the decoded articles (without any PAR2 check) has been tried in the past (#2160), but showed issues
* PAR2, however, does [also contain CRC32s](https://parchive.sourceforge.net/docs/specifications/parity-volume-spec/article-spec.html#i__134603784_796) that could be checked against, however they are block hashes, not file hashes
* a file's CRC32 can be obtained by *joining* the decoded articles' CRC32s (without the need to hash all the data again)
* similarly, the file's CRC32 can be obtained from the PAR2 by joining the block hashes. The last block's hash may need to have it's zero padding trimmed
* [sample Python code](https://gist.github.com/animetosho/942a96ab3299c98ee35663208e00737b) for CRC manipulation operations

Essentially this could eliminate the need to compute the MD5 hash of each file, whilst still verifying the data against what's specified in the PAR2. The downside is that CRC32 is a less reliable hash than MD5.

---

If we're going to experiment, I thought I'd point the following out, in case it's missed:

When assembling the CRC from decoded articles, they don't need to be assembled in order (unlike MD5). This is handy since articles don't necessarily arrive in order, and there's no need to keep the article CRCs around.

Example:

```python
file_crc = 0
once_an_article_is_decoded(article, file):
	article_end = article.offset + article.length
	bytes_til_file_end = file.size - article_end
	file_crc = crc_concat(article.crc, file_crc, bytes_til_file_end)

# once all articles have been decoded, file_crc will be the file's CRC32
```

When assembling CRC from PAR2's IFSC, the block size is constant across the entire PAR2. If you look at the definition of `crc_concat`:

```python
[...]

Issue #2169, opened 2022-05-03
Preallocate files


reallyuniquename:
Right now missing articles are simply getting skipped. Remaining data is misaligned and resulting filesize is different. 

This behaviour complicates a lot of simple ways of restoring missing chunks even if you can get them from other sources. You can't easily fill just the missing data from P2P networks. You can't use RAR's internal recovery record. You can't unpack existing data from multipart archives even when total number of lost articles is relatively small.

I know, I know. NZB files do not provide length of actual data, only a number of raw bytes. But you can still guess and I believe some clients already do that quite fine.

I've just had to pull 50GB of data once again while I was missing less than 1MB. This is really frustrating. I also see an increasing number of uploaders that intentionally do not provide any PARs and 1 missing article trashes the whole batch.

Anyway this has been brought up a few times on forums with a lof valid points but I don't think it got any traction: `/viewtopic.php?t=9851` & `/viewtopic.php?t=15373`.

Safihre:
It's easily possible, the information is inside the yEnc header. That's how nzbget does it. I already made a proof of concept a year or so ago. 
It's just quite a big investment and there was no real good reason to do it. Par2 repair works just as well in my tests with filled or unfilled data.
I understand your use case, but it's also a bit exotic.

A much better use case would be that this way we can actually implement a proper Retry where only the missing articles are tried again and filled in at the right spot in the file. That's the reason why nzbget implemented it.

thezoggy:
there is tons of obfuscation methods, purposely not including pars / leaving stuff incomplete on purpose (force you to get the actual nzb from the soruce) and many other reasons. some are to avoid dmca some are to avoid other sites from leeching their stuff.

writing files is slow and puts wear on hdd.. why do it with 0 then to just do it again with said data... i dont see any benefit, just more wear on a hdd and wasting time?

Safihre:
@thezoggy you're not actually writing data, you say `fp.seek(X)`. All the space between `0` and `X` will just not be filled with anything. So no extra writes.
Depends on the OS how it actually fills the space as far as I understand.

reallyuniquename:
True, implementing this would make proper Retry possible too.  

>the information is inside the yEnc header

I'm not familiar with NNTP. If article is missing where would the header come from? 

@thezoggy 
[...]

PR #2405, opened 2023-01-07
Use CRC32 for file check

puzzledsab:
Related to #2396


sabnzbd/par2file.py
```diff
@@ -106,40 +109,63 @@ def parse_par2_file(fname: str, md5of16k: Dict[bytes, str]) -> Tuple[str, Dict[s
     http://parchive.sourceforge.net/docs/specifications/parity-volume-spec/article-spec.html
     """
     total_size = os.path.getsize(fname)
+    metadata = {}
+    metadata["nr_files"] = -1
+    filedata = {}
```
Safihre, line 114: Could you make `metadata` and `filedata` into `dataclass`? It's easier to read than dict's.

sabnzbd/utils/crc32calc.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/python3 -OO
+
+"""
+Functions for doing CRC32 calculations
+Written by animetoshi
```
Safihre, line 5: animetoshO 
;)

sabnzbd/assembler.py
```diff
@@ -191,7 +188,10 @@ def assemble(nzo: NzbObject, nzf: NzbFile, file_done: bool):
                     # Could be empty in case nzo was deleted
                     if data:
                         fout.write(data)
-                        nzf.md5.update(data)
+                        if len(data) == nzo.article_size:
+                            nzf.crc32 = crc_multiply(nzf.crc32, nzo.crc32_coeff) ^ article.crc32
+                        else:
+                            nzf.crc32 = crc_concat(nzf.crc32, article.crc32, len(data))
```
Safihre, line 194: Maybe we can have an `nzf.update_crc32(data, article.crc)` function instead of doing this here?

Safihre (changes requested):
After these comments, do you prefer to merge as-is or first really get rid of the `md5` attributes etc?

Safihre:
I update sabyenc3 to return the article-CRC instead of `crc_correct`.
It's part of this branch: https://github.com/sabnzbd/sabyenc/tree/openssl_link
I verified that it returns the same value as `zlib.crc32(raw_data)`.

Maybe it's easier to then base this PR against https://github.com/sabnzbd/sabnzbd/tree/feature/buffered_reader since it already uses this newer version of `sabyenc3`?

sabnzbd/par2file.py
puzzledsab: Sorry, I don't see how to get rid of the filedata dict? How to connect the CRC32 values to the right file?

puzzledsab:
I'm not quite sure where to go with this patch. Have you considered keeping MD5 but making it optional? I think it should be good enough to be the only choice but I'm not 100% sure. The md5full value is even included in the database so I'm not sure about how much can break and if it needs some kind of special transition. I don't know enough git management stuff to know what kind of merging is preferable. I can submit a new PR to the other branch or I can wait until the ssl reader stuff is more settled. Whatever is easier.

thezoggy:
Don't we used the md5 for dupe checking.. pretty sure we shouldn't use crc32 for that

puzzledsab:
I looked at it a bit closer and the md5sum in the database seems to be the md5 of all the article IDs (nzo.md5sum), to identify duplicate nzbs. My patch only affects the file md5sum (nzf.md5sum), which is unrelated.

Safihre:
> I looked at it a bit closer and the md5sum in the database seems to be the md5 of all the article IDs (nzo.md5sum), used to identify duplicate nzbs. My patch only affects the file md5sum (nzf.md5sum), which is unrelated.

Exactly!
So we can safely transition. If it's good enough for nzbget, it should be good enough for us, right?
Or you'd like to offer both? Speed vs reliability?
We could have a testing period with both methods and let users report the cases where CRC failed.
@animetosho can we really expect CRC to give false-positive results, so that it approves a download even though it's broken? I don't care if it has some false-negatives, if a few times in 1000 downloads it incorrectly triggers a full par2 scan.

mnightingale:
> I update sabyenc3 to return the article-CRC instead of `crc_correct`.

The tests need updating :speak_no_evil:

Would there be any value in returning a custom type instead of a tuple?

animetosho:
> can we really expect CRC to give false-positive results, so that it approves a download even though it's broken? I don't care if it has some false-negatives

All hash checks are subject to false-positives, and false-negatives should never occur.  
What generally matters is the *likelihood* of false-positives.

Assuming no 'extreme' scenarios (like someone deliberately trying to engineer a collision), it's practically impossible MD5 will false-positive, given a 128-bit hash size.  
CRC32 only has 32 bits, so the probability there is much higher.

I don't know how to calculate the probability, but if CRC32 is a 'perfect' 32-bit hash (I imagine it shouldn't be too far off), the probability of a false-positive should be somewhere around one in four billion - low, but not impossible.

mnightingale:
I haven't looked at the implementation in detail but can or maybe it does already reduce the probability of collisions by checking multiple things?

Quick verify success = "CRC32 matches expected" and "All articles downloaded"
Otherwise do a full MD5 based check.

puzzledsab:
I think it's supposed to keep a count of bad/missing articles, @Safihre ? I was thinking that we could also compare the file size with the expected value in the PAR2. The 16KB MD5 is already checked implicitly. With all those checks the chance of not discovering an error should be negligible.

mnightingale:
> I was thinking that we could also compare the file size with the expected value in the PAR2.

I'd caution against using filesize as I think there is a good chance #2169 will land eventually :zipper_mouth_face:

animetosho:
> Quick verify success = "CRC32 matches expected" and "All articles downloaded"  
> Otherwise do a full MD5 based check.

Unfortunately you won't know when an MD5 check is required, as the CRC32 check won't ever give you a false-negative.  
Of course, if you know that there's missing articles, hash checking isn't even needed.

File size is specified in a number of places:

* NZB by summing segment sizes (I think Sabnzbd primarily gets it from here?)
* yEnc headers
* PAR2

You could do a check between all of these, along with the total number of bytes decoded (file preallocation shouldn't hinder this).

mnightingale:
> File size is specified in a number of places:
> 
> * NZB by summing segment sizes (I think Sabnzbd primarily gets it from here?)
> * yEnc headers
> * PAR2

You've  reminded me of a comment on #2396 about file offsets, where would be the best place to get the file offsets from if we wanted to do that approach?

**NZB**

```
<segment bytes="800000" number="1">...</segment>
<segment bytes="800000" number="2">...</segment>
```
Offsets derived from bytes/number

**yEnc headers**

```
=ypart begin=800001 end=1600000
```

---

I think SAB currently uses the NZB for the order to reconstruct the file so they are what should be used and perhaps it's possible that yEnc headers could lie or be constructed in a way to intentionally cause problems - such as tell the client this part is for position 1TB?

Safihre:
In my previous attempt I just checked indeed the filesize on disk, seems plenty accurate for now:
https://github.com/sabnzbd/sabnzbd/commit/2ac08dd0e6659b24e283ca2d3f9f62963d82a565#diff-4d3d76bbdc64028f5e5c1442647035ee0b14a7baef4a1f48d7caffb9e9992a6fL2120

I also kept a check of bad-articles-per-NZF:
https://github.com/sabnzbd/sabnzbd/commit/2ac08dd0e6659b24e283ca2d3f9f62963d82a565#diff-4d3d76bbdc64028f5e5c1442647035ee0b14a7baef4a1f48d7caffb9e9992a6fL2115
The reason was that some indexers add fake files to their NZB's for tracking. These files are "missing", however, they are not part of any par2 set. So even if we let par2cmdline try to repair it, it can't anyway.
So we only care for missing articles if they are part of a par2 set.

If you don't mind I would suggest to close this PR in favor of #2407 ?

PR #2407, opened 2023-01-09
CRC32 check

puzzledsab:
Same as the other PR, but to feature/buffered_reader. It needs more work but if this branch will be the new version then it's probably better if I fork it and continue after the merge.


sabnzbd/decoder.py
```diff
@@ -241,27 +247,31 @@ def decode_yenc(article: Article, raw_data: List[bytes]) -> bytes:
...
     # CRC check
     if not crc_correct:
         logging.info("CRC Error in %s", article.article)
         raise BadData(decoded_data)
 
+    article.crc32 = crc32(decoded_data)
+    # Determine part size and precalculate crc_2pow
+    if article.lowest_partnum and len(decoded_data) > nzf.nzo.article_size:
+        nzf.nzo.article_size = len(decoded_data)
+        nzf.nzo.crc32_coeff = crc_2pow(nzf.nzo.article_size * 8)
```
Safihre, line 273: Should we maybe do it on a per-NZF basis? Then for season-packs there's a much higher success ratio.

sabnzbd/decoder.py
```diff
@@ -241,27 +247,31 @@ def decode_yenc(article: Article, raw_data: List[bytes]) -> bytes:
...
         # For example when the first article was missing
-        article.nzf.nzo.verify_nzf_filename(article.nzf, yenc_filename)
+        nzf.nzo.verify_nzf_filename(nzf, yenc_filename)
 
     # CRC check
     if not crc_correct:
         logging.info("CRC Error in %s", article.article)
         raise BadData(decoded_data)
 
+    article.crc32 = crc32(decoded_data)
```
Safihre, line 269: Will you switch to using the output from `sabyenc3`? With the new branch that should work.

sabnzbd/assembler.py
```diff
@@ -207,7 +203,7 @@ def assemble(nzo: NzbObject, nzf: NzbFile, file_done: bool):
         # Final steps
         if file_done:
             set_permissions(nzf.filepath)
-            nzf.md5sum = nzf.md5.digest()
+            nzf.md5sum = nzf.crc32
```
Safihre, line 206: Can also probably be renamed to `crc32`? Or are there spots where you have doubts if it will be ok?

sabnzbd/decoder.py
puzzledsab: Considering that most nzbs are not season packs and a large share of the season packs will have content from mostly or only one poster, I think that the average number of CPU cycles used is lower when using a single precalculated value for the whole nzb. I could keep the nzo value and copy it to the nzf when it's the same but that seems overly complicated. Both are fast enough not to influence download speed for most users.

sabnzbd/newsunpack.py
```diff
@@ -1997,7 +1997,11 @@ def quick_check_set(setname: str, nzo: NzbObject) -> bool:
             # Do a simple filename based check
             if file == nzf.filename:
                 found = True
-                if (nzf.md5sum is not None) and nzf.md5sum == par2info.filehash:
+                if (
+                    (nzf.crc32sum is not None)
+                    and nzf.crc32sum == par2info.filehash
+                    and os.path.getsize(nzf.filepath) == par2info.filesize
```
Safihre, line 2003: The file can be gone from the disk (due to a previous rename if obfosufated files appear in multiple par2 sets), so then this would crash. That's why the old code is wrapped in Try/except IOError: https://github.com/sabnzbd/sabnzbd/commit/2ac08dd0e6659b24e283ca2d3f9f62963d82a565#diff-4d3d76bbdc64028f5e5c1442647035ee0b14a7baef4a1f48d7caffb9e9992a6fL2120

Safihre:
Exciting stuff, thanks a lot @puzzledsab for making it work!

PR #2408, opened 2023-01-09
Improve sample detection using filesize and extension

### LINKED ISSUE #2083: Shows/Movies with `sample` or `proof` in the name get ignored
Based on https://forums.sabnzbd.org/viewtopic.php?p=126985

Show name is: `The.Moment.of.Proof.S01E01.720p.HEVC.x265-MeGusta`
Because of the use of `is_sample`, the Sorter ignores all files of this job.

I think we can be a bit smarter: if `is_sample(NameOfTheDownload) == true`, then we allow samples.

Also @jcfp, do we need the `to_filepath` and `is_full_path` function really? Aren't the file paths always full-paths?
Probably missing something.

mnightingale:
Fixes #2083 (mostly)

I've taken inspiration from a script I've used in NZBGet for many years [DeleteSamples.py](https://github.com/clinton-hall/GetScripts/blob/master/DeleteSamples.py)

Before the current filename check I first check against a size cut off (100 MB - reduced from inspiration because files are getting smaller) and limit it to known sample extensions - some of these were based on the existing tests, I probably wouldn't have bothered with if not for them.

I removed `test_is_sample_known_false_positives` and moved them into test_is_sample but I think the tests should be simplified to testing the bounds of each condition (size, extension and name) - which I can do but I'd rather know if this is likely to be accepted.

I think there are probably still possibilities for false positive, but this will reduce the chance of them.

Perhaps I should include the suggested check if NameOfTheDownload contains sample|proof always return False?

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -44,7 +44,7 @@ from sabnzbd.constants import DEFAULT_PRIORITY, MEBI, DEF_ARTICLE_CACHE_DEFAULT,
 import sabnzbd.config as config
 import sabnzbd.cfg as cfg
 from sabnzbd.encoding import ubtou, platform_btou
-from sabnzbd.filesystem import userxbit, make_script_path, remove_file, is_valid_script
+from sabnzbd.filesystem import userxbit, make_script_path, remove_file, is_valid_script, get_ext
 
 if sabnzbd.WIN32:
     try:
@@ -68,11 +68,16 @@ TAB_UNITS = ("", "K", "M", "G", "T", "P")
 RE_UNITS = re.compile(r"(\d+\.*\d*)\s*([KMGTP]?)", re.I)
 RE_VERSION = re.compile(r"(\d+)\.(\d+)\.(\d+)([a-zA-Z]*)(\d*)")
 RE_SAMPLE = re.compile(r"((^|[\W_])(sample|proof))", re.I)  # something-sample or something-proof
+RE_SAMPLE_EXTENSIONS = re.compile(
+    r"^(\.mkv|\.avi|\.divx|\.xvid|\.mov|\.wmv|\.mp4|\.mpg|\.mpeg|\.ogg|\.jpg|\.jpeg|\.png|\.par2)$", re.I
+)
 RE_IP4 = re.compile(r"inet\s+(addr:\s*)?(\d+\.\d+\.\d+\.\d+)")
 RE_IP6 = re.compile(r"inet6\s+(addr:\s*)?([0-9a-f:]+)", re.I)
 
 # Check if strings are defined for AM and PM
 HAVE_AMPM = bool(time.strftime("%p"))
+
+SAMPLE_MAX_SIZE = 100 * MEBI
 
 
 def helpful_warning(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


mnightingale:
Well test_sorting.py is kinda complicated, how to resolve?

- Add a filesize for every file - before they just test against name
- Give is_sample a really large default size or None so it always return False
- Something else?

sabnzbd/misc.py
```diff
@@ -68,12 +68,17 @@
 RE_UNITS = re.compile(r"(\d+\.*\d*)\s*([KMGTP]?)", re.I)
 RE_VERSION = re.compile(r"(\d+)\.(\d+)\.(\d+)([a-zA-Z]*)(\d*)")
 RE_SAMPLE = re.compile(r"((^|[\W_])(sample|proof))", re.I)  # something-sample or something-proof
+RE_SAMPLE_EXTENSIONS = re.compile(
+    r"^(\.mkv|\.avi|\.divx|\.xvid|\.mov|\.wmv|\.mp4|\.mpg|\.mpeg|\.ogg|\.jpg|\.jpeg|\.png|\.par2)$", re.I
```
thezoggy, line 72: list of media files (does not include metadata files like nfo/png/jpg   or things like subtitles):
https://github.com/Radarr/Radarr/blob/develop/src/NzbDrone.Core/MediaFiles/MediaFileExtensions.cs

subtitles:
https://github.com/Radarr/Radarr/blob/develop/src/NzbDrone.Core/Extras/Subtitles/SubtitleFileExtensions.cs

can see this list can get pretty long..

thezoggy:
using the name is fine until its not. having a guard to just note delete the only media file, or delete if its under 100 MB is probably good enough of an improvement. and practically samples are only generally for media files not other stuff (unless badly named).  so having it only do it for mkv/avi/etc is fine.

now 'proof' is the other part of this, which you will see an image like jpg/jpng/etc of a scan of a disk or something. but honestly i dont think there is any reason why those should be nuked.

mnightingale:
Thanks for the review, I'll take a look at how sonarr/radarr detect samples and see if we could use ffprobe-python, biggest challenge is probably how you bundle that for all the platforms.

postproc would need changes because it wouldn't be able to skip since it would need a valid file to send to ffprobe.

I do agree that is_samples should only affect media files.

thezoggy:
> Thanks for the review, I'll take a look at how sonarr/radarr detect samples and see if we could use ffprobe-python, biggest challenge is probably how you bundle that for all the platforms.
> 
> postproc would need changes because it wouldn't be able to skip since it would need a valid file to send to ffprobe.
> 
> I do agree that is_samples should only affect media files.

ideally we would just do it like we do other 3rd party utilities (unrar/par/etc) for specific binaries (mac/win) but for other platforms we expect the package manager to install it.

Safihre:
I agree with @thezoggy's review. The list of extensions could probably be endless and the filesize is also a bit arbitrary..
We sort of given up on making really good Sample Detection, because that would require external tools which seems too much for a small option like this. Especially in the age of Sonarr/Radarr taking care of all these things.

mnightingale:
I agree I don't think it's worth the effort or bloat/maintenance from adding external tool dependencies which is the only way it could be achieved accurately.

It's probably quite a rare problem with such wide use of automation.

Perhaps consider removing the feature?

PR #2409, opened 2023-01-10
Call getsize in try and use CRC32 from sabyenc

puzzledsab:
I think it should be done after this commit. If you like I can remove article.decoded and use article.crc32 to signal status instead (None = not decoded, some value = decoded).

I looked into getting actual size on sparse files but I didn't find a good solution that works on all platforms. I guess we'll have to burn that bridge when we get there.

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -270,7 +270,7 @@ def decode_yenc(article: Article, raw_data: List[bytes]) -> bytes:
         logging.info("CRC Error in %s", article.article)
         raise BadData(decoded_data)
 
-    article.crc32 = crc32(decoded_data)
+    article.crc32 = crc_correct
     # Determine part size and precalculate crc_2pow
     if article.lowest_partnum and len(decoded_data) > nzf.nzo.article_size:
         nzf.nzo.article_size = len(decoded_data)
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -121,6 +121,14 @@ def is_writable(path: str) -> bool:
         return True
 
 
+def is_size(filepath: str, size: int) -> bool:
+    """Return True if filepath exists and is specified size"""
+    try:
+        return os.path.getsize(filepath) == size
+    except:
+        return False
+
+
 _DEVICES = (
     "con",
     "prn",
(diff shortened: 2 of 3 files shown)


mnightingale:
Not really part of the PR but I went poking around.

https://github.com/sabnzbd/sabnzbd/blob/c8dbf7e6e1b597f02f45f8f532650ac000b96c5d/sabnzbd/nzbstuff.py#L334

This should be Optional[int] right?

---

Relevant to this PR, I think CRC32=0 should be allowed and check for None here.

If I understand the other discussion correctly you just need [this](https://github.com/sabnzbd/sabyenc/blob/5b868ac87c7e1987d1c5de5b7231a4f46c6fbccf/src/sabyenc3.cc#L345-L347) to return Py_None?

Safihre:
Updated, it now indeed returns `None`. However, SABnzbd code still needs to be adjusted.
[42240683354c7b049fe3a490d8ed14eddb00851a](https://github.com/sabnzbd/sabyenc/commit/42240683354c7b049fe3a490d8ed14eddb00851a)

mnightingale:
> Updated, it now indeed returns `None`. However, SABnzbd code still needs to be adjusted. [42240683354c7b049fe3a490d8ed14eddb00851a](https://github.com/sabnzbd/sabyenc/commit/42240683354c7b049fe3a490d8ed14eddb00851a)

Isn't that mostly the same since `NULL == 0` will always be true?

I think you'd need it to be a PyObject, even default it to Py_None then when constructing the tuple check if it == Py_None and Py_INCREF.

@animetosho posted `python3 -c "import zlib; print(zlib.crc32(b'\x9d\n\xd9m'))"` which you can use to test

Safihre:
I had to make it an PyObject so it could be either a number or None, otherwise it wouldn't work.

mnightingale:
Oops I misread the lines since there are two very similar, one with uint and the other PyObject, looks good then.

animetosho:
Actually, [this](https://github.com/sabnzbd/sabyenc/blob/b4841abda95b849f02fd00e9e0830f821a211ba9/src/sabyenc3.cc#L311) and [this](https://github.com/sabnzbd/sabyenc/blob/b4841abda95b849f02fd00e9e0830f821a211ba9/src/sabyenc3.cc#L555) probably doesn't work the way you're hoping it to.

`NULL` in C is basically an alias for 0, so you're returning `Py_None` if the CRC is 0.  
You'll need an extra variable if you want to infer validity. You could do this with a 'valid' flag, or just send back both CRCs from `decode_buffer_usenet` and do the validity check in `decode_usenet_chunks` instead.

Safihre:
@animetosho oops, thanks! Will fix.

PR #2411, opened 2023-01-11
Stop using 0 as failed and use new crc32 value in SFV check

puzzledsab:
And fix declaration `self.crc32sum: Optional[int]`. I'm now setting nzf.crc32sum to -1 on invalid CRC because None means not finished writing (for direct unpack). It's a bit hackish so I'm open for other suggestions. I didn't want to set a value that could theoretically be correct if the decoder detects bad CRCs. If there are other ways of forcing a par2 check then maybe we should use that.

I tested the SFV update with an nzb which contained an sfv. When I removed the par2 files it correctly identified the files with the same CRC32 string.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -203,7 +203,10 @@ class Assembler(Thread):
         # Final steps
         if file_done:
             set_permissions(nzf.filepath)
-            nzf.crc32sum = nzf.crc32
+            if nzf.crc32 is None:
+                nzf.crc32sum = -1
+            else:
+                nzf.crc32sum = nzf.crc32
 
     @staticmethod
     def check_encrypted_and_unwanted(nzo: NzbObject, nzf: NzbFile):
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -203,10 +203,7 @@ class Assembler(Thread):
         # Final steps
         if file_done:
             set_permissions(nzf.filepath)
-            if nzf.crc32 is None:
-                nzf.crc32sum = -1
-            else:
-                nzf.crc32sum = nzf.crc32
+            nzf.assembled = True
 
     @staticmethod
     def check_encrypted_and_unwanted(nzo: NzbObject, nzf: NzbFile):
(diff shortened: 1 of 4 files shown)


Safihre:
I'm not against introducing new attributes, we don't have to infere from other attributes. Seems also bit weird that we have crc32 and crc32sum, they contain the same value in the end anyway?

puzzledsab:
It's true, I don't think it's necessary anymore now that crc32 is a real value and not a hashlib object. I'll look into replacing crc32sum with some sort of finished assembling flag and switch to reading the value from article.crc32.

thezoggy:
Will sab users need to have empty queue to prevent issues with the mf5->crc change.. what if a job started.. then sab users upgraded. Would the job just blow up

Safihre:
That's a good point, as assembled will not be set for already downloaded files of half downloaded jobs.

puzzledsab:
The CRC32 change won't be compatible with existing jobs so I would recommend telling users to upgrade with an empty queue. If the user does it anyway I think the new variables will be initialized to None. For CRC32 that should be OK, it will make the par2 check trigger. I'm not entirely sure with assembled. It is only used by direct unpack and will be interpreted as not assembled. md5sum is removed so we can't check what the status was. I don't know if it's better to make it believe that all the files are there or that none of the files are there.

Safihre:
Easy to test, start a download on develop. Pause it. Switch to new branch and continue.
Maybe it's not a problem at all, I don't think it should be. We might get extra par2 verification run (not an issue), also if users downgrade and the md5sum doesn't exist so the same thing would happen.

puzzledsab:
It seems to work. No crashing and the rar is extracted when the download has finished.

Safihre:
Great!

Safihre:
I did get a crash:
```
2023-01-15 12:41:15,461::INFO::[decoder:228] Traceback: 
  File "C:\Users\x\Documents\GitHub\sabnzbd\sabnzbd\decoder.py", line 154, in run
    decoded_data = decode_yenc(article, raw_data, raw_data_size)
  File "C:\Users\x\Documents\GitHub\sabnzbd\sabnzbd\decoder.py", line 271, in decode_yenc
    if article.lowest_partnum and len(data) > nzf.nzo.article_size:
TypeError: '>' not supported between instances of 'int' and 'NoneType'
```
Will fix in my buffer-sabyenc3 changes.

PR #2414, opened 2023-01-13
Use synoacltool to remove xbits on Synology

mreid-tt:
This is a further development on #2401 based on a https://github.com/sabnzbd/sabnzbd/pull/2401#issuecomment-1381743990 where a user wanted to ensure that the ACL bits are not stripped on Synology platforms. Using the same example on DSM7:

```
Sample archive contents:
Archive: ./rar_files/Rihanna - Loud.part1.rar
Details: RAR 4, volume, lock

 Attributes      Size    Packed Ratio    Date    Time   Checksum  Name
----------- ---------  -------- ----- ---------- -----  --------  ----
    ..A....      1907       686  35%  2014-03-16 18:10  C4966751  00._Rihanna_-_Loud.nfo
    ..A....   9800349   9308923  94%  2010-11-14 18:29  0F544CC5  01 S&M.mp3
[---snip---]
```

The existing merged code (https://github.com/sabnzbd/sabnzbd/pull/2401/commits/4c28ef21904124f58a2c1200308cdcf57aba053a) results in the extracted files looking like this:

```
$ ls -all Rihanna\ Loud/
total 129224
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      786 Jan  4 04:26  .
drwxrwxrwx+ 1 sc-sabnzbd synocommunity       24 Jan  4 04:26  ..
-rw-rw-rw-  1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rw-rw-rw-  1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

This PR proposes additional code for Synology which results in the extracted files looking like this:

```
$ ls -all Rihanna\ Loud/
total 129224
drwxrwxrwx+ 1 sc-sabnzbd synocommunity      786 Jan 13 12:09  .
drwxrwxrwx+ 1 sc-sabnzbd synocommunity       24 Jan 13 12:09  ..
-rw-rw-rw-+ 1 sc-sabnzbd synocommunity     1907 Mar 16  2014  00._Rihanna_-_Loud.nfo
-rw-rw-rw-+ 1 sc-sabnzbd synocommunity  9800349 Nov 14  2010 '01 S&M.mp3'
[---snip---]
```

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -31,6 +31,7 @@ import time
 import fnmatch
 import stat
 import ctypes
+import subprocess
 from typing import Union, List, Tuple, Any, Dict, Optional, BinaryIO
 
 try:
@@ -611,6 +612,41 @@ def set_chmod(path: str, permissions: int, allow_failures: bool = False):
             logging.debug("Could not change permissions of %s", path)
 
 
+def syno_deny_exec(path: str, allow_failures: bool = False):
+    """Add deny executable permissions to everyone on 'path'"""
+    try:
+        logging.debug("Restricting executable permission on %s", path)
+        subprocess.check_output(
+            ["synoacltool", "-add", path, "everyone:*:deny:--x----------:----"],
+            universal_newlines=True,
+        )
+    except subprocess.CalledProcessError as e:
+        if not allow_failures and not sabnzbd.misc.match_str(path, IGNORED_FILES_AND_FOLDERS):
+            logging.error(
+                T("Cannot restrict executable permission on %s for everyone"),
+                clip_path(path),
+            )
+            logging.info("Traceback: %s", e.output)
+        else:
+            logging.debug("Could not restrict executable permission on %s for everyone", path)
+    else:
+        # Deny executable also needs to be explicitly set for owner
+        try:
+            subprocess.check_output(
+                ["synoacltool", "-add", path, "owner:*:deny:--x----------:----"],
+                universal_newlines=True,
+            )
+        except subprocess.CalledProcessError as e:
+            if not allow_failures and not sabnzbd.misc.match_str(path, IGNORED_FILES_AND_FOLDERS):
+                logging.error(
+                    T("Cannot restrict executable permission on %s for owner"),
+                    clip_path(path),
+                )
+                logging.info("Traceback: %s", e.output)
+            else:
+                logging.debug("Could not restrict executable permission on %s for owner", path)
+
+
 def set_permissions(path: str, recursive: bool = True):
     """Give folder tree and its files their proper permissions"""
     if not sabnzbd.WIN32:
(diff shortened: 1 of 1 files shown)


thezoggy:
Why not just use a post processing script to do this rather than force all sab users to do it

mreid-tt:
@jpdsc based on the concerned you mentioned in https://github.com/sabnzbd/sabnzbd/pull/2401#issuecomment-1381743990, I created this test PR which should address your concern if I understood it correctly. Do test it and provide feedback including detailed logs.

jpdsc:
@mreid-tt so I have been looking into this again.
First of all, I'm not running the SABnzbd native Synology app. Next to this, my SABnzbd is running in a Docker container on a different machine.

Between the NAS and the machine where SABnzbd is running; I mounted a NFS share which is set up with permission as below.
Now, browsing this NFS share on the machine shows permissions:
image

On the Synology NAS, the same files have permissions:
image

Now, I think the issue I'm having specifically is that SABnzbd sees the files with execute and then chmod them for security reasons as the SABnzbd dev replied.

So, I'm not sure if above fix will actually solve my issue. I think the issue I'm having is because in the machine where SABnzbd is running and NSF share is mounted, the system sees it as 777 with x.
I assume then, the only to fix this issue for is to either run SABnzbd on the Synology NAS and use the fixes you provided or have a special option to ensure that x is ignored and therefore less security and will probably not be done as already mentioned in another issue.

mreid-tt:
@jpdsc you are correct in that this fix will have no effect on your issue. This is the reason why I suggested you open a new issue with the full details of the problem so it could be correctly diagnosed.

Based on your setup you seem to have files on the Synology with permission bits of '777' which at most would change to '666' when SABnzbd completes extraction processing. This should not prevent other software from accessing these files. Remember that the execution bit is for running the file as a program and it should still be readable unless the read bit is also absent.

To get started with some troubleshooting, additional information would be needed:
1. What are the permissions of the files when you issue a `ls -el [filename]` at the Synology terminal?
2. What is the user trying to access the file? Is there a log of any errors when trying to access the file?
3. What are the shared folder permissions set to for `Local users` or `System internal user` (Control Panel > Shared Folder > Edit > Permissions)? Is the user trying to access the file allowed `Read/Write` access?
4. What happens if you issue a command like `sudo -u [username] head [filename]` to simulate the user accessing the file?

These are some high-level steps which you can try to better diagnoase the permissions issue. Looking forward to your results so we could better understand the problem.

Safihre:
Would it maybe be possible to change the umask of SABnzbd (and the processes it spawns) during runtime so all created files don't have X bits set? @jcfp do you know?

jpdsc:
@mreid-tt 
As advised I checked the permissions on one of the files in the Synology NAS which comes out as:
`-rwxrwxrwx+ 1 admin users`

This is exactly the same being shown when looking at the NFS share in the machine where SABnzbd is running.
`-rwxrwxrwx 1 1024 users`

And how the GUI shows it in the NAS:
image

So, even with your fix, it will not work for the scenario where the storage is at the Synology NAS and SABnzbd is running on a different machine with NFS share to the Synology NAS.
NFS shows the same permissions as on the NAS, however, SABnzbd does not know this is a Synology (your fix above) and therefore will apply chmod due to the X bits.

This is how it looks like with chmod 666.
NAS:
`-rw-rw-rw- 1 admin users`

Machine where SABnzbd is running with NFS share to the Synology:
`-rw-rw-rw- 1 1024 users`

Permissions in the GUI:
image

So, the only for me in my scenario to fix this, is to ensure that SABnzbd does not remove the X bit. As Safihre pointed, adding such an option in Special can have a security risk if SABnzbd is without a password.

Safihre:
> Would it maybe be possible to change the umask of SABnzbd (and the processes it spawns) during runtime so all created files don't have X bits set? @jcfp do you know?

Hmm reading more about this I guess it won't work because we don't want to prevent the execute on folders, only on files.

mreid-tt:
I am closing this commit. I have asked for specific diagnostic information to help resolve the issue but folks seem to feel they know best so I'm done trying to help.

Safihre:
I'm happy you fixed the `-ai` problem for us, that should make Synology life already a lot better.

PR #2423, opened 2023-01-24
Immediately request new article after the previous was done

puzzledsab:
An article can now be downloaded in just one call to `recv_chunk`, and doing it in 4-8 is quite common in my tests. This means that several connections will be idle every round while `recv_chunk` is called on the other connections.


Safihre:
How about a `server.get_article()` that checks if there's anything in the queue and if there's nothing, then fills it and returns the first available one?
Or modify the NzbQueue one to do that instead?
We can use that at the top as well.

PR #2424, opened 2023-01-24
Make downloader use used buffer size to determine if it's ok to sleep

puzzledsab:
I believe this method is easier to understand and has the potential to be more reliable than the current `can_be_slowed`. Basically, keep track of what the biggest chunk received so far is and sleep in the following rounds if no connection received more than 1/3 of that number. The idea being that there should be plenty room left in the connection buffer if that is true.

The starting value of 64 KiB is a bit arbitrary but it seems very unlikely that any OSes uses a smaller number than this. 1/3 is also arbitrary but feels conservative enough. `max_chunk_size` is reset when the downloader is idle (and connections are reset) because the receive buffer starts small and grows on demand. It might be a bit safer to have an upper limit but I didn't want to complicate things more than necessary.

It can still be turned off completely by setting `downloader_sleep_time` to 0.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -55,6 +55,8 @@ _SERVER_CHECK_DELAY = 0.5
 _BPSMETER_UPDATE_DELAY = 0.05
 # How many articles should be prefetched when checking the next articles?
 _ARTICLE_PREFETCH = 20
+# Minimum expected size of TCP receive buffer
+_DEFAULT_CHUNK_SIZE = 65536
 
 TIMER_LOCK = RLock()
 
@@ -564,10 +566,9 @@ class Downloader(Thread):
         BPSMeter.update()
         next_bpsmeter_update = 0
 
-        # can_be_slowed variables
-        can_be_slowed: Optional[float] = None
-        can_be_slowed_timer: float = 0.0
-        next_stable_speed_check: float = 0.0
+        # Sleep check variables
+        last_max_chunk_size: int = 0
+        max_chunk_size: int = _DEFAULT_CHUNK_SIZE
 
         # Check server expiration dates
         check_server_expiration()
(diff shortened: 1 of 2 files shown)


Safihre:
Not against this approach, just wonder how it will work on those machines that had problems previously. Those on VM's or with special VPN's, no real way to test that unfortunately unless we just release it in the wild.

puzzledsab:
I don't know. My guesses about what might be causing it:
- For some reason, only one or a few connections are filled at a time. This could make the current version sleep between each call and only fetch 16 KB each time. If this is the problem then this new version would fix it.
- A very small TCP receive buffer. Depending on how small, this may or may not fix it. Lowering `_DEFAULT_CHUNK_SIZE` could help.
- Something makes it sleep much longer than it should. If this is the problem then it will probably not fix it.

It could also be a combination of several things. Someone has to test it to find out. It would be useful to know what the `max_chunk_size` grows to. I could also add some more debugging, for instance log every 10 seconds how many times it has slept.

sabnzbd/downloader.py
```diff
@@ -695,43 +700,43 @@ def run(self):
...
-                            if BPSMeter.get_stable_speed(timespan=10):
-                                can_be_slowed_timer = now + 8
-                                can_be_slowed = 1
-                            else:
-                                next_stable_speed_check = now + _BPSMETER_UPDATE_DELAY
-
-                        # Check 10 seconds after enabling slowdown
-                        if can_be_slowed_timer and now > can_be_slowed_timer:
-                            # Now let's check if it was stable in the last 10 seconds
-                            can_be_slowed = BPSMeter.get_stable_speed(timespan=10)
```
Safihre, line 722: Can you also remove the get_stable_speed? This way we have 1 commit that has all that we can revert if needed.

PR #2428, opened 2023-01-27
Use threadpool for recv_chunk()

puzzledsab:
Inspired by the Reddit thread about SABnzbd being slow on Rpi 4, I tried calling recv_chunk using threadpool. I think it might help on Rpi 4 because it doesn't have hardware AES decoding so SSL downloading uses a lot of CPU. The current version is hard coded to 2 threads because it's just a test.

The changes needed are very small. The diff only looks big because I put the old code into two new methods to avoid duplication for the single and multithreaded variations. However, there is a lot of CPU overhead when using threadpool compared fixed threads and queues (like the decoder). Does anyone know if there are any tricks to reduce the overhead without big code changes? If so, I think it could be very helpful for some users.

One issue I did notice while testing it on an Rpi 3 was that the decoder was unable to keep up and it saved cached parts to disk. I'm not sure why, I assumed the downloader thread would start sleeping before this happened. Maybe the `decoder_cache_article_limit` should be reduced. It may also help to increase `num_simd_decoders` to let decoding get enough CPU time to keep up, but it wasn't enough to increase it to 4 in my Rpi 3 test.

@mnightingale: You talked about using threadpool in your testing, do you have any input?

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -55,8 +55,12 @@ class Assembler(Thread):
     def process(self, nzo: NzbObject, nzf: Optional[NzbFile] = None, file_done: Optional[bool] = None):
         self.queue.put((nzo, nzf, file_done))
 
-    def queue_full(self):
-        return self.queue.qsize() >= MAX_ASSEMBLER_QUEUE
+    def queue_level(self) -> float:
+        size = self.queue.qsize()
+        if size:
+            return size / MAX_ASSEMBLER_QUEUE
+        else:
+            return 0
 
     def partial_nzf_in_queue(self, nzf: NzbFile):
         return (nzf.nzo, nzf, False) in self.queue.queue
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -110,9 +110,13 @@ class Decoder:
         sabnzbd.ArticleCache.reserve_space(raw_data_size)
         self.decoder_queue.put((article, raw_data, raw_data_size))
 
-    def queue_full(self) -> bool:
-        # Check if the queue size exceeds the limits
-        return self.decoder_queue.qsize() >= sabnzbd.ArticleCache.decoder_cache_article_limit
+    def queue_level(self) -> float:
+        # Return level of decoder queue. 0 = empty, >=1 = full.
+        size = self.decoder_queue.qsize()
+        if size:
+            return size / sabnzbd.ArticleCache.decoder_cache_article_limit
+        else:
+            return 0
 
 
 class DecoderWorker(Thread):
@@ -203,17 +207,20 @@ class DecoderWorker(Thread):
                     if not article_success:
                         # Convert the initial chunks of raw socket data to article lines,
                         # and examine the headers (for precheck) or body (for download).
-                        for line in b"".join(raw_data[:2]).split(b"\r\n"):
-                            lline = line.lower()
-                            if lline.startswith(b"message-id:"):
-                                article_success = True
-                            # Look for DMCA clues (while skipping "X-" headers)
-                            if not lline.startswith(b"x-") and match_str(
-                                lline, (b"dmca", b"removed", b"cancel", b"blocked")
-                            ):
-                                article_success = False
-                                logging.info("Article removed from server (%s)", art_id)
-                                break
+                        try:
+                            for line in b"".join(raw_data[:2]).split(b"\r\n"):
+                                lline = line.lower()
+                                if lline.startswith(b"message-id:"):
+                                    article_success = True
+                                # Look for DMCA clues (while skipping "X-" headers)
+                                if not lline.startswith(b"x-") and match_str(
+                                    lline, (b"dmca", b"removed", b"cancel", b"blocked")
+                                ):
+                                    article_success = False
+                                    logging.info("Article removed from server (%s)", art_id)
+                                    break
+                        except TypeError:
+                            article_success = False
 
                 # Pre-check, proper article found so just register
                 if nzo.precheck and article_success and sabnzbd.LOG_ALL:
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -545,8 +545,16 @@ class Downloader(Thread):
 
         # See if we need to delay because the queues are full
         logged_counter = 0
-        decoder_full = sabnzbd.Decoder.queue_full()
-        assembler_full = sabnzbd.Assembler.queue_full()
+
+        decoder_level = sabnzbd.Decoder.queue_level()
+        assembler_level = sabnzbd.Assembler.queue_level()
+
+        # Sleep a short while to release the GIL if the queue is starting to fill
+        if decoder_level > 0.5 or assembler_level > 0.5:
+            time.sleep(0.01)
+
+        decoder_full = decoder_level >= 1
+        assembler_full = assembler_level >= 1
         while not self.shutdown and (decoder_full or assembler_full):
             # Only log/update once every second, to not waste any CPU-cycles
             if not logged_counter % 10:
@@ -566,8 +574,8 @@ class Downloader(Thread):
             # Wait and update the queue sizes
             time.sleep(0.1)
             logged_counter += 1
-            decoder_full = sabnzbd.Decoder.queue_full()
-            assembler_full = sabnzbd.Assembler.queue_full()
+            decoder_full = sabnzbd.Decoder.queue_level() >= 100
+            assembler_full = sabnzbd.Assembler.queue_level() >= 100
 
     def run(self):
         # First check IPv6 connectivity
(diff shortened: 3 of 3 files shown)


mnightingale:
Interesting I didn't imagine such minor changes would be required, but even on a high speed device I ended up filling the cache.

I actually think this is because of the way you've setup adding to the pool and getting it's results, I'm not sure what the purpose of your range and step when calling map are but I think you may have just limited the pool to waiting until the queued jobs are completed that you are submitting in pairs.

Changing it to just submit all the sockets, removing the range call and using imap_unordered since the order of results doesn't matter. `for nw, bytes_received, done in dl_pool.imap_unordered(self.__recv, read)` it makes a pretty big difference for me, **"Downloaded in 4 seconds at an average of 2.4 GB/s"** I think that's nearly 1GB/s faster than I've managed previously and isn't filling the article cache - although using much more than I've seen previously.

I realise what you have so far is just a POC/draft but there are a bunch of error paths/types that aren't handled, and I'd suggest seeing if the newer ThreadPoolExecutor would have any advantages such as setting a timeout.
Maybe it's just me but I'd prefer if __recv received the read_fds[index] instead of the index, I'm also not sure if a thread accessing read_fds has any penalty?

In the other discussion I hypothesised that if the downloader could use multiple threads and it doesn't even need to be very many (I can't see more than 2-4 ever being needed, but I'd limit by cores/2 or something) then the decoder could also be inlined instead of the current system which only makes sense when the downloader is single threaded and the decoder isn't, there are probably a few changes to make to sabctools to make that as efficient as possible so we can pass in the newswrapper memoryview and return a new decoded bytes/bytearray.

puzzledsab:
I was hoping that only calling 2 at a time would make the bps limiter work better. I'm not sure if it actually works properly anyway now that the whole buffer is cleaned out every time, though. 

It will look smooth on the bps meter but internally it may be very jumpy because it can receive many MB to the receive buffers while the downloader is sleeping. Then all downloading can be completely stopped for several seconds while the bps limiter is catching up with all the sleeping and the data are dumped from `recv_chunk()`. If this happens then using the bps limiter to for instance keeping the ping time down while playing games won't work.

I think maybe it needs to have two modes. If the bps limit is enabled, only one download thread can be used and it will only read a limited number of bytes for each recv call. Then there can be a full speed mode where it reads everything and uses multiple threads. In this mode it would be no problem to call map only once.

puzzledsab:
I changed it to `imap_unordered` and adding everything at once. It didn't make much difference for me. The CPU usage in the Windows task manager is still between 1-2.5% instead of 0.3-0.5%. I'm obviously not limited by CPU speed, though.

mnightingale:
From what I could see the bandwidth limit appears to work and I don't think it'd be any different to the single thread 
approach. However at a lower level I'm not sure how well it actually works with socket buffers and TCP recieve windows.

In my opinion the limitter probably works as well as it can, the data recv_chunk is reading has already been recieved it's just being buffered by the socket and the only way I'm aware of to control how often more data is sent is to manage how often that buffer is being consumed which is all the limitter is doing.

For this reason I don't think it matters if you read a maximum of 64KB or 1MB from the underlying socket, either way you're consuming the same amount of data per second so you're not actually changing how fast the server sends data to you. We'd probably have to capture packets and see what the recieve windows are to determine if limited mode should read in smaller chunks.

mnightingale:
I found my testing script which I used to experiment with, it was mostly to try and identify what may be possible in SAB if the bottlenecks were removed, a few threads in a ThreadPoolExecutor certainly looked the best approach.

| Connections  | Threads | Only Downloading  | Downloading & Decoding |
|---|---|---|---|
| 8 | 1 | 1850 | 1600 |
| 8 | 2 | 3700 | 1580 |
| 16 | 1 | 1850 | 1540 |
| 16 | 2 | 3550 | 1600 | 
| 16 | 4 | 5550 | 2300 |

I managed to hook Go up for decoding/encoding a little while ago and it could do 4350 MB/s, therefore I want to investigate why I can't with a similar approach in Python but I wouldn't be surprised if it's just because of the GIL.

---

I think before this approach with multiple download threads is used we do need to investigate the download and decoder behaviour, downloading faster than can be decoded and being forced to write to disk is not desirable, I think it would be better to pause downloading until there is space available.

Possibly removing the decoder threads could help, it's less threads for Python to manage, probably less complex, but it probably just moves the bottleneck to the assembler so I think downloads should still be paused if the cache is full.

I want to get a RPi 4 so I can check some of these theories on slower hardware but had no luck with stock so far 😢

puzzledsab:
I noticed from the debug printing in #2424 that the granularity in the thread switching is not always very fine. It sometimes sleeps longer than it should. I've set it to 0.002 seconds but I've seen it sleep up to 0.13 seconds. It may be helpful to make sure each thread is running for a longer time so that the latency has less impact. Joining downloading and decoding is one solution but it could also help to increasing the number of threads.

I did some other testing with >100 threads doing md5 calculations. The performance didn't suffer noticeably compared to 8 threads on a 4 core CPU so the number of threads is not necessarily a problem.

Safihre:
Interesting approach, different than making the whole Downloader multi threaded.
Is that what you prose @mnightingale?
Because there's quite a lot of bookkeeping as you can see, much more than just a simple test script. Bit of a pain to make that work multi threaded with racing problems when dealing with server errors for example.

mnightingale:
> Interesting approach, different than making the whole Downloader multi threaded. Is that what you prose @mnightingale? Because there's quite a lot of bookkeeping as you can see, much more than just a simple test script. Bit of a pain to make that work multi threaded with racing problems when dealing with server errors for example.

I'd never considered doing it this way round, I had in mind that a thread per connection would be the simplest followed by grouping connections - but I don't really like such approach as there would be added complexity.
However I concluded that the amount of work restructuring how the downloader/newswrapper works just wasn't worth it.

I have just tried making each newswrapper a thread, it basically involves moving a lot of downloader to newswrapper and I missed out most of what wasn't essential (error handling, resets, etc.) and I'm not certain of the requirements for thread safety in Python.
However I found it to be more or less the same if not a bit slower than this approach - I haven't managed to profile it to find out why.

This approach does appear to improve efficiency, but I think because there is only one select and we have to wait until the pool has finished you still end up with sockets idle for longer than other approaches. I did notice on my system increasing the threads past 2 didn't make any difference.
On the other hand it does seem to have increased speed so that I'm actually limited by the decoder, so I do want to investigate why that might be.

I think this probably is an improvement worth merging since any other approach requires significant changes and maybe even replacing polling with events/observers.

puzzledsab:
My understanding is that imap_unordered will start returning results once they start becoming available, not wait until all are done. Is that not the case?

@mnightingale Have you tried changing the queue constants? LIMIT_DECODE_QUEUE = 100 seems high to me and MAX_ASSEMBLER_QUEUE = 5 seems low. I my last commit to this branch I changed it so that it will sleep 0.01 second if one of the queues is half full. I'm hoping that releasing the GIL will let the decoder and assembler start new processes more frequently and keep the cache from filling up.

Another alternative I thought about was to set a flag when it's half full and start sleeping 0.002 seconds after every recv call while it's true.

mnightingale:
> My understanding is that imap_unordered will start returning results once they start becoming available, not wait until all are done. Is that not the case?

I was referring to that we are limited to the frequency that the downloader thread loops.
For instance if you had 8 connections ready to read then the threadpool allows reading of 2 concurrently (because they release the GIL for the I/O bound part, but you still have to wait until all the `imap_unordered` results come in until you start reading from those earlier ones again.
Whereas a true multi-threaded approach would allow each to read more often.

I really don't like using time.sleep to encourage other threads to run, it really comes back to that a lot of SAB uses polling, using threading.Event would be better as I understand it because Python wouldn't need to schedule the thread again while it's waiting.

puzzledsab:
I think I may be missing something. `unlocked_ssl_recv_into` doesn't actually wait for data. It takes what there is and returns. If all the connections can poll instantly once there seem to be data available then I assume they will return little or no data every time (in normal use cases), leaving less time for other threads to run and causing higher CPU usage.

That result then (at least currently) has to be handled in Python, which generally takes the same amount of GIL time no matter how much data was returned. Sleeping both releases the GIL itself, makes it give up the GIL for longer periods while calling `unlocked_ssl_recv_into` and spends less GIL time on handling results.

mnightingale:
I was describing how I think this is different from a multithreaded approach, that the single thread of the downloader is the one calling `select.select` which is blocking until at least one socket is ready for reading for up to 1 second, then submitting work for each readable socket.

I think you're likely to find `select.select` returns when there are very few sockets ready which would explain why when I added more threads it made no difference.

As a stop-gap approach I think this is fine, I was just highlighting where I think this falls short compared to each socket having it's own thread doing the blocking.

puzzledsab:
I'll clean this up and resubmit.

PR #2431, opened 2023-01-29
Run assemble loop at most once per second for each file

puzzledsab:
As talked about previously. Not sure if maybe the limits should be changed a bit. LIMIT_DECODE_QUEUE = 100 seems very high because the decoding speed will generally be very smooth. MAX_ASSEMBLER_QUEUE is only 5, but the speed will vary a lot. Sometimes it will do nothing and sometimes it may have to write a full file before it returns and starts clearing the queue again.

I have added a variation that starts sleeping 0.01 second during decoding if one of the queues is half full in the threadpool PR. I'm hoping it will help the decoder and assembler to keep up with minimal impact on the download speed.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -170,11 +170,6 @@ class Assembler(Thread):
         1) Partial write: write what we have
         2) Nothing written before: write all
         """
-        # Run at most once per second unless the file is finished.
-        if nzf.next_assemble_time > time.time() and not file_done:
-            return
-        nzf.next_assemble_time = time.time() + 1
-
         # We write large article-sized chunks, so we can safely skip the buffering of Python
         with open(nzf.filepath, "ab", buffering=0) as fout:
             for article in nzf.decodetable:
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -23,6 +23,7 @@ import os
 import queue
 import logging
 import re
+import time
 from threading import Thread
 import ctypes
 from typing import Tuple, Optional, List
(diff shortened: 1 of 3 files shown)


sabnzbd/assembler.py
```diff
@@ -169,6 +170,10 @@ def assemble(nzo: NzbObject, nzf: NzbFile, file_done: bool):
         1) Partial write: write what we have
         2) Nothing written before: write all
         """
+        # Run at most once per second unless the file is finished.
+        if nzf.next_assemble_time > time.time() and not file_done:
+            return
+        nzf.next_assemble_time = time.time() + 1
```
Safihre, line 176: Shouldn't this be much better here? So it never even enters the queue?
https://github.com/sabnzbd/sabnzbd/blob/5e58fdf82141fcec6b7b379c957909d82aae3e51/sabnzbd/nzbqueue.py#L758-L765

puzzledsab: I didn't realize that things were put in the assembler queue so selectively. Now I see why 5 may be enough. However, doesn't that also mean that the timing between when it's put in the queue and when it's saved can be pretty skewed? Maybe it doesn't really matter.

puzzledsab:
I no longer think this is a good implementation because it doesn't scale with speed. One solution to that is to instead add an `nzf.saved_bytes` variable where the bytes of each article is added in `register_article`. It would only add it to the queue if it's over a certain threshold, for instance 20 MB, and then it's reset when an article is added.

Safihre:
But isn't that exactly what the current implementation does? Only trigger at a certain number of bytes? It's just a bit less correct because we assume an article size of ~750Kb.

puzzledsab:
Pretty much, I guess. Maybe just make DIRECT_WRITE_TRIGGER configurable as a special variable or let it scale with cache size? Those 4 MB articles can mess it up a bit so the scaling would have to be fairly conservative.

Issue #2425, opened 2023-01-25
Watched folder polling CPU usage


mnightingale:
I noticed a [reddit thread ](https://www.reddit.com/r/usenet/comments/10kddup/sabnzbd_much_slower_than_nzbget_on_my_setup_why/) earlier and a couple of people commented that the watched folder feature causes significant CPU usage even when it's not finding any changes.

By default SAB appears to poll the directory every 5 seconds, reading and comparing the directory tree.
I'm not familiar with the equivalent for all platforms but I have used inotify on linux in the past to monitor for changes without polling.

There does appear to be a Python package [watchdog](https://github.com/gorakhargosh/watchdog) which seems simple to setup, basically create a handler for FileSystemEventHandler events and start monitoring for "created" events.
It would probably also result in a simplification of a lot of the dirscanner logic.

Just wanted to raise awareness of the issue, I may take a look at it myself unless anyone else fancies investigating it.

sanderjo:
I don't know about CPU usage by current polling, but watchdog is cool!

With the example program .... watching my home dir just works:

```
<ObservedWatch: path=., is_recursive=True>
2023-01-25 13:19:40 - Modified file: ./ddddd.txt
2023-01-25 13:19:40 - Modified directory: .
2023-01-25 13:19:40 - Modified file: ./.cache/tracker/meta.db-wal
2023-01-25 13:19:40 - Modified file: ./.cache/tracker/meta.db

2023-01-25 13:19:51 - Modified file: ./ddddd.txt
2023-01-25 13:19:51 - Modified directory: .
2023-01-25 13:19:51 - Modified file: ./.cache/tracker/meta.db-wal
2023-01-25 13:19:51 - Modified file: ./.local/share/tracker/data/tracker-store.journal
2023-01-25 13:19:51 - Modified file: ./.cache/tracker/meta.db

2023-01-25 13:19:55 - Created file: ./dddddaaaaaaaaaa.txt
2023-01-25 13:19:55 - Modified directory: .
2023-01-25 13:19:55 - Modified file: ./dddddaaaaaaaaaa.txt
2023-01-25 13:19:55 - Modified directory: .
2023-01-25 13:19:55 - Modified file: ./.cache/tracker/meta.db-wal
2023-01-25 13:19:55 - Modified file: ./.local/share/tracker/data/tracker-store.journal
2023-01-25 13:19:55 - Modified file: ./.cache/tracker/meta.db
[...]

PR #2434, opened 2023-01-30
Improve dirscanner performance and reduce system calls

mnightingale:
This is related to the discussion on #2425 where it was identified that even if filesystem events could be used there is probably still a requirement for some users to use polling such as if their watched directory is on a remote network share.
Additionally a poll/scan would still be required on startup or whenever the directory changes.

Therefore to start off with I've looked at optimising the over 13 year old code to take advantage of some Python 3 features, reduce system calls and improve import speed.

Changes include:
- Added a few tests 🥳
  - This required a minor change to testhelper so I could inject pytest fixtures when using the set_config decorator
  - I'm not sure what you'll make of how I setup the tests, I wanted to black box test it because its internals will change a bit once I add events, so I have a bit of an odd thread wrapper so I can test files already being in the watched folder at start and those added after the startup scan cycle.
  - I've mocked a few things because I just wanted to know if add_nzbfile was being called correctly, some were necessary mocks because I'm not loading all of SAB.
- On to the fun stuff, I've replaced `os.listdir` with `os.scandir`, this won't have existed 13 years ago but it has a few advantages; it uses an iterator of `os.DirEntry` so reduced memory usage and populates `is_dir` and `is_file` usually without an additional system call. Windows doesn't populate `st_ino` in `entry.stat()` so it has to use `os.stat`.
- Since it's an iterator this change is what makes the majority of the other logic changes necessary and why I've split the code into a couple of functions and use recursion for the `catdirs` - fewer system calls.
- Previously for each possible path the code would sleep for at least one second before calling `os.stat` and comparing to the previous result, this is to ensure that we don't try and add files that are changing such as those still downloading or copying. This has the consequence that a watched directory with 60 items takes a minimum of 60 seconds to import! 
  - Therefore I've stuck a ThreadPoolExecutor in there so we can sleep for several of them at once, I've left it as the default max_workers (cpu_count + 4) which I think is probably fine, if anything it could probably be larger since the method basically just sleeps for up to 3 seconds - maybe asyncio would be better fit but I'm not familiar with it or how to use it in an existing non asyncio app
  - I've also used `loop_condition` with a wait of 1 second instead of `time.sleep` so it can return quicker on shutdown
- Minor change to `clean_file_list` to be more efficient using the sets difference method
- I wasn't sure why there are nested functions, I added a couple of new ones but didn't nest them and prefixed with `__` - not sure if there is a different style which would be preferred

### DIFF
--- a/sabnzbd/dirscanner.py
+++ b/sabnzbd/dirscanner.py
@@ -235,7 +235,6 @@ class DirScanner(threading.Thread):
                 files.add(path)
 
                 if path in self.ignored or path in self.suspected:
-                    await asyncio.sleep(0)
                     continue
 
                 if stat_tuple.st_size > 0:
--- a/sabnzbd/dirscanner.py
+++ b/sabnzbd/dirscanner.py
@@ -179,6 +179,7 @@ class DirScanner(threading.Thread):
         except:
             if not self.error_reported and not catdir:
                 logging.error(T("Cannot read Watched Folder %s"), filesystem.clip_path(folder))
+                logging.info("Traceback: ", exc_info=True)
                 self.error_reported = True
 
     async def when_stable_add_nzbfile(self, path: str, catdir: Optional[str], stat_tuple: os.stat_result):
--- a/sabnzbd/dirscanner.py
+++ b/sabnzbd/dirscanner.py
@@ -23,12 +23,15 @@ import os
 import time
 import logging
 import threading
+from concurrent.futures import ThreadPoolExecutor
 
 import sabnzbd
 from sabnzbd.constants import SCAN_FILE_NAME, VALID_ARCHIVES, VALID_NZB_FILES
 import sabnzbd.filesystem as filesystem
 import sabnzbd.config as config
 import sabnzbd.cfg as cfg
+
+VALID_EXTENSIONS = set(VALID_NZB_FILES + VALID_ARCHIVES)
 
 
 def compare_stat_tuple(tup1, tup2):
@@ -44,18 +47,10 @@ def compare_stat_tuple(tup1, tup2):
     return True
 
 
-def clean_file_list(inp_list, folder, files):
+def clean_file_list(inp_list, files):
     """Remove elements of "inp_list" not found in "files" """
-    for path in sorted(inp_list):
-        fld, name = os.path.split(path)
-        if fld == folder:
-            present = False
-            for name in files:
-                if os.path.join(folder, name) == path:
-                    present = True
-                    break
-            if not present:
-                del inp_list[path]
+    for path in set(inp_list.keys()).difference(files):
+        del inp_list[path]
 
 
 class DirScanner(threading.Thread):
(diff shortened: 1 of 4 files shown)


sabnzbd/dirscanner.py
```diff
@@ -124,90 +119,116 @@ def run(self):
             if self.dirscan_speed and not self.shutdown:
                 self.scan()
 
-    def scan(self):
-        """Do one scan of the watched folder"""
+    def __listfiles(self, folder, catdir=None):
```
Safihre, line 122: We don't really use the `__` prefix within SABnzbd. I know that it implies privateness in Python, but I'd like to avoid it.
Just a small nitpick.

sabnzbd/dirscanner.py
```diff
@@ -124,90 +119,116 @@ def run(self):
...
-
-                if filesystem.get_ext(path) in VALID_NZB_FILES + VALID_ARCHIVES:
-                    try:
-                        stat_tuple = os.stat(path)
-                    except OSError:
-                        continue
-                else:
-                    self.ignored[path] = 1
-                    continue
+        cats = config.get_categories() if catdir is None else {}
```
Safihre, line 125: Not all oneliners improve readability, let's just use the multi-line here.

sabnzbd/dirscanner.py
```diff
@@ -124,90 +119,116 @@ def run(self):
...
+                        yield path, catdir, None
                         continue
-                    else:
-                        del self.suspected[path]
 
-                if stat_tuple.st_size > 0:
-                    logging.info("Trying to import %s", path)
+                    # If the entry is a catdir then recursion
+                    if entry.is_dir():
+                        if not catdir and entry.name.lower() in cats:
```
Safihre, line 142: Surprised this hasn't caused problems for users, as far as I know the values of `cats` are not guaranteed to be lowercase.. But the old code also doesn't seem to care so I guess it's fine..

sabnzbd/dirscanner.py
```diff
@@ -124,90 +119,116 @@ def run(self):
...
+        else:
+            self.ignored[path] = 1
+
+    def scan(self):
+        """Do one scan of the watched folder"""
+
+        def run_dir(folder):
+            files = set()
+
+            with ThreadPoolExecutor() as executor:
```
Safihre, line 211: Does this not spawn new threads on every scan?
Is it worth it to just have a pool open and use that continuously? Or does it not really matter maybe?

tests/testhelper.py
```diff
@@ -71,6 +72,7 @@ def set_config(settings_dict):
     """Change config-values on the fly, per test"""
 
     def set_config_decorator(func):
+        @wraps(func)
```
Safihre, line 75: Why is this needed?

Safihre (changes requested):
Could you also add typing to all functions?

I do wonder if this change really will make a big difference for users that experience problems?

sabnzbd/dirscanner.py
mnightingale: It won't create threads until jobs are submitted but does have the overhead of creating the executor instance each loop.

I think a much better solution would be to use asyncio, monitoring the paths until stable would be in a single thread and it's managed the executor or a custom restricted one calling the synchronous code which adds them to the queue.
Any objections if I put together that kind of approach? I think it would actually fit in quite nicely to when I add the event based part.

tests/testhelper.py
mnightingale: As I understand it if its not there then decorators break introspection, which pytest uses to determine which fixtures to pass in.[^1]

```py
@set_config({"dirscan_dir": os.path.join(SAB_CACHE_DIR, "watched")})
def test_adds_valid_nzbs_on_startup(self, fs, mocker, dirscanner):
```

So without it the above sees the wrapped function and not the fixtures I need `fs, mocker, dirscanner`.

[^1]: Effective Python - Item 26

sabnzbd/dirscanner.py
Safihre: Sounds good to try asyncio!

mnightingale:
> I do wonder if this change really will make a big difference for users that experience problems?

I'll push the asyncio changes sometime today, for now I've possibly noticed why the current implementation could sometimes cause issues:

In `__init__` dirscan_speed is set to the users preference but 0 is set to None so that the wait in the main loop never times out resulting in no scanning as intended.

https://github.com/sabnzbd/sabnzbd/blob/90a4898dbde63928e7837fa23a1c3d828db9534f/sabnzbd/dirscanner.py#L86

But if the user than changes the directory, the same behaviour isn't applied so the wait timeout could be 0 and it'll start constantly scanning with no delay.

https://github.com/sabnzbd/sabnzbd/blob/90a4898dbde63928e7837fa23a1c3d828db9534f/sabnzbd/dirscanner.py#L91-L96

I noticed because I was removing it as I handle it differently with asyncio.

---

Apart from the above from a users point of view currently there will be minimal changes:
- The biggest change for users is it's more responsive because multiple files can be added concurrently.
- The use of generators means we won't load the full directory listing into memory and it takes care of some system calls (is_dir and stats) during it's listing

I think asynicio has made it much simpler to add the event based system which for most users will eliminate the need for polling entirely.
I'll see what you think when I push the next lot of changes but I may just add the event based system to this PR.

sabnzbd/dirscanner.py
```diff
@@ -115,99 +109,169 @@ def save(self):
...
+
+    async def shutdown(self):
+        """Cancel all tasks and stop the loop"""
+        loop = asyncio.get_event_loop()
+
+        # Get all tasks except for this one
+        tasks = filter(lambda task: task is not asyncio.current_task(), asyncio.all_tasks())
+
+        # Cancel them all
+        [task.cancel() for task in tasks]
```
mnightingale, line 272: I used filter instead of list comprehension, it's still a oneliner but I think easier to understand?

The second could be a for loop if you'd prefer, but that one seems simpler?

sabnzbd/dirscanner.py
```diff
@@ -115,99 +109,169 @@ def save(self):
...
+            if not (dirscan_speed := self.dirscan_speed):
+                break
+
+            if not (dirscan_dir := self.dirscan_dir):
+                break
+
+            try:
+                await self.scan_async(dirscan_dir)
+            except:
+                pass
```
mnightingale, line 260: I wasn't sure what to do with this, the thread never ends because of `loop.run_forever` so I need we need to be certain an unhandled exception (it catches OSError) won't stop the next scan.

There's also some code 'somewhere' that checks `is_alive` and restarts the thread (although I'm not sure if it actaully does) but again as far as I know that'll never happen.

Perhaps it should at least log it? logging.info or logging.debug?

mnightingale (commented):
@Safihre a couple of parts I'm interested in your opinion with.

mnightingale:
I've made the requested changes and updated it to use asyncio, it still creates its own thread but it would be a seperate task to update the main thread to be responsible for creating the loop used throughout the application. It's probably worth doing at some point then gradually threads can be replaced with asyncio if that's the direction you want to go in.

It would likely involve replacing the main() loop `while not sabnzbd.SABSTOP` with coroutines as there would be a `loop.run_forever()` there which blocks and we'd need a variable somewhere to allow other threads to get a reference to the main loop - like I do in scan for the tray icon.

Please let me know if you're happy with the changes, I'll probably want to add the event based system via watchdog before this is merged.

sabnzbd/dirscanner.py
Safihre: Second one seems better as a loop.

Safihre:
Look nice with asyncio!
No big comments so far.
So wait for the event based update?

mnightingale:
Asyncio does make it much easier to follow, it removes a lot of the boilerplate like needing to check for shutdown meaning the functions focus on just what they are responsible for. I also tried to reduce if nesting were reasonable and used a few walrus operators.
Eventually the new boilerplate like setting up the loop and handling shutdown could be moved up the chain.

There's just [one comment](https://github.com/sabnzbd/sabnzbd/pull/2434#discussion_r1103826049), it must not terminate the task, OSErrors are caught and reported this is really just for safety in case there is anyway an exception that hasn't been considered could be raised and ensure that the scanner task never ends.

From looking around how it's handled elsewhere maybe something like this:

```py
logging.info("Scan raised exception for Watched Folder %s", filesystem.clip_path(dirscan_dir))
logging.debug("Traceback: ", exc_info=True)
```

Once that's resolved it's probably best to merge this and I'll do a seperate PR for events it'll make it easier to review.

Safihre:
Is there's a crash, probably it should be a `warning`. So the user can act on it. I think we might already have a translation-text for that?
And then we probably also don't need to check anymore if the Dirscanner is alive in the periodic check, since it can't really crash anymore. Or we can just leave it, since it doesn't hurt.

sabnzbd/dirscanner.py
```diff
@@ -115,99 +109,171 @@ def save(self):
...
+            self.error_reported = False
+        else:
+            self.ignored[path] = 1
+
+    def scan(self):
+        """Schedule a scan of the watched folder"""
+        if not self.loop:
+            return
+
+        if not (dirscan_dir := self.dirscan_dir):
```
Safihre, line 220: Why do this and not just check `if self.dirscan_dir` and use `self.dirscan_dir` below in the scan-call?

sabnzbd/dirscanner.py
```diff
@@ -115,99 +109,171 @@ def save(self):
...
+                return
+
+            files: Set[str] = set()
+            futures: Set[asyncio.Task] = set()
+
+            for path, catdir, stat_tuple in self.get_suspected_files(dirscan_dir):
+                files.add(path)
+
+                if path in self.ignored or path in self.suspected:
+                    await asyncio.sleep(0)
```
Safihre, line 238: I'm not much versed in asyncio, so why is this 0-sleep needed?

mnightingale: It would probably be fine without that one actually, but the one just after the create_task is yielding the event loop to give other coroutines a chance to run, the intention is that that create_task we just started can start running straight away because the first thing it will do is asyncio.sleep for a second anyway.

Safihre: Right, but the parent-corotine is almost finished at that point (line below)? So it will give the other ones a change to run almost immediately? Or I'm reading it wrong?

sabnzbd/dirscanner.py
mnightingale: I think that change would be fine if `newdir` and `scan` will both be called from the same thread - I think it's the menu bar on Mac I didn't see a way to trigger it on Windows?

I do similar in other places such as `scanner` because we are in a different thread so as I understand it the directory could change between between when we do an `if` and actually use it so I make a copy.
I think it could be removed from scanner but then `scan_async` would have to check it as well.

sabnzbd/dirscanner.py
mnightingale: My thought process was that the user could reasonably want to scan a directory with 20+ nzbs but the directory might have 100 items in it which we won't see but it still has to traverse the directory and maybe it could be slow to list everything?

It's probably a really quick operation even on HDDs so potentially remove both?

mnightingale:
I've removed one of the `await asyncio.sleep(0)` paths, the other should stay because then we can start processing suspected files while we are still looping through the rest.

I'd leave the is_alive() part, I don't think it'll ever happen but better to leave it.

I think the only bit left is logging if a scan failed for an exception other than OSError.

It keeps the behaviour of catching OSError but only shows errors to the user once and resets if a file was subsequently added successfully.
I think I could reuse that logic - perhaps just making the `except OSError`  catch all exception and removing try/except in the scanner?

https://github.com/sabnzbd/sabnzbd/blob/f851f10ee14b8d0b2b247c251d59783610457eee/sabnzbd/dirscanner.py#L203-L205

Safihre:
Yes that would be my suggestion, just catch all.

mnightingale:
Changed the exception to catch all, I think that's all unless you think any changes are needed for https://github.com/sabnzbd/sabnzbd/pull/2434#discussion_r1106407318

sabnzbd/dirscanner.py
```diff
@@ -115,99 +109,167 @@ def save(self):
...
-                        self.ignored[path] = 1
+                    if path in self.suspected:
+                        if not compare_stat_tuple(self.suspected[path], stat_tuple):
+                            # Suspected file attributes have changed
+                            del self.suspected[path]
+
+                    yield path, catdir, stat_tuple
+        except:
+            if not self.error_reported and not catdir:
+                logging.error(T("Cannot read Watched Folder %s"), filesystem.clip_path(folder))
```
Safihre, line 181: Can you add this line below so we have some logging of the problem?
```
logging.info("Traceback: ", exc_info=True)
```

Safihre:
Small last comment, otherwise it's ready to merge.

PR #2435, opened 2023-01-30
Make direct_write_trigger scale with cache size

puzzledsab:
A cache size of 520 MB will give the same number as today (35). 1GB will be 68. Maybe it's a bit conservative but many half finished files or a large part size could fill the cache easily if it's too high. I'm thinking about creating some kind of function that adds all half finished files to the assemble queue if the cache is getting close to full. Maybe there already is such a function and I've missed it too? That would make it OK to use a larger percentage.


puzzledsab:
I'll try again tomorrow.

PR #2436, opened 2023-01-31
Let the assembler write trigger scale with the size of the cache

puzzledsab:
The +1 is not a hack to avoid division by zero. Setting it to 1 results in no caching (every article is added), 2 will cache one article before writing and so on.

### DIFF
--- a/sabnzbd/articlecache.py
+++ b/sabnzbd/articlecache.py
@@ -26,7 +26,7 @@ from typing import Dict, List
 
 import sabnzbd
 from sabnzbd.decorators import synchronized
-from sabnzbd.constants import GIGI, ANFO, MEBI, LIMIT_DECODE_QUEUE, MIN_DECODE_QUEUE
+from sabnzbd.constants import GIGI, ANFO, MEBI, LIMIT_DECODE_QUEUE, MIN_DECODE_QUEUE, ASSEMBLER_WRITE_THRESHOLD
 from sabnzbd.nzbstuff import Article
 
 # Operations on the article table are handled via try/except.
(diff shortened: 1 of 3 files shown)


sabnzbd/articlecache.py
```diff
@@ -68,6 +70,10 @@ def new_limit(self, limit: int):
         # The cache should also not be too small
         self.decoder_cache_article_limit = max(decoder_cache_limit, MIN_DECODE_QUEUE)
 
+        # Set assembler_write_trigger to be the equivalent of ASSEMBLER_WRITE_THRESHOLD %
```
Safihre, line 73: What if the limit is set very small or high?

puzzledsab: For smaller sizes it would write more often, which seems is sensible to me and probably what the user expects. Today it has a fixed limit of 35 and only one article from each file in the queue at a time. Unless I'm missing something that means that on a single big file it can have an indefinitely big backlog without being slowed down.

This will fill the article cache and then it will start swapping single parts to disk, which it will then have to read back again when it's time to assemble them. This will cause a much bigger slowdown than assembling more often. It could happen even without the limitation of one article from each file because it can use up to 5x35x750 KB = 130 MB for normal article sizes and 520 MB for 4 MB article sizes. If the cache is smaller it will save the individual parts.

For huge caches it will write more seldom and in larger chunks. It's not ideal if they set a 4 GB cache and download a few MB/s but that doesn't seem like a common use case. I've thought about adding a timestamp to each file which says when a part was last added. It would be set to 0 on each assemble. That way it would be possible to easily scan the nzb queue and look for files with unsaved articles that haven't been assembled in a while.

This could work much better with sparse files in cases where one server has downloaded some articles for each file and another server is lagging far behind so that none of the articles are saved to their proper files. There could be a flag in the assembler tasks which decides if it should save all articles or only those available in sequence. You said you had experimented with adding sparse files support already so I haven't really looked into it but I think it could be useful.

Did you have other concerns about this in mind than what I've thought about?

Safihre: This is a good explanation 👍
It just wasn't clear from the code right away, that was/is my concern.

Issue #2430, opened 2023-01-29
Profiling with line_profiler


mnightingale:
I wondered if anyone has managed to use https://github.com/pyutils/line_profiler with SAB?

The quick start instructions are:
- `pip install line_profiler`
- Decorate a function with `@profile` - don't import anything
- Run `kernprof -lv script_to_profile.py`

It should print the output and write it to a file, I tried it with a simple .py script, and it works.
However I can't seem to get it to work with SABnzbd, I've tried killing the process, a graceful shutdown and neither output anything.

puzzledsab:
It created a file when I ran it for a while with `kernprof -l -o profile.prof -i 5 SABnzbd.py` but it wasn't readable.
```
$ python -m pstats profile.prof
ValueError: bad marshal data (unknown type code)
Welcome to the profile statistics browser.
```

mnightingale:
😄 I got it to work

Windows/Mac combined changes:

```diff
diff --git a/SABnzbd.py b/SABnzbd.py
index a5e71ec1f..bd2b467a4 100755
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1630,12 +1630,10 @@ def main():
             AppHelper.stopEventLoop()
         except:
             # Failing AppHelper library!
-            os._exit(0)
+            pass
     elif sabnzbd.WIN_SERVICE:
         # Do nothing, let service handle it
         pass
-    else:
-        os._exit(0)
 
 
 ##############################################################################

```

- Decorate a function with `@profile`
- `kernprof -lv SABnzbd.py`

For Mac it doesn't work with the extra thread for the tray icon, presumably similar issue with the way it is shutdown so I just made a simple bootstrap script:

```py
import multiprocessing
import signal

import sabnzbd
from SABnzbd import main

if __name__ == "__main__":
    # Require for freezing
    multiprocessing.freeze_support()

    # We can only register these in the main thread
    signal.signal(signal.SIGINT, sabnzbd.sig_handler)
    signal.signal(signal.SIGTERM, sabnzbd.sig_handler)

    main()

```

- `kernprof -lv LineProfiler.py`

Should be useful for profiling specific functions.

@Safihre any idea why it's os._exit(0) calls? it works with pass which should fallthrough to an exit(0) or using sys.exit(0).
os._exit(0) seems an abrupt way of exiting without performing additional cleanup which I presume is what line_profiler was is hooking into.

Safihre:
[...]

PR #2438, opened 2023-01-31
Threaded polling of connections

puzzledsab:
Added special variable configuration so tests will probably fail because it's not in the wiki.


Safihre:
@mnightingale do you think we even need multiple decoder threads this way?

mnightingale:
Well I was confused for way too long until I noticed it defaults to 1 thread 🤣 - would suggest a minimum based on `os.cpu_count()`  - `min(4, int(os.cpu_count() / 2))` or something.
I think this is a neater approach, it does have the drawbacks that I outlined in the other PR just that most of the time you'll only get very few sockets ready to read at once so its unlikely to utilise many threads.

> @mnightingale do you think we even need multiple decoder threads this way?

I'm unconvinced multiple decoders or a separate thread for them is necessary regardless, from all I've played with encoding and decoding it's extremely fast, much faster than its possible to download. But it would need investigation, it's possible there is a lot of overhead calling the function and releasing the GIL - side thought I wonder if there is a way of creating a buckets of articles to pass to the decoder rather than one at a time to avoid any Python=>C or releasing and reacquiring the GIL overhead, which may actually be quicker than doing one at a time while downloading.

That is as long as SIMD is being used which even on slow hardware can do hundreds of MB/s, I presume the multiple workers was from before SIMD decoding was added. But then I've never tried it on RPi or anything so I could be wrong.

But to answer your question I think this way of using a ThreadPool is likely to often result in only reading from 1 or 2 sockets at a time so I think it would be very similar performance to the current queue/worker approach, it would probably be a good idea to analyse the frequency of sockets against a real server, if it's 1 a lot of the time the queue/worker approach may be better.

The decode would need to be handled within the pool, i.e. not in the `__handle_recv_result` so you'd probably want to request the next article before you start decoding, so the server can start sending it while you decode the previous one.

@puzzledsab looking at `__handle_recv_result` I noticed the `BPSMeter` sleeping, I think it should come after all the `imap_unordered` results are all in not for each one.

Additionally and even if this PR isn't merged for some reason maybe I've misunderstood it's purpose but if you're hitting the bandwidth limit it's going to wait until it's below before queuing finished articles for decoding, I think it'd be better to queue it for the decoder to work on while we sleep, perhaps the intention was to sleep before it requests the next article? - in which case I think the "if done ... decode" part should be before the possible sleeps.

puzzledsab:
1 thread is default because then it will work like before, not using Threadpool at all. I was hoping that would make it easier to accept and I also think it may work better with the speed limiter, although maybe not if the sockets have huge receive buffers.

The threading is primarily intended for Rpi 4, possibly other cheap multi core CPUs and anyone with extremely fast connections and a good number of cores. On the Rpi 4 it's possible to get a decent speedup by using stunnel (SSL proxy) and unencrypted transfer in SABnzbd. It may be partly because then it can request more than 16 KB but I think it's also because it will spread the decryption load on more cores. Using multiple threads for reading in combination with the new receive code will hopefully do the same.

I don't think overhead related to GIL switching is a big part for the decoder threads. The current method for sleeping doesn't work all that well. Usually it will trigger at most every other round, probably less, and receive 16KB or less from every socket like it did in the old version. `unlocked_ssl_recv_into()` may be called 25-50 times for decoder call so that's where the overhead would be. I did see someone report improved performance by increasing the decoder threads to 3 or 4 with the current optimized decoder.

I hope it will at least be included in the beta so we can get some real world tests to see if it works.

mnightingale:
I finally managed to get a RPi 4, testing with my fake server over the local network initially.
The ThreadPool does appear to make a noticeable difference, from 30MB/s to 50MB/s with 2 threads but only 60 MB/s with 4.
I only test with 8 connections though for consistency so it'll be rare that many threads are used.

I also tried against a real server and went 33-42ish (jumps around a bit), but that's more or less my line speed, hence I prefer testing with a fake server.

It's really just convincing me that a rework with a thread per connection may be worth the effort.

@puzzledsab I believe you use a RPi 3? what kind of difference have you noticed?

puzzledsab:
Cool! The Rpi 3 only supports 100 Mbit/s so it could always fill the maximum bandwidth of about 11 MB/s. The CPU usage according to top with the current develop branch has been reduced by just over 20%, though.

I merged the develop branch to fix a conflict so now it will also have the new sleep method. That might help slightly, particularly if you increase the `downloader_sleep_time` to maybe 50 (= 5 ms) or more. On my Rpi there seems to be some extra overhead when sleeping the default of 2 ms or less.

Did you try to increase the decoder threads? Did the cache fill so that it started swapping parts to disk? Did you actually write the data to disk or disable that part? Have you tried a different number of connections? What is your CPU usage according to top? If it's close to 400 with the new sleep then I don't think there's much more to do with the current design.

This is quite exciting. It would be nice to get even more than 60 MB/s but a doubling is still pretty cool. Hopefully it works in the real world as well. It's all coming together with the new SSL receive method just like we hoped :)

Safihre:
Great work, will look at this in more detail how we can structure it best. 

> It's really just convincing me that a rework with a thread per connection may be worth the effort.

You really mean a (Python) thread per connection? Not the previous approach of keeping it around a number of threads similar to the number of cores? Again those examples of users actually putting in the 50-100 connections that providers nowadays offer.

puzzledsab:
Come to think of it, maybe it's better to set `downloader_sleep_time` to 0 in your case if the downloader thread is the bottleneck and it's already filling the receive buffers every time.

mnightingale:
@puzzledsab I forgot about the 3's limitations, I have my 4 booting and downloading to an SSD via USB, seems to manage 200-300 MB/s in benchmarks (SABs test is nearer 100 MB but I don't think it's at all accurate for fast drives), cipher set to AES128.
It barely hits the cache ~50 articles at most but perhaps eventually I can look at HDDs performance, just focussing on CPU for now.

This is with 16 connections without any other changes from the defaults except thread count:
Using `top -b -c -p $(pgrep -d',' -f SABnzbd) > single.log`

**Single thread:** 
Barely using multiple cores ~130%
Downloaded in 1 min 11 seconds at an average of 28.5 MB/s
https://gist.github.com/mnightingale/870a3a07fd1c97f1a7ed1a0d77ddc6d3

**Four threads:**
Hitting all cores ~330%
Downloaded in 29 seconds at an average of 68.7 MB/s
https://gist.github.com/mnightingale/964195210d79d542c0a58c154913d38c

**Master/3.7.1:**
Just for comparison to the current release branch
Similar to develop, 130% with one core at 100%
Downloaded in 1 min 18 seconds at an average of 26.2 MB/s
https://gist.github.com/mnightingale/b6302f3433b0e084526cde73f972a157

It's quite an improvement but I want to get it to 1 Gbps 😄

---

> You really mean a (Python) thread per connection? Not the previous approach of keeping it around a number of threads similar to the number of cores? Again those examples of users actually putting in the 50-100 connections that providers nowadays offer.

Honestly I'm not sure; if I put together a POC I'll probably do a thread per connection, it's by far the simplest to implement, the threads would need to be very tight, [selectors](https://docs.python.org/3/library/selectors.html) would be used with no timeout so threads return as soon as possible, maybe some waiting events to avoid listing on known idle threads at all but the select calls are system/kernel levels calls so extremely fast anyway.

From what I've read Python is good at scheduling threads, forcing a GIL holding thread to release every 5ms and the actual time consuming part of download/decode releases the GIL anyway and requests it again when needed.

Certainly I think pools of connections have potential, the complicated bit would be balancing those connections - probably round robin each priority and handling changes, but the actual logic of the thread/worker would be very similar whether it's dealing with 1 connection or many, hence implementing a thread per connection first then investigating if it's causing issues and possibly create pools seems like the way to go.

---

I highly recommend checking out the book [High Performance Python, 2nd Edition](https://www.oreilly.com/library/view/high-performance-python/9781492055013/) I only started reading it over the weekend and I'm only a few chapters in but so far it's excellent and there are a lot of topics that are very relevant to SAB.

Safihre:
Why not just use blocking sockets in each thread instead of selectors?

mnightingale:
Not sure to be honest, I think you’re probably right that threads could be too expensive though.
I’m going to investigate other approaches, Gevent looks interesting.

Safihre:
> Not sure to be honest, I think you’re probably right that threads could be too expensive though. I’m going to investigate other approaches, Gevent looks interesting.

Just to note that I discarded gevent previously due to not such great performance on Windows. A quick Google search seems to still bring up that it works less well on Windows.

Safihre:
So what to do with this PR? Really focus on getting this specific change ready for production or work on a complete rework of the internals as @mnightingale is suggesting?

mnightingale:
On a RPi 4 changing the cipher to ECDHE-RSA-CHACHA20-POLY1305 helped a lot.

This is with 16 connections:

Master: 79.0 MB/s
(no image)

Develop: 102.7 MB/s
image

Threadpool: 107.7 MB/s (2 threads)
image

CPU usage goes from ~230 to ~260 so still plenty to spare, the bandwidth graph is a bit more stable, and it's pretty much hitting 1 Gbps which is quite impressive for such a device.[^1]

Therefore I think for at least the medium term this approach is a good improvement.

For the longer term more investigation into possible approaches in Python are probably required; threads, threadpools or something else however I think for most users it could just be a lot of work for little gain.[^2]
Because I presume you want to get to a beta release stage I think it's best to use this PR and maybe rework internals for a future release, I wouldn't want a beta/release to be blocked by such a change.

[^1]: It did make me wonder if we could detect if we are running on a Raspberry Pi and override the default cipher, it's pretty complicated for users to go into `openssl speed` and make any sense of it.

[^2]: My current line of thinking is:
**threads** - probably fastest but would need mechanisms for back-pressure, probably similar to issues that are dealt with currently.
**threadpools** - likely similar performance to threads, more complicated to manage (spread out the sockets), less threads for Python to manage/schedule.
**others** - I don't know what I don't know, I tried asyncio but it seems to be more for concurrency than parallel, it seemed really efficient and you can write a really clean protocol for interacting with nntp. But its was just managing one thread so it's slower than threading even more so when I added decoding, but this could just be because I don't know how to use it well.
**tl;dr** - I think threads are probably where I'd start.

puzzledsab:
I wasn't aware of the performance advantage of that particular codec on hardware without AES support and it appears neither are any of those who give advice on r/usenet. I see that @Safihre added it to the SSL page on the wiki in december last year, though. I'm considering rearranging the layout of the high performance tip page so that each tip first has a short summary of what to do specifically and then a longer description of how and why.

In my view this PR is fairly ready so if I'm missing anything that could cause problems then please let me know. I'm not sure if we should make the performance tweaks somewhat easier to find. Maybe a separate tab for performance tweaks which would contain cache size, "Only get articles for top of the queue", downloader_sleep_time, number of downloader, decoder and direct unpack threads, bps limiter and possibly some more. @Safihre?

sabnzbd/downloader.py
```diff
@@ -477,6 +485,18 @@ def sleep_time_set(self):
         self.sleep_time = cfg.downloader_sleep_time() * 0.0001
         logging.debug("Sleep time: %f seconds", self.sleep_time)
 
+    def recv_threads_set(self):
```
Safihre, line 488: Why do we need this?
No need to add support for live-changes for Special's settings.

sabnzbd/cfg.py
```diff
@@ -434,6 +434,7 @@ def validate_notempty(root, value, default):
 local_ranges = OptionList("misc", "local_ranges", protect=True)
 max_url_retries = OptionNumber("misc", "max_url_retries", 10, minval=1)
 downloader_sleep_time = OptionNumber("misc", "downloader_sleep_time", 10, minval=0)
+receive_threads = OptionNumber("misc", "receive_threads", 1, minval=1)
```
Safihre, line 437: Should we not have a smarter default? Or at least 2?

sabnzbd/downloader.py
```diff
@@ -756,96 +776,110 @@ def run(self):
...
 
-            for selected in read:
-                nw = self.read_fds[selected]
-                article = nw.article
-                server = nw.server
+            if self.recv_threads > 1:
+                for nw, bytes_received, done in self.recv_pool.imap_unordered(self.__recv, read):
+                    self.__handle_recv_result(nw, bytes_received, done)
+                if self.bandwidth_limit:
+                    self.__check_speed()
```
Safihre, line 783: I think we don't need 2 paths, with and without the pool. To 'disable' it can just be set to have 1 pool.
This way we also don't need the `__handle_recv_result`  and `__check_speed` functions, as they can just be inlined.

Safihre:
Just some small comments, then we can merge,
What about a default of 2? Or we want to be smarter with number of cores available?

sabnzbd/cfg.py
puzzledsab: The threading adds a ton of overhead when the CPU is powerful enough to be mostly idle. For me the CPU load in Windows task manager bounces between 0.0 and 0.5% with 1 thread and the new sleep method during small downloads. With 2 threads it's between 2.5 and 4%. According to Yappi it is all threadpool stuff. I think it should only be increased from 1 if it is actually helpful for the speed.

sabnzbd/downloader.py
puzzledsab: I wasn't sure. I'll remove it.

sabnzbd/downloader.py
puzzledsab: Because of the overhead and also because I think it may be better for the bandwidth limiter if can sleep after each recv it checks instead of letting the threadpool clear all the buffers and then do a lot of sleeping.

sabnzbd/cfg.py
mnightingale: I'm not saying you're wrong certainly there is an overhead with threads, I just wondered if you could share any detail of what you saw in profiling to draw the conclusion that you've gone 0-0.5% to 2.5-4%, could it be attributed to being more efficient and calling the `recv_chunk` more often when using threads?

I don't really see any change in CPU usage and the cheroot webserver threadpool is taking up the most time.

puzzledsab: Here's a dump from Yappi after having downloaded for a while: https://gist.github.com/puzzledsab/94f137f406edf550f669a3d36a80d429

The profiling adds some overhead per function call but for most of the threading related calls on the top of the list it's a very small share and the increase in CPU usage is consistent with and without Yappi.

The new sleep time logging shows that it sleeps noticeably more frequently when using only 1 thread so it's most likely doing more downloader loops per second when not using threadpool. That's down from sleeping 0.00292 seconds about 6650 times per 20 seconds with one thread to about 6180 times with 2 threads.

mnightingale: Would you mind saving a callgrind file out, `stats.save("callgrind.yappi", type="callgrind")`
I want to take a look at the callers.

I'm not sure if you can do that directly in PyCharm, the script I use is https://gist.github.com/mnightingale/f4fb48af38bed12def73ebcea679736e with the line_profiler fix I found on https://github.com/sabnzbd/sabnzbd/issues/2430#issuecomment-1414328294

puzzledsab: https://gist.github.com/puzzledsab/03d88c191f35cd5003509f356c890bcc

I shortened the fixed parts of the paths a bit.

Safihre: I'm not convinced that the added code-complexity is worth it if it requires manually setting a Special setting. It should work out of the box..

mnightingale: > https://gist.github.com/puzzledsab/03d88c191f35cd5003509f356c890bcc
> 
> I shortened the fixed parts of the paths a bit.

Could you share one were you've set it to one thread so not using a pool?
At least 25% of the time is the webservers threads/pool.
What CPU is this running on?

I did always assume we'd have an initial based on `os.cpu_count()`.
e.g. if `os.cpu_count()` > 1 then default to 2 and probably restrict the max somewhere with `os.cpu_count()` or a lower limit.

puzzledsab: 1 thread: https://gist.github.com/puzzledsab/e472590c73d413f13dbd76af99b76acb

@Safihre An alternative could be to do a speed test to see if it increases the speed when the number of threads is increased. In that case I think it should still be possible to set it manually in some way because in my experience the auto detection is not very reliable. I really believe it should be available considering how it can drastically increase the speed for some users. On Rpi 4 it seems to be much faster than the speeds I've seen reported for nzbget (~30 MB/s), even with AES128. I suspect it can increase the speed on other CPUs with 4 cores or more as well, although maybe not quite as much.

Safihre: I tried it on my i5, but I cannot really see a difference.. Both develop and this branch with 2 threads are almost always using 0-1% CPU. Maybe on develop it's more often 0 than 1, but it's very hard to judge.

puzzledsab: It's great if it's just me but I have no idea why. I'm using Windows 10 and Python 3.11.2 (just updated from 3.11.1 and 3.10 a week or so ago) from python.org. I've tried various settings for connections, downloader_sleep_time, ciphers and different servers but the difference when switching between 1 and 2 threads is always very distinct.

thezoggy: you running sab on ssd or spinner? wonder if threading has more impact when you have slower disks involved or combo of higher speed + other io going on during the event

puzzledsab: SSD only, different disks for incomplete and complete.

mnightingale: Maybe something to do with it being 4 core 4 threads? since downloader/assembler/decoder will frequently all have something to do, does your decoder have the default 2 threads?

Possibly as a default for os.cpu_count <= 4 we use **1** otherwise **2** - from my testing I never saw any meaning full improvement with more than 2.

As for keeping two paths, it's not ideal but @puzzledsab profiling does seem to show quite an impact.

Safihre: SABnzbd is about working straight out of the box, so no tweaking for regular users.
Either we default to 2 threads, or if that has a negative performance impact, we look for a different approach like the thread-per-connection.

puzzledsab:
I kept the code for not using threads when it's set to 1. I think it should at least be available during the testing period. It's just a few lines and it probably works better with the bandwidth limiter. Regarding number of threads I think it's better with slightly too many than too few, both for the receiver and decoder. There is hardly any overhead for one or two extra threads.

Safihre:
I just don't like how the code looks at the moment, so I'll probably do a refractor before 4.0 stable.
But I guess it's good enough.

One thing to note is that when there's an uncaught exception in a ThreadPool/Executor worker, it closes the thread. So we should handle that gracefully and make sure the error is logged. We had this problem in the other ThreadPool we already have for the Status window.

And could you take a look at the excessive logging from the download sleeper? I commented on the merged PR. We can't launch it like that.

puzzledsab:
Isn't everything that can crash already inside the try in `__recv()`? I moved getting the newswrapper as well but something has to be very wrong for that to fail. My understanding is that it basically works like before and logs the same errors it would have. Am I missing anything?

Safihre:
Oh yes sorry, my bad, indeed it's already all inside try/accept!

Safihre:
I changed the default to 2 threads because on my i5 the cpu_count based one resulted in 3 SIMD threads and 5 recv_threads, way too many. 
Let's start with 2 and see during testing if it needs adjusting.

puzzledsab:
Right, it's because of hyperthreading so it thinks you have 8 CPUs. I think we'd need to include psutil to detect the actual number. It would allow us to set ionice priority for Windows as well, though.

mnightingale:
I think it's certainly worth users testing, but once I was using the 'correct' cipher the RPi 4 was only 5% faster and I've tested the develop branch against real servers at >450MB/s on a VPS - and I'm sure it could go faster as it wasn't an optimal setup for testing the speed, it wasn't a CPU specialised system, it was writing to disk and not as close to the servers as possible.

However as @puzzledsab has suggested a few times I do think where this could really show it's worth is if parts of download/decode/assemble could be combined within the C extension to make better use of time when SAB is mostly just waiting for more data.

There are a few ways I think that could work:

1. Adding to `sabctools.unlocked_ssl_recv_into` where we've already released the GIL
    1. incremental decoding of whatever we have read into the buffer so far
    1. decode the whole article if end of article pattern found - similar to now but avoid some overhead - if we still wanted a copy for the article cache then the decoded buffer will be smaller anyway
2. I'm not sure how this would fit in and would likely need to be optional because it would probably only be beneficial for SSD users; but investigate writing the decoded buffer straight to file without creating a copy to store in the article cache.

Safihre:
> I think it's certainly worth users testing, but once I was using the 'correct' cipher the RPi 4 was only 5% faster and I've tested the develop branch against real servers at >450MB/s on a VPS - and I'm sure it could go faster as it wasn't an optimal setup for testing the speed, it wasn't a CPU specialised system, it was writing to disk and not as close to the servers as possible.

You mean that using multiple receive threads was only 5% beneficial? I thought you previously reported a big increase here: https://github.com/sabnzbd/sabnzbd/pull/2438#issuecomment-1417461603

mnightingale:
Yes it helped a lot when it was bottlenecked by the CPU AES decoding, there is no hardware accelleration on RPi's but switching to ECDHE-RSA-CHACHA20-POLY1305 in https://github.com/sabnzbd/sabnzbd/pull/2438#issuecomment-1422317526 brought the 'without this PR' speed way up and the threadpool only gave 5% on top of that.

Safihre:
Ah I see. And how about on your super-duper-fast test-system? There it did have some effect, right?

puzzledsab:
The CHACHA20 is only supported by some providers so it can still be very useful. Also more than 1 Gbps is becoming more common. I hear in Asia there are some places where you can get 20 Gbit/s at home.

It might be useful to be able to adjust the switch interval, though. The downloader thread will be locking the GIL for longer continuous stretches of time than before now that it probably won't have to wait for the results (except for the first one). I'm not sure if this is how it works but I assume that the assembler thread releases the GIL each time it's writing a block. If it has to wait 0.005 seconds between each time it gets it back then that could limit the write speed to 200 write calls per second or 150 MB/s with 750 KB blocks. On a normal system it probably won't be that bad but the VMs seem to work in mysterious ways.

Safihre:
@mnightingale if you have the time, could you test if this change (compared to the commit before it) has substantial impact on your very high performance test setup?

mnightingale:
50GB, 800KB parts with 16 connections using AES128.

Pre ThreadPool 14223d239a64a5610e6f151ea14ae2b89d5e70d4:
Downloaded in 45 seconds at an average of 1.1 GB/s

ThreadPool PR 73fce52df1a5180aeeb1146c502bbcdb0a2af607
Downloaded in 33 seconds at an average of 1.5 GB/s

It's not particularly representative though because it's just local sockets, with real servers the round trip time of requesting articles would have an impact which can be overcome somewhat with more connections.

---

I was actually writing a comment to the effect of "is the added complexity and potential CPU usage worth it" when you merged it as I can't see it helping many users; without it a user with enough connections and close to the server should be able to hit 5-10 Gbps and I did verify that at least 5 Gbps was possible on a VPS in the same country as a couple of severs.

Most users are going to have at most 1 Gbps which quite modest hardware can already manage and there will be very few on 2.5/5 or over 10 Gbps and a system capable of utilising it.

We are already at quite a massive performance increase  and reduction in CPU usage from the unlocked_ssl and crc32 changes.

Don't get me wrong @puzzledsab I'm grateful for all the work you did exploring different options and it has helped me way up threads vs pools.

Safihre:
That's a nice theoretical improvement! But indeed I also wonder if it's worth it.. We can keep it for now during alpha testing and see what happens.

PR #2439, opened 2023-01-31
Gradual slowdown on filling queues

puzzledsab:
When one of the queues is > 50% full it will gradually start to slow down by adding a sleep time which depends on how full it is.

I have removed `partial_nzf_in_queue()` because I realized that it breaks the full queue detection if only one big file is being written. In that case there will never be more than one article in the queue. I believe it could be the reason why the cache was filling and it started saving cached parts temporarily to disk in @mnightingale's test.  To compensate for that I increased the limit to 10. If it's combined with #2436 then the impact will be limited on fast systems with lots of cache.


puzzledsab:
New black rules? I didn't modify those files.

sabnzbd/downloader.py
```diff
@@ -55,6 +55,8 @@
 _BPSMETER_UPDATE_DELAY = 0.05
 # How many articles should be prefetched when checking the next articles?
 _ARTICLE_PREFETCH = 20
+# Minimum decoder or assembler queue level before slowing down download. 1 = full
+_QSLOWDOWN_LEVEL = 0.6
```
Safihre, line 59: Could you give it a less cryptic name? Like `QUEUE_THRESHOLD` or `MAX_QUEUE_LEVEL`?

sabnzbd/downloader.py
```diff
@@ -545,29 +547,33 @@ def decode(self, article, raw_data: Optional[bytearray] = None, raw_data_size: O
...
-            time.sleep(0.1)
-            logged_counter += 1
-            decoder_full = sabnzbd.Decoder.queue_full()
-            assembler_full = sabnzbd.Assembler.queue_full()
+        decoder_level = sabnzbd.Decoder.queue_level()
+        assembler_level = sabnzbd.Assembler.queue_level()
+
+        # Sleep for an increasing amount of time, depending on queue sizes.
+        if decoder_level > _QSLOWDOWN_LEVEL or assembler_level > _QSLOWDOWN_LEVEL:
+            time.sleep((decoder_level + assembler_level - _QSLOWDOWN_LEVEL) / 2)
```
Safihre, line 556: The `/2` is arbitrary, right? Why did you chose it?

sabnzbd/nzbqueue.py
```diff
@@ -756,11 +756,7 @@ def register_article(self, article: Article, success: bool = True):
 
         # Write data if file is done or at trigger time
         # Skip if the file is already queued, since all available articles will then be written
-        if file_done or (
-            articles_left
-            and (articles_left % DIRECT_WRITE_TRIGGER) == 0
-            and not sabnzbd.Assembler.partial_nzf_in_queue(nzf)
-        ):
+        if file_done or (articles_left and (articles_left % DIRECT_WRITE_TRIGGER) == 0):
```
Safihre, line 759: I wonder how this will play out, if it doesn't just freeze because the queue is full of first articles from the other PR.
For example a huge season pack with 500 files would block the assembler queue for quite a while.
Not really related to this PR, just thought of it while seeing this change.
Will have to test the effects on slow devices with small cache, but with fast internet..

sabnzbd/downloader.py
puzzledsab: Mostly because a maximum sleep of 0.2 seconds (for one full queue) seemed fair. If it triggers every time a part completes it will slow down the speed to 3.75 MB/s before it's full and the absolute limit kicks in. For the assembler queue the minimum sleep will be 0.05 seconds with a limit of 10 elements. It won't slow it down very significantly if it's only needed once in a while to make the processing catch up, but at the same time it's not so short that it's insignificant if it happens on every call. It hopefully gives the assembler queue some granularity in the slowing down even though it has a fairly low max limit.

If it's higher then it should probably be split up in multiple shorter sleeps with a check for shutdown and other things in between. I wanted to avoid that for simplicity. If it's much smaller then the cases where it's not enough will increase.

If it should be changed it's probably downwards a bit (for a shorter sleep). Maybe /4. I think anything less than that would be mostly useless for the assembler queue.

sabnzbd/downloader.py
puzzledsab: I struggled to come up with anything because it's not an absolute threshold. I changed it to `SOFT_QUEUE_LEVEL` and put it with the other queue limit variables in `constants.py`.

sabnzbd/nzbqueue.py
puzzledsab: I think they should be written quickly enough to mostly keep up because it's only a single part and it's the first in the article list. Writing them one at a time in the assembler instead of concurrently through the decoder might even be an advantage, at least on spinning disks. I guess we'll find out.

puzzledsab: If it does cause problems then an alternative could be to add a simple accounting of how many articles the queue actually contains. For instance, first article would only add 1, the trigger would add the specified number and done could add trigger / 2. The numbers would be included in the queued elements and subtracted by the assembler. An better alternative but less simple would be to have sparse files and force some flushing if the cache is getting full.

PR #2440, opened 2023-01-31
Don't crash on invalid yenc footer

### LINKED ISSUE #2427: ValueError: Invalid yEnc footer not handled well
I mentioned this error in the segfault issue. I got it again now:
```
  File "sabnzbd\sabnzbd\decoder.py", line 158, in run
    decoded_data = decode_yenc(article, raw_data, raw_data_size)
  File "sabnzbd\sabnzbd\decoder.py", line 252, in decode_yenc
    yenc_filename, crc_correct = sabctools.yenc_decode(data)
ValueError: Invalid yEnc footer

During handling of the above exception, another exception occurred:

  File "sabnzbd\sabnzbd\decoder.py", line 206, in run
    for line in b"".join(raw_data[:2]).split(b"\r\n"):
`TypeError`: sequence item 0: expected a bytes-like object, int found
```
I fixed it locally by catching the `TypeError` but I think the yenc decoder should ideally decode what it can and return it with an empty `crc_correct`. The broken post had a proper start and seemed almost complete but did not have any part of the footer. It might be useful for par2 if it were decoded.

puzzledsab:
Fixes #2427. Aren't anyone else experiencing this with the dev branch? I've woken up to a crashed SABnzbd 3 times already. I would prefer if it was fixed by making the decoder decode what it has but either way it shouldn't crash here if the data unexpected.

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -205,7 +205,6 @@ class DecoderWorker(Thread):
                         # and examine the headers (for precheck) or body (for download).
                         for line in raw_data[:2000].split(b"\r\n"):
                             lline = line.lower()
-                            logging.debug("Line %s", lline)
                             if lline.startswith(b"message-id:"):
                                 article_success = True
                             # Look for DMCA clues (while skipping "X-" headers)
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -187,7 +187,7 @@ class DecoderWorker(Thread):
 
             except (BadYenc, ValueError):
                 # Handles precheck and badly formed articles
-                if nzo.precheck and raw_data and raw_data[0].startswith(b"223 "):
+                if nzo.precheck and raw_data and raw_data.startswith(b"223 "):
                     # STAT was used, so we only get a status code
                     article_success = True
                 else:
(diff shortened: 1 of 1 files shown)


mnightingale:
Unfortunately this isn't really fixing the problem; there are a few parts of this file that don't handle that raw_data is now a bytearray (except for uu which also confuses the flow).

For example if BadYenc is thrown `raw_data[:2]` used to mean the first two chunks of bytes, now it's the first two bytes.
Similar with `raw_data[0].startswith(b"223 ")`.

As for the underlying issue in #2427 as suggested we need to update the decoder to ignore that we dont have a footer and decode to the end (probably as you suggested find the last \r\n) - and maybe add null bytes up to the expected lengt).

Safihre:
Thanks for fixing it! I just moved to my new house so my office spot isn't really up yet, so couldn't fix it myself.
But, maybe you can also (locally) add a breakpoint or save the raw fats to pickle to inspect these invalid footers? Are they really invalid?
This is also how we created the test pickle files for sabyenc3 long ago, just saving to disk all failures.
Thanks!

puzzledsab:
In the one I checked previously it was invalid because it was missing a few KB on the end of the block. The footer wasn't there.

Safihre:
Maybe you can check a few more if you have the time? :)
I'll merge this for now indeed.

puzzledsab:
I checked 3 more, it's the same. It's the yenc header, random data and then \r\n.\r\n at the end. They were all missing at most a few hundred bytes compared to the segment bytes values in the nzb.

PR #2441, opened 2023-02-02
par2 arm64 update

goodpizzaman:
par2 for Apple M series chips for macOS. Based on par2SL.
Verification speeds up to 1GB+/sec.


Safihre:
Why did you close it?

goodpizzaman:
Verification is faster but the repair process is still slow.

PR #2443, opened 2023-02-02
Write first article directly

puzzledsab:
Yet another mini optimization. Write the first article directly to the correct file instead of temporary file if the correct filename is available. I didn't try to reserve space before checking free because it's just one article and it will usually be saved quickly. It saves a write and a read so if the cache is full then this will most likely help clear it faster because of reduced disk trashing.

### DIFF
--- a/sabnzbd/articlecache.py
+++ b/sabnzbd/articlecache.py
@@ -92,15 +92,11 @@ class ArticleCache:
         # Register article for bookkeeping in case the job is deleted
         nzo.add_saved_article(article)
 
-        if article.lowest_partnum and not article.nzf.import_finished:
-            # If the filename is verified and there is room left then the first part
-            # will be saved to the correct file immediately, so it's kept in cache.
-            # Otherwise it will be saved temporarily to save cache space.
-            if article.nzf.filename_checked and self.space_left():
-                self.reserve_space(len(data))
-                self.__article_table[article] = data
-            else:
-                self.__flush_article_to_disk(article, data)
+        if article.lowest_partnum and not (article.nzf.import_finished or article.nzf.filename_checked):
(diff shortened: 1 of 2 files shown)


Safihre:
Hmmm that seems a bit ugly, calling Assembler there. 
I think it would be much cleaner to trigger a write here?
https://github.com/sabnzbd/sabnzbd/blob/9ab5e86c81b176ca2c3da565143fc902e13451af/sabnzbd/nzbqueue.py#L756-L762
And then in ArticleCache only write to disk if the filename isn't verified?

sabnzbd/articlecache.py
```diff
@@ -93,9 +93,14 @@ def save_article(self, article: Article, data: bytes):
...
 
         if article.lowest_partnum and not article.nzf.import_finished:
-            # Write the first-fetched articles to disk
-            # Otherwise the cache could overflow
-            self.__flush_article_to_disk(article, data)
+            # If the filename is verified and there is room left then the first part
+            # will be saved to the correct file immediately, so it's kept in cache.
+            # Otherwise it will be saved temporarily to save cache space.
+            if article.nzf.filename_checked and self.space_left():
+                self.reserve_space(len(data))
```
Safihre, line 100: Shouldn't this just skip to the code below to follow the regular path? Otherwise we could also exceed cache limits.

puzzledsab: But only slightly. I commented on why I did it like that in the PR description. If you still disagree I will change it.

Safihre: I think it's wise to chose the safe route. Especially since it's just a minor optimization. It also prevents adding another code-path in case of problems, it just goes the normal path.

PR #2447, opened 2023-02-02
Replace article_queue list with a deque

mnightingale:
# Why

Almost all operations on the article_queue which is a list are `pop(0)` and the one that isn't probably should be.

Popping the first item on a list has complexity O(n) - it has to shift every following item along.

It's a pretty minor issue but its a lot more CPU operations than are necessary.

# How

Replaced with `collections.deque` which is O(1) at either end and still O(1) to append.

https://wiki.python.org/moin/TimeComplexity

# Extra

I wanted to keep this PR simple, but what do we think about reusing the same deque?

This PR currently leaves it as `self.article_queue = collections.deque()`

So `reset_article_queue` could do:

```py
def reset_article_queue(self):
    logging.debug("Resetting article queue for %s", self)
    while self.article_queue:
        article = self.article_queue.popleft()
        sabnzbd.NzbQueue.reset_try_lists(article, remove_fetcher_from_trylist=False)
```

Or to stay like it is currently:

```py
def reset_article_queue(self):
    logging.debug("Resetting article queue for %s", self)
    for article in self.article_queue:
        sabnzbd.NzbQueue.reset_try_lists(article, remove_fetcher_from_trylist=False)
    self.article_queue.clear()
```

`sabnzbd.NzbQueue.get_articles` could be a generator which we loop over to append to the existing deque.

# Future

I could do with seeing if similar happens anywhere else that is worth changing, `NzbFile.articles` seems a possible candidate, `remove_article` does an x in s O(n) but it'll be near the beginning so not too bad, then it does the actual remove which is O(n) and near worst case since it'll be near the beginning. 
Perhaps a dict would be more appropriate for the types of calls it's used for.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -28,7 +28,8 @@ import socket
 import random
 import sys
 import ssl
-from typing import List, Dict, Optional, Union
+import collections
+from typing import List, Dict, Optional, Union, Deque
 
 import sabnzbd
 from sabnzbd.decorators import synchronized, NzbQueueLocker, DOWNLOADER_CV
@@ -152,7 +153,7 @@ class Server:
         self.request: bool = False  # True if a getaddrinfo() request is pending
         self.have_body: bool = True  # Assume server has "BODY", until proven otherwise
         self.have_stat: bool = True  # Assume server has "STAT", until proven otherwise
-        self.article_queue: List[sabnzbd.nzbstuff.Article] = []
+        self.article_queue: Deque[sabnzbd.nzbstuff.Article] = collections.deque()
 
         # Skip during server testing
         if threads:
@@ -223,17 +224,17 @@ class Server:
         """Get article from pre-fetched and pre-fetch new ones if necessary.
         Articles that are too old for this server are immediately marked as tried"""
         if self.article_queue:
-            return self.article_queue.pop(0)
+            return self.article_queue.popleft()
         elif self.next_article_search < time.time():
             # Pre-fetch new articles
             self.article_queue = sabnzbd.NzbQueue.get_articles(self, sabnzbd.Downloader.servers, _ARTICLE_PREFETCH)
             if self.article_queue:
-                article = self.article_queue.pop(0)
+                article = self.article_queue.popleft()
                 # Mark expired articles as tried on this server
                 if self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
                     sabnzbd.Downloader.decode(article)
                     while self.article_queue:
-                        sabnzbd.Downloader.decode(self.article_queue.pop())
+                        sabnzbd.Downloader.decode(self.article_queue.popleft())
                 else:
                     return article
             else:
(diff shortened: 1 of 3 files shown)


puzzledsab:
It's clearly faster in my test but I didn't see much to the O(1) vs O(n) effect until the list was fairly large.
Code:
```python
import timeit

times = 100000
setup = """
import collections

l = []
q = collections.deque()
"""

testcode = """
for i in range(1000):
    l.append("abcddddddddddd")
while(l):
    l.pop(0)
"""
print("list", timeit.timeit(stmt=testcode, setup=setup, number=times))

testcode = """
for i in range(1000):
    q.append("abcddddddddddd")
while(q):
    q.popleft()
"""
print("deque", timeit.timeit(stmt=testcode, setup=setup, number=times))
```

range(100), times = 100000:
list 1.3113736999221146
deque 0.7135395999066532

range(1000), times = 100000:
list 18.926564099965617
deque 7.86598280002363

range(10000), times = 10000
list 77.11450539994985
deque 7.664715199964121

puzzledsab:
I suspect that it doesn't work quite like that in practice, though. As you can see from my tests, deque is about twice as fast for both 100 and 1000 items. Only at 10 000 items does it become 10x faster. Maybe it just moves a pointer or something for a certain amount of data before it moves and frees it in bulks.

mnightingale:
I don't think there is anything to debate about it, Python's own documentation explains the time complexity of the functions.
I wasn't trying to suggest that there was some massive win to be had, indeed these are actually tiny 20 item lists, just that moving list elements in memory in this way is more intensive then moving a couple of pointers.

List pop - https://github.com/python/cpython/blob/0675b8f032c69d265468b31d5cadac6a7ce4bd9c/Objects/listobject.c#L1033-L1038 - memmove all the other items hence O(n) 

Deque pop - https://github.com/python/cpython/blob/0675b8f032c69d265468b31d5cadac6a7ce4bd9c/Modules/_collectionsmodule.c#L200-L208 - update the pointers so O(1)

Likewise the suggestion of reusing the deque, saves all the constant object allocations and garbage collection.

Safihre:
I wonder if we should just have a constant for the empty deque's that we return when nothing was found? Or just return None?

mnightingale:
> I wonder if we should just have a constant for the empty deque's that we return when nothing was found? Or just return None?

I described it in the "Extra" part, I didn't want to complicate this PR with it initially but I'd suggest a single allocation in \_\_init\_\_ and within the get_articles calls remove fetch_limit as an argument and make them generators - just break the loop once the deque is at capacity or the generator stops returning new values (for the special cases such as only first articles), this avoid extra memory usage from returning a new a list/deque.

I can add it to this PR or make a new one?

puzzledsab:
> I don't think there is anything to debate about it, Python's own documentation explains the time complexity of the functions.

I'm not pretending I'm an expert on how it works. My point was only that it doesn't matter much if the list contains 20 or 200 elements, the real world savings from using deque are about the same.

mnightingale:
I'm going to leave this change for now; its too insignificant and there are a few more time consuming functions that are better focussed on.

PR #2450, opened 2023-02-06
include https config files in backup

### LINKED ISSUE #2338: sab backup - include cert?
perhaps we should include the relevant sab cert/key so it makes the backup that much easier to restore..

- HTTPS Certificate
- HTTPS Key
- HTTPS Chain Certificates

jcfp:
Files are included in the backup zipfile under their default basename (with "server.chain" used for the one that doesn't have a usable default value). On restore, all files are placed in the admin dir, the affected setting(s) modified to point to the restored file(s) and the config saved.

To prevent backing up a https config broken by some last-minute change as well as the inclusion of arbitrary (possibly large or sensitive) files into the backup by manipulating the settings just prior to creating a backup, the cert/key/chain files are only included if https was enabled at startup and files with a matching full path successfully loaded at that time.

Closes #2338

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -77,6 +77,7 @@ from sabnzbd.constants import (
     DEF_LOG_FILE,
     DEF_STD_CONFIG,
     DEF_LOG_CHERRY,
+    CONFIG_BACKUP_HTTPS,
 )
 import sabnzbd.newsunpack
 from sabnzbd.misc import (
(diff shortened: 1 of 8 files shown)


SABnzbd.py
```diff
@@ -1446,6 +1447,12 @@ def main():
         logging.error(T("Failed to start web-interface: "), exc_info=True)
         abort_and_show_error(browserhost, cherryport)
 
+    # Create a record of the active cert/key/chain files, for use with config.create_config_backup()
+    if enable_https:
+        for setting in CONFIG_BACKUP_HTTPS.values():
+            if full_path := getattr(sabnzbd.cfg, setting).get_path():
```
Safihre, line 1453: Nice 🦭

Safihre:
Looks good!
Maybe we can do something with using `cfg.X.default`? Instead of hardcoded values?

And can you update the tests?

Safihre:
Problem was that the `cfg.X.get_path` returns long-path notation on Windows. So the `full_path in sabnzbd.CONFIG_BACKUP_HTTPS_OK` check in `create_config_backup` failed.
Fixed that.
Anything else that needs to be added still or you think it's ready?

PR #2457, opened 2023-02-13
Queue based threaded receiver

puzzledsab:
I mocked up a simple queue based test to see how it performed. It doesn't use as little CPU as no threading but a lot less than threadpooling. I'd still prefer the threadpool version with 1 thread as default because I think the BPS limiter would work better, the CPU usage is lower and it's less code. The main known exception which can definitely use more than 1 thread is Raspberry Pi 4. We could try to auto detect that and default to 2 - 4 threads.


Safihre:
Interesting, it's basically a ThreadPool but then we wrote our own.
But still defaults to 1 thread and not 2?
 @mnightingale what do you see in terms of speed?

puzzledsab:
The code needs some polishing before it's ready for merging. It's only meant as a demonstration of an alternative. I intended to change the default to scale with the number of cores and maybe the number of decoders should too.

mnightingale:
Performance seems more or less the same as a ThreadPool, but it does look like it's very similar to how the ThreadPool class works (noticed they used queue.SimpleQueue) so I can't see why it would make much difference to CPU usage.

@puzzledsab maybe try `concurrent.futures.ThreadPoolExecutor` in the other PR and see if you notice any CPU difference, I don't think there will be, I think it's just a newer api that has some other uses.

I think I'd probably encourage sticking with the Python builtins, I don't think the trade off off all the complexity is worth it.

puzzledsab:
I didn't think there would be a difference either and the map function seems more limited but the overhead practically disappeared. It's lower than this branch so no point in keeping it. I'll commit an update to the other branch in a sec.

PR #2458, opened 2023-02-14
Increase threshold for logging excessive sleep time

puzzledsab:
It seems your system sleeps much longer than `downloader_sleep_time` requests much more frequently than mine. That's one of the things I wanted to know. I only rarely get those messages. Before you switch to this code, could you try setting
```
import sys
sys.setswitchinterval(0.001)
```
in `SABnzbd.py` and see if it helps? Is there a lot of other load on your system?

The new code increases it from 0.01 to 0.05 seconds longer than `downloader_sleep_time`. If it sleeps longer then that frequently then it will seriously impact the performance.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -717,7 +717,7 @@ class Downloader(Thread):
                     time.sleep(self.sleep_time)
                     now = time.time()
                     # Debugging code for v4 test release
-                    if now - time_before > self.sleep_time + 0.01:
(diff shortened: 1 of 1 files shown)


Safihre:
I see I only have it when I run in PyCharm's debugger. 
Also only happens when I'm on slower wifi (2.4Ghz instead of 5Ghz) so my PC is just sitting around waiting.
Still seems good to increase the limit a bit.

PR #2459, opened 2023-02-16
Flush oldest nzf data when cache is full

puzzledsab:
If the user has multiple servers and there are missing articles then some servers are sometimes far ahead of the others in the queue. The assembler will only be called for every 71th article or something like that if the cache is 1 GB so there's a good chance that each unfinished file has articles that can be written. This patch logs the time for when the nzf last got new data from the decoder and calls the assembler on the nzf with the oldest data if the cache becomes full.

Unfortunately it will only write up to the first unavailable article. A better solution would be to write all available articles in the correct place but to do that we need to know the starting index of each article. All subsequent writes to that file would also need to use seek. 

I considered doing it when the cache is 90% full instead but chose the solution with the minimal impact compared to how it works today. It usually won't be triggered either way.

The easiest way to check that it triggers is to disable the call to process in register_article.


Safihre:
Yes, I agree that the real solution is to let sabctools report the start and end position. Don't know if this intermediate step is then worth it.
With the new yEnc stuff in sabctools it's really just a few lines of code to extract start and end.

puzzledsab:
I'm uncertain about how well the current code works. It seems to sometimes generate broken files when I set `sparse_mode = True` in `assemble()` but I haven't been able to track down why yet. It's usually ok even though the articles are not written in sequence.

Do you think it's already getting too complex? We'd need to detect if sparse files are supported and I haven't looked into a way to retry nzbs with missing articles.

puzzledsab:
It's become rather complicated because there are so many situations that need to be handled slightly different, depending on what write mode is currently used, if sparse mode is enabled and if it can be enabled.
- Normal mode is append so that there is no need for seek.
- seek only seems to work with w+ and r+, not a+, so if sparse mode is enabled while looping then we need to stop and reopen the file in one of those modes.
- w+ will overwrite the file if it already exists, but r+ can't be used if it doesn't exist.
- crc32 is not calculated until file_done is set so that it can be done in order

I'm sure the current logic can be simplified. My main priority has been to figure out how to make it work at all.

PR #2460, opened 2023-02-16
Only allocate disk speed random data when required

mnightingale:
I was having a play with https://github.com/bloomberg/memray and noticed 10MB of random data is allocated at startup and kept around forever to be used for testing disk speed.

Better to just allocate it when required.

## Before
before

## After:
after


Safihre:
Nice, not very smart of us.

Issue #2171, opened 2022-05-04
Support Season Packs in Series Sorter


ItsME6969:
Feel free to delete this issue as its a dupe of https://github.com/sabnzbd/sabnzbd/issues/2112
I commented on that issue a little over a month ago and the issue was already closed so maybe my comment wasn't seen but I'll just copy-paste my comment.
```
I thought I was also trippin but it happens quite alot with season packs.
I have sorting as Show Name/Season 1/Show Name - S01E05.mkv
When this error happens, mostly the Show Name also fails to rename properly and gets truncated and what was Show.Name.S01.S01.1080p.AMZN.WEB-DL.DDP5.1.H.264 ends up as Show Name S01 1080P Amzn Web-Dl Ddp5 1 H.
The folder name clearly has S01 but the resulting structure is Show Name S01 1080P Amzn Web-Dl Ddp5 1 H/Season 2/...
Note the Season 2 and how the H.264 lost the .264.
And then the last episode of the pac is the one that mostly fails to rename for me and usually gets labelled as S02E64. Presumably the 2 and 64 came from H.264 and SAB took that as the season and episode numbers. So it ends up as
Show Name S01 1080P Amzn Web-Dl Ddp5 1 H - S02E64.mkv without the episode number and I have to rename that file with the last episode number.
The same nzb with nzbget and post processing enable works with no problem.
```

Safihre:
Can you share an example NZB? Here or at safihre@sabnzbd.org
And your exact Sorting string.

ItsME6969:
I just sent you a random NZB that I actually just tried right now before sending. I dont do alot of season packs but all the ones that I get have the same error even this one that I just randomly from the indexer so presumably it bugs out on all season packs. Single episodes work perfectly tho.

I'm on SAB 3.2.1
`%sn/Season %s/%sn - S%0sE%0e.%ext`

Safihre:
Wait, you are on 3.2.1? Why?

ItsME6969:
😱 lmao its what comes with the box man. Anyway the bug still persists in 3.5.3 which I have locally installed and I just tested and confirmed that with a different season pack. Same bug.

Safihre:
Oke, so I downloaded your example NZB and I get the result I expected.
As far as I know we don't support season-packs and never have, right @jcfp?
So it only renames the first one, and that's it.
[...]

PR #2461, opened 2023-02-17
Replace series/date/movie sorters with a generic one + season pack handling

### LINKED ISSUE #2143: Create a General sorter
Based on discussion in #1928.
Only if GuessIt agrees that something is the right category ("Movie" or "Serie"), the sorting will be applied.
For the general sorting like ` %dn.%ext` this is not important.

We should probably refactor Sorting to no longer have separate sorters, but to just have 1 type of sorter that can have a Movie or Serie hint. 
Then we should also allow to have any number of this general sorters, in case people want different sorter per category.

jcfp:
This changeset replaces the current setup of series, date, and movie sorters with a generic sorter. Users may define up to 100 different sorters, picking the categories and/or guessed job types each applies to. Sorters are tried in order of appearance: the first one that matches all requirements in terms of job type and categories is used, no further sorters considered. Sorters can be switched on and off individually and re-ordered by dragging and dropping in the interface.

* Any existing sorter configuration is auto-converted on startup to the new settings (and should be fully compatible);
* All sorter definitions can be defined, modified, and re-ordered through the API as well;
* Extra information is supplied as arguments to pre-queue scripts: resolution, year, month, day, and the guessed job type;
* Season pack handling is limited to single seasons, triggered automatically for jobs that appear to qualify, and can be turned off globally in config->specials;
* Sequential file handling now uses guessit as well ('cd' and 'part' properties); the old code used regexp but malfunctioned for anything other than 'cd';
* TestDownloadFlow was moved out of test_functional_downloads.py into testhelper and modified to support loading custom ini files, in order to facilitate testing the settings conversion code.

Note that test_consistency currently fails over the missing wiki entry for the new season sorting switch. Tests for the sorter renaming work reliably locally but seem to have random failures on the github infra across all Python versions (cause unknown, I suspect something to do with the use of fake filesystem).

Closes #2143 #2171


Safihre:
I rebased it on my change in develop that ignores the wiki test.
The whole code is a bit hard for me to judge, have to just try it and rely on the tests.

However, the UX needs to be a bit polished. The Sorting interface was already a bit cumbersome, but with the current change it becomes very cluttered.
The ability to drag and drop is nice, but does force the row-style display. In other places (except Categories) we just rely on having the user set a number, like for Servers and RSS-feed-rules. That way we can use a bit better the space.

thezoggy:
what about those that used sorting before

do we reset it all and make people setup fresh with new?
or try and migrate to equal settings?

Safihre:
@thezoggy please read the PR description first.

Safihre:
For now I will create a 4.0.0Alpha1 without this, but the goal is of course to include it for 4.0.0.
Hopefully I will finally have time to take a look at this peacefully.

Safihre:
Nice, much better!

Safihre:
Did the debugging info get you any hints why the tests keep randomly failing?

thezoggy:
if you need some recent examples of CD# as its pretty common with music releases:
```
Bodzza-Parcecs_EP-(CD177)-WEB-2023-PTC
Roger_Shah_and_Tom_Benscher-Dinner_For_Two-(MAGIC_ISLAND_CD11)-REPACK-CD-2022-OTTS
George_Palmer-Working_Man-(II_CD23)-CD-2022-YARD
```

and for games or disc rips:
```
LORD.OF.THE.RINGS.THE.RETURN.OF.THE.KING.PROPER.CD3-DEViANCE
Split.Second.1992.Disc1.1080p.2.Disc.Limited.Edition.GBR.Blu-ray.AVC.LPCM.2.0
Love.in.the.Moonlight.S01.DISC15.1080p.KOR.Blu-ray.AVC.LPCM.2.0-HDAsia
Goodfellas.1990.REMASTERED.DiSC2.COMPLETE.BLURAY-JRP
Highlander.The.Search.for.Vengeance.2007.DC.DELUXE.EDiTiON.DiSK1.COMPLETE.JPN.DVD9
```

Safihre:
@jcfp Should we report a bug to Guessit?

Safihre:
I pushed some small CSS changes.
Only thing that I would like to see added is the Preview option for existing sorters, so you can see the effect of a change.

Safihre:
And could you update the wiki? :)
Soring, but also Special page: https://github.com/sabnzbd/sabnzbd/actions/runs/4348005183/jobs/7596006734#step:7:15598

thezoggy:
it looks like the old sorting stuff is still shown on the api when pulling from get_config

`/api?mode=get_config&apikey=(removed)&output=json`
```
         "enable_tv_sorting":0,
         "tv_sort_string":"%sn/Season %s/%sn - S%0sE%0e - %en.%ext",
         "tv_categories":[
            "sonarr"
         ],
         "enable_movie_sorting":0,
         "movie_sort_string":"%title (%y)/%title (%y).%ext",
         "movie_sort_extra":" CD%1",
         "movie_categories":[
            "movies"
         ],
         "enable_date_sorting":0,
         "date_sort_string":"%t/%t - %y-%0m-%0d - %desc.%ext",
         "date_categories":[
            "*"
         ],
```

which *arr apps would be looking at to know if sorting is enabled/warn to disable. i've let them know that for sab 4.x to not look at those values

PR #2473, opened 2023-02-22
Make switchinterval configurable

puzzledsab:
I'd like to find out if it can make a difference.


mnightingale:
Could it use callbacks to set it at runtime?

Safihre:
@puzzledsab could you add the wiki entry so the tests don't fail on develop?

Safihre:
Can you please update the wiki? The tests on develop fail because of the missing options that you added. I think also recv_threads still needs to be added.

PR #2476, opened 2023-02-23
Decode articles as they are downloaded

mnightingale:
Decode articles within the pool at the same time as downloading.

It simplifies the main download > decode > assemble process and removes the possibility of downloading faster than decoding without needing to check queue lengths or sleep.

## Tasks

I will probably add more to this list it is mostly just a reminder for myself of things I've noticed but not investigated yet.

- [x] Check `__recv` logic
  - [x] Can it ever return `nw = None`
  - [x] Pass `NewsWrapper` as argument
  - [x] ~Consider moving exception handling into `NewsWrapper.recv_chunk`~
  - [x] Merge into `__process_server`
- [x] Check what will need locks and add them or more likely move some logic back out of the pool
  - [x] server.get_article
  - [x] BPSMeter.update
  - [x] __finish_connect_nw
  - [x] ~__request_article (read_fds - maybe atomic? - should probably use add_socket helper)~
  - [x] server.*
    - [x] ~have_stat~
    - [x] ~have_body~
    - [x] ~bad_cons~
    - [x] ~errormsg~
    - [x] ~active~
    - [x] busy_threads
    - [x] idle_threads
  - [x] article.nzf.nzo.update_download_stats
  - [x] ~remove_socket~
- [x] ~Maybe request next article before decode~

Strikethough means I didn't implement those ideas, or decided locks were not required.


sabnzbd/downloader.py
```diff
@@ -771,45 +740,32 @@ def run(self):
                 continue
 
             if self.recv_threads > 1:
-                for nw, bytes_received, done in self.recv_pool.map(self.__recv, read):
+                futures = {self.recv_pool.submit(self.__process_nw, self.read_fds.get(fds)) for fds in read}
```
Safihre, line 743: This line is already a bit too cryptic for me ;)

mnightingale: Hopefully clearer now 😄

I'm pretty close to being happy with the PR, I'll hopefull finish it tomorrow.

mnightingale:
image

image

Definitely a race condition, I'm missing a "Decoding finished" log for at least one of the files so that would indicate the assembler never received a `file_done = True` queue item.

I'm testing against a real server with 8 recv_threads and lots of connections to try and trigger it more often but I'm not sure if its actually a new issue, I think the same could happen from multiple decoder threads and the tighter download/decode/assemble just makes it happen more often.

I'm a bit stumped at the moment on what causes it.

puzzledsab:
I assume it's because you're running a lot more of the code concurrently than the threadpool version. It's is only multiprocessing `recv_chunk()` while yours is doing everything in `process_nw()`. There are multiple things there that probably shouldn't be run concurrently, particularly anything related to the bpsmeter and the whole get_article chain.

mnightingale:
> I assume it's because you're running a lot more of the code concurrently than the threadpool version. It's is only multiprocessing `recv_chunk()` while yours is doing everything in `process_nw()`. There are multiple things there that probably shouldn't be run concurrently, particularly anything related to the bpsmeter and the whole get_article chain.

I was aware that I'd created threading problems to solve but needed to go through each call to check them, I had already identified the bpsmeter would need locks or more likely just moving it back to the main download thread. I do want to keep the requesting of the next article in the pool because it can takes a while and other connections can carry on with their own work.

For the current issue I think you're right and it's actually missing requesting some articles because of race conditions in `server.get_article()` call that's causing me problems, I got too focussed on the end state (articles left in cache and no file_done) to look further up the chain.
Most of the other calls are just dealing with the individual NewsWrapper and it's state so that's fine without locks.

From some early testing of about 20 downloads putting a lock on `get_article` seems to have resolved it, it was happening almost every time previously.

I'll push a commit using `@NzbQueueLocker` which is used in several other places in downloader so I think it's the correct existing lock to use.

Thanks for your input.

Safihre:
@mnightingale I'm also surprised to see 104%. Did you fix that? As far as I know our counts have been correct for quite some versions now (there were some problems in the past with 100%+ being displayed).

mnightingale:
> @mnightingale I'm also surprised to see 104%. Did you fix that? As far as I know our counts have been correct for quite some versions now (there were some problems in the past with 100%+ being displayed).

Yes, I think because of the lack of lock it may have managed to download the same article multiple times throwing off the totals, with the lock and a breakpoint on queue end I've verified multiple times it's correctly at 100%.

sabnzbd/downloader.py
```diff
@@ -818,7 +781,7 @@ def __handle_recv_result(self, nw: NewsWrapper, bytes_received: int = 0, done: b
         if nw.status_code != 222 and not done:
```
Safihre, line 781: GitHub won't let me comment on the line above, so I'll do it here:
We probably need a lock around `sabnzbd.BPSMeter.update` since otherwise the counts could get out of balance due to threading.

sabnzbd/downloader.py
```diff
@@ -875,11 +838,49 @@ def __handle_recv_result(self, nw: NewsWrapper, bytes_received: int = 0, done: b
...
+                        "Delayed - %d seconds - Assembler queue: %d",
+                        logged_counter / 10,
+                        sabnzbd.Assembler.queue.qsize(),
+                    )
+
+                # Wait and update the queue sizes
+                time.sleep(0.1)
+                logged_counter += 1
+
     def __finish_connect_nw(self, nw: NewsWrapper) -> bool:
```
Safihre, line 884: I think we will also need locks for error handling, there's all kinds of setting error messages and increasing/decreasing server threads that need to be atomic to work correctly.

sabnzbd/downloader.py
mnightingale: I've moved this one back to the main downloader thread

sabnzbd/downloader.py
mnightingale: I've added the things I've identified to handle to the list.

I think some operations are atomic and don't need locks like `server.bad_cons = 0`.

What about the read_fds dictionary?
`self.read_fds[nw.nntp.fileno] = nw`
I could just use synchronized on add_socket and remove_socket but as far as I can tell they are safe without?

Some like `__finish_connect_nw` I'll just stick a synchronized decorator around since they are rarely called anyway.

`article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, nw.data_position)`
I think a global lock in downloader or a lock per nzo would probably also work but likely not worth it.

sabnzbd/downloader.py
```diff
@@ -861,25 +829,68 @@ def __handle_recv_result(self, nw: NewsWrapper, bytes_received: int = 0, done: b
             server.bad_cons = 0
             server.errormsg = server.warning = ""
 
-            # Update statistics and decode
-            article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, nw.data_position)
-            self.decode(article, nw.get_data_buffer(), nw.data_position)
+            # Update statistics
+            with sabnzbd.decorators.NZBQUEUE_LOCK:
+                article.nzf.nzo.update_download_stats(sabnzbd.BPSMeter.bps, server.id, nw.data_position)
```
mnightingale, line 834: Do you think this can move this above the status_code check where a lock is already acquired, with an `if done:`

`done` is true if `\r\n.\r\n` is found which I think is only at the end of of an article?

Safihre: Yes I think so.

mnightingale:
I've added locks where I think they are required, I've use `with lock` but can swap to try finally if you'd prefer the decorators seem to think it's faster (not sure if still true).

The main ones I haven't added locks for are what I think are atomic/thread-safe and in some cases are already being accessed from other threads (each connection starts a thread and eventually calls add_socket)

Added a comment above for a change to avoid an extra lock.

Safihre:
I'll take another look.
So how does this change perform? Is there improvement? 😊

Safihre:
Great. Indeed the simpler data flow of this approach without Decoder queue is also what appeals to me.

sabnzbd/downloader.py
```diff
@@ -771,49 +747,44 @@ def run(self):
...
-                    nw, bytes_received, done = self.__recv(selected)
-                    if bytes_received > last_max_chunk_size:
-                        last_max_chunk_size = bytes_received
-                    self.__handle_recv_result(nw, bytes_received, done)
-                    if self.bandwidth_limit and bytes_received:
-                        self.__check_speed()
-
-    def __recv(self, selected):
-        nw = None
+                    if nw := self.read_fds.get(selected):
```
Safihre, line 762: Why do you use `get` and not just `self.read_fds[selected]`? Can it be `None`? If so, maybe it needs a comment how that could be possible.

sabnzbd/downloader.py
```diff
@@ -771,49 +747,44 @@ def run(self):
...
+                        self.process_nw(nw)
+
+            self.__check_speed()
+            self.__check_assembler()
+
+    def process_nw(self, nw: NewsWrapper) -> None:
```
Safihre, line 768: Nit-pick: I don't think we add typing for when a function returns `None`.

sabnzbd/downloader.py
```diff
@@ -771,49 +747,44 @@ def run(self):
...
-            if done:
-                self.__reset_nw(nw, "server closed connection", wait=False)
+            self.__reset_nw(nw, "server closed connection", wait=False)
             return
 
         article = nw.article
         server = nw.server
-        sabnzbd.BPSMeter.update(server.id, bytes_received)
+
+        with sabnzbd.decorators.NZBQUEUE_LOCK:
```
Safihre, line 781: I think we need a different lock for this.
The NZBQUEUE_LOCK is meant for things that interact with the queue, like getting articles or removing NZO's. 
Seems we just need a lock "local" to the Downloader for these things?

sabnzbd/downloader.py
```diff
@@ -861,25 +832,64 @@ def __handle_recv_result(self, nw: NewsWrapper, bytes_received: int = 0, done: b
...
             # Request a new article immediately if possible
             if nw.connected and server.active and not (self.paused or self.shutdown or self.paused_for_postproc):
                 nw.article = server.get_article()
                 if nw.article:
                     self.__request_article(nw)
                     return
-            server.busy_threads.remove(nw)
-            server.idle_threads.append(nw)
+
+            with sabnzbd.decorators.NZBQUEUE_LOCK:
```
Safihre, line 851: Same here, we could do with a Downloader-specific lock.

sabnzbd/downloader.py
mnightingale: It's keeping the same behaviour @puzzledsab added in https://github.com/sabnzbd/sabnzbd/pull/2438/commits/1edbe38859850577aa21d623f206670a76c1ff24 since that was added I presumed they found throwing KeyError was possible - I'm not sure how but presumably a server can be disabled or something when the downloader thread has already selected it?

So I preferred to push that check into the downloader and use `get` which returns None instead of needing to add try/except in the main downloader thread.

sabnzbd/downloader.py
```diff
@@ -221,6 +223,7 @@ def request_info(self):
             self.request = True
             Thread(target=self._request_info_internal).start()
 
+    @NzbQueueLocker
     def get_article(self):
```
mnightingale, line 227:
```suggestion
    @synchronized(DOWNLOADER_LOCK)
    def get_article(self):
```

Change this one to use the new lock also?

sabnzbd/downloader.py
mnightingale: Perhaps just a comment "# The connection may no longer be available, such as if the server is disabled" or something?

sabnzbd/downloader.py
mnightingale: Added DOWNLOADER_LOCK for both and added a missing one on `__reset_nw`

sabnzbd/downloader.py
puzzledsab: I don't think it should be possible. I just did it to close absolutely every possibility of the try not catching anything that could fail inside the thread.

sabnzbd/downloader.py
Safihre: Indeed!

sabnzbd/downloader.py
Safihre: It can't happen, it can just use `self.read_fds[selected]`!

Safihre:
@puzzledsab / @mnightingale I wonder if we need to change the `sleep_time` stuff. Because indeed this new multi-threaded approach uses more CPU than setting 1 recv_thread. However, I can increase my `downloader_sleep_time` bij 10x and it still downloads at the same speed and uses almost no CPU at all.

@mnightingale It's ready for review right? Since there's still some checkboxes in the start post.

puzzledsab:
The default can probably be increased somewhat but it can be risky to increase it too much. Now that we read the entire buffer every time it's no problem on a fast CPU or a slow connection but if it sleeps to long then the buffers can fill up and the speed is reduced. It depends on the speed of the CPU (clearing the buffers), the transfer speed (filling the buffers) and the number of connections. More connections results in more total buffer space that can be filled before the next loop.

A higher sleep time will be a bigger problem with this PR because it doesn't decode while the downloader loop is sleeping.

mnightingale:
> @puzzledsab / @mnightingale I wonder if we need to change the `sleep_time` stuff. Because indeed this new multi-threaded approach uses more CPU than setting 1 recv_thread. However, I can increase my `downloader_sleep_time` bij 10x and it still downloads at the same speed and uses almost no CPU at all.

A lot of the sleep code is a bit of mystery to me, I get that one reason it's there is to encourage other threads to run but I'm not sure sleeping so much is actually necessary - without it the download thread can spend it's time waiting on the select.select call without CPU usage (it also releases the GIL).

In some cases could it not just make CPU usage more spikey, e.g. reading 10MB in 10 calls vs 100 is the CPU not doing a similar number of operations over the same span of time just less spread out?

> @mnightingale It's ready for review right? Since there's still some checkboxes in the start post.

Pretty much, there are just a couple of minor things for me to address from past reviews.

I've also just had an idea for perhaps a simple way to achive a checkpoint I'd ruled out "Maybe request next article before decode", instead of decoding in `process_nw` I could just submit a separate decoder task to the pool, that way the request for the next article is sent immediately and I'd expect all the download tasks will complete before the decode task so it keeps the network busy. I just need to test its performance before pushing.

I've updated the checkboxes, the strikethroughed points are things I didn't change such as where I don't think locks are required - usually because I think they are atomic (often other threads access them without locks currently).

The only other thing is coming up with a way to keep a CPU bottleneck wrench message as discussed in the A2 discussion.
I'm hoping I can have a check for something like if for X number of consecutive articles spend >50% of time was spent on decoding then we are CPU limitted, because that's kind of similar to the current behaviour (decode slower than download) but I need to test it on a CPU limited device.

> A higher sleep time will be a bigger problem with this PR because it doesn't decode while the downloader loop is sleeping.

This is probably the biggest difference, when the downloader thread sleeps there isn't much else to be processed anymore; assembly will be IO bound since crc calculations are cheap similarly with unpacking.

thezoggy:
why do we sleep at all? is it to try and mask the need to have a lock or deal with timing issues on 'slow' hardware?

Safihre:
No, to save CPU on fast systems actually. Otherwise it's just looping checking for new stuff while there's nothing much there yet. It's been like that for a couple of years already, just the system changed.

thezoggy:
@Safihre yes which is why I was asking about sleeping now we know when stuff is active and if cache is full and then have the user configured one. My comment was in regards of this change with the decoding. Wouldn't we want to not sleep and just let the locks/queue handle slowing thing down when it needed to? (We would sleep in the main polling loop)

mnightingale:
I think using `time.perf_counter_ns()` kind of works as an estimate of the decoder speed.

```py
# Decode
decode_start_ns = time.perf_counter_ns()
self.decode(article, nw.get_data_buffer())
decode_duration_ns = time.perf_counter_ns() - decode_start_ns

if nw.data_position and decode_duration_ns:
    logging.info(
        "Decode %.2f ms = %.0f MB/s",
        decode_duration_ns / 1_000_000,
        nw.data_position / decode_duration_ns * 1_000
    )
```

I was thinking of comparing `BPSMeter.bps` to the calculated decode speed, if they are within a set distance then download speed is probably being bottlenecked by the decoding speed.

@puzzledsab this seems to be your kind of thing, do you think that works?

Perhaps I could measure an article every few seconds? which would also avoid increasing the count very quickly.

This is on RPI 4 downloading at ~100 MB/s, so even though it has one decode at 48 MB/s I don't think that should be flagged. Maybe if there are >=3 in 5 seconds or something?

>2023-03-02 13:26:19,589::INFO::[downloader:844] Decode 2.75 ms = 301 MB/s
2023-03-02 13:26:19,601::INFO::[downloader:844] Decode 4.82 ms = 171 MB/s
2023-03-02 13:26:19,602::INFO::[downloader:844] Decode 4.09 ms = 202 MB/s
2023-03-02 13:26:19,612::INFO::[downloader:844] Decode 4.64 ms = 178 MB/s
2023-03-02 13:26:19,614::INFO::[downloader:844] Decode 4.28 ms = 193 MB/s
2023-03-02 13:26:19,624::INFO::[downloader:844] Decode 4.75 ms = 174 MB/s
2023-03-02 13:26:19,625::INFO::[downloader:844] Decode 3.98 ms = 207 MB/s
2023-03-02 13:26:19,636::INFO::[downloader:844] Decode 4.68 ms = 176 MB/s
2023-03-02 13:26:19,637::INFO::[downloader:844] Decode 3.94 ms = 210 MB/s
2023-03-02 13:26:19,656::INFO::[downloader:844] Decode 4.68 ms = 177 MB/s
2023-03-02 13:26:19,659::INFO::[downloader:844] Decode 5.22 ms = 158 MB/s
2023-03-02 13:26:19,659::INFO::[downloader:844] Decode 17.21 ms = 48 MB/s
2023-03-02 13:26:19,674::INFO::[downloader:844] Decode 7.82 ms = 106 MB/s
2023-03-02 13:26:19,675::INFO::[downloader:844] Decode 6.31 ms = 131 MB/s
2023-03-02 13:26:19,675::INFO::[downloader:844] Decode 4.07 ms = 203 MB/s

I've dismissed the "Maybe request next article before decode" idea, decoding is so fast that submitting more tasks has too much overhead and as you can see ~5ms is a tiny delay before requesting an article which will be even smaller on faster devices.

Safihre:
Let's not include this CPU load detection in the current PR, so we can get it merged :)
If @mnightingale could make the last tiny changed, it's done!

mnightingale:
Understood, those last couple of pushes should be everything.

puzzledsab:
I'm not sure if it's useful for end users to measure the ydecode speed or other indirect CPU usage measurements. It might be in some cases as part of a debugging process but if it is the only measurement that normal users see then I think could be more confusing than helpful. I think the decrypting and reading of the data from the sockets usually use more CPU than ydecode, particularly when using default encryption. It's not all that useful the way it is in develop either but it's already there and doesn't use much CPU.

If users aren't getting maximum speed or want to minimize CPU usage then they'll have to use try and fail with the settings. The process CPU usage can be seen using top or task manager.

mnightingale:
I'm just trying to come up with a way to reintroduce a `sabnzbd.BPSMeter.delayed_decoder += 1` call, I wouldnt actually log the above messages.

As you identify if there is a CPU bottleneck it would manifest as either SSL encyption or yEnc decoding, the previous behaviour would increment the counter if the decode queue was large so it never flagged the download speed as a potential issue.

Because download + decode will now be so linked it's quite difficult to come up with a meaningful measurement that's actaually identifying a CPU bottleneck and not the network. Perhaps I could measure the call to read the socket, but I'm not sure what that's actaully measuring.

With the approach above I was trying to identify if the average overall speed via BPSMeter is close to just the decode speed then we could flag that as a CPU bottleneck, e.g. if downloading+decode = 50MB and decode = 55 MB/s they are close so potentially the decode speed is limitting the overall speed.

Maybe just something simple like checking `psutil.cpu_percent()` periodically, but I'm not sure how well that'd work with the way we or Python use threads.

puzzledsab:
But you're only measuring the speed for a particular thread. As I understand it, all the connections are being handled concurrently in this PR so they may all try to decode at the same time, sharing the CPU time. If they do then it doesn't matter if the decode speed in a single thread is much lower than the total download speed.

Issue #2484, opened 2023-03-04
Proposal: "Deobfuscate final filenames" on by default


sanderjo:
How about setting "Deobfuscate final filenames" on by default? For example in SAB 4.0.0

There was one reason to not turn it on yet, but I can't remember it, so I'm not sure if it is still a valid blocker.

@Safihre do you know it?

Safihre:
Seems like a valid option.
You can submit a PR to enable it by default.

PR #2493, opened 2023-03-09
Deobfuscate filenames ON by default

sanderjo:
Solves https://github.com/sabnzbd/sabnzbd/issues/2484


sanderjo:
Main thing: I created a file My_Test_Download.bin (instead of testfile.bin), which is not obfuscated, to avoid debofuscating and getting an unpredictable name (because NZB name is random). That file is in the new data test sets, and needed changes in other test that test raw things.

@shypike is the OK for you?

I have temporarily commented out test_par2_file -> test_parse_par2_file because I wanted to the main tests to go right.

Safihre:
Yes that sounds like a good solution.

Safihre:
Looks good. If you fix the last test that you commented-out we can merge.

PR #2496, opened 2023-03-10
only save series_info in the history db for job type "tv"

jcfp:
This restores previous behaviour, when analyse_show called the series sorter directly. The series info in the db is only used to detect series dupes based on name/season/episode, so storing it for other job types serves no purpose.

### DIFF
--- a/sabnzbd/database.py
+++ b/sabnzbd/database.py
@@ -474,8 +474,8 @@ def build_history_info(nzo, workdir_complete="", postproc_time=0, script_output=
 
     # Analyze series info only when job is finished
     series = ""
-    if series_info:
-        seriesname, season, episode = sabnzbd.newsunpack.analyse_show(nzo.final_name)[:3]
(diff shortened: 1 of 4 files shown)


sabnzbd/database.py
```diff
@@ -474,8 +474,8 @@ def build_history_info(nzo, workdir_complete="", postproc_time=0, script_output=
 
     # Analyze series info only when job is finished
     series = ""
-    if series_info:
-        seriesname, season, episode = sabnzbd.newsunpack.analyse_show(nzo.final_name)[:3]
+    if series_info and (show_analysis := sabnzbd.newsunpack.analyse_show(nzo.final_name))[10] == "tv":
```
Safihre, line 477: I'm a bit allergic to these `[10]`. Isn't there a better way? Maybe return some dataclass or something.

Safihre:
I guess it will never be very pretty code..

PR #2497, opened 2023-03-10
Make cleanup_list use scandir and keep main nzb directory even if it's empty

puzzledsab:
https://forums.sabnzbd.org/viewtopic.php?t=26160


puzzledsab:
It's not strictly necessary to keep it but I think it's a somewhat less confusing result for the user. It's an indication that it finished, even if it's empty.

thezoggy:
Might be good for those that symlink to another dir to not have it removed in case other apps expect it to be there to dump into it

PR #2499, opened 2023-03-10
change regex for hash from md5 to more generic to catch more hashes, and hide email_from as well

thezoggy:
in response to: https://forums.sabnzbd.org/viewtopic.php?p=129264#p129264

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -629,10 +629,10 @@ LOG_API_JSON_RE = re.compile(rb"'(apikey|api)': '[\w]+'", re.I)
 LOG_USER_RE = re.compile(rb"(user|username)\s?=\s?[\S]+", re.I)
 LOG_PASS_RE = re.compile(rb"(password)\s?=\s?[\S]+", re.I)
 LOG_INI_HIDE_RE = re.compile(
-    rb"(email_pwd|email_account|email_to|pushover_token|pushover_userkey|pushbullet_apikey|prowl_apikey|growl_password|growl_server|IPv[4|6] address)\s?=\s?[\S]+",
+    rb"(email_pwd|email_account|email_to|email_from|pushover_token|pushover_userkey|pushbullet_apikey|prowl_apikey|growl_password|growl_server|IPv[4|6] address)\s?=\s?[\S]+",
     re.I,
 )
-LOG_HASH_RE = re.compile(rb"([a-fA-F\d]{25})", re.I)
+LOG_HASH_RE = re.compile(rb"([a-zA-Z\d]{25})", re.I)
 
 
 def _api_showlog(name, kwargs):


sabnzbd/api.py
```diff
@@ -629,10 +629,11 @@ def _api_warnings(name, kwargs):
 LOG_USER_RE = re.compile(rb"(user|username)\s?=\s?[\S]+", re.I)
 LOG_PASS_RE = re.compile(rb"(password)\s?=\s?[\S]+", re.I)
 LOG_INI_HIDE_RE = re.compile(
-    rb"(email_pwd|email_account|email_to|pushover_token|pushover_userkey|pushbullet_apikey|prowl_apikey|growl_password|growl_server|IPv[4|6] address)\s?=\s?[\S]+",
+    rb"(email_pwd|email_account|email_to|email_from|pushover_token|pushover_userkey|pushbullet_apikey|prowl_apikey|growl_password|growl_server|IPv[4|6] address)\s?=\s?[\S]+",
     re.I,
 )
 LOG_HASH_RE = re.compile(rb"([a-fA-F\d]{25})", re.I)
+LOG_FEED_API_RE = re.compile(rb"(uri\s=\s.*r=)([a-z0-9]{32})", re.I)
```
Safihre, line 636: Why would these not be caught by the line above it?

thezoggy:
Different hash format. The current one does match most things but failed on this one because geek doesnt do hex so you need a-z to cover it. Since I didn't want to change the existing hash one to be even more zealous on matching I figure it was low stakes to just target rss uri as a last resort catch

sabnzbd/api.py
thezoggy: Different hash format. The current one does match most things but failed on this one because geek doesnt do hex so you need a-z to cover it. Since I didn't want to change the existing hash one to be even more zealous on matching I figure it was low stakes to just target rss uri as a last resort catch

Safihre: Can you give an example of the API key?

thezoggy: `mRu0nMyGsm5wyBFrprcU7L71GUBF2Qcg`

Safihre: Ah, now I see. Then let's just update the existing one to include `a-z` so it's universal.

thezoggy: will do

thezoggy: changing existing hash format to a-z makes it a bit over zealous for my taste...   thoughts?
```
2023-03-09 10:52:42,050::DEBUG::[deobfuscate_filenames:111] Checking: <HASH>WENLE
2023-03-09 10:52:42,050::DEBUG::[deobfuscate_filenames:167] Obfuscated (default)
2023-03-09 10:52:42,054::DEBUG::[deobfuscate_filenames:111] Checking: <HASH>ggm3d
2023-03-09 10:52:42,055::DEBUG::[deobfuscate_filenames:167] Obfuscated (default)
2023-03-09 10:52:42,057::DEBUG::[deobfuscate_filenames:111] Checking: <HASH>donjx
2023-03-09 10:52:42,057::DEBUG::[deobfuscate_filenames:167] Obfuscated (default)
2023-03-09 10:52:42,060::DEBUG::[deobfuscate_filenames:111] Checking: <HASH>GQ
2023-03-09 10:52:42,060::DEBUG::[deobfuscate_filenames:167] Obfuscated (default)
2023-03-09 10:52:42,062::DEBUG::[deobfuscate_filenames:111] Checking: <HASH>6aa163e
...
2023-03-09 19:19:37,392::INFO::[nzbstuff:1740] Detected filename based on par2: We.Baby.Bears.S01E37.Lighthouse.1080p.CN.WEB-DL.AAC2.0.x264-TCV.part07.rar -> <HASH>5U0Cn.part07.rar
2023-03-09 19:19:37,393::DEBUG::[assembler:79] Decoding part of \\?\C:\Users\<USERNAME>\downloads\incomplete\We.Baby.Bears.S01E37.1080p.CN.WEB-DL.AAC2.0.x264-TCV\<HASH>5U0Cn.part07.rar
...
2023-03-12 01:30:39,760::DEBUG::[newsunpack:2006] Quick-check of file <HASH>8.part22.rar OK
2023-03-12 01:30:39,760::DEBUG::[newsunpack:2006] Quick-check of file <HASH>8.part13.rar OK
2023-03-12 01:30:39,761::DEBUG::[newsunpack:2006] Quick-check of file <HASH>8.part15.rar OK
```

Safihre: That’s fine, they are hashes anyway.

mnightingale:
Have alternatives to regular expression ever been considered for cleaning logs? they are not particularly easy to decipher or identify excess matching and there are bound to eventually be cases not covered.

For instance a regular expression to extract http/https URLs, parse them with urllib.parse or similar then either remove the whole path or query parts or attempt to strip sensitive parts as now?

thezoggy:
> Have alternatives to regular expression ever been considered for cleaning logs? they are not particularly easy to decipher or identify excess matching and there are bound to eventually be cases not covered.
> 
> For instance a regular expression to extract http/https URLs, parse them with urllib.parse or similar then either remove the whole path or query parts or attempt to strip sensitive parts as now?

not sure that would be good for performance to try and do that. but i wouldnt be opposed to more explicit log cleanse like sonarr does:
https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Common/Instrumentation/CleanseLogMessage.cs#L13

just for our purposes, the log cleanse actually does two parts.. the log and the config. 
the config bits we easily just say nuke out the values for x keys. 
but like in the case of the rss uri, I didnt want to nuke out the whole value just a piece of it (although prob could just nuke out the key and force user to share explicitly that if needed for triage.. as the url itself may also not be wanted to be shared as some sites are anti them being mentioned)

originally i had regex to do just the part but it was forgone to just leverage existing 'hash' to also take care of that. which that hash is what does the heavy lifting and takes care of some apkeys some passwords and then of course all the 'hashes' in filenames and such. since it isnt the full length of the hash there is some bits that remains so can still somewhat follow along with the reduced version of the hash. 

so now that ive said all that, do we want to keep the current hash match to md5 like we had.. and i just nuked out the `uri` field rather than worrying about just a segment of it?

mnightingale:
I like the way Sonarr breaks them up by application/service, it keeps them individually more readable and easier to maintain.

I think even their first pattern is matching query arguments a bit better than SAB, e.g. ?apikey=... or &apikey=... - as far as I can see SAB is just looking for apikey=
By requiring it starts with ? or & I think that would prevent the overmatching.

The URL shared on the forum link I think is a newznab indexer which seem to default to using 'r' for the key so we should search for `?r=...` and `&r=...` to mask.

I do think alone just matching any a-z characters of length 25 is too eager.

mnightingale:
Example of Sonarr' first expression converted to Python (with r= added)

```python
LOG_URL_RE = re.compile(
    r"""(?<=\?|&)(apikey|token|passkey|auth|authkey|user|uid|api|r|[a-z_]*apikey|account|passwd)=(?P<secret>[^&=""]+?)(?=[ ""&=]|$)""",
    re.I,
)

test = LOG_URL_RE.sub('\\1=<APIKEY>', 'https://api.nzbgeek.info/rss?t=-2&limit=200&dl=1&del=1&r=supersecretkey&apikey=anotherkey')

print(test)
```

```bash
https://api.nzbgeek.info/rss?t=-2&limit=200&dl=1&del=1&r=<APIKEY>&apikey=<APIKEY>
```

Safihre:
There's no need to over complicate things like this. The current regex satisfies the needs very well.

PR #2501, opened 2023-03-13
Corrected msgids -  msgid "Pause * prioirty jobs" spelling error

jakepez:
Corrected spelling errors in “Prioirty” msgid definitions. This pull request should update all definition language definitions and calls to them.

### DIFF
--- a/po/main/SABnzbd.pot
+++ b/po/main/SABnzbd.pot
@@ -1746,32 +1746,32 @@ msgstr ""
 
 #. Config->Scheduler
 #: sabnzbd/skintext.py
-msgid "Pause low prioirty jobs"
+msgid "Pause low priority jobs"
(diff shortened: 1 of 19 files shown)


Safihre:
We already fixed this by patching the English "translation" `en.po`. Probably you didn't run the compiling of the languages? 
However, it would be good to really fix it indeed. For context, this wasn't possible before because the translation platform wouldn't pick up the source changes of msgid in language files. But now it should do that due to a recent change I made.

Could you remove the strings you fixed from `en.po`?

jakepez:
I've reverted the change to en.po.

I didn't run the compiling of the languages - just happened to be poking around looking at changes for the 4.0 release and this stuck out to me as something I could submit an update for.

PR #2509, opened 2023-03-20
Check speedlimit after each recv

mnightingale:
https://forums.sabnzbd.org/viewtopic.php?p=129310

I don't think the pooled approach ever handled this correctly and from recent refactoring it was also changed on the single thread way.

Currently we are receiving from up to all the connections before applying speed limits, resulting in spikey traffic above the users defined limit.


mnightingale:
Already taken care of, it's within a `with DOWNLOADER_LOCK` block.

PR #2510, opened 2023-03-21
Set maximum soft sleep time to 0.15 seconds

puzzledsab:
Another minor tweak as I saw some reports that downloads start up slower in 4.0. I suspect it's because of that other change which saves the first part to the correct file instead of a temporary file in combination with the soft sleep that would sleep 0.1 second for each item. If the user has a lot of active connections then a lot of first parts may finish in a single loop. The change to sleep less than 1/4th as long for each item probably helps a lot but there's no point in sleeping longer because it only happens if the queue is full. In that case the sleep loop kicks in if it doesn't clear during the soft sleep.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -763,7 +763,7 @@ class Downloader(Thread):
 
             # Check the Assembler queue to see if we need to delay, depending on queue size
             if (assembler_level := sabnzbd.Assembler.queue_level()) > SOFT_QUEUE_LIMIT:
-                time.sleep((assembler_level - SOFT_QUEUE_LIMIT) / 4)
+                time.sleep(min((assembler_level - SOFT_QUEUE_LIMIT) / 4, 0.15))
                 sabnzbd.BPSMeter.delayed_assembler += 1
                 logged_counter = 0
 


sabnzbd/downloader.py
```diff
@@ -763,7 +763,7 @@ def run(self):
 
             # Check the Assembler queue to see if we need to delay, depending on queue size
             if (assembler_level := sabnzbd.Assembler.queue_level()) > SOFT_QUEUE_LIMIT:
-                time.sleep((assembler_level - SOFT_QUEUE_LIMIT) / 4)
```
Safihre, line 766: Seems we can just make this sleep `0.15`?
Because it's only triggered if `>0.5`.
So for example it's 0.8, then we get `(0.8-0.5) / 4 = 0.075`. Seems like minor difference?

Safihre: And in that regard, what's then even still the point of this extra sleep? 
In the loop below we sleep `0.1` anyway?
Should we just trigger the loop below before `1`, but already at `0.8` or `0.9`?

puzzledsab: Triggering the loop earlier would be the same as setting a smaller assembler queue limit.

The intention is to increase the delays more gradually. At the moment the queue size is 12 and SOFT_QUEUE_LIMIT = 0.5, so it triggers first at 7 elements for (7/12 - 0.5) / 4 = 0.021 seconds and increases in steps of that size. I'm hoping this will let it find a sweet spot for delay size where the download speed matches the write speed without hitting the hard limit. I think this will reduce the chance of stalling the download speed.

For instance, if the soft limit normally keeps the queue at 8 items then there is still room for 4 more items. If the disk is a bit busy and writes slow down for a short time then it can fill up the remaining slots before stopping completely. With no soft limit it will continuously staying just below the maximum limit and sleep for at least 0.1 sec, which may be a bit long. If the writing stalls then so will instantly the downloading. When the downloading continues it takes a little bit of time before the server starts sending at full speed again.

Setting the bandwidth limit seems to let the average speed be higher in some cases where the disk speed is the limiting factor, presumably because it keeps the assembler queue from filling up. Hopefully the soft limit will give the same result.

Issue #2515, opened 2023-03-27
TypeError: expected str, bytes or os.PathLike object, not bool


uhthomas:
```
ERROR a few seconds ago Error while adding <redacted> removing
Traceback (most recent call last):
  File "/app/sabnzbd/sabnzbd/nzbparser.py", line 319, in process_single_nzb
    nzo = nzbstuff.NzbObject(
  File "/app/sabnzbd/sabnzbd/nzbstuff.py", line 744, in __init__
    admin_dir = os.path.join(self.download_path, JOB_ADMIN)
  File "/usr/lib/python3.10/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not bool
```

Seems like a permission error, but thought I'd open this issue anyway as it is probably a good idea to not panic.

```
ERROR a minute ago Failed making (/media/downloads/incomplete/<redacted>)
Traceback (most recent call last):
  File "/app/sabnzbd/sabnzbd/filesystem.py", line 717, in create_all_dirs
    os.mkdir(path_part_combined)
PermissionError: [Errno 13] Permission denied: '/media/downloads/incomplete/<redacted>'
```

uhthomas:
Whilst Googling, I found https://forums.sabnzbd.org/viewtopic.php?t=26097. The comments suggest that `JOB_ADMIN` is the problem here?

May be unrelated, but I did change the paths in the config UI from the default to something else. Maybe a culmination of permissions and this is enough to get SABnzbd into this state?

Safihre:
The problem is that when there's something wrong with creating the temporary directly, this function returns False:
https://github.com/sabnzbd/sabnzbd/blob/d22f2296c24ea2c705ad5109a9e53214bdb32068/sabnzbd/nzbstuff.py#L746

We just don't check the result of that. We could do that, but it would still result in some form of exception. However, we would be able to display a nicer error.

uhthomas:
Yeah, I see it now.

https://github.com/sabnzbd/sabnzbd/blob/d22f2296c24ea2c705ad5109a9e53214bdb32068/sabnzbd/filesystem.py#L733

PR #2516, opened 2023-03-27
Check for errors when creating the download path

uhthomas:
Fixes https://github.com/sabnzbd/sabnzbd/issues/2515.

I've checked other usages of the function with `create_dir=True` and this seems to be the only place which doesn't check the return value. Additionally, added a unit test for `get_unique_dir` to verify behaviour and prevent regressions.

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -743,7 +743,10 @@ class NzbObject(TryList):
         else:
             # Determine "incomplete" folder
             self.download_path = os.path.join(cfg.download_dir.get_path(), self.work_name)
-            self.download_path = long_path(get_unique_dir(self.download_path, create_dir=True))
+            self.download_path = get_unique_dir(self.download_path, create_dir=True)
+            if not self.download_path:
+                raise NzbEmpty
+            self.download_path = long_path(self.download_path)
             set_permissions(self.download_path)
 
         # Always create the admin-directory, just to be sure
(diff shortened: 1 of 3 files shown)


sabnzbd/nzbstuff.py
```diff
@@ -743,7 +743,11 @@ def __init__(
         else:
             # Determine "incomplete" folder
             self.download_path = os.path.join(cfg.download_dir.get_path(), self.work_name)
-            self.download_path = long_path(get_unique_dir(self.download_path, create_dir=True))
+            self.download_path = get_unique_dir(self.download_path, create_dir=True)
+            if not self.download_path:
+                logging.warning(T("Could not create download path: %s"), self.download_path)
```
Safihre, line 748: Doesn't create_all_dirs already throw AN ERROR? 
https://github.com/sabnzbd/sabnzbd/pull/2516/files#diff-db363ac9b90f46184c420b28c372136cff39bb5de103238aae125e0d1f4a1471R732

uhthomas: Yeah. I was following the convention established elsewhere where an error log is written if the value is false.

https://github.com/sabnzbd/sabnzbd/blob/d22f2296c24ea2c705ad5109a9e53214bdb32068/sabnzbd/postproc.py#L701-L709

Safihre: I don't want to introduce a new translation just for this problem. Let's keep it simple:
```suggestion
            self.download_path = long_path(get_unique_dir(self.download_path, create_dir=True))
            if not self.download_path:
                raise NzbEmpty
```
Also I noticed it should be `raise` and not `return`.

Issue #2500, opened 2023-03-11
Updating Internet Speed Test tool


D0cR3d:
I noticed the Speed Test tool is consistently on the low side and not reporting anywhere near the top speed that my ISP provides. After some debugging, it seems the files you host do not have the capability for handling extremely high-speed connections (1+ GB/s). 

### Speed Test using current URL files

```
Starting speed test:
DEBUG:root:Log level is DEBUG
DEBUG:root:Basic measurement, with small download:
DEBUG:root:URL is https://sabnzbd.org/tests/internetspeed/5MB.bin
DEBUG:root:Downloaded bytes: 5242880
DEBUG:root:Duration in seconds: 0.694500
DEBUG:root:Speed in MB/s: 7.20
DEBUG:root:https://sabnzbd.org/tests/internetspeed/20MB.bin
DEBUG:root:URL is https://sabnzbd.org/tests/internetspeed/20MB.bin
DEBUG:root:Downloaded bytes: 20971520
DEBUG:root:Duration in seconds: 0.717502
DEBUG:root:Speed in MB/s: 27.87
```

### Speed Test using a higher speed provider

```
Starting speed test:
DEBUG:root:Log level is DEBUG
DEBUG:root:Basic measurement, with small download:
DEBUG:root:URL is http://speedtest.tele2.net/10MB.zip
DEBUG:root:Downloaded bytes: 10485760
DEBUG:root:Duration in seconds: 0.203999
DEBUG:root:Speed in MB/s: 49.02
DEBUG:root:http://speedtest.tele2.net/1GB.zip
DEBUG:root:URL is http://speedtest.tele2.net/1GB.zip
DEBUG:root:Downloaded bytes: 1073741824
DEBUG:root:Duration in seconds: 12.032498
DEBUG:root:Speed in MB/s: 85.10
```

I'm in Europe, and so is Tele2, so I know geo location wise this is going to provide a better result for me. I'm not sure if any geo location has been setup for the sab test files, and if those in other locations would still get improved results using the files from tele2.

You can read more about their testing feature [here](http://speedtest.tele2.net/). They do indicate that anyone in the world can use their speedtest.

thezoggy:
the sab website and content is served up by cloudflare cdn. if you are having slow xfer to one of their servers you can open a ticket with them to investigate. but most likely you prob are just seeing some sort of congestion from your isp to them

sanderjo:
[...]

PR #2524, opened 2023-04-04
bigger files (50 and 100MB) for internet speed testing

sanderjo:
solves https://github.com/sabnzbd/sabnzbd/issues/2500


Safihre:
@sanderjo  you have to use "closes" and not "solves" for the Github to auto close.

PR #2525, opened 2023-04-06
add per-category extension cleanup list

jcfp:
This change adds a per-category extension cleanup list (comma separated, empty by default), allowing for more fine-grained control of unwanted extensions.

If a cleanup list is defined for a given category, it overrules the global cleanup list for jobs in that category. In case no cleanup list is defined at category level, the global extension cleanup list is applied. Existing configurations should thus continue to work as-is and no settings conversion is needed either.

### DIFF
--- a/interfaces/Config/templates/config_cat.tmpl
+++ b/interfaces/Config/templates/config_cat.tmpl
@@ -19,6 +19,7 @@
                             <th>$T('mode')</th>
                             <th>$T('script')</th>
                             <th>$T('catFolderPath')</th>
+                            <th>$T('Cleanup List')</th>
                             <th colspan="2">$T('catTags')</th>
                         </tr>
                         <!--#end if#-->
@@ -75,10 +76,13 @@
                                 <!--#end if#-->
                             </td>
                             <td class="nowrap">
-                                <input type="text" name="dir" class="fileBrowserSmall" value="$slot.dir" size="20" data-initialdir="$defdir" data-title="$T('catFolderPath')" title="$T('explain-catTags2')" />
+                                <input type="text" name="dir" class="fileBrowserSmall" value="$slot.dir" size="12" data-initialdir="$defdir" data-title="$T('catFolderPath')" title="$T('explain-catTags2')" />
                             </td>
                             <td>
-                                <input type="text" name="newzbin" value="$slot.newzbin" size="20" />
+                                <input type="text" name="cleanup" value="$slot.cleanup" size="12" />
+                            </td>
+                            <td>
+                                <input type="text" name="newzbin" value="$slot.newzbin" size="12" />
                             </td>
                             <td class="nowrap">
                                 <button class="btn btn-default">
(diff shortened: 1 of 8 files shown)


Safihre:
Hmmm... is this really something that is useful for the regular user?
If someone really wants this fine grained control, wouldn't we normally just advice a post-processing script?

thezoggy:
this is one of those things i think is a great value add for those that would benefit with having different categories without having to resort to making a post queue script.. which people seem to really struggle with on windows

Safihre:
If so, I think I would like it hidden under Advanced Settings. And also we need to add explanation to the regular Cleanup List that there's also a category version.

And what's next? People will want a Unwanted Extensions per category?

Safihre:
Another concern: for all the other Category settings if you leave it empty, it will take the setting in the Default row. But that's not the case for this setting.. The "default" setting is listed in Switches.

And why not be able to Deobfuscate Final Filenames per category? Or Category specific Duplicate Detection? And so on....

thezoggy:
We already have several things per cat, and recently added sorting per cat.. prob not hard to imagine we extend the unwanted extensions in a similar way... Where you do it on the option page rather than the category page.

Now would it make sense to have every option be per cat, no. But are there some that would make sense to add flexibility.. yes.

Even if it's just to limit scope of option to x cat(s) rather than it be all cats or none. That way you could say, I want the unwanted extensions to the cats I use for automation (radarr/sonarr/etc).

albino1:
Even just a checkbox to exclude any cleanup on a category would go a long ways.  Pretty much I find that you either want cleanup or you don't.  While I'm not sure I would ever go through and make individual category cleanup lists, I would just go all or nothing, I'm sure other people would find ways to make use of it.

For example, if you normally have things like nfos or images removed, but you wanted them for a specific release, it would be nice to be able to just switch that job on the fly to a category called something like No Cleanup.  This is especially true for things like music, bluray discs, or apps, where there's a lot of files that many people would normally have on their cleanup list that you might want to keep.

thezoggy:
in case it wasnt clear, I was advocating that instead of add stuff to the categories page (because its already cluttered and wont really scale to keep adding things unless we redo that whole thing which is a bigger ask).

I was saying that for the options like extension cleanup list / unwanted extensions / whatever else.. we add a multi-select list box (like we have on the sorter page) to apply scope of command for the command to the selected categories. With default being 'default' selected. With us not showing the list unless option is shown, and an advanced setting.

firefox_2023-04-11_15-12-26


This seems like a good compromise of not having per category setting which may be overkill but still offer flexibility to apply the logic to one or more categories of the user choosing and not to others.

Safihre:
Will postpone this to 4.1, and hopefull I will have time to think of a redesign of the Categories page to include topics like this.

Safihre:
At this point I don't want to add this feature, I just can't think of a proper way to integrate this and make it useable for the 99% of regular users.

PR #2526, opened 2023-04-06
Basic direct write implementation

Safihre:
I created a naive implementation of direct writing article data to their correct spot in the file and it works.
However, it does not pre-allocate files like suggested in #2169.
@animethoso will it be faster to pre-allocate files or just use this seek-ahead approach?

Still need to come up with a way to deal with jobs that were started before this commit as they lack `article.begin/end`.

@mnightingale @puzzledsab What do you think of this approach? It will make the assembler loop over every article for each write, however, the cache is kept as empty as possible.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -187,14 +187,22 @@ class Assembler(Thread):
                     # Could be empty in case nzo was deleted
                     if data:
                         # Seek ahead if needed
-                        if article.data_begin > file_position:
-                            fout.seek(article.data_begin, 0)
-                        file_position = article.data_begin + article.data_size
+                        if article.data_begin != file_position:
+                            fout.seek(article.data_begin)
+                        file_position = article.data_begin + len(data)
                         fout.write(data)
                         nzf.update_crc32(article.crc32, len(data))
                         article.on_disk = True
                     else:
                         logging.info("No data found when trying to write %s", article)
+                else:
+                    # If the article was not decoded but the file
+                    # is done, it is just a missing piece, so keep writing
+                    if file_done:
+                        continue
+                    else:
+                        # We reach an article that was not decoded
+                        break
 
         # Final steps
         if file_done:
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -160,13 +160,18 @@ class Assembler(Thread):
 
     @staticmethod
     def assemble(nzo: NzbObject, nzf: NzbFile, file_done: bool):
-        """Assemble a NZF from its table of articles
-        1) Partial write: write what we have
-        2) Nothing written before: write all
-        """
+        """Assemble a NZF from its table of articles"""
+
+        # When a file exists, we cannot use "w+b"
+        open_mode = "w+b"
+        if os.path.exists(nzf.filepath):
+            open_mode = "r+b"
 
         # We write large article-sized chunks, so we can safely skip the buffering of Python
-        with open(nzf.filepath, "ab", buffering=0) as fout:
+        with open(nzf.filepath, open_mode, buffering=0) as fout:
+            # Track position, so we can prevent a seek if writing continuous
+            file_position = 0
+
             for article in nzf.decodetable:
                 # Break if deleted during writing
                 if nzo.status is Status.DELETED:
@@ -181,19 +186,15 @@ class Assembler(Thread):
                     data = sabnzbd.ArticleCache.load_article(article)
                     # Could be empty in case nzo was deleted
                     if data:
+                        # Seek ahead if needed
+                        if article.data_begin > file_position:
+                            fout.seek(article.data_begin, 0)
+                        file_position = article.data_begin + article.data_size
                         fout.write(data)
                         nzf.update_crc32(article.crc32, len(data))
                         article.on_disk = True
                     else:
                         logging.info("No data found when trying to write %s", article)
-                else:
-                    # If the article was not decoded but the file
-                    # is done, it is just a missing piece, so keep writing
-                    if file_done:
-                        continue
-                    else:
-                        # We reach an article that was not decoded
-                        break
 
         # Final steps
         if file_done:


thezoggy:
is this similar to if one just set article cache to 0 and articles are just directly written to said file?

puzzledsab:
I think it will be slower than forcing append or nothing. In my tests on Win 10/NTFS it didn't seem to use sparse files when using fseek. When I wrote the last part first it would take twice as long so I assume it zero filled it first. I don't know for sure how to check if it is using sparse files.

Safihre:
Might be worth to test if you first create a sparse file (found some Stackoverflow links how to do it) and then write to end of file, vs the current seek approach.

mnightingale:
Firstly, the approach appears to work. I tested editing a real nzb removing some articles in the middle of a file and it appeared to be constructed correctly although I didn't confirm data positions where unchanged after repair.

Just a couple of things I noticed:

Consider `fout.seek(article.data_begin)` without the whence 0 - it's the default and I had to look up the docs to see what it means.
Alternatively use the io.SEEK_SET constant?

---

I couldn't see any difference performance wise just always calling fseek everytime or using fout.tell() instead of tracking the position ourselves, NZBGet always seeks.
I don't know if this is because I only have SSDs in my desktop but checking with profiling fseek/tell calls where insignificant.

---

This comes back to something @puzzledsab said on the sabctools discussion.

Apparently it's possible for `article.data_size != len(data)` so maybe we need to slice the write as we may write less than article.data_size but never more?

I think such case could trip up on the current `file_position` tracking, maybe it'd need `len(data[: article.data_size])`
and with the write sliced to at most the data_size `fout.write(data[: article.data_size])`

Basically if the file_position tracking is kept then I think there are cases where len(data) is less than or greater than article.data_size which could throw off some of the logic including that it currently only seeks ahead.

Alternatively if we always seek since we always write in order it wouldn't matter if parts overlap:

---

Regarding sparse files, I haven't managed to figure out how to create sparse files at least on Windows.

```
if open_mode is "w+b":
    fout.truncate(nzf.bytes)
```

This is doing as @puzzledsab observed, creating a zero filled file which takes a while and isn't what we want, confirmed by:

```
$ fsutil sparse queryFlag .\data_2gb.bin
This file is NOT set as sparse
```

I think the only time sparse files would benefit usenet would be avoiding writing zeros where there are a lot of missing articles, but that's rare and if there are a lot of failures the nzb will be cancelled anyway. However since we are now seeking compared the current behaviour without sparse files this may result in apparently slow assembly because if we needed to write mostly to the end of a file we might need to write lots of zeros first.

I can't find for instance any uses of FSCTL_SET_SPARSE in CPython, maybe it doesn't support it on Windows, maybe a job for sabctools.

There are also things to consider like users with directories on network shares, if we can detect it's not supported we don't want be zero filling files.

Safihre:
> Consider `fout.seek(article.data_begin)` without the whence 0 - it's the default and I had to look up the docs to see what it means. Alternatively use the io.SEEK_SET constant?

👍

> I couldn't see any difference performance wise just always calling fseek everytime or using fout.tell() instead of tracking the position ourselves, NZBGet always seeks. I don't know if this is because I only have SSDs in my desktop but checking with profiling fseek/tell calls were insignificant.
> 
> This comes back to something @puzzledsab said on the sabctools discussion.
> 
> Apparently it's possible for `article.data_size != len(data)` so maybe we need to slice the write as we may write less than article.data_size but never more?

Why care aboit all this? Next article we just write over the tail of the data that was too long?
Our data tracking is updated every time because we don't do `current+new`, we juste `article.begin + article.data_size`. So it would be a much bigger problem if the `article.begin` was corrupt? And even if it is corrupt, then so what? Clearly something is wrong wrong with this uploader.

mnightingale:
> > I couldn't see any difference performance wise just always calling fseek everytime or using fout.tell() instead of tracking the position ourselves, NZBGet always seeks. I don't know if this is because I only have SSDs in my desktop but checking with profiling fseek/tell calls were insignificant.
> > This comes back to something @puzzledsab said on the sabctools discussion.
> > Apparently it's possible for `article.data_size != len(data)` so maybe we need to slice the write as we may write less than article.data_size but never more?
> 
> Why care aboit all this? Next article we just write over the tail of the data that was too long?

I don't think we actually would:

```py
if article.data_begin > file_position:
    fout.seek(article.data_begin, 0)
file_position = article.data_begin + article.data_size
fout.write(data)
```

| # | data_begin | data_size | len(data) | file_position after | actual position after fout.tell() |
|---|------------|-----------|-----------|-----------------|-----------------|
| 0 | 0          | 100       | 50        | 100               | 50         |
| 1 | 100        | 100       | 100       | 200              | 150         |

Hopefully this demonstrates what I mean, if the amount of data we write doesn't match data_size subsequent parts will be written to the wrong location because file_position is what we expect it should be, so the actual position and file_position wouldn't match.
There is also the case where len(data) > data_size, if one part is invalid it will throw the rest of the file out which could be irreparable if the par2 options are ever changed not to deal with misaligned blocks.

So if you want to keep file_position and not always seek or use tell calls but avoid the difficult to read slicing then I'd suggest:

```py
if article.data_begin != file_position:
    fout.seek(article.data_begin, 0)
file_position = article.data_begin + len(data)
fout.write(data)
```

So also check if we are before the expected position, and increment the position by how much is actually going to be written and 99.99% of the time it will be the same as incrementing by the expected length but guards against there being something wrong with an article.

> So it would be a much bigger problem if the `article.begin` was corrupt? And even if it is corrupt, then so what? Clearly something is wrong wrong with this uploader.

I'm advocating for trusting the headers which we determined is the same behaviour as nzbget, I don't disagree if the header value doesn't match the actual length then something has gone wrong with the upload but I'd still ensure whatever data we do get is positioned at the header begin offsets.

Safihre:
Thanks for that explanation, I was indeed wrong.
I confirmed indeed that NZBGet always [seeks](https://github.com/nzbget/nzbget/blob/5e26d52d706f129769e1d620a595c78498ca8cff/daemon/nntp/ArticleWriter.cpp#L139-L142). If that's the case, probably the performance impact is negligible. 

In that regard, if NZBGet finds it worthwhile to pre-allocate the file, shouldn't we also?
For reference, the pre-allocate [code](https://github.com/nzbget/nzbget/blob/5e26d52d706f129769e1d620a595c78498ca8cff/daemon/util/FileSystem.cpp#L255-L333).

mnightingale:
> In that regard, if NZBGet finds it worthwhile to pre-allocate the file, shouldn't we also? For reference, the pre-allocate [code](https://github.com/nzbget/nzbget/blob/5e26d52d706f129769e1d620a595c78498ca8cff/daemon/util/FileSystem.cpp#L255-L333).

It's probably worth doing but as I said previously unless I'm misinformed I think it's only useful when a file is missing lots of articles because we skip writing the zeros.

I've tested it with a method in sabctools and it did create a sparse file but I still need to figure out also handling Linux as it'd be nicer if the single method handled both.

If you have a look at it I had to use  `msvcrt.get_osfhandle(fout.fileno())` to get the handle for DeviceIoControl but it looks like for Linux we'd just need to call truncate, so we'd probably want the method to take fout or fout.fileno() but then would need to use private api - I think _Py_get_osfhandle

Safihre:
@mnightingale I think I missed your code example? Importing `msvcrt` doesn't seem to be a problem since it's part of the stdlib?

I hope I have time soon to do some speed testing with this, if indeed it's better to include sparse file support.

puzzledsab:
I would be surprised if it isn't slower to write parts as they arrive, particularly if it has to zero fill. When using for instance 40 connections, a lot of the times the articles are not going to be downloaded in perfect sequence. If we are writing the articles as they arrive then there will be a fair amount of zero filling. I've always suspected that to be part of the reason why some report that nzbget is writing slower than SABnzbd.

puzzledsab:
@thezoggy : Currently SAB will only write articles that are available in sequence, or all available parts when we know that no more parts can be downloaded. If you have multiple servers and some are lagging behind then the articles will be stored temporarily in RAM or on disk until all the servers have been checked. If the cache is 0 then they will be stored in temporary files until they can be written to the correct file in sequence. With the new system it would write each part directly to the correct file.

mnightingale:
I hadn't noticed that this PR also allows writing out of order, I didn't think we wanted that.
Perhaps to better handle when we currently have to wait until all servers have been tried it would be better to only write out of order if an article has been downloaded for some time, but I expect that would require more significant changes.

Safihre:
Why not also write out of order? We only write every X number of articles so it's not like we really write every article as it comes in.

puzzledsab:
It may not be very noticeable if the user has a big cache but there will be some holes. If they have to be zero filled then that will take time. Even if sparse files are used it's faster to write large chunks of data sequentially, particularly on spinning disks.

mnightingale:
I agree with what @puzzledsab has said and add that with sparse files and spinning disks the possible disadvantage is more fragmented files especially if data is being written out of order.

The current assembler max queue size and other triggering logic may already be sufficient, I think it could just do with a test counting how many sequential writes are done vs random in both the case where articles are complete on the primary server vs when some must be retried.

Some getting written out of order isn't a problem but lots of them would be undesirable.

Safihre:
Thanks @mnightingale for the work in `sabctools`! 💯 
So with that in place, can we create some speed-test to see if indeed it's faster to pre-allocate the file instead of repeatedly extend the file (which is what this PR does)? Both for the almost-consecutive-writes case and the lots-of-holes case?
To confirm if indeed NZBGet is right for always doing it.

mnightingale:
Results of some testing I've done on Windows with a HDD, mostly looking at what effect the changes have had on file fragmentation, testing against a local server.

Repeated 3 x 5GB downloads.
1GB cache limit, 16 connections
Using [contig](https://learn.microsoft.com/en-us/sysinternals/downloads/contig) to check fragmentation.

**NTFS**
Branch/options = Numbers of fragments
develop = 15, 15, 16
feature/directwrite = 17, 16, 17
feature/directwrite + sparse = 8, 6, 8
feature/directwrite + sparse + in-order = 7, 7, 12
feature/directwrite + SetFilePointerEx (as if sparse isn't supported) = 1, 1, 1 but terrible performance, presumably it has to zero fill up to the file size on first write.

**exFAT (doesn't support sparse)**
feature/directwrite = 3, 2, 2
feature/directwrite + sparse = terrible performance
feature/directwrite + sparse + in-order = terrible performance
feature/directwrite + sparse (fixed - conditional SetFilePointerEx) = 2, 3, 3
feature/directwrite + sparse (fixed - conditional SetFilePointerEx) + in-order = 3, 2, 3

I've pushed a commit to the sparse method to call SetFilePointerEx only if the file could be made sparse - I did have it like previously but didn't appreciate its necessity, it resolves the terrible performance I noted above.

Local testing probably means articles are received mostly in-order but it does appear to produce less fragmented files, I think that is because setting the filesize means the file can be better allocated on disk and the same should apply Linux/Mac with its call to ftruncate.

I see while I was writing this you've added the in-order writing back, above was tested before that commit but I was testing with that change anyway.

mnightingale:
> I've pushed a commit to the sparse method to call SetFilePointerEx only if the file could be made sparse - I did have it like previously but didn't appreciate its necessity, it resolves the terrible performance I noted above.

I have noticed the sparse method isn't setting the file position back to the start which I probably can, maybe that could have been related to my problem.
I'm sure you're aware but the actual file size will need to come from an article, so if open_mode is "w+b" and nothing has been written yet.

mnightingale:
I've fixed the file position after calling sparse so it returns to the original position, I think size will need extracting from the ybegin line before it can be used.

Safihre:
So these fragmentation measures, how to interpret them? Lower is better or worse?

Safihre:
I did notice that when I allowed out-of-order writing, it would actually do that quite a lot with 40 connections. I had expected maybe once or twice, but it was easily 5-10 times per 100MB file. So that's why I'm trying to understand from your testing if sparse files make that better or worse?

Safihre:
> > I'm not sure why first articles of all files are downloaded but I'm sure there is reason for it.
> 
> That's a quick health check on a newly started job, most useful in cases where only the last part is missing/removed.

Not just that, we also use the md5 of the first 16k of data to already give files the right filename based on par2. Which is needed for Direct Unpack.
This is also something I copied exactly from nzbget that does it for the same reason 😃

puzzledsab:
> I wonder if it could be possible to get into a deadlock situation due to the sabnzbd.Assembler.queue_level() checks in the downloader thread, I think we could be retrying downloading an article but can't because we are also waiting for the assembler

The assembler queue entry will be removed even if no new parts could be written this shouldn't cause a deadlock?

Safihre:
I think I will postpone this to 4.1, to not further delay the great work that was already put into 4.0.
But I will try to include reading article start/length in 4.0 so more jobs will have it when upgrading to 4.1.

mnightingale:
I agree with shifting this to 4.1, I think it's more important to get a 4.0 release out - I've seen a few users reporting improved performance.

For uuencode do we just need to make the seeks conditional on data_begin not being None so we continue appending to the end?

I think we may need to keep track of how much data we have written on an NzbFile object so we can seek to the end since open_mode "r+b" will go to the beginning of the file and if the sparse method which sets file size is used opening in append mode would be the wrong position anyway.

Safihre:
@mnightingale do you maybe have interest in implementing the sparse mode code into this?
I feel you are already deeper into all this!

mnightingale:
Sure, do you happen to have a UU post you test with?
I've written most of the required changes and I'm pretty sure the logic will work for UU but I'd prefer to be able to test it.

Safihre:
@jcfp do you still have good test files? I only have broken ones it seems

Safihre:
Ah yes, I forwarded the big one to @mnightingale

Safihre:
Superseded by #2574

PR #2538, opened 2023-04-23
add bootstrap multiselect

thezoggy:
add bootstrap multiselect v0.9.15 (have to use older version for bootstrap 3.x support)
disable sortable on pattern key to allow users to select text to copy


thezoggy:
I forget when i did the config way back in the day it was before grid systems, so ran into overflow issue and had to allow it for the places I want to show a popup...  Also due to legacy bootstrap had to use older version of plugin and css but got it working.

some example:

when expanded, and picking:

firefox_2023-04-23_05-01-26

when collapsed (disabled):

firefox_2023-04-23_05-01-45


only did it for the categories section as quick example.

Safihre:
> I'm not a fan of being able to click anywhere on the collapsed sorter to toggle showing it, makes it a little jarring when stuff pops in and the thing you were clicking on moves about. So might want to limit that to really just the toggle button?

 That's on purpose, because otherwise for new users it might not be clear enough that you need to toggle to edit.

Safihre:
But not sure if this is really much better than a multi-select?

thezoggy:
It reduces how much space is needed to show the car selector (could be expanded for other things as well). And since it's on the page multiple times it adds up. Just like if we added affected cat option to various switches (unwanted extensions/cleanup).

interfaces/Config/templates/_inc_header_uc.tmpl
```diff
@@ -33,6 +33,7 @@
 
     <link rel="stylesheet" type="text/css" href="${root}staticcfg/bootstrap/css/bootstrap.min.css?v=$version" />
     <link rel="stylesheet" type="text/css" href="${root}staticcfg/css/chartist.min.css" />
+    <link rel="stylesheet" type="text/css" href="${root}staticcfg/css/bootstrap-multiselect.min.css" />
```
Safihre, line 36: Move it 1 line up to be right behind bootstrap.

interfaces/Config/templates/config_sorting.tmpl
```diff
@@ -47,7 +47,7 @@
                 </div>
                 <!--#if $cur != 0#-->
                 <div class="col1">
-                    <fieldset>
+                    <fieldset overflow>
```
Safihre, line 50: Should this not be `class="overflow"`?

interfaces/Config/templates/config_sorting.tmpl
```diff
@@ -101,7 +101,7 @@
                                 <option value="4" <!--#if "4" in $selected_types then 'selected="selected"' else ""#-->>$T('guessit-type-other')</option>
                             </select>
                         </div>
-                        <div class="field-pair">
+                        <div class="field-pair overflow">
```
Safihre, line 104: Same here.

interfaces/Config/templates/config_sorting.tmpl
```diff
@@ -435,6 +435,12 @@
     }
 
     \$(document).ready(function() {
+
+        \$("[name='sort_cats'], [name='sort_cats_display']").multiselect( {
+            buttonTextAlignment: 'left',
+            maxHeight: 200,
+        });
+
```
Safihre, line 443: Is there a way to give it a minum width? It's a big ugly with the different sizes depending on what is selected.

interfaces/Config/templates/staticcfg/css/style.css
```diff
@@ -378,6 +378,25 @@ tr.separator {
...
+}
+div.btn-group.open input[type='checkbox'] {
+    top: 0px !important;
+    margin-left: -22px !important;
+}
+div.btn-group.open label {
+    padding-left: 30px !important;
+    padding-bottom: 5px !important;
+}
+.multiselect.disabled {
```
Safihre, line 394: Why this extra disabled styling? I prefer the default Bootstrap disabled styling much better.
Hacked example:
image

Safihre (changes requested):
Can't say I'm a huge fan of this, but if you and @jcfp prefer it who am I to resist 🤣

interfaces/Config/templates/config_sorting.tmpl
thezoggy: i can, theres a few options let me see if you like this one..

interfaces/Config/templates/staticcfg/css/style.css
thezoggy: it was to combat sab's style.css where we override bootstrap's .disabled with #aaa which i find makes it a little hard to read (and was to help jcfp's issue). but i can tweak it so its not so stark..  

rather than use disabled, shouldnt we just use readonly instead?

thezoggy:
updated example:

firefox_2023-04-27_00-37-25

selection:
firefox_2023-04-27_00-39-14

selection expanded:
firefox_2023-04-27_00-38-24

Safihre:
At this point I don't want to include this and stick to browser-native elements for now!

thezoggy:
Ok then should cherry pick the fixes for other stuff then

PR #2539, opened 2023-04-23
Add releases tag

francoism90:
This seems to be the only thing missing for Flatpak.

I believe this can be automated, but I haven't found a GitHub hook yet.

Thanks

### DIFF
--- a/linux/org.sabnzbd.sabnzbd.appdata.xml
+++ b/linux/org.sabnzbd.sabnzbd.appdata.xml
@@ -29,6 +29,9 @@
   <url type="help">https://sabnzbd.org/wiki/</url>
   <url type="faq">https://sabnzbd.org/wiki/faq</url>
   <url type="contact">https://sabnzbd.org/live-chat.html</url>
+  <releases>
+    <release version="3.7.2" date="2023-02-05" type="stable"></release>
+  </releases>
   <launchable type="desktop-id">sabnzbd.desktop</launchable>
   <provides>
     <mediatype>application/x-nzb</mediatype>
--- a/linux/org.sabnzbd.sabnzbd.appdata.xml
+++ b/linux/org.sabnzbd.sabnzbd.appdata.xml
@@ -30,6 +30,7 @@
   <url type="faq">https://sabnzbd.org/wiki/faq</url>
   <url type="contact">https://sabnzbd.org/live-chat.html</url>
   <releases>
+    <release version="4.0.0" date="2023-01-01" type="stable"></release>
     <release version="3.7.2" date="2023-02-05" type="stable"></release>
   </releases>
   <launchable type="desktop-id">sabnzbd.desktop</launchable>


thezoggy:
Shouldn't this be directed towards master not dev?

francoism90:
@thezoggy I just remember this does not work, you need to include the upcoming version as well.
Maybe this should wait for the next SABnzbd release? As Flatpak will download the latest tar and checks appdata, which is now empty at the moment and already tagged as 3.7.2.

Sorry for the noise, I didn't know the <releases> tag was required for Flatpak.

Safihre:
You will update this every release?

francoism90:
Bit disappointed this hasn't been merged for the 4.0.0 release.

Safihre:
You didn't answer my question about who is responsible for updating this..

francoism90:
I already stated I would gladly maintain the SABnzbd Flatpak. This would indeed be a manual step, however maybe it can be automated as well, since you are already seem to be bumping versions.

This would be the final showstopper of getting this upstream. Flatpak has been getting more attention and most distributions are using it, so it would benefit your project as well.

Not saying it's a must, but not everybody runs Snap for example.

Safihre:
Can you put the right release date? Then I'll merge. Do you also need this in master?

You can take a look at the build script how we could incorporate automatically adding new versions. Should be possible with some extra logic steps.

francoism90:
@Safihre I can correct the date, but it doesn't make much difference. The release/version should be set before creating a new (git) tag.

Safihre:
That doesn't make sense to me. How can we put a release in there that isn't released yet? For example you already put 4.0.0 before it was out. So now we should put 4.0.1 before we even release it?
Seems more logical to do it afterwards when the release is done?

francoism90:
Because Flatpak downloads by commit/tag, so yeah, you should bump your version in appdata before creating the release, otherwise it will use the previous version. In most cases I expect this will be on the same day.

It's the same for Snap right? Or does this work differently?

I don't understand why Flatpak just follows the GitHub release link, but it seems only XML is supported and it should be added to appdata.

Safihre:
Right, so it should be present in the code at the moment the tag is created. That makes more sense.

Snap is a mess 🥲 so not a good comparison.

francoism90:
Yep, sorry if I'm pushing too much btw.

If you could let me know when you're tagging 4.0.1 (or any next release), I would be happy to create a PR for appdata. For now it doesn't matter, since I need to update the current Flatpak for 4.0.0

Thanks.

Safihre:
I added it for 4.0.1, so now you can release it, right?

francoism90:
Let's say you guys release 4.0.1 next month/week, you should add this again to the appdata.xml with the release date of next month/week. Next and final step, is tagging the release, so the Flatpak can point to this tarball (which includes the changelog of appdata.xml).

I don't think you need to include actual changes, the release date seems to be key.

PR #2542, opened 2023-04-25
test writing long and unicode filenames

### LINKED ISSUE #2534: Very long filename problem
Hi again, did have an issue I thought I had fixed with max folder length being reduced to 100 but im still having problems with long filenames. 

The one is question is 199 characters being returned by the indexer. 

Can we have an option to truncate the filename after say X characters?

If not how is it best to deal with these as it seems to happen a lot on foreign grabs.

Thanks

sanderjo:
Modern filesystems should be able to write long and unicode filenames

Closes #2534

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -1291,8 +1291,8 @@ def directory_is_writable(test_dir: str) -> bool:
         allgood = False
 
     # unicode in filename
-    unicodename = "🚀" * 20  # Note: 4 byte per char
-    if directory_is_writable_with_file(test_dir, unicodename):
+    # Note: 4 byte per char
+    if directory_is_writable_with_file(test_dir, "🚀" * 20):
         logging.debug("OK: I can write a unicode filename to %s", test_dir)
     else:
         sabnzbd.misc.helpful_warning("Cannot write a unicode filename to %s", test_dir)
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -1279,9 +1279,15 @@ def directory_is_writable(test_dir: str) -> bool:
     if not directory_is_writable_with_file(test_dir, "sab_test.txt"):
         sabnzbd.misc.helpful_warning(T("%s is not writable at all. This blocks downloads."), test_dir)
         return False
+    return True
 
-    # basic writing OK, so let's check further
-    allgood = True  # default return value
+
+def test_filesystem_capabilities(test_dir: str) -> bool:
+    """Checks if we can write long and unicode filenames to the given directory.
+    If not on Windows, also check for special chars like \ and :
+    Returns True if all OK, otherwise False"""
+
+    allgood = True  # default return value: all OK
 
     # long filename; normal filesystems accept 255 byte filenames
     if not directory_is_writable_with_file(test_dir, "A" * 245 + str(random.randrange(10000, 99999))):
@@ -1293,7 +1299,7 @@ def directory_is_writable(test_dir: str) -> bool:
         sabnzbd.misc.helpful_warning(T("Cannot write a unicode filename to %s. This can cause problems."), test_dir)
         allgood = False
 
-    # if not on Windows, check strange characters
+    # if not on Windows, check special chars like \ and :
     if not sabnzbd.WIN32 and not directory_is_writable_with_file(test_dir, "sab_test \\ bla :: , bla.txt"):
         sabnzbd.misc.helpful_warning(
             T("%s is not writable with special character filenames. This can cause problems."), test_dir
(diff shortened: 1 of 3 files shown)


sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
 
 
 def directory_is_writable(test_dir: str) -> bool:
```
Safihre, line 1275: I want a super basic test for this. 
These should **not** return `False` when run in the testing suite, otherwise the test is wrong.

sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
...
         sabnzbd.misc.helpful_warning(T("%s is not writable at all. This blocks downloads."), test_dir)
         return False
-    return True
+
+    # basic writing OK, so let's check further
+    allgood = True  # default return value
+
+    # long filename; normal filesystems accept 255 byte filenames
+    longname = "A" * 245 + str(random.randrange(10000, 99999))
+    if directory_is_writable_with_file(test_dir, longname):
```
Safihre, line 1288:
```suggestion
    if directory_is_writable_with_file(test_dir, "A" * 245 + str(random.randrange(10000, 99999))):
```

sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
...
+        logging.debug("OK: I can write a long filename  to %s", test_dir)
```
Safihre, line 1289: No need for the debug message.

sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
...
+    else:
+        sabnzbd.misc.helpful_warning("Cannot write a long filename to %s", test_dir)
```
Safihre, line 1291: It needs to be a translatable text:
```suggestion
        sabnzbd.misc.helpful_warning(T("Cannot write a long filename to %s. This can cause problems."), test_dir)
```

sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
...
+        allgood = False
+
+    # unicode in filename
+    unicodename = "🚀" * 20  # Note: 4 byte per char
+    if directory_is_writable_with_file(test_dir, unicodename):
+        logging.debug("OK: I can write a unicode filename to %s", test_dir)
```
Safihre, line 1297: No need for the debug.

sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
...
+    longname = "A" * 245 + str(random.randrange(10000, 99999))
+    if directory_is_writable_with_file(test_dir, longname):
+        logging.debug("OK: I can write a long filename  to %s", test_dir)
+    else:
+        sabnzbd.misc.helpful_warning("Cannot write a long filename to %s", test_dir)
+        allgood = False
+
+    # unicode in filename
+    unicodename = "🚀" * 20  # Note: 4 byte per char
+    if directory_is_writable_with_file(test_dir, unicodename):
```
Safihre, line 1296:
```suggestion
    # Note: 4 byte per char
    if directory_is_writable_with_file(test_dir, "🚀" * 20):
```

sabnzbd/filesystem.py
```diff
@@ -1272,15 +1273,37 @@ def directory_is_writable_with_file(mydir, myfilename):
...
+        logging.debug("OK: I can write a unicode filename to %s", test_dir)
+    else:
+        sabnzbd.misc.helpful_warning("Cannot write a unicode filename to %s", test_dir)
```
Safihre, line 1299:
```suggestion
        sabnzbd.misc.helpful_warning(T("Cannot write a unicode filename to %s. This can cause problems."), test_dir)
```

Safihre (changes requested):
Thanks! Just a few changes ;)

sabnzbd/filesystem.py
sanderjo: > I want a super basic test for this.

So ... move the unicode and long file testing to a different location ... ?

> These should **not** return `False` when run in the testing suite, otherwise the test is wrong.

So ... return nothing ... ? Return True when run in the testing suite? Not run from testing suite? I don't understand. In the testing suite it should be OK because it's being tested on a modern FS without FAT-like limitations, so I don't understand how the test can be wrong.

Safihre: No, I just want you to add this to a `test_filesystem`. Just a simple call to a test-folder.

PR #2556, opened 2023-05-09
fix pattern key

thezoggy:
Disable sorting on pattern key so you can select text on it / prevent it from moving.


Safihre:
Thanks!

PR #2566, opened 2023-05-15
postproc: Print last line in case of error

Ram-Z:
When a postproc fails, provide last line in addition to exit code.

before:
image

after:
image

### DIFF
--- a/sabnzbd/postproc.py
+++ b/sabnzbd/postproc.py
@@ -556,7 +556,7 @@ def process_job(nzo: NzbObject):
                     if script_ret and cfg.script_can_fail():
                         script_error = True
                         all_ok = False
-                        nzo.fail_msg = T("Script exit code is %s") % script_ret
+                        nzo.fail_msg = T("Exit(%s) %s") % (script_ret, script_line or "")
 
         # Email the results
         if not nzb_list and cfg.email_endjob():
--- a/sabnzbd/postproc.py
+++ b/sabnzbd/postproc.py
@@ -577,18 +577,18 @@ def process_job(nzo: NzbObject):
         if script_output:
             # Can do this only now, otherwise it would show up in the email
             if script_ret:
-                script_ret = "Exit(%s) " % script_ret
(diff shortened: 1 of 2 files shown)


thezoggy:
to note that the things you see with the T() is stuff thats translated, so changing that line like that would invalidate the translation. although im sure we could prob cleanup the translation to make a little more sense with the line, `Script (exit) lastline: (#) ####`

Ram-Z:
Ah. Thanks. I hadn't given that T() much thought.                        
                                                                         
I went with consistency with the string in the popup when clicking on a  
download.                                                                
                                                                         
There's not necessarily much screen estate in that field, so I'd rather  
avoid unnecessary words.                                                 
                                                                         
Does "Exit" in this context need translating?

Ram-Z:
OK. Here's an updated patch.

I first made the `Exit(%s) %s` string in the popup translatable. This should not introduce any changes besides adding a new translatable text.

Then I reuse that new string in the main history view.

sabnzbd/postproc.py
```diff
@@ -577,18 +577,18 @@ def process_job(nzo: NzbObject):
...
                 nzo.set_unpack_info(
                     "Script",
-                    '%s%s <a href="./scriptlog?name=%s">(%s)</a>' % (script_ret, script_line, script_output, T("More")),
+                    '%s <a href="./scriptlog?name=%s">(%s)</a>' % (script_msg, script_output, T("More")),
                     unique=True,
                 )
             else:
                 # No '(more)' button needed
-                nzo.set_unpack_info("Script", "%s%s " % (script_ret, script_line), unique=True)
+                nzo.set_unpack_info("Script", "%s " % script_msg, unique=True)
```
Ram-Z, line 591: I kept this extra space in ``%s "` as I wasn't sure if it was needed for anything.

Safihre:
I'm working on some improvements for the post processing scripts and will include this improvement. So I haven't forgotten this PR :)

Ram-Z:
Great, thanks.

PR #2573, opened 2023-05-27
Fix uu decode workaround

mnightingale:
Replaced with line from https://github.com/python/cpython/blob/3.12/Lib/uu.py#L165

Tested with "audio fidelity", no more warnings and skipped articles.

PyCharm warns with `ord(line[0])` and `Expected type 'str | bytes', got 'int' instead ` hence exception `TypeError('ord() expected string of length 1, but int found')`

### DIFF
--- a/sabnzbd/decoder.py
+++ b/sabnzbd/decoder.py
@@ -292,7 +292,7 @@ def decode_uu(article: Article, raw_data: bytearray) -> bytes:
             except binascii.Error as msg:
                 try:
                     # Workaround for broken uuencoders by Fredrik Lundh
-                    nbytes = (((ord(line[0]) - 32) & 63) * 4 + 5) / 3
+                    nbytes = (((line[0] - 32) & 63) * 4 + 5) // 3
                     decoded_line = binascii.a2b_uu(line[:nbytes])
                 except Exception as msg2:
                     logging.info(
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -184,7 +184,6 @@ class TestUuDecoder:
     @pytest.mark.parametrize(
         "bad_data",
         [
-            b"MI^+0E\"C^364:CQ':]DW++^$F0J)6FDG/!`]0\\(4;EG$UY5RI,3JMBNX\\8+06\r\n$(WAIVBC^",  # Trailing junk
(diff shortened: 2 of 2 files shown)


mnightingale:
The test failure is fun, first failure in `test_broken_uu`.

`bytearray(b'$(WAIVBC^')` is the last line which would fail the current workaround implementation because it passes the wrong type in and never tries the workaround, with this PR it gets decoded to `b'#xi\xda'`.

But the test fails because it reaches `article.nzf.type = "uu"` instead of the expected raised BadData.

Maybe someone more familiar can verify that this is correct and the tests will need updating?

Safihre:
@jcfp :) is the expert

sabnzbd/decoder.py
```diff
@@ -292,7 +292,7 @@ def is_uu_junk(line: bytes) -> bool:
             except binascii.Error as msg:
                 try:
                     # Workaround for broken uuencoders by Fredrik Lundh
-                    nbytes = (((ord(line[0]) - 32) & 63) * 4 + 5) / 3
+                    nbytes = (((line[0] - 32) & 63) * 4 + 5) // 3
```
thezoggy, line 295: this now looks to align with:

https://github.com/python/cpython/blob/main/Lib/encodings/uu_codec.py#L61
>             nbytes = (((s[0]-32) & 63) * 4 + 5) // 3

mnightingale: I found the same in the deprecated uu library, which is why I think the test is currently testing for a failure that the workaround is supposed to be able to handle.

I noticed cpython doesn't expect an exception to be thrown by the workaround.

With the sample nzb @Safihre supplied dropping the flac files into foobar200 they all verify, I'm not familiar with the FLAC format but it appears they use CRC8 and can contain an md5 which the files do.

Safihre: Cool. In that case let's change the test.

mnightingale:
I've removed the test case which expects a failure but with the working workaround decodes successfully.

POC https://gist.github.com/mnightingale/2af3aab61bb65c8e28499de62ce2ab65

Tested with audio fidelity, the work around is triggered 4 times due to "Trailing garbage" and succeeds.
A "verify integrity" test passes in foobar2000

Safihre:
Thanks 🥳

PR #2574, opened 2023-05-27
Write articles to correct offsets and use sparse files

mnightingale:
Continuation of #2526 with added support for creating sparse files.

Sparse call failures are logged to debug, on Windows when sparse files are not supported it'll be a silent failure.
UU behaviour is as before, it doesn't use sparse files and always appends writes to the end of the file.


Safihre:
Cool. Looking at nzbget, I think if I understand the code correctly that at the end it also truncates the file to it's actual size if there's less data at the end than originally indicated. Would involve more bookkeeping so wonder if it's worth it?

mnightingale:
Where does it do the truncate at the end? I can't seem to find it.

I'm not sure it should be required, but I guess we could track the furthest offset written to.
Maybe get a likely more reliable file length from the PAR2 data.

Safihre:
@mnightingale Did you maybe have time to look at this? 😊

mnightingale:
I don't see any discernible performance impact from using sparse files, but I never expected to.

My primary interest was to write articles at the correct offsets, to enable exotic manual repairs and theoretically it may speed up PAR repairs.
To that end sparse files do help because broken downloads occupy less space since "empty" areas of the file don't need to be written and there is less seeking of the file.

We still write files in sequence; this helps produce files with fewer fragments which is good for performance especially on HDDs and I think is a good design choice. But the use of sparse files does mean we could write out-of-order without significant penalty, but I think that should only happen sparingly.

There are a couple of areas out of scope of this PR where it may be beneficial to allow writing articles out-of-order in certain cases:
- `articlecache.flush_articles` could flush to the destination file in the correct position
- I haven’t investigated the current behaviour but with a small cache it might sometimes be optimal to write out of order rather than slow download or cache to disk - I'm thinking of a situation where article 1-50 fail and are waiting to be retried on a different server with slower or fewer connections, in the meantime we've downloaded 51-100 but[^1] due to sequential writing have to keep them around until the others are retried or failed at which point the cache could be full so `articlecache.save_article` is out of memory and writes to disk, then the assembler finds a decoded article and reads it from disk only to write it to destination file - I haven't looked into this sufficiently but I do want see how it behaves with low memory limits, maybe it could be as simple as a "force assemble" that writes out-of-order if the cache is rather full or a NZF hasn’t written for some time.

[^1]: Or maybe another file is available, but cache is full so it has to temporarily write to disk, read and write to destination - I'd hope this can be avoided but if not maybe zero-copy so avoid some overhead

thezoggy:
a common thing is people that put their incomplete folder on a mount/drive that has io overhead (shfs/mergerfs) or remote file system. so any additional read/writes just taxes things more. would sparse files be beneficial to those?

mnightingale:
@Safihre I noticed this isn't included in 4.0.3 are you keeping it back for the next minor release?

Safihre:
Yes, I am trying to adhere to [semantic versioning](https://semver.org/) and only allow bugfixes in patch-releases such as 4.0.x.
Sparse files are a significant change in behavior that it will have to wait for 4.1.x.

Issue #2251, opened 2022-07-26
Post Processing Failed when Quick-check renaming occurs


alexmaras:
I'm using version 3.6.0 of sabnzb in docker. For some downloads, I'm seeing a post-processing error. With debug loggin on, I get the following:

```
DEBUG::[newsunpack:2096] Quick-check will rename 188275-3.0 to (new-file-name-here).jpg
      1 2022-07-26 21:06:38,038::INFO::[notifier:123] Sending notification: Error - Post Processing Failed for (download name here) (see logfile) (type=error, job_cat=
        None)
      2 2022-07-26 21:06:38,038::ERROR::[postproc:596] Post Processing Failed for (download name here) (see logfile)
      3 2022-07-26 21:06:38,038::INFO::[postproc:597] Traceback:
      4 Traceback (most recent call last):
      5   File "/app/sabnzbd/sabnzbd/postproc.py", line 388, in process_job
      6     par_error, re_add = parring(nzo)
      7   File "/app/sabnzbd/sabnzbd/postproc.py", line 751, in parring
      8     need_re_add, res = par2_repair(nzo, setname)
      9   File "/app/sabnzbd/sabnzbd/newsunpack.py", line 1123, in par2_repair
     10     qc_result = quick_check_set(setname, nzo)
     11   File "/app/sabnzbd/sabnzbd/newsunpack.py", line 1980, in quick_check_set
     12     new = quick_check_set_new(setname, nzo)
     13   File "/app/sabnzbd/sabnzbd/newsunpack.py", line 2100, in quick_check_set_new
     14     new_path = renamer(
     15   File "/app/sabnzbd/sabnzbd/decorators.py", line 36, in call_func
     16     return f(*args, **kw)
     17   File "/app/sabnzbd/sabnzbd/filesystem.py", line 857, in renamer
     18     oldpath, _ = os.path.split(old)
     19   File "/usr/lib/python3.9/posixpath.py", line 103, in split
     20     p = os.fspath(p)
     21 TypeError: expected str, bytes or os.PathLike object, not NoneType
```

From what I can see, the `old` filename (in this case `188275-3.0`) is coming through as NoneType. Not sure if this is because it's being misunderstood as a number or not.

Safihre:
I sounds the same as #2204. Are you able to run the develop version?

alexmaras:
I've updated to `develop` from the current git repo to see if I still have the same issue. Now I'm getting:


```
[...]

PR #2584, opened 2023-06-03
Fix sorting lowercasing

### LINKED ISSUE #2578: Series Sorter Season Pack Bug
{%sn} (%y)/{%sn} (%y) - s%0se%0e.%ext
The series sorting works perfectly for individual episodes but on season packs with the above it creates a file name with the curly braces.

instead of `showname/showname - s01e01` it ends up as `showname/{ShowName} - s01e01` the filename still capitalized

only affects packs. running 4.0.1

jcfp:
Also fixes lowercasing for sort strings using %fn in sequential and regular sorting.

Closes #2578

### DIFF
--- a/sabnzbd/sorting.py
+++ b/sabnzbd/sorting.py
@@ -432,7 +432,7 @@ class Sorter:
                         ("%ext", f_ext.lstrip(".")),
                     ],
                 )
-                f_new = f_name_new + f_ext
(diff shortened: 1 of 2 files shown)


Safihre:
Thanks!
Not #2251?

Safihre:
Ok!

PR #2595, opened 2023-06-16
Add versions to appdata

francoism90:
The last time you guys forget to bump the releases. :)


Safihre:
Told you I would forget 😂

Safihre:
I'll try to build a check in package.py that only during building of stable release checks if that release is present in the xml.

francoism90:
@Safihre That would be great, thanks. :+1:

PR #2598, opened 2023-06-22
Fix sorting for 2551

### LINKED ISSUE #2551: Change in how the release is saved?
Before v4.0, an NZB file named "big_buck_bunny_2000_1080p_h264.nzb" containing "big_buck_bunny_1080p_h264.mov" would be saved into "completed/Big Buck Bunny (2000)/big_buck_bunny_2000_1080p_h264.mov" with the sorting pattern `%sn (%y)`.

However, from v4.0, it is saved into "completed/Big Buck Bunny (2000)/big_buck_bunny_2000_1080p_h264/big_buck_bunny_2000_1080p_h264.mov" instead.

Was this an intentional change with the new generic Sorter?

jcfp:
Turned out the problem wasn't in `postproc.prepare_extraction_path` after all, but with a call that was deleted by mistake when introducing the generic sorter. As a result, the temporary workdir in the sorter-constructed path was kept around if the sort string didn't rename files (but only defined a directory path).

Changes:
* restore call to `move_to_parent_directory` in case the Sorter does not rename any files + functional tests for that;
* add tests for `postproc.prepare_extraction_path` (had already written those anyway...);
* make sorting.Sorter.rename use `get_unique_filename` to avoid an unnecessary fallback to deobfuscate that showed up during testing.

Fixes #2551


tests/data/SINGLE_sort_s23e06_480i-SABnzbd/abc.mov
Safihre, line 1: Could you move the new and existing test files to a sub-directory "tests/data/sorting"?

Safihre:
I looked at the code but don't think I fully understand it. So I'll just trust you and the unit tests 😃

Safihre:
Something about the new tests is not working stable enough on macOS, it's failing intermittently:
```
FAILED tests/test_functional_sorting.py::TestDownloadSorting::test_download_sorting_single[single-ep_sort_s06e66_4k_uhd-SABnzbd-result3] - AssertionError
```
https://github.com/sabnzbd/sabnzbd/actions/runs/5398675209/jobs/9804786427
https://github.com/sabnzbd/sabnzbd/actions/runs/5386700498/jobs/9777056950
https://github.com/sabnzbd/sabnzbd/actions/runs/5359968446/jobs/9724245539

PR #2604, opened 2023-06-27
add random import back to fixup

thezoggy:
resolves regression from https://github.com/sabnzbd/sabnzbd/commit/786d5b066758c25a3e97e7f7383170f44f2ed968


Safihre:
Yeah this was stupid cherry picking, on develop this import could be removed...

Safihre:
Note that people can fix the problem by setting Address Selection (in Config Switches) to Happy Eye Balls or First.

PR #2606, opened 2023-06-27
better docker detections: works for older and newer docker versions

sanderjo:


Safihre:
That is indeed why the CI is failing on Windows.

PR #2607, opened 2023-06-28
Add multi-select to history

### LINKED ISSUE #2603: Feature Request: Multiple Selection in "Completed"
Could a "multiple selector" be added in completed history?

User-case, occasionally I have had the case where I've downloaded something in error and before it processed [big queue] I've realised and could stop this and delete it. In a big queue it would save a few strokes if I could multi-select and delete.

Such already exists to the primary queue, so it feels a no-brainer to offer such a facility to the "completed" history queue.

ricci2511:
Closes #2603.

It should look consistent with the queue's multi-select. Some screenshots:

selected-all selected-one selected-none

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.history.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.history.js
@@ -15,6 +15,8 @@ function HistoryListModel(parent) {
     self.totalItems = ko.observable(0);
     self.ppItems = ko.observable(0);
     self.pagination = new paginationModel(self);
+    self.isMultiEditing = ko.observable(false).extend({ persist: 'historyIsMultiEditing' });
+    self.multiEditItems = ko.observableArray([]);
 
     // Download history info
     self.downloadedToday = ko.observable();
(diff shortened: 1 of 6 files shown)


interfaces/Glitter/templates/static/javascripts/glitter.history.js
```diff
@@ -257,6 +259,89 @@ function HistoryListModel(parent) {
...
+            self.multiEditItems.remove(function(inList) { return inList.nzo_id == item.nzo_id; })
+        }
+
+        setCheckAllState('#history-options #multiedit-checkall', '.history-table input[name="multiedit"]')
+        return true;
+    }
+
+
+    // Check all
+    self.checkAllJobs = function(item, event) {
```
Safihre, line 288: Is there a way we can generalize this function so we can also use it for the queue? There is a lot of code duplication here, some of which is needed, but I think this function could be generalized.

Safihre (changes requested):
Thanks for taking the time and I apologize for the delay in review!

I added a review comment and noticed that it needs a fix for the tabbed layout:
image

interfaces/Glitter/templates/static/javascripts/glitter.history.js
ricci2511: Sure, I'll take a look at it again once I have time and will try to reduce duplication as much as possible.

ricci2511:
I ended up moving `addMultiEdit`, `checkAllJobs` and `doMultiDelete` up to `glitter.main.js` which are now reused by both the queue and history. This should reduce the duplication by quite a bit.

interfaces/Glitter/templates/static/javascripts/glitter.history.js
```diff
@@ -270,7 +279,7 @@ function HistoryModel(parent, data) {
     // If we update the full set every time it uses lot of CPU
     // The Status/Actionline/scriptline/completed we do update every time
     // When clicked on the more-info button we load the rest again
-    self.nzo_id = data.nzo_id;
+    self.id = data.nzo_id;
```
Safihre, line 282: Why did you change this? Because it should all work based on `nzo_id`.

interfaces/Glitter/templates/static/javascripts/glitter.history.js
```diff
@@ -41,15 +43,15 @@ function HistoryListModel(parent) {
         var newItems = [];
         $.each(data.slots, function(index, slot) {
             var existingItem = ko.utils.arrayFirst(self.historyItems(), function(i) {
-                return i.historyStatus.nzo_id() == slot.nzo_id;
+                return i.historyStatus.nzo_id() == slot.id;
```
Safihre, line 46: I don't fully understand this change?

Safihre (commented):
Looks good visually, just a few quesions.

Safihre:
Thanks!

Safihre:
Found a small bug, could you fix it?
If both queue and history have multi-select opened and I click "Check all" for the Queue, it also checks the checkbox for the History.
Same if I then uncheck 1 item, the History checkbox also updates.
image

ricci2511:
Sure, I'll take a look at it tomorrow.

ricci2511:
Just opened a new pr for the fix, should work fine now. If you have any other suggestion feel free to reach out to me.

Issue #1307, opened 2019-07-08
Apply IP filtering permissions to X-Forwarded-For


ssa3512:
In #884 logging was added for the X-Forwarded-For header, but it doesn't appear that the access check function looks at this header. When I set internet access to "no access" I am still able to access as cherrypy.request.remote.ip is still 127.0.0.1

https://github.com/sabnzbd/sabnzbd/blob/b12a723a44335d5b98bfd77858b8b658cbe76aab/sabnzbd/interface.py#L131-L147

Can we add a check for the xff header in this function and using that value instead of remote.ip if it exists?

Safihre:
@jeffgeorge and @sanderjo maybe you can help me with this one?
What is the risk? Can this header me manipulated by an attacker? Or is it configured by the user?
What if I just add a `X-Forwarded-For` header with `localhost` myself to any request? Won't it just accept it then?

jeffgeorge:
My understanding is that you can shove whatever you want into an `X-Forwarded-For` header. It's only there as an informative header, not something to do any kind of authentication check against. If we auth'ed against it what's stopping someone from passing in `X-Forwarded-For: 127.0.0.1`?

If you're running SABnzbd in a reverse-proxy situation, SABnzbd should be listening on an internal-only IP. Yes, it'll still technically be available from the host it's running from, and yes that means this access control no longer works, but by setting up a reverse proxy in the first place you're taking access control into your own hands.

ssa3512:
@jeffgeorge Is there any harm in checking this header in addition to the source IP if it is present? I believe if properly configured, the reverse proxy will override anything on an incoming request with the ACTUAL source IP in the xff header. Adding this additional check is more restrictive than the current behavior so it wouldn't lead to any new unauthorized access scenarios but should properly restrict them in the scenario where a reverse proxy is being used.

jeffgeorge:
Oh that's a good point. If you've got a reverse proxy set up `cherrypy.request.remote.ip` would be some flavor of localhost, and the `X-Forwarded-For` header should be the real ip. If we've already checked and confirmed that `cherrypy.request.remote.ip` is some flavor of localhost then yes it should be safe to *then* check `X-Forwarded-For`. You wouldn't be able to forge *both* headers at the same time.

sanderjo:
> @jeffgeorge and @sanderjo maybe you can help me with this one?
> What is the risk?

Sorry, I don't understand: is this about a security risk, or a nice-to-have regarding access control.

[...]

PR #2611, opened 2023-07-04
add optional xff header verification to check_access

jcfp:
As the "X-" at the start of the name suggests, no formal specification exists for the X-Forwarded-For header. Consensus seems to be that it's supposed to provide one or more IP addresses (with multiple entries comma-separated, spacing optional). A legitimate reverse proxy setup would typically rewrite the header though, so there will often only be a single IP address present even for connections that passed through a chain of proxies. Cherrypy seems to only pass on a single header even if multiple xff headers were sent with the request.

This changeset adds optional verification of the address(es) in the header. Verification is off by default, in which case access is determined solely on the basis of the remote IP - same as before - so existing setups relying on this behaviour continue to work.

With verification turned on, every IP address in the XFF header is checked **in addition to** the remote IP, and access only granted if all checks pass. If the header contains multiple addresses, they must all qualify for access to be granted. Invalid entries (hostnames, IP addresses with port numbers attached, random junk, etc.) aren't filtered out and may lead to access being denied.

Thus, for a typical connection coming in through a reverse proxy (remote IP 127.0.0.1 with XFF IP 12.34.56.78, as seen from the perspective of sabnzbd):
- allowed with verification off, as the remote IP is a loopback address (same as before);
- denied with verification on (fixing the reported issue), unless 12.34.56.78 is in a subnet listed in local_ranges.

For a connection made by "evil1337" to a sabnzbd instance directly accessible from the interwebs (not behind a reverse proxy; remote IP 12.34.56.78 with or without a fake XFF header):
- always denied regardless of XFF verification setting and header content, based solely on the remote IP, unless 12.34.56.78 is in a subnet listed in local_ranges (same as before).

Closes: #1307

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -49,6 +49,7 @@ from sabnzbd.misc import (
     is_ipv6_addr,
     get_server_addrinfo,
     is_lan_addr,
+    is_local_addr,
     is_loopback_addr,
     ip_in_subnet,
     helpful_warning,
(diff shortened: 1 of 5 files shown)


Safihre:
Great, could you also add it to the wiki for 4.1? Then I can merge it.

sabnzbd/interface.py
```diff
@@ -206,15 +207,14 @@ def check_access(access_type: int = 4, warn_user: bool = False) -> bool:
...
 
-    if not cfg.local_ranges():
-        # No local ranges defined, allow all private addresses by default
-        is_allowed = is_lan_addr(remote_ip)
-    else:
-        is_allowed = any(ip_in_subnet(remote_ip, r) for r in cfg.local_ranges())
+    # Never check the XFF header unless access would have been granted based on the remote IP alone!
+    if is_allowed and cfg.verify_xff_header() and (xff_header := cherrypy.request.headers.get("X-Forwarded-For")):
+        xff_ips = [ip.strip() for ip in xff_header.split(",")]
+        if not (is_allowed := all(is_local_addr(ip) or is_loopback_addr(ip) for ip in xff_ips)):
```
Safihre, line 216:
```suggestion
        is_allowed = all(is_local_addr(ip.strip()) or is_loopback_addr(ip.strip()) for ip in xff_header.split(",")))
        if not is_allowed:
```
Or something like that, sometimes the walrus operator makes things harder to read instead of easier (in my opinion :P )

Safihre:
Thanks again!

PR #2621, opened 2023-07-16
Add dark mode for wizard, config, and login

### LINKED ISSUE #2386: Dark mode for SABnzbd Config
Since the main UI has a light/dark mode

It would probably be a good idea to provide a dark mode for the config section also.

alavry:
Closes #2386

Uses `prefers-color-scheme: dark` as per https://github.com/sabnzbd/sabnzbd/issues/2386#issuecomment-1365721634

Some screenshots attached

Screenshot 2023-07-15 at 10 15 27 PM
Screenshot 2023-07-15 at 10 15 25 PM
Screenshot 2023-07-15 at 10 15 24 PM

Let me know if any tweaks are needed


interfaces/Config/templates/staticcfg/css/style.night.css
```diff
@@ -0,0 +1,212 @@
...
+.Servers .ct-series-a .ct-point {
+  stroke: #EBEBEB;
+}
+
+.col2 p,
+.col2-cats {
+  color: #AAA;
+}
+
+#content>div.colmask>div:nth-child(3) {
```
Safihre, line 196: What is this one?

interfaces/wizard/static/style.night.css
```diff
@@ -0,0 +1,120 @@
+body {
```
Safihre, line 1: I see quite some overlap, can we not combine the login/wizard/config dark skins?

Safihre (changes requested):
Thanks for the work! 

Can you make the buttons the same as in Glitter?
I did a quick walkthrough (still need to do a thorough one) and noticed the exit button here:
image

interfaces/Config/templates/staticcfg/css/style.night.css
alavry: config_rss.tmpl has an div with an inline style/without a class (`<div style="padding-left: 10px; border-bottom: 1px solid #DFDEDE;">` @ line 379) so I used nth-child to refer to it

Recognize it's fragile but felt like fixing the inline style was out of scope

I don't think it catches anything else based on my testing, but the selector could be expanded to include .RSS first in case it does

interfaces/wizard/static/style.night.css
alavry: We could - which directory should the combined skin go in?

Safihre: The `staticcfg`.

thezoggy: Also might want to use how glitter does loading the night theme based on config setting (dark or auto based on os) rather than assume by os only.

alavry:
Made all the changes - let me know if any more are needed

Are all the additional commits on this branch fine btw? Assumed it would get squashed in the end

thezoggy:
> Made all the changes - let me know if any more are needed
> 
> Are all the additional commits on this branch fine btw? Assumed it would get squashed in the end

you can squash+rebase on your side to make things cleaner

Safihre:
I'm on vacation so hope to take a look at it again next week!

Safihre:
Made some changes so it's more similar to main theme, because different shades of grey were used.
Can't say I'm a huge fan of how extra dark it is compared to regular interface, but just these small changes took me a while so I'm giving up for now.

Safihre:
Also not happy yet with the look of the wizard and login.
Hopefully soon I have a half day again to work on it :)

alavry:
Understandable, thanks for reviewing, making tweaks, and merging this anyway. I'm glad I could provide an initial implementation, even if it didn't fully achieve the overall vision.

thezoggy:
so setting theme in sab to night, does not yeild the dark mode for wizard/config/login. 
you have to also change the browser to prefer night (default looks to be set to auto, so inheirts what os is set).

so in my case, my os is light. so browser then with auto is light. so even if i set sab to use night.. it doesnt.. but only for glitter. this makes things a bit little inconsistent.

i would expect this behavior if its set to 'auto' which then inherits whatever the os is set to.. 
I would expect it to be night (just like glitter does) if I specifically set that. 

so imho, it looks like `media="(prefers-color-scheme: dark)"` needs to be dropped and just let auto.css' media query handle it like we do for glitter
https://github.com/sabnzbd/sabnzbd/blob/develop/interfaces/Glitter/templates/static/stylesheets/colorschemes/Auto.css

Safihre:
Agreed. Want to create a PR @thezoggy?

thezoggy:
> Agreed. Want to create a PR @thezoggy?

submitted. included a few dark fixes as well.. There are still a few things that look a little funky to me but its decent enough for now. one of these days we can use bootstrap 5.x which has dark mode support built in :x

PR #2634, opened 2023-07-27
Ensure all data is written to file

### LINKED ISSUE #2632: Random issue where unpack fails due to corrupt rar
Random issue where unpack fails due to corrupt rar but sab fails to catch it before it tries to unpack.

This has happened randomly on various nzbs, even on the built-in test nzbs.

I've manually run par2 checks on the archive and 1 or 2 archives are corrupted which par2 repairs, but sab never detected the error before it tried the unpack.

This is on 4.0.3 on macOS [[9fafe64](https://github.com/sabnzbd/sabnzbd/commit/9fafe64)], but it happened in previous 4.x versions.

Like the issue can happen, I remove the job, resubmit the same nzb and it unpacks correctly this time.

Does sab run a full par2 check after the download is complete before unpack?

I know that this is somewhat of a recent thing as I don't remember ever seeing this before.

I check the logs and quickcheck passes and everything before the unpack happens, then unpack fails and the webUI shows "Corrupt RAR file".

mnightingale:
Related #2632 but probably not the cause

TL;DR in raw/non-buffered mode `fout.write` may not write everything requested https://docs.python.org/3/library/io.html?highlight=write#io.RawIOBase.write

---

> Write the given [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object), b, to the underlying raw stream, and **return the number of bytes written. This can be less than the length of b in bytes, depending on specifics of the underlying raw stream**, and especially if it is in non-blocking mode. None is returned if the raw stream is set not to block and no single byte could be readily written to it. The caller may release or mutate b after this method returns, so the implementation should only access b during the method call.

Without looping the quick check CRC32 could pass but file on disk could be corrupt.

I was made aware of this while investigating some mergerfs issues a while ago; there was a bug in append mode but it was also discovered that the application wasn't handling short writes which could/will happen when passing a disk boundary depending on mergefs options.

I've now idea how it works at a low level, but I'd presume it can be related to the OS enforcing some kind of IO fairness and writing less than requested if there is high IO pressure.

It should probably be ported to a v4.0.x release.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -206,7 +206,11 @@ class Assembler(Thread):
                             file_position = article.data_begin + len(data)
                         else:
                             fout.seek(0, os.SEEK_END)
-                        fout.write(data)
+
+                        written = 0
+                        while written < len(data):
+                            written += fout.write(data[written:])
+
                         nzf.update_crc32(article.crc32, len(data))
                         article.on_disk = True
                     else:


Safihre:
For many many SABnzbd releases we never did this, even when still using MD5:
https://github.com/sabnzbd/sabnzbd/blob/efaffb829827c4e20f45a895db819e2a76f11df9/sabnzbd/assembler.py#L193-L194

Could this really be a problem now because we write faster maybe?

mnightingale:
You changed your mind?

Anyway my response I was about to post:

I'm not sure on the release history but I think `buffering=0` i.e. RawIOBase has only been used since around 3.6.0

Behaviour before that is different https://docs.python.org/3/library/io.html?highlight=write#io.BufferedIOBase.write

> Write the given [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object), b, and return the number of bytes written (always equal to the length of b in bytes, since if the write fails an [OSError](https://docs.python.org/3/library/exceptions.html#OSError) will be raised)

The chances of this ever being a problem are very low, but writing less than requested is a valid response for the underlying IO operations.

Safihre:
> You changed your mind?

My previous comment was just wondering why we didn't experience it before :)
I double checked Python's C code if there's no `writeall` or something, but there isn't in non-buffered mode.
So this check will just be needed indeed, since I still think that we don't need the overhead of Python's buffered mode.

PR #2639, opened 2023-08-03
Fix check-all checkbox state

ricci2511:
Fixes a small bug related to the multi-select checkboxes state, reported in this [comment](https://github.com/sabnzbd/sabnzbd/pull/2607#issuecomment-1662063256) from pr #2607.

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.main.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.main.js
@@ -134,7 +134,7 @@ function ViewModel() {
     })
 
     self.updateCheckAllButtonState = function(section) {
-        setCheckAllState('.multioperations-selector #multiedit-checkall', `.${section}-table input[name="multiedit"]`)
+        setCheckAllState(`[data-section="${section}"] #multiedit-checkall`, `.${section}-table input[name="multiedit"]`)
     }
 
     // Add queue or history item to multi-edit list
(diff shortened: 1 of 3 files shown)


interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -134,7 +134,7 @@ function ViewModel() {
     })
 
     self.updateCheckAllButtonState = function(section) {
-        setCheckAllState('.multioperations-selector #multiedit-checkall', `.${section}-table input[name="multiedit"]`)
+        setCheckAllState(`[data-section="${section}"] #multiedit-checkall`, `.${section}-table input[name="multiedit"]`)
```
Safihre, line 137:
```suggestion
        setCheckAllState('[data-section="${section}"] #multiedit-checkall', '.${section}-table input[name="multiedit"]')
```
I think you are on a mac and the quotes got converted!

But maybe there's a better way? We now have duplicate ID's: `multiedit-checkall`. Why not make that `multiedit-checkall-queue` and `multiedit-checkall-history` and use that to select on instead of adding `data-section`?

ricci2511: You're right, for some reason I treated `multiedit-checkall` like a class instead of id, my bad.
The quotes thing is necessary because of string interpolation.

PR #2647, opened 2023-08-06
unrar 6.23

thezoggy:
fixes "a security issue involving out of bounds write is fixed in RAR4 recovery volumes processing code."


puzzledsab:
I think this should be released soon. It looks like a serious bug for SABnzbd because unrar is executed automatically for automatically downloaded files. The reason why it doesn't have a higher CVSS score is that in normal use cases it would require user interaction. https://www.zerodayinitiative.com/advisories/ZDI-23-1152/

PR #2653, opened 2023-08-12
don't call get_unique_filename in Sorter.rename

### LINKED ISSUE #2619: Sorting in 4.0.3 results in filenames ending in .1, issue not present in 4.0.2
Recently after upgrading to 4.0.3 I noticed that many of the sort function renamed files were now ending in '.1'.

I have moves and series sorting set to the following; `%dn/%dn.%ext`.

In 4.0.3 this has started to result in `Show.Name.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb.nzb` being sorted to `/downloadFolder/Show.Name.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb/Show.Name.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb.1.mkv`.

In 4.0.2 the same `Show.Name.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb.nzb` is sorted to `/downloadFolder/Show.Name.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb/Show.Name.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb.mkv` with no other changes in settings.

Rolling back the same installation from 4.0.3 to 4.0.2 fixes the issue. Going back to 4.0.3 brings the renaming issue back.

I am running from source on Ubuntu 20.04 in a Python 3.11.4 virtualenv. All Python dependencies are updated from the requirements.txt file.

jcfp:
Remove leftover call to get_unique_filename that IIRC was only added for debugging purposes.

Fixes #2619

### DIFF
--- a/sabnzbd/sorting.py
+++ b/sabnzbd/sorting.py
@@ -558,11 +558,9 @@ class Sorter:
         # Rename it
         f_name, f_ext = os.path.splitext(largest_file.get("name"))
         filepath = self._to_filepath(largest_file.get("name"), base_path)
-        new_filepath = get_unique_filename(
-            os.path.join(
-                base_path,
-                to_lowercase(path_subst(self.filename_set, [("%fn", f_name), ("%ext", f_ext.lstrip("."))]) + f_ext),
-            )
+        new_filepath = os.path.join(
+            base_path,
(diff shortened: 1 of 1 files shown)


Safihre:
Indeed this way it would generate unnecessary an unique filename if the file already had the right name!

PR #2658, opened 2023-08-17
preserve lowercasing pattern in season sorting

### LINKED ISSUE #2657: Sorter capitalization of Season Packs
I bundled 2 bugs in this one report but the Season Packs still don't get proper capitalization.

```2023-08-16 06:54:06,106::DEBUG::[filesystem:877] Renaming "....Show.Name.S01.1080p.../Show Name - S01e08.mkv" to "show name/Show Name - S01e08.mkv".```

{%sn (%y)/%sn (%y) - S%0se%0e}.%ext
this is the sort string I have so everything so should be lowercase, and it shows as lowercase in the preview, but when it actually processes the files, they end up mixed case.

_Originally posted by @ItsME6969 in https://github.com/sabnzbd/sabnzbd/issues/2654#issuecomment-1679972863_

jcfp:
Insert extra lowercasing markers before and after the last directory separator in the sort string when doing season pack handling, to avoid breaking the lowercasing for the filename part.

Fixes #2657

### DIFF
--- a/sabnzbd/sorting.py
+++ b/sabnzbd/sorting.py
@@ -280,7 +280,7 @@ class Sorter:
                 sort_string = sort_string[:-5]  # Strip '.%ext' off the end; other %ext may remain in sort_string
             if self.is_season_pack:
                 # Create a record of the filename part of the sort_string
-                _, self.season_pack_setname = os.path.split(sort_string)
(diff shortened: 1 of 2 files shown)


Safihre:
In this case I am starting to wonder if instead of having this quite complicated code, we should just put in the manual that `{}` should be applied separately for each `/` separated part..

PR #2674, opened 2023-09-04
WIP: Add par2cmdline as option for windows, still default to multipar.

thezoggy:
Add par2cmdline as option for windows, still default to multipar.
- 32binary for multipar was still included in 64bit zip due to bad path, fixed.
- added enable_multipar so that can still be used in this transition phase/testing.. to go back and forth for now

This is WIP, just re-did my previous stuff of just getting the exe added + bits wired up to just use it. Have not done any testing with jobs that needed repair (I almost never have any.. so will have to go hunting to find some samples to use to test to verify with)


sabnzbd/interface.py
```diff
@@ -858,6 +858,7 @@ def saveSwitches(self, **kwargs):
     "api_warnings",
     "helpful_warnings",
     "ampm",
+    "enable_multipar",
```
Safihre, line 861: The unit tests are correct. You added it here, but on the actual Config page. Also, shouldn't it be a Special setting? Then we also don't need a new skintext.

sabnzbd/newsunpack.py
```diff
@@ -1034,7 +1036,7 @@ def par2_repair(nzo: NzbObject, setname: str) -> Tuple[bool, bool]:
             joinables, _, _, _ = build_filelists(nzo.download_path, check_rar=False)
 
             # Multipar on Windows, par2cmdline on the other platforms
-            if sabnzbd.WIN32:
+            if cfg.enable_multipar and sabnzbd.WIN32:
```
Safihre, line 1039: You forgot the ()

Safihre (commented):
Could you not squash your commits? Then I can see what you changed.

Also, we'll need to add it to the wiki. That's why the unit test to check that is there.

sabnzbd/newsunpack.py
Safihre: I think because of that it will always use Multipar instead of turbo.

sabnzbd/cfg.py
```diff
@@ -406,6 +406,7 @@ def validate_notempty(root, value, default):
 preserve_paused_state = OptionBool("misc", "preserve_paused_state", False)
 enable_par_cleanup = OptionBool("misc", "enable_par_cleanup", True)
 process_unpacked_par2 = OptionBool("misc", "process_unpacked_par2", True)
+enable_multipar = OptionBool("misc", "enable_multipar", True)
```
Safihre, line 409: Can you try having it disabled by default? Then it can run the unit tests on ut.

sabnzbd/interface.py
thezoggy: It is a special setting, but added just like the rest where they do have a skin text entry.. are you saying they shouldn't have them?

Safihre: Oh yes sorry, it's indeed a Special. But Special settings don't have a skintext.

thezoggy: yeah, plan was up in the air if we would keep multipar support or just remove. but as I dont think ill add it on config landing page like the 7zip/multipar one does.. ill drop here

sabnzbd/newsunpack.py
thezoggy: when I tested it worked fine to do one or the other if its enabled. but yes out of the box multipar would be enabled.. figured that would be by design until multipar is removed

thezoggy:
> 

actually only have the one commit since i redid everything. mainly was just adding par and wiring the bits up to use it instead of multipar. i just stumbled on fixing the path for the 32bit multipar when i was looking at references of where it was used

sabnzbd/cfg.py
thezoggy: sure thing

thezoggy:
> Also, we'll need to add it to the wiki. That's why the unit test to check that is there.

yep, plan was to do so after the wip stuff. but i can do so now if your good with the plan to have multipar to be toggled to keep it around for atleast a minor release version before we rip it all out

sabnzbd/newsunpack.py
Safihre: It needs to be `enable_multipar()`, not just `emable_multipar`.
I really don't know how it could work without.

thezoggy: aye, was reading it as something else (mobile). yep I've fixed up

Safihre:
Tested it and as expected, par2cmdline still doesn't work well with UTF8 on Windows.
Not supplying it with long-path notation already helps a bit for filenames with a limited number of UTF characters. But of course will fail if it ever encounters a long-path file.
Opened an issue with original par2cmdline (strangely never done before), as requested by animetosho.
Parchive/par2cmdline/issues/189 
Multipar handles both cases fine.

Not much we can do about that right now. So will switch `enable_multipar=true` and merge so people can test if they want.

Issue #2675, opened 2023-09-06
On Windows, install start menu shortcuts globally


anthonyvdotbe:
SABnzbd itself is installed globally in `$env:ProgramFiles`, but its start menu shortcuts are placed in the user-specific location (`$env:APPDATA\Microsoft\Windows\Start Menu\Programs`) instead of the global location (`$env:ProgramData\Microsoft\Windows\Start Menu\Programs`). In my experience, this is uncommon: either applications install both their program files and shortcuts globally (most applications), or they install both in user-specific locations (e.g. VS Code).

Safihre:
Seems logical. Probably leftover from ancient times when we made the NSIS script.
Do you want to submit a PR to fix it?

PR #2677, opened 2023-09-07
Use "all users" locations for shortcuts on Windows

anthonyvdotbe:
Fixes https://github.com/sabnzbd/sabnzbd/issues/2675

See [$SMPROGRAMS](https://github.com/NSIS-Dev/Documentation/blob/main/docs/Variables/SMPROGRAMS.md) documentation, which refers to [SetShellVarContext](https://github.com/NSIS-Dev/Documentation/blob/main/docs/Commands/SetShellVarContext.md).

Disclaimer: I haven't actually tested this

### DIFF
--- a/builder/win/NSIS_Installer.nsi
+++ b/builder/win/NSIS_Installer.nsi
@@ -66,6 +66,9 @@ Unicode true
   RequestExecutionLevel admin
   FileErrorText "If you have no admin rights, try to install into a user directory."
 
+;------------------------------------------------------------------
+; Makes sure shell folders (e.g. $SMPROGRAMS and $DESKTOP) use the "all users" location
+  SetShellVarContext all
 
 ;------------------------------------------------------------------
 ;Variables


builder/win/NSIS_Installer.nsi
```diff
@@ -226,6 +233,10 @@ Section $(MsgAssoc) assoc
 SectionEnd ; end of file association section
 
 Section /o $(MsgRunAtStart) startup
+  ;------------------------------------------------------------------
+  ; Makes sure shell folders (e.g. $SMPROGRAMS and $DESKTOP) use the "all users" location
+    SetShellVarContext all
```
Safihre, line 238: From the documentation I read that it needs to be set seperatly for installer and Uninstaller, so is it really needed 4x?

Safihre:
I realize this will probably be more impact full than expected. How to deal with upgraded installations: we used to check if the link is installed to see if user wants that again, but now it's in a different location.
Additionally, the uninstaller will not remove the old links. For example if you installed with 3.x but then uninstall with this adjusted version.

Safihre:
I realize this will probably be more impact full than expected. How to deal with upgraded installations: we used to check if the link is installed to see if user wants that again, but now it's in a different location.
Additionally, the uninstaller will not remove the old links. For example if you installed with 3.x but then uninstall with this adjusted version.

builder/win/NSIS_Installer.nsi
anthonyvdotbe: Most likely not, but I erred on the side of caution. I've reviewed the occurrences and eliminated some.

anthonyvdotbe:
> I realize this will probably be more impact full than expected. How to deal with upgraded installations: we used to check if the link is installed to see if user wants that again, but now it's in a different location. Additionally, the uninstaller will not remove the old links. For example if you installed with 3.x but then uninstall with this adjusted version.

These concerns should be addressed by the latest commit.

thezoggy:
Would be better to offer the option rather than change. As changing locations only really works without issues for brand new installs..

Safihre:
@thezoggy he added logic now for new and old version. I'm on vacation at the moment, so can't test it yet.

PR #2679, opened 2023-09-12
Fix when downloader paused only download force priority items

### LINKED ISSUE #2676: Downloads non-force priority articles while paused
## Steps
- Pause queue
- Add a couple of items
- Set first item to force priority

## Outcome
Some of the normal/paused items download.

## Observations
It depends on how many connections there are but with 50 at 1 Gbps I downloaded 1GB of a paused item before it idled again.

## Why
I think it'll be hitting [this section](https://github.com/sabnzbd/sabnzbd/blob/667ffec667aaf85356220f15863a02d1240fd5a9/sabnzbd/downloader.py#L614-L645) and there is an active job until the force one finishes post-processing, so it'll execute line 645.

So it can get a normal non-force article to request here:

https://github.com/sabnzbd/sabnzbd/blob/667ffec667aaf85356220f15863a02d1240fd5a9/sabnzbd/nzbqueue.py#L720-L722

I'm not sure what logic is needed to prevent it downloading a non-force nzo while paused, but I think those are the relevant parts.

mnightingale:
Fixes #2676 

A pretty simple change that seems to fix the problem.

Strangly the comment `# Not when queue paused and not a forced item` wasn't what it is doing, digging into the history a bit this is what it used to do!

It was changed in https://github.com/sabnzbd/sabnzbd/commit/6bc1c51013bc31169eba491fd57d6aeeb549280c#diff-910de0ac3edf4c8b48165397a18d3afc8b4d93dd2a09db16e92ae4963019193e - I'm not familiar with 2.x but I don't think it should have been removed, doesn't seem related to the categories to server mapping that commit was removing.

nzo.status is Status.QUEUED even though the UI shows Paused - it also checks if the downloader is paused

### DIFF
--- a/ISSUES.txt
+++ b/ISSUES.txt
@@ -48,6 +48,3 @@
   You can make SABnzbd wait for a mount of the "temporary download folder" by setting
   Config->Special->wait_for_dfolder to 1.
   SABnzbd will appear to hang until the drive is mounted.
-
-- If the queue is paused but one or more jobs have the Force priority, SABnzbd might still
-  download some data from other jobs in the queue if active servers have unused connections.
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -719,7 +719,9 @@ class NzbQueue:
         propagation_delay = float(cfg.propagation_delay() * 60)
         for nzo in self.__nzo_list:
             # Not when queue paused and not a forced item
-            if nzo.status not in (Status.PAUSED, Status.GRABBING) or nzo.priority == FORCE_PRIORITY:
+            if (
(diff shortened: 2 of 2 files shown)


mnightingale:
Spotted something related with "Request a new article immediately" that I want to fix

mnightingale:
> Spotted something related with "Request a new article immediately" that I want to fix

@Safihre changed my mind on that, self.paused could be self.no_active_jobs() like elsewhere but I'll leave it out of this PR.

Safihre:
Looks good! You can also remove the item from the ISSUES file👍

mnightingale:
> Looks good! You can also remove the item from the ISSUES file👍

Done, it'll also need removing from the wiki

PR #2694, opened 2023-09-27
Add option to force config to use light theme

### LINKED ISSUE #2691: 4.1.0 - disable dark theme for config
user complaint on discord:
>  Just upgraded to 4.1.0 - is there a way to set dark for the main front end but light for the settings page? I prefer dark for the front end but the back end is very text heavy and dark mode makes that very hard to read. I can't see an option or switch for it, but just wondered if there was a way.

thoughts on us adding a checkbox or something to force using light theme for config?

thezoggy:
add option to force web config to use light theme:
firefox_2023-09-27_17-55-46

this way if user does glitter-auto/night they can have night for the login/app but still have light theme for the config. is also an advanced option. 

Fixes #2691


thezoggy:
I tried to figure out a way to just do this via css (media queries) but could not :/
An alternative route would be to do a whole css file for night-noconfig but then would have to pollute skins to add two options (glitter - auto (noconfig) / glitter - night (noconfig) 

Let me know if you are good with this and I'll submit the config>general wiki entry

Safihre:
I'm sorry, but I just think this is unnecessary. It's a case of black or white, we're not doing halfsies.

thezoggy:
> I'm sorry, but I just think this is unnecessary. It's a case of black or white, we're not doing halfsies.

upto now, the glitter-night didnt cover config, so it has been halfsies until recently. this was just restoring that.

PR #2728, opened 2023-11-10
Implement more sophisticated duplicate handling

### LINKED ISSUE #2698: Duplicate RSS sourced downloads not obeying paused state
### SABnzbd version

4.1.0 [b7e3401]

### Operating system

Windows 10

### Using Docker image

None

### Description

I have a range of RSS feeds that load jobs, which are supposed to in a paused state. 
If sab spots a duplicate, and if duplicates are set to only "tag" rather than pause, delete etc, then the job is loaded as an active job and starts downloading, even though the RSS settings would have been to load the job as paused.

This is not the expected behaviour.
The DUPLICATE setting, if set to ignore or tag, should ensure the job is in the state a non-duplicate would be in, which in this case is paused.

This only impacts duplicate jobs, but the majority of jobs loaded by RSS feeds.

_Per [safihre] request, reporting here._

Safihre:
@mnightingale / @jcfp I was hoping you might have time to take a look at this invasive change and give me feedback or things that can maybe be improved.

### Goal
More sophisticated duplicate handling

### Requirements

1. Detect duplicates also when the first job is still in the queue or being post-processed.
2. Defer taking action on the 'duplicate' until we know for sure the first download is a success.

### Implementation details:
Duplicate action will still be applied directly if there is already a successful download in the history that matches the new job.
If the first job is still in the queue or being post-processed, apply a label `ALTERNATIVE` and pauses the job.
Duplicate detection can also work for URL's, if a custom NZB name is supplied. For example when the URL's are from a RSS feed.

### Open questions
1. Maybe allow pre-queue script to also set the `nzo.duplicate_series_key`?
2. Can we do something about race-conditions? E.g. a job jumps from post-processing to history right in between our checks.
3. Do we even need to separate regular duplicate detection from Series duplicates? Starting to wonder that.. @thezoggy maybe you have a strong opinion about that?

### TODO's

1. Change the pre-queue script to allow it to set the `nzo.duplicate` flag (since `DUP_PRIORITY` is removed to also fix #2698).
2. ~Handle the `is_proper` in the new setting.~


mnightingale:
I’ve played around with it for a while, it seems to work as intended.

I did notice the behaviour @jcfp noted above; it doesn’t take any action if you delete an active download.

In handle_duplicate_alternative:

- Case-insensitive comparison may benefit from casefold() instead of lower() to better normalise Unicode, but everywhere else in the code uses lower and the only time I can imagine it would help would be naming issues from indexers such as converting Unicode to lowercase themselves
- A really minor thing but I’d suggest using early returns, i.e., the computationally cheaper conditions first, so move lower/casefold to the end
- From an implementation point of view, handle_duplicate_alternatives is called on every postproc, it copies the nzo list, loops through them all but only takes action for duplicates.
This probably isn’t a problem for most users, but it’s my understanding there are already limitations with large queues?
In most cases it’ll be really quick to exit anyway (no duplicates).
Instead of looping through them all maybe the active NZO could have a flag that it has duplicates in the queue? so they're not all looped though every time.
There is potentially a better solution involving SQLite for the queue and history (single table + BLOBs) which would probably also resolve the large queue issues but I don’t image there is much interest in it.
- Code style preference but the main `if nzo.duplicate and ( …` is a bit confusing to read, I’d suggest splitting it into two statements `if not nzo.duplicate: continue`

> 1. Maybe allow pre-queue script to also set the nzo.duplicate_series_key?

Seems it would be quite a niche use-case?
But of that same thought process maybe the api should also allow it so an external tool can specify a dupe key rather than SAB trying to guess it - I don't know if there is much of a use case for it but you could act on add requests without needing to download the .nzb

> 2. Can we do something about race-conditions? E.g. a job jumps from post-processing to history right in between our checks.

I haven’t really looked at the code around this, but I guess you’re suggesting a lock would be required around querying the history and queue. I think that seams reasonable but there is likely potential for deadlocks.
Maybe it's just me but I wouldn't knowingly allow a race to be introduced.

thezoggy:
Somewhat related, with the more and more refugee that we get from nzbget the number one thing I hear people miss was how nzbget had the ability to hide items in history which allowed them to keep stuff for dupe detection but keep things lean in the UI/API responses to apps.

As apps like sonarr/radarr usually recommend removing completed items (on nzbget it would mark them as hidden).

Thoughts on us adding the extra column to history for that visibility?

Safihre:
@thezoggy we can consider it, but not as part of this PR or probably this release of SABnzbd.
@DarrenPIngram please use the interface of Github to reply and not replying to email, it doesn't work.

Safihre:
@jcfp Help, I refactored the pre-queue script but I made a mistake somewhere that I don't fully get.
It reports for example:
```
FAILED tests/test_functional_adding_nzbs.py::TestAddingNZBs::test_adding_nzbs_priority_sample[None-0-None--2-2--1] - AssertionError: assert 'Normal' == 'Low'
```
So that would mean:
```
prio_meta_cat 	None
prio_def_cat 	0
prio_add 	None
prio_add_cat 	-2
prio_preq 	2
prio_preq_cat 	-1
```
But why is the expected priority `Normal: 0`? I thought pre-queue takes precedence over anything?

Safihre:
I decided to leave the pre-queue script for now.
It will just not have the ability to mark a job as duplicate.
@jcfp Still hope maybe you can take a look at my pre-queue problem? I tried a bit more and clearly there's a mistake in my logic, but the whole logic that we currently have just doesn't make sense to me how it all comes together 🥲
It's in: https://github.com/sabnzbd/sabnzbd/tree/feature/prequeue

PR #2736, opened 2023-11-24
Add smarter duplicate detection

Safihre:
With #2728 I realized we can make our duplicate detection a bit smarter, why should we limit ourselves to just Series? 
Since we use GuessIt, we can also detect `movie` and `date`.
For lack of better name, I called it _Smart_ Duplicate detection 😂

I intend to also create a dedicated wiki page for it, since it's too much information for regular Switches help page.

Still need to fix the pre-queue script.


Safihre:
Wiki page: https://sabnzbd.org/wiki/extra/duplicate-detection.html

RealTracat:
Will this also check the nzb-backup for duplicates? I've noticed that files downloaded via RSS only check the history for duplicates, but not the nzb-backup folder. While manually added nzb files are checked against both?

Safihre:
Yeah but I might have broken it on the refactor 🤗
Will investigate.

Safihre:
@RealTracat The NZB-backup is checked for RSS-feeds, but only after the NZB is fetched. Because beforehand we can't really know the filename, only the `title` that was provided by the RSS-feed.

RealTracat:
So from being fetched it goes straight to being downloaded? Not fetched, checked for dupes, then put in download queue?

RealTracat:
So yes, sounds like something got broken as it's not detecting the duplicate with the nzb file in the backup folder?
(Just fyi i do have it set for checking duplicate nzb files and have an appropriate folder setup)

Safihre:
Can you set logging to Debug and then open a new issue where you show the logging of the duplicate being added but not being detected?

PR #2753, opened 2023-12-19
Improved IPv6 address detection 

sanderjo:
Delivers:

* in happyeyeballs() you can specify the family: ipv4 or ipv6. Default is still: not specified, so quickest across ipv4 and ipv6
* only report IPv6 addres if really working end-to-end (not just IPv6 address on LAN interface)
* preparation for speedtest for IPv4 resp IPv6

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -103,8 +103,7 @@ import sabnzbd.config as config
 import sabnzbd.cfg
 import sabnzbd.notifier as notifier
 import sabnzbd.zconfig
-from sabnzbd.getipaddress import localipv4, publicipv4, ipv6, dnslookup
-from sabnzbd.utils.getperformance import getpystone, getcpu
+from sabnzbd.getipaddress import local_ipv4
 import sabnzbd.utils.ssdp as ssdp
 
 try:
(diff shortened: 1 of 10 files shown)


Safihre:
I don't see the need for separate IPv4/6 speedtest. Especially because we don't show  the result to the user separetly.
If you can remove that part, I can merge it.

sanderjo:
> I don't see the need for separate IPv4/6 speedtest. Especially because we don't show the result to the user separetly. 

If ipv4 versus ipv6 speeds are different (which should not happen, and thus indicates a fault situation), it does get shown:

https://github.com/sabnzbd/sabnzbd/pull/2753/files#diff-00fa54540fa04483e6d2b9cdd86d97aa69f143b6d141c954ccb5a01634122d30R1294-R1297

> If you can remove that part, I can merge it.

Is it OK to keep the option to be able to specify the family type (INET or INET6)?

https://github.com/sabnzbd/sabnzbd/pull/2753/files#diff-f3e412127992005701400843e8db256e5a3e4728aa1b24afcd77fd13910d8cd2R78

Safihre:
> Is it OK to keep the option to be able to specify the family type (INET or INET6)?

Yes.

PR #2760, opened 2023-12-27
par2cmdline-turbo v1.1.1

thezoggy:
upgrade to latest, reduction in size of binaries due to removal of unused code


Safihre:
Thanks, didn't notice the size reduction so didn't update it so far.

PR #2761, opened 2023-12-27
refactor: display more accurate uptime

H4ckerxx44:
I have seen that the uptime only gets shown as `1d` if you are past 24 hours and thought this is pretty inaccurate and I do not really like it.

There is well more than enough space to show "0d 0h 0m" for the uptime and that would be way more accurate for not much of an overhead in terms of performance cost or whatever you might think off.

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -134,7 +134,7 @@ def format_time_left(totalseconds: int, short_format: bool = False) -> str:
 
 def calc_age(date: datetime.datetime, trans: bool = False) -> str:
     """Calculate the age difference between now and date.
-    Value is returned as a formatted string of days, hours and minutes
+    Value is returned as a formatted string of days, hours and minutes.
     When 'trans' is True, time symbols will be translated.
     """
     if trans:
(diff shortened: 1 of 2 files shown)


H4ckerxx44:
Will fix the test failures.

thezoggy:
Not sure of hand if any apps would be impacted as changing API output usually causes issues for 3rd party apps that would need to handle such change...


Quick glance I see nzb360 does show sab uptime.

H4ckerxx44:
> Not sure of hand if any apps would be impacted as changing API output usually causes issues for 3rd party apps that would need to handle such change...
> 
> Quick glance I see nzb360 does show sab uptime.

Good question, the maintainers probably know this, we will see what they say.

Safihre:
This code is used in multiple places indeed. I oppose against changing it, just have to figure out the test failures :)

H4ckerxx44:
> This code is used in multiple places indeed. I oppose against changing it, just have to figure out the test failures :)
> 
> Alternatively, we could also just change it for the uptime and not for all places where this function is used? For example I don't see the point in showing an NZB is 744d 11h 8s old, that doesn't really matter and is probably why the current output is like this.

Well, my entire motivation for this PR is just the "Uptime" display of sabnzbd, nothing related to actual NZB files.
I just find the `Uptime: 1d`  on the config page to be kinda "meh" if you get what I mean.

Safihre:
Closing due to inactivity. Unfortunately this is a bit more complicated than it might seem at the beginning!

PR #2774, opened 2024-01-09
Move UI theme selection out of advanced settings

sbalbrecht:
I think most users would agree that choosing a light/dark theme is a basic, non-advanced setting, and after discovering others had made posts trying to find the setting I thought I'd make this simple change.

I was wondering though, the field name in the template is referred to as `web_dir`, the description below the field displays the server URL, and the field label was "Web Interface" instead of something like "Web Interface Skin". All of these seemed like odd choices to me for this field, so I was wondering if a more advanced setting was meant to be here in its place?

general-settings


sabnzbd/skintext.py
```diff
@@ -265,7 +265,7 @@
     "explain-host": TT("Host SABnzbd should listen on."),
     "opt-port": TT("SABnzbd Port"),
     "explain-port": TT("Port SABnzbd should listen on."),
-    "opt-web_dir": TT("Web Interface"),
+    "opt-web_dir": TT("Web Interface Skin"),
```
Safihre, line 268: Let's use Theme instead of Skin.

Safihre:
They were moved to Special settings when we were phasing out the old skin. It doesn't need to be Advanced anymore.

sabnzbd/skintext.py
sbalbrecht: Good call

Safihre:
If you could also submit a PR to sabnzbd.github.io repo to change the naming on the 4.2 General settings wiki page, that would be great!

sbalbrecht:
Sure thing, I will tonight 👍

Safihre:
Thanks!

thezoggy:
> I was wondering though, the field name in the template is referred to as `web_dir`, the description below the field displays the server URL, and the field label was "Web Interface" instead of something like "Web Interface Skin". All of these seemed like odd choices to me for this field, so I was wondering if a more advanced setting was meant to be here in its place?

its an artifact of how we used to do stuff 15+ years ago. as a web template was loaded from a directory.
you would have your main skin choice (web_dir) then a second one (web_dir2) which you would set for your mobile/remote browsing experience. this of course was well before responsive web design or 3rd party apps that leveraged the api.
additionally how it was done back in the day was to allow people to make their own skins easily and just toss in a folder and modify existing to their desire. this over the years have been walked back to make it less of a support issue since everyone had skins that did everything differently resulting in additional functions/translations and upkeep.

sbalbrecht:
Thanks for the context @thezoggy. I can see how a bunch of third-party skins would be a nightmare to support.

Is there any demand for improving the current UI? Not sure I'd change any features (maybe besides tackling outstanding Issues) but I've been toying with the idea of modifying the color palettes and trying to give it more polish. I also notice it's using a 9 year old version of bootstrap, so I thought it may need some TLC.

thezoggy:
I more modern UI wouldn't be opposed, that's how glitter came about back in the day by saf 😁

Safihre:
That's indeed exactly how my Sab adventure started! 😄
Besides Bootstrap, the Knockout technique used is also ancient, long superceded by Angular and the likes. 
I don't think I would prefer to replace the UI with a totally new look, but it could use some refreshing.

sbalbrecht:
Yeah, it might not be too bad to update the existing templates to use a modern framework, keeping the same look and feel for now. I know Svelte + Tailwindcss is a very popular pair these days for its ease of development, and should be compatible with a Python backend.

If I were to start working on something, where would be the best way to ask for feedback or questions about the existing implementation? Sab forum, github issue, discord?

PR #2781, opened 2024-01-16
Ipv6 staging

sanderjo:
"ipv6_staging" as special switch. Goal: put in stuff that we (or others) consider not yet suitable ipv6.

First feature: measure IPv4 and IPv6 seperately. They should be (about) the same. But in case of big differences, SABnzbd will give a warning


Safihre:
Should we also add the automatic mapping of `news6` hosts?

sanderjo:
> Should we also add the automatic mapping of `news6` hosts?

Yes. That's the real added value.

PR #2783, opened 2024-01-18
use local ipv6 if routable

jcfp:
Only connect to the remote test host to discover the systems ipv6 address if there's reason to believe it will be different from the local address, i.e. if the latter is in an address range such as fc00::/7 (ipv6 "unique local addresses") or fe80::/10 (ipv6 "link-local").

Saves the time and effort of doing a remote request, and allows for displaying the ipv6 address in the status and interface options dialog even if remote lookups are disabled and/or the test host removed in config->special.

### DIFF
--- a/sabnzbd/getipaddress.py
+++ b/sabnzbd/getipaddress.py
@@ -168,8 +168,8 @@ def local_ipv6():
 
 
 def public_ipv6():
-    if ipv6_address := local_ipv6():
-        if sabnzbd.misc.is_lan_addr(ipv6_address):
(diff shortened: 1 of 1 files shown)


Safihre:
I didn't know that was possible. However, part of the external lookup is also to be able to track sever side which percentage of users have a working ipv6 connection.
This is after Newshosting kindly requested us to remove the code that automatically adds 'news6' alternatives. This way we can track statistics about it's relevance.

Safihre:
Sounds good!

PR #2808, opened 2024-02-16
Handle error code 451.

### LINKED ISSUE #2807: Error code 451 should be treated like 430
### SABnzbd version

4.2.2RC1-35-gb1a9ff708

### Operating system

Linux

### Using Docker image

None

### Description

Some providers such as giganews return error code 451 instead of 430 for some removed articles (https://support.giganews.com/hc/en-us/articles/9952660678285-What-are-errors-430-and-451)

At the moment, sabnzbd just ignores this response and the article request eventually times out and is retried a few times before being abandoned and allowing other servers to try the article.

This code should be treated the same as 430.

It may also be worth having some better handling for unknown error codes rather than just timing out despite the server actually responding.

ShaneMcC:
This is used by some servers to show that an article was intentionally removed.

Fix #2807

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -764,6 +764,18 @@ class Downloader(Thread):
                 nw.reset_data_buffer()
                 self.__request_article(nw)
 
+            else:
+                logging.warning(
+                    "%s@%s recieved unknown status code %s for article %s: %s",
+                    nw.thrdnum,
+                    nw.server.host,
+                    nw.status_code,
+                    article.article,
+                    nw.nntp_msg,
+                )
+                done = True
+                nw.reset_data_buffer()
+
         if done:
             # Successful data, clear "bad" counter
             server.bad_cons = 0
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -766,12 +766,12 @@ class Downloader(Thread):
 
             else:
                 logging.warning(
-                    "%s@%s recieved unknown status code %s for article %s: %s",
+                    T("%s@%s recieved unknown status code %s for article %s: %s"),
                     nw.thrdnum,
                     nw.server.host,
                     nw.status_code,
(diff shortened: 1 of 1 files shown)


sabnzbd/downloader.py
```diff
@@ -764,6 +764,18 @@ def process_nw(self, nw: NewsWrapper):
                 nw.reset_data_buffer()
                 self.__request_article(nw)
 
+            else:
+                logging.warning(
```
Safihre, line 768: Could you enclose the warning in the T function like I showed in my comment? That's needed to make the text translatable for other languages. And  I think we should not include the nntp_msg as this could be a huge binary blob if somehow article data was included.

ShaneMcC: I had somehow missed your previous comment entirely.

I didn't wrap it in the `T()` function as nothing else in that file is wrapped.

For the nntp_msg can we just include the first line? The whole line may provide a context-clue as to why the server responded how it did and how it should be handled in future (eg 451 is `451 0 <articleid> cancelled`)

Safihre: > I didn't wrap it in the `T()` function as nothing else in that area of the file was wrapped.

Only warnings/errors need that, since those are shown to the user. This is a warning, so it needs it.

> For the nntp_msg can we just include the first line? The whole line may provide a context-clue as to why the server responded how it did and how it should be handled in future (eg 451 is `451 0 <articleid> cancelled`)

In that case we will just have to manually debug it by connecting to that server. 
No need to complicate this, just the code will be enough for now :)

ShaneMcC: This is now wrapped with marked as translated in `9256c87e657ea0202fd717f6d6b4f9859c8192b3`, and I've made it only show the first line of the output, which will be the header line.

Safihre:
Thanks!

PR #2819, opened 2024-03-09
Update de.po

mathiasslawik:
Fix typo / style


Safihre:
You can't translate like this. Translations are handled by Transifex.
See here how to contribute help: 
https://sabnzbd.org/wiki/translate

PR #2831, opened 2024-04-01
Added tests cases to wrap calls to the Apprise integration

caronc:
Refs #2796 

Added test coverage for Apprise Notifications. Also set up framework to adapt to other notification services.

### DIFF
--- a/tests/test_notifier.py
+++ b/tests/test_notifier.py
@@ -1,0 +1,212 @@
+#!/usr/bin/python3 -OO
+# Copyright 2007-2024 by The SABnzbd-Team (sabnzbd.org)
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+tests.test_notifier - Testing notification functionality
+"""
+import sys
+from sabnzbd.notifier import send_notification, send_apprise, NOTIFICATION_TYPES
+from unittest import mock, skipIf
+
+
+class TestNotifier:
+
+    @skipIf((sys.version_info[0] != 3 or sys.version_info[1] <= 11), ">= Python v3.11 only")
+    @mock.patch("threading.Thread")
+    @mock.patch("sabnzbd.notifier.check_classes")
+    @mock.patch("sabnzbd.cfg.apprise_enable")
+    @mock.patch("sabnzbd.cfg.apprise_urls")
+    def test_to_thread_send_apprise_notification(self, mock_urls, mock_cfg, mock_check_classes, mock_thread):
+        """
+        Test up to threading call as mock does not carry forward inside; this is
+        done in another test
+        """
+
+        #
+        # Underlining Notifier functions that should be tested in another unit test
+        #
+        # A return of >1 implies we have configuration ready
+        mock_check_classes.return_value = 1
+
+        #
+        # Mocks below pertain to testing
+        #
+        mock_cfg.return_value = True
+        mock_urls.return_value = [
+            "json://localhost",
+            "xml://localhost",
+            "pbul://credentials",
+        ]
+        for t in NOTIFICATION_TYPES.keys():
+            # send notification doesn't return anything so there is nothing to test here other
+            # than the fact we didn't throw an exception if we made it this far and each
+            # iteration succeeds.
+            send_notification("title", "body", t)
+
+    @skipIf((sys.version_info[0] != 3 or sys.version_info[1] <= 11), ">=Python v3.11 only")
+    @mock.patch("sabnzbd.notifier.get_target")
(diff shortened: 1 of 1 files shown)


tests/test_notifier.py
```diff
@@ -0,0 +1,183 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+tests.test_notifier - Testing notification functionality
+"""
+import sys
+from sabnzbd.notifier import send_apprise, NOTIFICATION_TYPES
+from unittest import mock
+
+if sys.version_info[0] == 3 and sys.version_info[1] < 11:
```
Safihre, line 25: What is the difference between the 2 tests?

tests/test_notifier.py
```diff
@@ -0,0 +1,183 @@
...
+
+    class TestNotifier:
+
+        @mock.patch("sabnzbd.notifier.get_target")
+        @mock.patch("sabnzbd.DIR_PROG")
+        @mock.patch("apprise.Apprise.notify")
+        @mock.patch("apprise.Apprise.add")
+        @mock.patch("sabnzbd.cfg.apprise_urls")
```
Safihre, line 33: For `cfg` things we have our own wrapper:
https://github.com/sabnzbd/sabnzbd/blob/develop/tests/test_misc.py#L46

Safihre:
Thanks for remembering my comment about testing :)
In general, what do we really test here? Since the actual `Apprise` calls are mocked?

tests/test_notifier.py
caronc: `@mock.patch` in Python v3.11 has different behavior then earlier versions. `apprise.Apprise.add` wouldn't work as it does in previous versions.

The only workaround i found was to do the exact same test with the `@mock.patch` handling for pre-py311 and post-py311. Post versions needed `apprise.Apprise.Apprise.add` (which doesn't with the earlier versions)

The `unittest.skipIf` doesn't work in Python 3.12; it kept running the pre-py311 test which failed because of the `@mock.patch`

I also tried doing the same if block as this and just swapped the strings to be passed into the patch call, and that didn't work either (looked a whole lot cleaner). After 3.5hrs, i just got frustrated and what you saw worked so i just left it.

The problem is that Apprise works in all these environments, but the evolution of the `unittests` module has not been seamless between Python versions of the same major.

Perhaps in this case, there is something i need to do in my `apprise/__init__.py` to resume compatibility with Python 3.11 and forward for unit test references...

tests/test_notifier.py
caronc: I mean for the purpose of this test, mocking the response here handles everything. But i can try to switch it up for something more complicated if you want to leverage what you've already created. I won't be able to look into this unfortunately until a week or 2 as I'm traveling for work.

caronc:
> Thanks for remembering my comment about testing :) In general, what do we really test here? Since the actual `Apprise` calls are mocked?

The tests created here don't do much, but they still verify exceptions are caught and handled. They test all of the edge cases in the threaded call Sabnzbd makes under the hood in the `sabnzbd/notify.py` file. 

Just my personal view: Sabnzbd should see Apprise as a black box (as it would `requests` and any other framework it builds on or leverages). You should never need to mock beyond it's highest level to test your own application. Trust that these modules already cover there own set of test cases. 🙂 
Besides, for Apprise, we've already wrapped it's entry points in `try/except` so the overhead to completely mask you concern of it taking out Sabnzbd should be handled, or no?  Thoughts?

I need to find out why `@mock.patch` path breaks with Apprise and >=Python 3.11.  This one has me stumped 😞

Safihre:
What I mostly want to catch is that when Apprise adds new sub-dependencies, that we can catch it somehow and not just when users start complaining.
 
You don't put your imports in a try/except, right? So then when Renovate-bot updates Apprise, the CI-tests should just fail on our side because the imports of unknown modules fail.

caronc:
@Safihre , that's correct, there are already try/except checks in place.
[
1. [Here is the `catch all`](https://github.com/caronc/apprise/blob/6cd528cd5c706bd9cb41e6c433868bbf17ab2de5/apprise/Apprise.py#L184-L204) for the `Apprise.add()` where URL's are added
1. [Here is the sequential `catch all` ](https://github.com/caronc/apprise/blob/6cd528cd5c706bd9cb41e6c433868bbf17ab2de5/apprise/Apprise.py#L580-L593) for the `Apprise.notify()` call
1. [Here is the parallel `catch all` ](https://github.com/caronc/apprise/blob/6cd528cd5c706bd9cb41e6c433868bbf17ab2de5/apprise/Apprise.py#L623-L635) for the `Apprise.notify()` call

We're wrapping exception cases that are already handled,but I respect your concern so the try/catch has just been double downed as per your request.

The Test cases added do test a random exception being thrown [here](https://github.com/sabnzbd/sabnzbd/blob/1cebb5ccf0f7ad5f897674c10567d209183494b1/tests/test_notifier.py#L165-L172) while impossible, still allow you to test your concern and enter into the `except` block that would have never otherwise been entered.

caronc:
I created a repository [here](https://github.com/caronc/117860-cpython-issue/) demonstrating the issue along with an accompanied cpython issue.

The current argument is the issue you're seeing is related to the project file naming i chose. That i should use all lower case filenames and not ones that align with the objet within it.  This will involve some heavy lifting (possibly breaking changes to Apprise).  So i'll have to investigate how much effort this is.  The true bug imo is with Python `unittest` which broke in Python v3.11.

these comments are both outside the contents of this ticket, but also they relate too.  the work-around in place on this commit looks messy, but it's the only means I had to work around the Python (reported) bug. :slightly_frowning_face:

Safihre:
While I appreciate the effort very much, I think right now it's not  really needed to add this. 
I might give the suggestion that the notifier is unit-tested, even though it's not really. Especially since we couldn't catch things like #2837.

PR #2854, opened 2024-05-07
External internet access denied: show IP address of client

sanderjo:
To avoid ongoing confusion for users getting "External internet access denied - https://sabnzbd.org/access-denied". Often happens with docker, virtual machine, VPN and tunnel users: it's not considered local by SABnzbd, because of a different-than-local IP address. 
EDIT: and, yes, IPv6 users, even when accessing SABnzbd on (host/LAN) locally via the public IPv6 address.

Error now includes the IP address of the client. This should make clearer why SABnzbd considers it external access.

Example:

"External internet access denied - https://sabnzbd.org/access-denied --- Request from 2001:abcd:4300:200:c99:30cb:b7db:8e80"

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -164,7 +164,8 @@ def secured_expose(
         if not check_access(access_type=access_type, warn_user=True):
             cherrypy.response.status = 403
             if cfg.api_warnings():
-                return _MSG_ACCESS_DENIED
+                ipaddress = cherrypy.request.remote_label.split()[0]  # first item is the ip address
+                return f"{_MSG_ACCESS_DENIED} --- Request from {ipaddress}"
             return
 
         # Verify login status, only for non-key pages


sanderjo:
> Externally we want to give as little information as possible.

So users getting this message come to the forum asking why it happens.

Your own IP address is not a secret to yourself. 

> These users cannot see their Docker/etc logs?

Apparently. And even if so: more work. Logging into your docker and then find logging is not easy for normal sab-docker users.

sanderjo:
> Note that the address reported here is the client IP _as seen from the perspective of sab_. Depending on the setup, that may not match the client's own public address and thus could leak info, for example in case of a reverse-proxy.

That would be my goal; setups like that cause the confusion, so let's print the IP-address.

Safihre:
Users with such setups should be able to open the logs from command line...

sanderjo:
https://github.com/sabnzbd/sabnzbd.github.io/commit/47a59f8198e884692cd9c24c8beab3c787800d06

PR #2856, opened 2024-05-11
Added tests cases to wrap calls to the Apprise integration

caronc:
I fixed the `import` and `mock.patch()` issues with Apprise tied to [this very large merge request](https://github.com/caronc/apprise/pull/1119) and bumped its version to v1.8.0.

It allows testing with Apprise to be a lot more seamless now.  I was able to properly test using the `cfg` section of your library.

More tests can be added for all other notification services, and the one with Apprise can be expanded as well.  But this is a much cleaner implimentation of #2831


Safihre:
This looks good.

PR #2865, opened 2024-05-22
set basic systemd hardening options

jcfp:
Set a bunch of hardening options in the systemd service file that are unlikely to get in any legitimate user's way. Most are rather self-descriptive; `ProtectSystem=full` prevents modifications to critical parts of the filesystem (`/boot` `/efi` `/etc` and `/usr`).


Safihre:
Isn't `/usr` also where the default download location is `/usr/home/xx`?

sanderjo:
> Isn't `/usr` also where the default download location is `/usr/home/xx`?

No: /home/<user>/

PR #2873, opened 2024-05-28
fix checkdir argument handling

### LINKED ISSUE #2872: checkdir util fails when the directory path contains spaces
### SABnzbd version

4.3.2rc2

### Operating system

linux, mac; windows unknown

### Using Docker image

None

### Description

The code runs `df` by passing both the command and the directory argument in a single string, which may result in lack of proper quoting. If the directory path contains spaces (etc...), this causes the util to fail at detecting FAT32 or in some circumstances even report info for another partition/directory.
https://github.com/sabnzbd/sabnzbd/blob/b1d42c7c222b059fb3fdb0a5cd1fd36996c31dfe/sabnzbd/utils/checkdir.py#L43-L44
https://github.com/sabnzbd/sabnzbd/blob/b1d42c7c222b059fb3fdb0a5cd1fd36996c31dfe/sabnzbd/utils/checkdir.py#L86-L87
both calling `os.popen` which wraps [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen):
https://github.com/sabnzbd/sabnzbd/blob/b1d42c7c222b059fb3fdb0a5cd1fd36996c31dfe/sabnzbd/utils/checkdir.py#L13-L16

jcfp:
This should fix the issue with the spaces (and a bunch of other chars) breaking the fat32 check, but for lack of disks formatted that way, let alone access to a macos system, my options to test any of my changes were really limited. Noticed another `os.popen` in the macos part of the code that could also suffer from the same problem, tried to fix that too. I did not look at any of the other logic, or the windows part at all.

Fixes #2872

### DIFF
--- a/sabnzbd/utils/checkdir.py
+++ b/sabnzbd/utils/checkdir.py
@@ -11,7 +11,7 @@ from typing import List
 
 
 def getcmdoutput(cmd: List[str]) -> List[str]:
-    """execectue cmd, and give back output lines as array"""
+    """execute cmd, and return a list of output lines"""
     subprocess_kwargs = {
         "bufsize": 0,
         "shell": False,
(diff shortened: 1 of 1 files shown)


sabnzbd/utils/checkdir.py
```diff
@@ -80,35 +74,37 @@ def isFAT(check_dir):
...
-                        print(thisline)
                     # Starts with /, so a real, local device
                     device = thisline.split()[0]
-                    mountcmd = "mount | grep " + device
-                    mountoutput = os.popen(mountcmd).readline().strip()
-                    if debug:
-                        print(mountoutput)
+
+                    # Run the equivalent of "mount | grep $device"
+                    p_mount = subprocess.Popen(["mount"], stdout=subprocess.PIPE)
```
Safihre, line 84: This cannot be done through ` getcmdoutput`?
Like `getcmdoutput(["mount","|","grep","device"])?

Safihre: Oh I see that stackoverflow says _no_ 😔

Issue #2884, opened 2024-06-17
4.3 - backup_for_duplicates -- default not in sync with docs


thezoggy:
### SABnzbd version

4.3.2

### Operating system

unraid

### Using Docker image

linuxserver

### Description

- backup_for_duplicates

in config we default to off:
https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/cfg.py#L453

in docs we say its default to on:
https://sabnzbd.org/wiki/configuration/4.3/special

imho docs is good, we just need to fixup cfg

PR #2885, opened 2024-06-17
default backup_for_duplicates to on to align to docs

thezoggy:
closes https://github.com/sabnzbd/sabnzbd/issues/2884


Safihre:
This is disabled on purpose. We don't need this anymore for new installations due to the new Archive.

Issue #2892, opened 2024-06-24
Windows on ARM / Snapdragon: unrar and par2 not found ... because of uname machine CPU arch detection


sanderjo:
### SABnzbd version

4.3.2, Windows x86_64 exe version

### Operating system

Windows on ARM / Snapdragon Elite X

### Using Docker image

None

### Description

See https://forums.sabnzbd.org/viewtopic.php?p=132134#p132134

Cause of the problem is https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/newsunpack.py#L119 and https://github.com/sabnzbd/sabnzbd/blob/develop/sabnzbd/__init__.py#L40

Workaround: use the 32-bit (or if possible: 64-bit) x86 exe, just like we do a loose check for 7za.

Real solution: find out what Windows on ARM / Snapdragon Elite X reports for `platform.uname().machine` (pity: not specified in https://docs.python.org/3/library/platform.html#platform.machine )

EDIT: it's in the logging provide by the user
```
2024-06-24 05:39:35,746::INFO::[SABnzbd:1161] CPU architecture = ARM64
2024-06-24 05:39:35,764::INFO::[SABnzbd:1164] Platform = nt - Windows-11-10.0.26100-SP0
```
from
`logging.info("CPU architecture = %s", platform.uname().machine)`

/EDIT

 and handle it a bit more clever. And/or start using unrar for Windows on ARM (as soon as it's there; not yet on https://www.rarlab.com/download.htm).

```
2024-06-24 05:39:35,746::INFO::[SABnzbd:1159] Python-version = 3.12.3 (tags/v3.12.3:f6650f9, Apr 9 2024, 14:05:25) [MSC v.1938 64 bit (AMD64)]

2024-06-24 05:35:44,890::INFO::[notifier:157] Sending notification: Fout - par2-programma niet gevonden. (type=error, job_cat=None)
2024-06-24 05:35:44,890::ERROR::[SABnzbd:453] par2-programma niet gevonden.
2024-06-24 05:35:44,891::INFO::[notifier:157] Sending notification: Fout - unrar-programma niet gevonden. (type=error, job_cat=None)
2024-06-24 05:35:44,891::ERROR::[SABnzbd:468] unrar-programma niet gevonden.

2024-06-24 05:35:44,891::INFO::[SABnzbd:473] 7za binary... found (C:\Program Files\SABnzbd\win\7zip\7za.exe)
```

[...]

PR #2893, opened 2024-06-24
make WIN64 also true if on ARM64

sanderjo:
closes https://github.com/sabnzbd/sabnzbd/issues/2892

Note: this is for the x86_64 version of SABnzbd.exe with x86_64 python/unrar/par2/sactools emulated on Windows on ARM aka Snapdragon Elite X


Safihre:
Thanks!

Issue #2896, opened 2024-06-27
Replace Par2.exe with ARM native build on Windows ARM Platform


jgershgorin:
### Description

By default on Windows SABnzbd uses multipar instead of par2 command line. So first, we have to turn off multipar by going to:

Settings->Special, uncheck "enable_multipar" and press the "Save Changes" button at the bottom.

Then download the ARM64 native par2 from:

https://github.com/animetosho/par2cmdline-turbo/releases 

Latest is "par2cmdline-turbo-v1.1.1-win-arm64.7z" as of this post. Uncompress the par2.exe in this file to the \SABnzbd\win\par2\x64 folder, replacing the existing par2.exe

Safihre:
Closing this, we will make a proper ARM64 release when it's possible.

Issue #2613, opened 2023-07-07
Switching to par2cmdline-turbo


Tristar7:
Currently the fastest par2 verify/repair tool

https://github.com/animetosho/par2cmdline-turbo

Safihre:
I think you are not asking this in the right place. We don't include par2cmdline, only for macOS but I doubt that's the OS you are on.
It's the package manager for the OS you are on that determines which one to use.
We could ask @jcfp if he could update the Ubuntu repo to use this one?
Or I could use it for Synology package.
I could use it for snapcraft but I hate snapcraft so don't want to do that.
Maybe @francoism90 wants to use it for the flatpack?

francoism90:
@Safihre I would like to stay as close as possible to the SABnzbd defaults. Yeah, the lib versions may differ, but in most cases it's the same as building it from source.

thezoggy:
the solution for linux has just to use docker mod to load par2cmdline-turbo   instead of normal one.

i have a branch for par2cmdline-turbo for mac/win i just havent gotten around to tossing pr to add it upstream as I had been working with animetosho to work out the kinks last few months for mac (mainly arm stuff). but its been good for a bit.

my plan is to depref multipar and just have us use par2cmdline-turbo across the board (mac/win) with included binaries. also works for win arm but not going to include that one for now

Tristar7:
> I think you are not asking this in the right place. We don't include par2cmdline.. It's the package manager for the OS you are on that determines which one to use.

ok... such an answer seriously confuses me.

I'm on win, and **you are including multipar there**.

I'm saying to throw that snail out and replace it with par2cmdline-turbo instead.

I'd expect command-line parameters to be different so it would require some changes on Sab's side, but if not I'll just switch it myself..

Safihre:
You even want to replace Multipar? I guess that's possible indeed.
I didn't see that it's even faster than Multipar nowadays.
@thezoggy you want to do this or want me to do it?

Tristar7:
Multipar could stay as an option, but I see no point in keeping it.

Turbo is overall twice as fast / needs half the time in some benches, so it should become the default, or at least a choice to switch to.

thezoggy:
> You even want to replace Multipar? I guess that's possible indeed. I didn't see that it's even faster than Multipar nowadays. @thezoggy you want to do this or want me to do it?

[...]

PR #2900, opened 2024-06-30
use native Win-on-ARM64 par2 exe

sanderjo:
closes https://github.com/sabnzbd/sabnzbd/issues/2896

the included par2.exe for Windows-on-ARM64 is from https://github.com/animetosho/par2cmdline-turbo/releases : par2cmdline-turbo-v1.1.1-win-arm64.7z


sanderjo:
Interesting:

On Windows-ARM64, the native arm64\par2.exe is found

`2024-06-30 11:09:43,982::INFO::[SABnzbd:452] par2 binary... found (C:\Program Files\SABnzbd\win\par2\arm64\par2.exe)
`
... but multipar\par2j64.exe is used / preferred?

`2024-06-30 11:13:44,762::INFO::[misc:1249] [N/A] Running external command: ['C:\\Program Files\\SABnzbd\\win\\multipar\\par2j64.exe', 'r', '-vl2', '-uo', '-vs2', '-vd\\\\?\\C:\\Users\\<USERNAME>\\Downloads\\incomplete\\missing_articles_test_download_100MB\\__ADMIN__', '\\\\?\\C:\\Users\\<USERNAME>\\Downloads`

I guess that is how it works on Windows-X86 too. Is that what we want @Safihre  @animetosho , or prefer par2 from @animetosho ?

mnightingale:
> Having played with this, the major drawback is that the file handling is still using the original par2cmdline which sucks at unicode filenames. Multipar handles that without problems on Windows.

_Originally posted by @Safihre in https://github.com/sabnzbd/sabnzbd/issues/2613#issuecomment-1665630486_
           
I think there is a switch `enable_multipar` you can use to prefer par2cmdline, but it seem it has problems with unicode.

Safihre:
Exactly! It doesn't handle unicode at all on Windows 🥲 so it's just not a good option for regular users.

thezoggy:
I'll just start off saying that I think at some point we prob should do a proper winarm installer/build once there is a enough of a user base and applications to justify it.

From my previous testing I did not see any meaningful value in the arm64 variant par2cmdline-turbo compared to the x64/x86 versions (with regards to speed/being cpu bound):

firefox_2024-06-30_17-52-53

Now I do not have cpu % at the time of my test (as I was mainly focused on time of execution). 
Maybe Sander has better testing that shows another picture though.

Safihre:
Will decline this one until it can handle UTF8 correctly.

Issue #2880, opened 2024-06-12
Some smaller files not correctly renamed / deobfuscated


swindmill:
### SABnzbd version

4.3.2

### Operating system

Fedora 40

### Using Docker image

linuxserver

### Description

When retrieving at least one nzb, the largest / main `.mp4` file is named correctly, but smaller additional image and subtitle files are left with names like the following:

- `'[N3wZ] \SAA8Ya311417\::[PRiVATE] \4c1364e8a9\::a1dcc7cc5a5706.c0c06e530978b33ede7057745e3505.d34f94fe::+c1774e456437+ [newzNZB] [11+12] - yEnc  52127 (1+1).jpeg'`
- `'[PRiVATE] \073152052d\::94e87fcdd83495.a8abae491d4c709b054020ebffcaed.75234e79::+bc9fb05a653d+ [newzNZB] [7+12] - yEnc  92807 (1+1).jpeg'`
- `'[PRiVATE] \f506747fd0\::e0fa915c514b52.855ee665aa804562bee98a15f6fbed.215a90d5::+e80ad65074d9+ [newzNZB] [2+12] - yEnc  100531 (1+1).txt'`

Other downloaders like the [Newshosting newsreader](https://www.newshosting.com/newsreader/) or [NZBGet](https://github.com/nzbgetcom/nzbget) more correctly rename these files to look like the following:

- `i_8.jpg`
- `i_4.jpg`
- `4_English.srt` (notice the correct `.srt` extension vs the `.txt` extension SABnzbd is using

I enabled debug logging and retrieved this nzb, here is a log excerpt: [sabnzbd.log](https://github.com/user-attachments/files/15797710/sabnzbd.log)

thezoggy:
there is a long sorid history of newznzb/wtfnzb/xpost stuff that is purposely done in a way that if you get the nzb from their site its good but if you get it from a site like .su or sites that 'index' it it will be  pretty crappy. you can look at reddit for countless things about it. most newznab sites have blacklisted them for a reason.. but sadly .su isnt really maintained so it never did. so you prob see it slipping in from there.  

now generally sab is actually better than other stuff on handling them compared to nzbget and others..
nzbget used to not handle it at all, but the recent updated fork can handle some but still
https://github.com/nzbgetcom/nzbget/issues/105

[...]

PR #2903, opened 2024-07-06
deobfuscate_subtitles

sanderjo:
renames loose .srt files so they get the same basename as the largest file in the directory

Closes https://github.com/sabnzbd/sabnzbd/issues/2880

Test .NZBs with short .srt files that will get renamed:

obfuscated mainfile:

https://raw.githubusercontent.com/sanderjo/NZBs/master/Reftestnzb_Subtitles_Obfuscated_Mainfile_405938efd30b.nzb

non_obfuscated mainfile:

https://raw.githubusercontent.com/sanderjo/NZBs/master/Reftestnzb_Some_Subtitles_1234_ebcb671ed6f6.nzb

### DIFF
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -322,16 +322,18 @@ def deobfuscate_subtitles(directory):
     """
     Find .srt subtitle files, and rename to match biggest file
 
-    Some_Big_File_2024.mp4       #### biggest file
+    Some_Big_File_2024.mp4       #### largest file
     Some_Big_File_2024.srt
+    Some_Big_File_2024.ger.srt   # no renaming wanted!
     14_English.srt
     dut.srt
     Something.else.txt
 
     will result in
 
-    Some_Big_File_2024.mp4       #### biggest file
+    Some_Big_File_2024.mp4       #### largest file
     Some_Big_File_2024.srt
+    Some_Big_File_2024.ger.srt   # no renaming, because same basename / start as largest file
     Some_Big_File_2024.14.English.srt
     Some_Big_File_2024.dut.srt
     Something.else.txt
@@ -346,8 +348,8 @@ def deobfuscate_subtitles(directory):
     # get srt files:
     srt_files = glob.glob(os.path.join(directory, "*.srt"))
     for srt_file in srt_files:
-        if without_extension(srt_file) == without_extension(largest_file):
-            # already the same, so skip
+        if without_extension(srt_file).startswith(without_extension(largest_file)):
+            # already the same start as the largest file, so skip
             continue
         # not the same, so rename the srt file
         filename_only = os.path.basename(srt_file)  # like "14_English.srt", so no path
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -27,7 +27,6 @@ files to the job-name in the queue if the filename looks obfuscated
 Based on work by P1nGu1n
 
 """
-import glob
 import hashlib
 import logging
 import os
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -322,7 +322,7 @@ def deobfuscate_subtitles(directory):
     """
     Find .srt subtitle files, and rename to match largest file
 
-    Some_Big_File_2024.mp4      #### largest file
+    Some_Big_File_2024.mp4      # largest file
     Some_Big_File_2024.srt      # no renaming wanted
     Some_Big_File_2024.ger.srt  # no renaming wanted
     14_English.srt              # to be renamed
@@ -346,7 +346,7 @@ def deobfuscate_subtitles(directory):
     largest_without_ext = without_extension(largest_file)  # get full path base name of largest file
     logging.debug(f"Largest_file {largest_file} with base {largest_without_ext}")
 
-    # get srt files:
+    # get srt files and handle them one by one
     srt_files = glob.glob(os.path.join(directory, "*.srt"))
     for srt_file in srt_files:
         if without_extension(srt_file).startswith(largest_without_ext):
--- a/sabnzbd/deobfuscate_filenames.py
+++ b/sabnzbd/deobfuscate_filenames.py
@@ -317,7 +317,7 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str) -> List[str]:
 
 def without_extension(fullpathfilename):
     """Returns full file path, without extension
-    So '/some/dir/somefile.bin' resulst in '/some/dir/somefile'"""
+    So '/some/dir/somefile.bin' results in '/some/dir/somefile'"""
     return os.path.splitext(fullpathfilename)[0]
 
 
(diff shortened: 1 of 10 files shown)


sabnzbd/postproc.py
```diff
@@ -574,6 +574,7 @@ def process_job(nzo: NzbObject) -> bool:
                     # Deobfuscate the filenames
                     logging.info("Running deobfuscate")
                     deobfuscate.deobfuscate(nzo, newfiles, nzo.final_name)
+                    deobfuscate.deobfuscate_subtitles(workdir_complete)
```
Safihre, line 577: I just can't be used like this.
For users that use the `folder/*` notation for their Category settings, all jobs will be unpacked to the same folder without sub-folders for each job. So the `workdir_complete` could contain 10 different movies and assosiated files.
And a lot of users have this, since it's the solution when Googling for "stop SABnzbd from creating job folders".

You will need to change `deobfuscate.deobfuscate` to update the `newfiles` list, and only use that list as your input for `deobfuscate.deobfuscate_subtitles`.

If you want, I can make that change (the updating of `newfiles` in `deobfuscate.deobfuscate`)?

sanderjo: Yes, great if you can do that.

sanderjo:
> What happens in case a job contains multiple larger files, each accompanied by already correctly named srt files?

Good point. In the existing main deobfuscate(), there is already a check for that, via `first_file_is_much_bigger(filelist)`. I'll investigate & solve.

Thanks

sabnzbd/postproc.py
Safihre:
```suggestion
                    deobfuscate.deobfuscate_subtitles(newfiles)
```
You didn't use the output of the previous one?

sabnzbd/deobfuscate_filenames.py
```diff
@@ -225,90 +233,151 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+
+    # if directory, create list from that directory
+    if type(input) == str and os.path.isdir(input):
+        all_files = [os.path.join(root, file) for root, dirs, files in os.walk(input) for file in files]
+    elif type(input) == list:
+        # already a list, so nothing to do
+        all_files = input
+    else:
+        logging.debug(f"{input} no directory and no list")
+        return None
```
Safihre, line 356: This whole section seems unnecessary, since we always use a list as input.
Just assume that the input is a list and then modify the tests to input a list instead of a directory.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -225,90 +233,151 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+        logging.debug("No .srt files found, so nothing to do")
+        return None
+
+    # check there is a clearly biggest file
+    if not (biggest_file := get_biggest_file(all_files)):
+        logging.debug("No clearly biggest file found, so no subtitle renaming feasible")
+        return None
+
+    biggest_file_without_ext = without_extension(biggest_file)  # get full path base name of biggest file
+    logging.debug(f"biggest_file {biggest_file} with base {biggest_file_without_ext}")
```
Safihre, line 369: Logging line should be made a bit more human-friendly or be removed.

sabnzbd/deobfuscate_filenames.py
```diff
@@ -225,90 +233,151 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+
+    # handle srt files one by one
+    for srt_file in srt_files:
+        if without_extension(srt_file).startswith(biggest_file_without_ext):
+            # already the same start as the biggest file, so skip
+            continue
+        # not the same start, so rename the srt file
+        filename_only = os.path.basename(srt_file)  # like "14_English.srt", so without path
+        filename_only = filename_only.replace("_", ".")  # replace underscore with dot
```
Safihre, line 378: Why? This seems unrelated functionality?

sabnzbd/deobfuscate_filenames.py
```diff
@@ -225,90 +233,151 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+        # now put that name after the base name of the biggestfile:
+        new_full_name = f"{biggest_file_without_ext}.{filename_only}"  # put (renamed) srt behind that
+        unique_filename = get_unique_filename(new_full_name)  # make sure it's really unique
+        logging.debug(f"Renaming subtitle {srt_file} to {unique_filename}")
```
Safihre, line 382: This is already logged by the `renamer` function, so it can be removed.

sabnzbd/deobfuscate_filenames.py
sanderjo: It makes VLC and thus the user happy ... which is the goal of this PR.

See https://github.com/sabnzbd/sabnzbd/issues/2880#issuecomment-2168680195 : if we keep the "_" in the filename, it won't show up nicely in VLC. If we replace the "_" with ".", it does appear nicely in VLC.

sanderjo: I've added a comment to make that clearer

Safihre (changes requested):
I think we should also update the description on the Switches page to include this new functionality.
image

sabnzbd/deobfuscate_filenames.py
```diff
@@ -225,90 +232,140 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+        if without_extension(srt_file).startswith(biggest_file_without_ext):
+            # already the same start as the biggest file, so skip
+            continue
+        # not the same start, so rename the srt file
+        filename_only = os.path.basename(srt_file)  # like "14_English.srt", so without path
+        filename_only = filename_only.replace("_", ".")  # replace underscore with dot; better for VLC
+        # now put that name after the base name of the biggestfile:
+        new_full_name = f"{biggest_file_without_ext}.{filename_only}"  # put (renamed) srt behind that
+        unique_filename = get_unique_filename(new_full_name)  # make sure it's really unique
+        renamer(srt_file, unique_filename)  # ... and rename actual file on disk
```
Safihre, line 371: Should we maybe also include a History update?
```
nzo.set_unpack_info("Deobfuscate", T("Deobfuscate renamed %d subtitle file(s)") % nr_files_renamed)
```

sabnzbd/deobfuscate_filenames.py
```diff
@@ -225,90 +232,140 @@ def deobfuscate(nzo, filelist: List[str], usefulname: str):
...
+    logging.debug(f"Using as base filename: {biggest_file_without_ext}")
+
+    # handle srt files one by one
+    for srt_file in srt_files:
+        if without_extension(srt_file).startswith(biggest_file_without_ext):
+            # already the same start as the biggest file, so skip
+            continue
+        # not the same start, so rename the srt file
+        filename_only = os.path.basename(srt_file)  # like "14_English.srt", so without path
+        filename_only = filename_only.replace("_", ".")  # replace underscore with dot; better for VLC
```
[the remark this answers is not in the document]
sanderjo: I do not agree, but to avoid discussion, and get the PR done, I'll remove the line.

sabnzbd/deobfuscate_filenames.py
sanderjo: Good idea. Done

PR #2922, opened 2024-08-16
remove pyfakefs umask workarounds

jcfp:
Remove the temporary umask workarounds introduced in #2905, and put pyfakefs' new apply_umask parameter to good use.


Safihre:
Thanks 👍

PR #2927, opened 2024-08-30
Add doctype

bt90:
The login page is missing a doctype, which causes Chrome and Firefox to render it in quirks mode:

image

https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode


Safihre:
Thanks!

PR #2934, opened 2024-09-17
add additional colors for server priorities

thezoggy:
5 colors before, 8 colors now. added teal, orange, brown
firefox_2024-09-17_12-16-09 sab-servers


sanderjo:
Sidenote: I put my servers at prio's with "room" / "distance" between them, so it's easier to put one server before / in between other servers.

image

PR #2951, opened 2024-10-26
Better analysis & better user info if newsserver connection fails

sanderjo:
If a user fills out a weird port for a correct newsserver (which happens to speak on port 80 too), SABnzbd does say 
image
... which is confusing.

So improved that message to
image

I rewrote the python code to make the logic more clear.

Feedback wanted, for example `port119working` ... is that the correct variable naming? Maybe better: port_OK_119 ?

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1355,14 +1355,18 @@ def test_nntp_server_dict(kwargs: Dict[str, Union[str, List[str]]]) -> Tuple[boo
                 "possibly an indexer, not a usenet server. You have to fill a usenet server."
             ) % (host, port, host)
         elif port563working and port not in [119, 563]:
-            # it's a newsserver (good), but the user has specified a weird port
-            return False, T("Could not connect to %s on port %s. Use the default usenet settings, with SSL") % (
+            # it's a newsserver (good), on port 563, but the user has specified a weird port
+            return False, T(
+                "Could not connect to %s on port %s. Use the default usenet settings: port 563 and SSL turned on"
+            ) % (
                 host,
                 port,
             )
         elif port119working and not port563working and port not in [119, 563]:
-            # it's a newsserver (good), but the user has specified a weird port
-            return False, T("Could not connect to %s on port %s. Use the default usenet settings, without SSL") % (
+            # it's a newsserver (good), only on port 119, but the user has specified a weird port
+            return False, T(
+                "Could not connect to %s on port %s. Use the default usenet settings: port 119 and SSL turned off"
+            ) % (
                 host,
                 port,
             )
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1392,7 +1392,7 @@ def test_nntp_server_dict(kwargs: Dict[str, Union[str, List[str]]]) -> Tuple[boo
     if nw.status_code:
         if nw.status_code == 480:
             return_status = (False, T("Server requires username and password."))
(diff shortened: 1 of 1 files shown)


Safihre:
You made black a bit too happy, please run it again with --line-length=120

sanderjo:
Yeah, fresh OS & Pycharm install. Pycharm now has `--line-length=120`, was still overacting, but correct now?

image

Safihre:
Correct now 👍

sabnzbd/api.py
```diff
@@ -1317,15 +1317,39 @@ def test_nntp_server_dict(kwargs: Dict[str, Union[str, List[str]]]) -> Tuple[boo
...
+
+        test_server.timeout = DEF_NETWORKING_TEST_TIMEOUT  # force a short timeout
+        # let's try well-known ports: HTTP and NTTP(S)
         test_server.port = 80
-        test_server.timeout = DEF_NETWORKING_TEST_TIMEOUT
         test_server.request_addrinfo_blocking()
-        if test_server.addrinfo:
+        port80working = bool(test_server.addrinfo)
+
+        test_server.port = 119  # NNTP
```
Safihre, line 1331: We should only try 119 if the port we originally tested wasn't 119. Otherwise we are testing twice. Yes, I know the HappyEyeBalls has caching, so could also just note that in a comment.
Same for 563.

sabnzbd/api.py
```diff
@@ -1317,15 +1317,39 @@ def test_nntp_server_dict(kwargs: Dict[str, Union[str, List[str]]]) -> Tuple[boo
...
+
+        if (not port119working) and (not port563working) and port80working:
+            # That's a webserver, not a newsserver!
             return False, T(
                 "Could not connect to %s on port %s. It appears that %s operates as a web server (port 80), "
                 "possibly an indexer, not a usenet server. You have to fill a usenet server."
             ) % (host, port, host)
+        elif (port119working or port563working) and port not in [119, 563]:
+            # it's a newsserver (good), but the user has specified a weird port
+            return False, T("Could not connect to %s on port %s. Use the default ports 119 (NNTP) or 563 (NNTPS) ") % (
```
Safihre, line 1347: Maybe use "secure" insteds of nntps? And leave out the nntp in general, very few regular users will know those terms.

sabnzbd/api.py
sanderjo: I thought this simple logic was better ... but I'll change.

sabnzbd/api.py
```diff
@@ -1317,15 +1317,57 @@ def test_nntp_server_dict(kwargs: Dict[str, Union[str, List[str]]]) -> Tuple[boo
...
+
+        if (not port119working) and (not port563working) and port80working:
+            # That's a webserver, not a newsserver!
             return False, T(
                 "Could not connect to %s on port %s. It appears that %s operates as a web server (port 80), "
                 "possibly an indexer, not a usenet server. You have to fill a usenet server."
             ) % (host, port, host)
+        elif port563working and port not in [119, 563]:
+            # it's a newsserver (good), but the user has specified a weird port
+            return False, T("Could not connect to %s on port %s. Use the default usenet settings, with SSL") % (
```
Safihre, line 1359: Should we not add the port number in the message? Since they added  custom port, they might not know 119/563 is the default?

sanderjo: I thought then turning "SSL" on and off would fill it out automatically ... but not so.

So, yes, I'll add

PR #2956, opened 2024-11-01
fix(fs): only apply custom_permissions to new dirs

OffBy0x01:
I was running into an issue with path creation failing on some systems but not others. No matter what I tried in the config I could not get it to behave. 

It seems that the root of the problem stems from the `set_permissions` function being applied to all parts of the path, differing from the behaviour of `mkdir -p`. This is problematic for mounts & symlinks where existing `path_part` permissions should not be modified as they may be shared by other services/apps.

My proposed change works similarly to the existing code but only applies the `custom_permissions` to the final directory - parents are created with default permissions. 

I've not seen any repeats of the issue I was facing since (locally) deploying this change in august. I intended to raise it here but forgot to do so 😅


Safihre:
I don't see why the current code doesn't work?
It skips the existing folders and only creates new ones and only applies permissions to those new parts of the path. 

I don't want to use pathlib, since we don't use it anywhere.

OffBy0x01:
Apologies, I had misremembered some of the details. I went back and had a look over my notes.

Starting from root doesn't always work as you may not have permissions to operate on (for example) `/mnt`. Under the current model, it won't know the difference between a permissions error vs the file not existing. 

If I configure `/mnt/data/publications` it'll start from `/`, then check if `/mnt` exists (no permissions so it fails -> assumes this means it doesn't exist) & tries to create, which also fails. 

I think it might be possible to replace pathlib with pure os.mkdir provided we can supply an absolute path.

OffBy0x01:
Why? Because it worked prior to using sab. 

Given that making this change resolved it, I'm inclined to believe I was correct about it being a sab issue even if my understanding of why it was occurring was slightly off 😅

I'll switch back to the official release to get a stack trace of the issue. 

In short I'm using the docker image with a (docker) volume mounted NFS share.

Safihre:
> In order to work with /mnt/data/publications, a minimum level of access to all path elements is required. That's not a sabnzbd thing, it's just how unix works. If you lock the user running sabnzbd completely out of /mnt, anything underneath that becomes inaccessible - including any files and directories that already exist and have permissions that do allow access.

I always assumed this, but is it really true? Reading online I bit I find things like https://unix.stackexchange.com/a/13891

>  current directory is /foo/bar, you can access files in /foo/bar through a relative path but not through an absolute path.

Safihre:
@jcfp Hoping you had time to look at my question above? :)

Safihre:
Closing this due to lack of response from @OffBy0x01.

PR #3014, opened 2025-01-11
Directly decompress gzip responses

opello:
The URLGrabber sets Accept-encoding: gzip on requests, so a server may respond with gzipped compressed data.  That data is tested without having been decompressed which then can result in unnecessary errors.

I tested a few ways of resolving the problem, the quick-and-dirty first pass was:

```py
if b"<nzb" in data and sabnzbd.filesystem.get_ext(filename) != ".nzb":
    filename += ".nzb"
elif b"\x1f\x8b" == data[:2] and sabnzbd.filesystem.get_ext(filename) != ".nzb.gz":
    filename += ".nzb.gz"
```

which did solve the problem, but it seemed like actually handling the gzip response body was the more correct solution after reading through [this StackOverflow question](https://stackoverflow.com/questions/3947120/does-python-urllib2-automatically-uncompress-gzip-data-fetched-from-webpage).

I'm not sure I followed all of the applicable coding styles or if this is the way the project would care to handle this detail.  I also saw 15ad60f8a99e5f6ad0f70b8432b646a926dcc1c5 which is where some of the implementation guidance originated, but also presents some doubt about how you would like to address this.


opello:
There was a decent amount of discussion on IRC/Discord.

There's risk associated with making a change to accommodate one site.  The only way to know if that site is misbehaving is to wait and hear back from the site's admins about the intended (and implemented) behavior.

The current behavior is that when the client presents the Accept-Encoding: gzip header, a gzipped xml body is returned.  And when the client does not present the header, a plain xml body is returned.

My understanding is informed by [this Python issue discussion](https://bugs.python.org/issue1508475) which basically says when a urllib2 consumer adds the Accept-Encoding: gzip (or deflate, br, zstd, whatever) header they take on the responsibility of inflating it before further handling of that response body.  It seems that prior to 15ad60f8a99e5f6ad0f70b8432b646a926dcc1c5, the work of handling compressed responses was divided between urlgrabber.py and nzbparser.py and afterwards it was contained within nzbparser.py.  However, urlgrabber.py still interacts with the response body (i.e. `if b"<nzb" in data`) and when that is compressed the chance to reject a failed request (say one that returns an html error page, or a login page) is lost.  This PR re-adds the decompression step, when the response indicates it with the Content-Encoding: gzip header, so that any subsequent handling of the response body makes sense.

The site also doesn't present a Content-Disposition header with the filename, shown above.  The currently released urlgrabber.py ultimately fails `if sabnzbd.filesystem.get_ext(filename)` and logs an error: `Unknown filetype when fetching NZB, retry after 30s <url>`.  The reason there's no extension though is because earlier the `if b"<nzb" in data` failed because it was still compressed.  Otherwise it would have been added:

```py
if b"<nzb" in data and sabnzbd.filesystem.get_ext(filename) != ".nzb":
    filename += ".nzb"
```

sabnzbd/urlgrabber.py
```diff
@@ -231,6 +236,12 @@ def run(self):
                         continue
                 fetch_request.close()
 
+                if gzipped:
+                    try:
+                        data = gzip.GzipFile(fileobj=io.BytesIO(data)).read()
```
Safihre, line 241: Why not use `gzip.decompress`?
https://docs.python.org/3/library/gzip.html#gzip.decompress

opello:
Once Content-Disposition is added (assuming it passes filename.nzb) I expect the data will be handled without this change. However, I don't think that means this change is superfluous, because even though that response will be described as filename.nzb, it will come as a gzipped response body (assuming no other changes) and nzbparser.py will handle it with this:

```py
if check_bytes == b"\x1f\x8b":
    # gzip file or gzip in disguise
    filename = filename.replace(".nzb.gz", ".nzb")
    nzb_fp = gzip.GzipFile(path, "rb")
```

And while I expect that with both Content-Disposition and a gzipped response body the likelihood of receiving or even detecting an erroneous response in urlgrabber.py decreases to inconsequential, I still think that the gzipped response body (vs. an intentionally gzipped nzb payload) should be inflated at the point the response is received instead of as a robustness case of the file parser. If that isn't the architecturally correct answer I'd be curious as to why.

Safihre:
Agree, it's a simple change and it makese sense to do it.

Safihre:
Thanks! I didn't see that you already modified the code as requested :)

PR #3039, opened 2025-02-19
Update unrar to 7.10

thezoggy:
Update to unrar 7.10 - https://www.rarlab.com/rarnew.htm


Safihre:
They can just use the Process Priority setting for that. That's exactly why we have that on Windows.
What other options could be useful to set per user?

Safihre:
Agreed, it's only CPU priority.
Feel free to add a Special setting for this, I don't think it it should be a Switch setting (at least at the moment).

PR #3041, opened 2025-02-22
Fix typos

luzpaz:
Found via `codespell -q 3 -S "*.po,*.pot,*.min.js" -L allo,collet,fo,parm,ciph,datas,parms,reenabled,re-use,re-used,re-using,sav,tage`

### DIFF
--- a/interfaces/Config/templates/config_server.tmpl
+++ b/interfaces/Config/templates/config_server.tmpl
@@ -628,7 +628,7 @@
                             })
                         }
                         if(event.data === 'hide_server') {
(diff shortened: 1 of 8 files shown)


Safihre:
Could you revert the changes to skintext? Those are the keys to the translation files.
If those change, all translations are invalidated.
We fix spelling errors by adding them to the English "translation".

luzpaz:
Sure, the issue is I thnk Github is throttling my connection ATM and so I can't push from my local branch. I'll trying to use the WebUI and see if I can do it.

luzpaz:
Done.

PR #3059, opened 2025-03-31
Fix Platform Detection on Linux

mreid-tt:
**Description:**

This patch improves platform detection on Linux by directly attempting to run `systemd-detect-virt` while suppressing error output. This approach enhances robustness by avoiding unnecessary checks and handling potential errors gracefully.

**Changes:**

- Directly executes `systemd-detect-virt` with error suppression using `stderr=subprocess.DEVNULL`.
- Ensures compatibility with environments where the command might not be present.

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -818,11 +818,9 @@ def get_platform_description() -> str:
         else:
             # Check for other forms of virtualization
             try:
-                # Check if the command exists by searching in the PATH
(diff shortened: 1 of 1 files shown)


mreid-tt:
Thanks, @jcfp, for the valuable feedback! I've implemented the simplified change as suggested. Let me know your thoughts on the updated PR.

Safihre:
Awesome!

PR #3061, opened 2025-04-05
fix root dir check in make_mo

jcfp:
looking for the "builder" dir causes make_mo to fail when run from the root of the -src tarball. This change makes it look for "po" instead, which is present in releases and also the relative path make_mo actually wants.


Safihre:
Thanks!

PR #3062, opened 2025-04-09
stop catching BaseException

jcfp:
Any bare `except:` also catches BaseException (KeyboardInterrupt and SystemExit among others, see [exception hierarchy](https://docs.python.org/3/library/exceptions.html#exception-hierarchy) in the python documentation) which is almost always unintended and undesirable.

This PR changes all bare excepts to only catch regular exceptions.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -311,7 +311,7 @@ def identify_web_template(key, defweb, wdir):
     if wdir is None:
         try:
             wdir = fix_webname(key())
-        except:
+        except Exception:
             wdir = ""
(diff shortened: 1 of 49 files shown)


sanderjo:
> also catches BaseException (KeyboardInterrupt

Does this solve that sometimes CTRL-C doesn't work for SABnzbd? In such cases, pressing CTRL-C results in a message on console about that, but SABnzbd does not stop at all.

If so: Great!

Safihre:
Thanks!

Issue #3069, opened 2025-04-17
pushover + prowl priority values not loading correctly


thezoggy:
### SABnzbd version

4.5.1

### Operating system

unraid

### Using Docker image

binhex

### Description

reported by user on discord and I could verify myself as well. you can set pushover priority for events and save..which does write the correct value. but on reloading the page it defaults to off rather than selecting the actual value

![Image](https://github.com/user-attachments/assets/6a0606a0-d8b1-48be-8791-ea62d9adcd40)

saving, can see `pushover_prio_startup` is set to 1 now
```
[pushover]
pushover_token = 11111
pushover_userkey = 54321
pushover_device = 12345
pushover_emergency_expire = 3600
pushover_emergency_retry = 60
pushover_enable = 1
pushover_cats = *,
pushover_prio_startup = 1
pushover_prio_download = -3
pushover_prio_pause_resume = -3
pushover_prio_pp = -3
pushover_prio_complete = -3
pushover_prio_failed = -3
pushover_prio_disk_full = -3
pushover_prio_new_login = -3
pushover_prio_warning = -3
pushover_prio_error = -3
pushover_prio_queue_done = -3
pushover_prio_other = -3
```

refresh page, now back to Off

![Image](https://github.com/user-attachments/assets/0cd8d3c8-1951-4a8b-b838-f1e1f2c54887)

PR #3070, opened 2025-04-17
fixup pushover/prowl pri settings

thezoggy:
as values returned are int not strings now, this fixes up config page on selecting them.
closes https://github.com/sabnzbd/sabnzbd/issues/3069


Safihre:
Good catch

Issue #1633, opened 2020-10-09
Files double unpacked due to different UTF-8 normalizations


Safihre:
For a few days the tests on Travis have been failing for macOS, for some reason the `test_download_unicode_made_on_windows` was unpacking the resulting file twice.
After much debugging I found out that the string `frènch_german_demö` that is printed in the logs is actually presented in 2 different ways. This caused the unpacker not to detect a set was already unpacked, basically because:
```python
>>> "frènch_german_demö" == "frènch_german_demö"
False
```
The reason is that they are obtained from 2 different sources:
Output of `os.listdir`:
```
b'fre\xcc\x80nch_german_demo\xcc\x88'
```
Output of the par2 files:
```
b'fr\xc3\xa8nch_german_dem\xc3\xb6'
``` 
https://stackoverflow.com/a/26733055

Now just need to find a way to fix this..

puzzledsab:
I'm not sure where to apply it so I haven't tested but maybe this will help:
```
#!/usr/bin/python

import unicodedata as ud
str1 = b'fre\xcc\x80nch_german_demo\xcc\x88'
str2 = b'fr\xc3\xa8nch_german_dem\xc3\xb6'

str1u = str1.decode("utf-8")
str2u = str2.decode("utf-8")

print(str1 == str2)
print(str1u == str2u)
print(ud.normalize('NFC',str1u) == str2u)
print(str1u == ud.normalize('NFC',str2u))
print(ud.normalize('NFC',str1u) == ud.normalize('NFC',str2u))
```

False
False
True
False
True

Safihre:
I tried that indeed: I made a unicode-friendly `os.listdir` wrapper, but didn't work.

sanderjo:
And how about this: hard-encode to pure ASCII?

```
>>> ud.normalize('NFKD', str1u).encode('ascii','ignore')
b'french_german_demo'

>>> ud.normalize('NFKD', str2u).encode('ascii','ignore')
b'french_german_demo'


>>> ud.normalize('NFKD', str1u).encode('ascii','ignore') == ud.normalize('NFKD', str2u).encode('ascii','ignore')
True
```

Safihre:
@sanderjo Indeed that fails for the Chinese-download test.
So since this problem is so rare, I haven't give it much priority. The result is just a double unpack, which isn't that bad.

[...]

PR #3074, opened 2025-04-18
Fix #2947: Handle surrogate characters in history paths

### LINKED ISSUE #2947: Adding history entry fails if `path` contains unicode surrogate code points
### SABnzbd version

4.3.3

### Operating system

Arch Linux

### Using Docker image

None

### Description

If the `path` returned by `one_file_or_folder()` contains unicode surrogate code points, adding the history entry will fail on download completion with `SQL Command Failed`. It looks like utf-8 encoding is performed on the string query parameters, which fails due to the special character:

https://github.com/sabnzbd/sabnzbd/blob/fb0ef21768a0ac07bec0520e4840dbf2b6d1ddc2/sabnzbd/postproc.py#L669

(`workdir_complete` is later used as `path` in the SQL query)

```
2024-10-15 14:40:20,591::INFO::[notifier:157] Sending notification: Error - SQL Command Failed, see log (type=error, job_cat=None)
2024-10-15 14:40:20,591::ERROR::[database:154] SQL Command Failed, see log
2024-10-15 14:40:20,591::INFO::[database:155] SQL: INSERT INTO history (completed, name, nzb_name, category, pp, script, report,
            url, status, nzo_id, storage, path, script_log, script_line, download_time, postproc_time, stage_log,
            downloaded, fail_message, url_info, bytes, duplicate_key, md5sum, password)
            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
2024-10-15 14:40:20,592::INFO::[database:156] Arguments: (1729003220, 'test', 'test.nzb', '*', 'D', 'Default', '', None, 'Completed', 'SABnzbd_nzo_mfg7581n', '/mnt/test.1/weird.characters.\udc96.txt', '/var/lib/sabnzbd/Downloads/incomplete/test.1', '', '', 0, 0, 'Servers:::server=288 B\r\nDownload:::Downloaded in 0 seconds at an average of 918 B/s<br/>Age: 2m\r\nSource:::test.nzb\r\nRepair:::[test] No par2 sets;[test] Trying RAR-based verification;[test] RAR files verified successfully\r\nUnpack:::[test] Direct Unpack - Unpacked 1 files/folders in 0 seconds', 230, '', '', 230, None, '4da44f19d4f2410a09356163191507b3', None)
2024-10-15 14:40:20,592::INFO::[database:157] Traceback:
  File "/usr/lib/sabnzbd/sabnzbd/database.py", line 128, in execute
    self.cursor.execute(command, args)
UnicodeEncodeError: 'utf-8' codec can't encode character '\udc96' in position 50: surrogates not allowed
2024-10-15 14:40:20,592::INFO::[database:308] Added job test to history
```
(line numbers are off due to me adding debug statements everywhere)

Here, the filename contains the special character `\udc96` (as reported by `os.listdir()`). Interestingly, the filename reported by the unpacker contains `\ufffe\ue096` instead:
```
2024-10-15 14:40:20,524::INFO::[postproc:482] Unpacked files ['/mnt/_UNPACK_test.1/weird.characters.\ufffe\ue096.txt']
```

A possible solution would be to extend the functions responsible for sanitizing filenames/foldernames to strip such characters.

### Reproduction Steps:
1. Create a file with `$'\226'` in its name
2. Create a rar archive containing this file
3. Post it to usenet
4. Download it via sabnzbd

```
touch weird.characters.$'\226'.txt
rar a test weird.characters.$'\226'.txt
nyuu [...] -o test.nzb test.rar
```

ShivamB25:
**Description:**

Resolves #2947.

**Problem:**
Paths derived from `os.listdir` (via `postproc.one_file_or_folder`) can contain surrogate escape characters if filenames have invalid UTF-8 byte sequences. When these paths are inserted into the history database (`database.add_history_db`), it causes a `UnicodeEncodeError` as the SQLite driver tries to encode them using UTF-8, which disallows surrogates.

**Fix:**
Sanitize the `workdir_complete` and `nzo.download_path` strings in `database.build_history_info` using `.encode('utf-8', errors='replace').decode('utf-8')`. This replaces invalid characters before the path is passed to the database driver, preventing the encoding error.

**Considered:**
Applying the fix earlier in `postproc.py`, but decided sanitizing just before the database insertion in `database.py` was more targeted to the error source.


jkhsjdhjs:
Can confirm that with this fix the history entry can be added successfully and doesn't error anymore. However, I'm also not sure if this is the correct way to fix it. However, I also don't think it's an issue with the setup since the SQLite driver disallows these unicode surrogate characters, but the filesystem doesn't. Did anyone else ever try to reproduce the issue?

jkhsjdhjs:
With the fix, the downloaded path is `test/weird.characters.?.txt`. Without the fix the history entry cannot be added, so I don't know the path.
The path on the filesystem is `test/weird.characters.$'\226'.txt` in both cases. So yes, with the fix the path in the database doesn't match the path on the filesystem anymore, but the filesystem path cannot be stored in the database. So I think the correct fix would be to sanitize the filenames on the filesystem, and add the history entry afterwards.

jkhsjdhjs:
I think the file on the disk is doing unicode just fine, it's just that the downloaded file really contains this special unicode character. As I mentioned in the issue description, it's very easy to create a file containing such characters and download it again. Any archive may contain such files, and sabnzbd should be able to handle it. The problem is that SQLite doesn't allow such characters.

Safihre:
It's not the Indexer, but the OS that serves them in the listdir. See #1633.

ShivamB25:
i feel like this would be good. but i do hope external sonarr. n stuff doens't breaks the other integrations (help needed)

jkhsjdhjs:
Yep, the indexers themselves mostly don't support utf8, so the nzb filenames aren't a problem. The RAR archives, however, can contain files with these characters. After extraction, sabnzbd has a look at the extracted files via `one_file_or_folder` and attempts to store the result in the history db, which is where the problem occurs. It rarely happens, I have only seen a few examples so far.

Issue #2590, opened 2023-06-12
quota/left_quota missing suffix


thezoggy:
user on discord called out inconsistent on how sab reports sizing.. sometimes single letter suffix sometimes two letter, example:
```
"cache_size":"30.8 MB"
"left_quota":"-14.6 G"
"cache_size":"30.8 MB"
```

I see there are various ones where we do single letter or multiple letter..  usually if we want to show Bytes as lowest value.

for cache size we pass a post suffix to append:
```
    header["cache_size"] = to_units(anfo.cache_size, "B")
```

for quota/left_quota we do not...
```
    header["quota"] = to_units(sabnzbd.BPSMeter.quota)
    header["left_quota"] = to_units(sabnzbd.BPSMeter.left)
```

Testing on latest dev which I have no quota, I do see queue result shows:
```
    "quota": "0 ",
    "have_quota": false,
    "left_quota": "0 ",
```

Looked back at git history, looks like the quota stuff has always been a single letter prefix.. so yeah we should prob fix that up as "0 B" would be better for apps to handle than the lack of suffix.

Also I see our speed one is also missing suffix...
```
    "speed": "0 ",
```
But uncertain if that would break things if we 'fixed' that one.

Safihre:
Yeah it's true, it's a mess. Relic from the times when we made our interface fully in Cheetah..
We should just use `int` everywhere and let the browser/client figure out the notation. But that just breaks stuff for others.

PR #3076, opened 2025-04-24
Update to_units

mwmiller:
This is mostly intended to address #2590.

It also makes some minor adjustments to `to_units` to improve its performance, correctness and maintainability.  Of course, some of that may be both subjective and wrong.

My local testing shows some broken tests, but they also appear broken on my machine on the up-to-date `develop` branch.  If I have, in fact, broken anything with these adjustments, I am happy to iterate in this PR.

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1798,9 +1798,9 @@ def build_header(webdir: str = "", for_template: bool = True, trans_functions: b
     header["have_warnings"] = str(sabnzbd.GUIHANDLER.count())
     header["finishaction"] = sabnzbd.QUEUECOMPLETE
 
-    header["quota"] = to_units(sabnzbd.BPSMeter.quota)
+    header["quota"] = to_units(sabnzbd.BPSMeter.quota, "B")
     header["have_quota"] = bool(sabnzbd.BPSMeter.quota > 0.0)
-    header["left_quota"] = to_units(sabnzbd.BPSMeter.left)
+    header["left_quota"] = to_units(sabnzbd.BPSMeter.left, "B")
 
     anfo = sabnzbd.ArticleCache.cache_info()
     header["cache_art"] = str(anfo.article_sum)
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -38,6 +38,7 @@ import ctypes
 import html
 import ipaddress
 import socks
+import math
 from threading import Thread
 from collections.abc import Iterable
 from typing import Union, Tuple, Any, AnyStr, Optional, List, Dict, Collection
@@ -604,15 +605,18 @@ def to_units(val: Union[int, float], postfix="") -> str:
 
     if val < 0:
         sign = "-"
+        val = abs(val)
     else:
         sign = ""
 
     # Determine what form we are at
-    val = abs(val)
-    n = 0
-    while (val > 1023) and (n < 5):
-        val = val / 1024
-        n = n + 1
+    # Anything beyond petabytes is outside of scope
+    if val == 0:
+        n = 0
+    else:
+        n = min(5, math.trunc(math.log2(val) / 10))
+
+    val = val / 2 ** (10 * n)
 
     if n > 1:
         decimals = 1
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -618,12 +618,19 @@ def to_units(val: Union[int, float], postfix="") -> str:
 
     val = val / 2 ** (10 * n)
 
+    # Showing the single decimal per doc string
     if n > 1:
         decimals = 1
     else:
         decimals = 0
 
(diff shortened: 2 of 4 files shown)


sabnzbd/api.py
```diff
@@ -1798,9 +1798,9 @@ def build_header(webdir: str = "", for_template: bool = True, trans_functions: b
     header["have_warnings"] = str(sabnzbd.GUIHANDLER.count())
     header["finishaction"] = sabnzbd.QUEUECOMPLETE
 
-    header["quota"] = to_units(sabnzbd.BPSMeter.quota)
+    header["quota"] = to_units(sabnzbd.BPSMeter.quota, "B")
     header["have_quota"] = bool(sabnzbd.BPSMeter.quota > 0.0)
-    header["left_quota"] = to_units(sabnzbd.BPSMeter.left)
```
Safihre, line 1803: It isn't that simple, otherwise I would have fixed it long ago.

You also need to change it in the front end where we use this value to give a visual indicator if the quote was exceeded.

sabnzbd/misc.py
```diff
@@ -604,22 +605,32 @@ def to_units(val: Union[int, float], postfix="") -> str:
...
-    while (val > 1023) and (n < 5):
-        val = val / 1024
-        n = n + 1
+    # Anything beyond petabytes is outside of scope
+    if val == 0:
+        n = 0
+    else:
+        n = min(5, math.trunc(math.log2(val) / 10))
+
+    val = val / 2 ** (10 * n)
```
Safihre, line 619: What's the formula you're using here?

And what speedup does it achieve? 
The previous code was really Pythonic code and easy to read and understand. It's a mantra that I would like to keep unless major speedups can be achieved.

sabnzbd/api.py
mwmiller: > You also need to change it in the front end

Apologies. I tried to follow the discussion on the issue because it did indeed seem too simple.  I apparently did not follow it as well as I thought.  I took away the fact that it produces strings instead of numbers was a hassle, but meant that its output would be usable.

I will look into this modulo the other comments which have come in.

sabnzbd/misc.py
mwmiller: Thanks for the assist, @jcfp. You've described well what is happening.  It seems it's not as obvious as I thought, so I will add some commentary.

> the function is ever exposed to small enough numbers

I did consider this, but thought it out of the domain since these are seemingly always bytes.  (Thus the `1024` instead of `1000` for more SI-like units.) That being said, it does also sometimes operate on rates which might go very low.  Also, the fix is simple:  I am already mapping `0` to `0` above; I'll simply extend that to `1024`.

mwmiller:
> Problem with us changing our API responses affect not just sab but 3rd party apps that use said info. So some of the crust has stuck around to prevent breakage. One of those things we also need better API test sets that would come with something like pydantic

Obviously, I'm not sure how to address this.  If this whole exercise is more trouble than it's worth for a slightly more meaningful quota display, then feel free to bin this PR.  I was certainly looking to help, not hurt.

mwmiller:
> no worries, just saying that this whole stuff is just legacy stuff from a few decades ago

Completely understandable.  I apologize if my message sounded bitter in any way.  I truly meant that you can bin this PR with no hard feelings.  It was an attempt to improve around the edges where I might not break things.  And somehow... 😀

In any case, based on a quick `richbench` it seems like the `to_units` code is about 10% faster.  If I have some time, I might hook up `pyperf` and see if I can get a better report.   I don't have a good sense of the distribution of input values, so I'm just checking across a wide range.  Also, I see there are a lot of calls thereto, but I don't know if that's enough to offend the codebase style.

I can remove the API-changing commit if this other stuff is desirable.

Safihre:
Thanks, the comments make the code clear 👍
I don't see any Glitter changes though?

mwmiller:
>  I don't see any Glitter changes though?

Yeah, that's my bad.  The changes were breaking the CI builds.  I futzed around until I could figure out how to tell that those tests were breaking here as well.  I'm going to restart on the changes... probably after better understanding the KO templates.

Apologies on the false hope.  I expect to work out the difficulties today.

sabnzbd/api.py
mwmiller: @Safihre This commit now actually (and finally) includes the UI changes.  The tests seem to be passing.

I searched my best for places where the values are used and updated them as necessary. I've looked in the UI and not noticed any anomalies. I do not, however, doubt my human ability to have missed something.

Safihre: Nice!

Safihre (approved):
Seems good to me. 
Only during the release candidate phase we will really get feedback if anyone is upset about these breaking chances.
Right now I would assume this shouldn't be a problem!

PR #3083, opened 2025-05-07
Performance Improvement: Add Conditional Parallel Unrar Functionality to DirectUnpacker

lincmba:
## Description
This pull request introduces a conditional implementation of parallel Unrar functionality to the `DirectUnpacker` class in SABnzbd. The changes aim to optimize post-processing by enabling parallel extraction of RAR files when specific conditions (e.g., large file sizes or high volume counts) are met. This feature is configurable via new options added to the configuration system.

## Key Changes

1. Parallel Unrar Implementation:
    - Introduced multi-threaded Unrar functionality using `concurrent.futures.ThreadPoolExecutor`.

2. New Configuration Options:
    - `use_parallel_unrar`: Specifies whether to use parallel unrar or not.

3. Updated `DirectUnpacker` Class:
    - Modified the add method to conditionally initiate parallel Unrar functionality based on the thresholds.
    - Fallback to single-threaded extraction if conditions are not met.

4. Unit Tests and Benchmarking:
    -  Added tests to validate the functionality of parallel Unrar under different scenarios.
    -  Included a benchmark test to compare the performance of single-threaded and parallel extraction modes.

## Benefits
- Performance Optimization: Improves post-processing speed for large or complex RAR sets.
- Flexibility: Users can configure thresholds to suit their environment.
- Backward Compatibility: Default behavior remains unchanged for users who do not meet the thresholds or configure them.

## Testing
- Verified the functionality with mock RAR sets of varying sizes and volume counts.
- Benchmark results demonstrate significant performance improvement for large RAR sets when parallel Unrar is enabled.


Safihre:
Thanks for checking this.
So when would this add extra speed?
Say we have an NZB with 2 Rar sets, each containing 10 files. We download things in sequence, so only when we finished downloading almost all raf files of set 1 we start on set 2. 
Isn't the direct unpack of set 1 basically finished by the time we get data for set 2?

We can already Direct Unpack multiple jobs at the same time.

Issue #2952, opened 2024-10-26
Age Filtering in RSS


Helipad166:
### Description

I have recently moved from NZBGet to SABnzbd. One thing I cannot figure out is how to filter RSS by age. My use case is I get my NZBs from RSS and on this particular website, the NZBs do not get generated until a few hours later, but the RSS will contain torrent links. In NZBGet, I set the age > 6h, to filter out the really new feeds. But I can't do this in SABnzbd and it ends up with errors in the download page. 

I investigated the RSS link and they can do maxage but not the other way around :( 

Is there a way around this or can age filtering be added as a feature?

sanderjo:
http://127.0.0.1:8080/config/switches/#propagation_delay

Posts will be paused until they are at least this age. Setting job priority to Force will skip the delay.

Safihre:
> http://127.0.0.1:8080/config/switches/#propagation_delay
> 
> Posts will be paused until they are at least this age. Setting job priority to Force will skip the delay.

That's not what he wants I think.

But indeed, we don't support age based filtering indeed. And interestingly, it wasn't requested so far even though it seems like a good feature.

Helipad166:
> http://127.0.0.1:8080/config/switches/#propagation_delay
> 
> Posts will be paused until they are at least this age. Setting job priority to Force will skip the delay.

No, propagation delay does not help in this case as SABnzbd will still try to retrieve NZB from an RSS link that is not available. 

The RSS link is https://feed.animetosho.org/rss2, somehow new feed do not generate NZB until a few hours later but torrent link is active. I think an easy fix (which worked on NZBget) is to put an age limit on this. 

> > http://127.0.0.1:8080/config/switches/#propagation_delay
> > Posts will be paused until they are at least this age. Setting job priority to Force will skip the delay.
> 
> That's not what he wants I think.
> 
> But indeed, we don't support age based filtering indeed. And interestingly, it wasn't requested so far even though it seems like a good feature.

If you can implement this, it would be awesome - conceptually it shouldn't be too hard to add another item in the drop to filter by age field - but then again I am not a programmer.

Safihre:
Re-opening because age filtering is a valid feature request.

PR #3089, opened 2025-05-18
feat: Add a delay for feed addition to job queue for download

bytemerger:
Provides a fix for #2952

Screenshot 2025-05-18 at 01 32 21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 *.py[cod]
 __pycache__
 /venv
-
+myenv/
 # Snapcraft
 parts/
 prime/
--- a/interfaces/Config/templates/config_rss.tmpl
+++ b/interfaces/Config/templates/config_rss.tmpl
@@ -373,6 +373,17 @@
                     <!--#end if#-->
                 </div>
             </form>
+            <form action="update_rss_feed_delay" method="post">
+                <input type="hidden" name="apikey" value="$apikey" />
+                <input type="hidden" name="feed" value="$active_feed" />
+                <div class="field-pair">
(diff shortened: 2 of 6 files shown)


bytemerger:
> why not just use switches > queue > propagation delay ?

There is a case where new feed do not generate NZB until a few hours later, #2952 this issue discusses it.
A more robust fix would be a proper filter entry with age

Safihre:
Don't we just need a new Filters "Age at least" and "Age at most"?
Then the user can use natural language like "6h" and we can parse that.
@mnightingale

mnightingale:
I think that is a reasonable place for the limits, an alternative is the per feed edit screen where name and URL are as.

I don't use those filters a huge amount, do the rules follow the order so can you do multiple rules i.e. 'RequiresCat' and 'Age at least' or is it one or the other?

I'm not sure if it's very clear that 'age' refers to rss entry age rather than article age, but I don't see how that could be improved.

Also take care of the `remove_obsolete` call in rss.py I haven't checked exactly what it does, but it may already enforce a max age of 3 days.

---

I'm not really sure of the need for any of this anyway since we now check the `application/x-nzb` mimetype, seems like an XY problem.
Perhaps the max age does make sense when adding a new feed for the first time; but that may be what the existing 3 day limit is for.

Safihre:
Oh I didn't know that... Expected index and post to be the same!

Safihre:
What is your opinion @thezoggy what to do with this? Add the feature, or not? And which date to use?

Safihre:
Closing due to inactivity

Issue #3073, opened 2025-04-18
Tab to Rename Next/Previous


ittroll:
### Description

When renaming an entry in the queue, it would be great if tab and shift-tab could be used to quickly move between the other entries in rename mode. Similar to the behaviour in Windows File Explorer.

Safihre:
Resolved in #3091

PR #3091, opened 2025-05-20
Add tab and shift+tab navigation between rename fields in queue

ArunRawat404:
### Description
This PR implements the feature requested in issue [#3073](https://github.com/sabnzbd/sabnzbd/issues/3073), allowing users to navigate between rename fields in the queue using the Tab and Shift+Tab keys — similar to the behavior in Windows File Explorer.

#### Changes Made

Enhanced the `editName` function to:
  - Detect Tab and Shift+Tab key presses.
  - Move focus to the next or previous renamable queue item.
  - Automatically trigger rename mode for the target row and focus the input


Safihre:
Nice. Thanks!!

Issue #3016, opened 2025-01-13
Support extracting .tar


benhaney:
### Description

This has been brought up a couple times before (#1967, #2644) and not implemented because it didn't seem like a common enough use case. I recently started using Lidarr, and a pretty decent portion of albums I've found have been in a .tar (just tar, no compression). Like, >10% of them. It's getting really annoying to manually untar all these since sabnzbd won't, so I'm throwing my hat in the ring asking for tar support. A previous issue brought up that "Linux people" might not like their tarballs being extracted, so maybe a switch is warranted if that's a thing? I'd be cool with it defaulting to off since I don't know how often stuff is shipped as a tarball intended to not be extracted, but there's definitely a legitimate and seemingly pretty common use of tar as an intermediary archive that should be recursively unpacked.

Safihre:
Sure, it's the third time it's requested so now it's often requested 😄

We could just 7zip or unrar to do the work, almost no extra code would be needed.

So these tar files are inside Rar/zip files? Or they are the main files of the NZB's?
You can maybe send me an example nzb at safihre@sabnzbd.org?

Safihre:
There is a bigger risk with tar though, since it can have hard-links and `..` etc in the files.
So it could be a security issue.

benhaney:
> these tar files are inside Rar/zip files?

No, so far are the examples I've found are just a bare .tar next to a .nfo. I suppose they _could_ be nested in other archives though. Are you asking because supporting them as a main file is different than supporting them during recursive unpacking?

> You can maybe send me an example nzb

Sure. Sent a couple.

> There is a bigger risk with tar though, since it can have hard-links and `..` etc in the files

I don't know about 7zip and unrar, but every implementation of the `tar` utility I've ever seen will force all extracted files (including absolute paths and paths with `..`) to stay within the target extraction directory without an explicit flag to allow them to creep into parent dirs. Do 7zip and unrar not do that?
It's also not immediately obvious to me why preserving hard links is a security concern. As far as I'm aware tar can only represent links to other files in the same archive, not arbitrary files on the filesystem.

davidnewhall:
You could always try Unpackerr. Works great with Lidarr.

ggibby:
So far today 100% of my 25 downloads were .tar files.

thezoggy:
> So far today 100% of my 25 downloads were .tar files.

Where they all from the same poster / on same indexer?

ggibby:
[...]

Issue #2644, opened 2023-08-05
Extracting tar.*z archives resulting in tarballs not usable for other tools like sonarr


theniwo:
I just experienced this behaviour and It would be better imho when compressed tarballs would be uncomressed using tar and not gunzip

sanderjo:
Can you post the relevant part from sabnzbd.log? NZB name, and gunzip command

theniwo:
Sure. It looks like I misinterpreted the outcome without looking at the logs first. My bad. It actually was a rar file, but it contained a tarball

```
2023-08-04 23:53:08,661::INFO::[newsunpack:293] Unrar starting on /localdata/DATA/Downloads/usenet/incomplete/For.All.Mankind.S02.German.DL.720p.WEB.h264-WvF
2023-08-04 23:53:08,662::INFO::[newsunpack:559] Extracting rarfile /localdata/DATA/Downloads/usenet/incomplete/For.All.Mankind.S02.German.DL.720p.WEB.h264-WvF/87da74e9c8bf0d3e7849363f.part01.rar (belonging to 87da74e9c8bf0d3e7849363f) to /downloads/_UNPACK_For All Mankind S02 German DL 720p WEB h264-WvF
2023-08-04 23:53:08,664::INFO::[misc:1175] [N/A] Running external command: ['/bin/nice', '-n10', '/bin/ionice', '-c2', '-n4', '/usr/local/bin/unrar', 'x', '-idp', '-scf', '-o-', '-or', '-ai', '-puemd3cwoaxeiybkt.onion-snogrk5f879hswy5krz154xAhvcln7vd-distribution-not-allowed', '/localdata/DATA/Downloads/usenet/incomplete/For.All.Mankind.S02.German.DL.720p.WEB.h264-WvF/87da74e9c8bf0d3e7849363f.part01.rar', '/downloads/_UNPACK_For All Mankind S02 German DL 720p WEB h264-WvF/']
2023-08-05 00:16:38,524::INFO::[newsunpack:297] Unrar finished on /localdata/DATA/Downloads/usenet/incomplete/For.All.Mankind.S02.German.DL.720p.WEB.h264-WvF
2023-08-05 00:16:38,529::INFO::[postproc:437] Unpacked files ['/downloads/_UNPACK_For All Mankind S02 German DL 720p WEB h264-WvF/For.All.Mankind.S02.German.DL.720p.WEB.h264-WvF.tar']
[...]

Issue #1967, opened 2021-10-20
Decompress other formats supported by `unrar`


Safihre:
https://forums.sabnzbd.org/viewtopic.php?f=4&t=25602
Should we also reclusively extract files like `tar`, `tar.gz`  and maybe others?

Safihre:
Good question, this is the first time I saw the request in the forums.. So guess not?
If you both agree that we shouldn't do it, let's close this then.

PR #3092, opened 2025-05-20
Add TarFile class for secure .tar inspection and extraction

G4te-Keep3r:
Implements a TarFile class for safely inspecting and extracting .tar files, including protection against tarbombs, absolute paths, path traversal, and symbolic/hard links.

Includes:
- is_tar() check to validate archive format
- is_tar_safe() with strict path and link checks
- namelist() and open() methods for safe usage
- Unit test coverage for valid and malicious tar scenarios

Note:
This implementation has passed synthetic tests but has not been tested with real-world SABnzbd download jobs due to local environment constraints. Future validation is required.

Related: #3016
Also references: #1967, #2644 (for historical context)


G4te-Keep3r:
my interpretation from issue comments was that more direct analysis was needed first for tars and not to just hand it over. I didn't see a direct way to check contents and pass it off to user program of choice (should be a user option).

Might have an idea now of a direction to try. Or even maybe having all archive operations handled in 1 place. I need to dig deeper and get a working instance running local to test (and local servers to provide various archives).

sabnzbd/newsunpack.py
```diff
@@ -1918,3 +1919,92 @@ def open(self, name: str) -> BinaryIO:
...
+
+
+    def namelist(self) -> List[str]:
+        """Return list of names in tar"""
+        names = []
+        try:
+            with tarfile.open(self.path, 'r:*') as tar:
+                for member in tar.getmembers():
+                    # Avoid files with absolute paths or .. (path traversal)
+                    if member.name.startswith("/") or ".." in member.name:
```
Safihre, line 1983: Did you see the validations that Python itself implements for safe tar's?
https://github.com/python/cpython/blob/main/Lib/tarfile.py#L769
On a quick glance I can see they also check `os.sep` and they use `os.path.isabs` checks.
Maybe more to explore?

sabnzbd/newsunpack.py
```diff
@@ -1918,3 +1919,92 @@ def open(self, name: str) -> BinaryIO:
...
+
+            return names
+        except tarfile.TarError:
+            # Not a valid tar file
+            return None
+
+    def open(self, name: str) -> BinaryIO:
+        """Read named file from tar and return data"""
+        data = None
+        with tarfile.open("example.tar", "r|") as tar:
```
Safihre, line 2002: `example.tar` 🙂

PR #3093, opened 2025-05-20
Filter rss enclosures by application/x-nzb mimetype

### LINKED ISSUE #2952: Age Filtering in RSS

mnightingale:
I think this resolves #2952 although the PR does not provide the functionality requested, I think this is a better solution which achieves the same outcome.

Currently RSS processing takes the first link from the entry, but if there are any [enclosure elements](https://www.rssboard.org/rss-specification#ltenclosuregtSubelementOfLtitemgt) it takes the first url.

However, it doesn’t handle that an item may have multiple enclosures, and type is a required field with the file’s MIME type so the first enclosure may not be the nzb.

Using the RSS example given https://feed.animetosho.org/rss2 you can often see items have multiple enclosures with types `application/x-bittorrent` or `application/x-nzb` but the nzb is often not initially available, the current code appears to try adding the torrent.

Although type is a required field, I did allow it to be missing, maybe there are bad implementations out there. But happy to make it require the expected type if you would prefer.

The only issue I can see is it may fall through to the `Empty RSS entry found` warning more often which could be annoying, maybe add a special option to ignore it?

@bytemerger you may wish to see if this resolves your need for #3089

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -649,19 +649,26 @@ def _get_link(entry):
     """Retrieve the post link from this entry
     Returns (link, category, size)
     """
+    link = None
     size = 0
     age = datetime.datetime.now()
 
     # Try standard link and enclosures first
-    link = entry.link
-    if not link:
-        link = entry.links[0].href
     if "enclosures" in entry:
         try:
-            link = entry.enclosures[0]["href"]
-            size = int(entry.enclosures[0]["length"])
+            for enclosure in entry["enclosures"]:
+                if "type" in enclosure and enclosure["type"] != "application/x-nzb":
+                    continue
+
+                link = enclosure["href"]
+                size = int(enclosure["length"])
+                break
         except Exception:
             pass
+    else:
+        link = entry.link
+        if not link:
+            link = entry.links[0].href
 
     # GUID usually has URL to result on page
     infourl = None
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -654,7 +654,7 @@ def _get_link(entry):
     age = datetime.datetime.now()
 
     # Try standard link and enclosures first
-    if "enclosures" in entry:
+    if "enclosures" in entry and entry["enclosures"]:
         try:
             for enclosure in entry["enclosures"]:
                 if "type" in enclosure and enclosure["type"] != "application/x-nzb":
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -724,7 +724,7 @@ def _get_link(entry):
 
         return link, infourl, category, size, age, season, episode
     else:
-        logging.warning(T("Empty RSS entry found (%s)"), link)
+        logging.info(T("Empty RSS entry found (%s)"), link)
         return None, None, "", 0, None, 0, 0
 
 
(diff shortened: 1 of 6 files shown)


Safihre:
Will this work? Because when no matching `enclosure` is found, it falls back to `link` which in the case of https://feed.animetosho.org/rss2 just points to the overview page on their website.
For example I just saw this one at the top of the feed.

```xml
		<item>
			<title>Kakushite! Makina-san!! S01E04 UNCENSORED English Dub 1080p WEB x264 AAC - TMD-Group [OceanVeil] (Makina-san&#039;s a Love Bot?!)</title>
			<description>
				<![CDATA[XXX]]>
			</description>
			<link>https://animetosho.org/view/kakushite-makina-san-s01e04-uncensored-english-dub-1080p.n1973921</link>
			<comments>https://animetosho.org/view/kakushite-makina-san-s01e04-uncensored-english-dub-1080p.n1973921</comments>
			<enclosure url="http://storage.animetosho.org/torrent/314b402d1bb7d7307ce172a4b7e0349e158d4901/Kakushite%21%20Makina-san%21%21%20S01E04%20UNCENSORED%20English%20Dub%201080p%20WEB%20x264%20AAC%20-%20TMD-Group%20%28OceanVeil%29.torrent" type="application/x-bittorrent" length="0"/>
			<enclosure url="http://storage.animetosho.org/nzbs/000a961d/Kakushite%21%20Makina-san%21%21%20S01E04%20UNCENSORED%20English%20Dub%201080p%20WEB%20x264%20AAC%20-%20TMD-Group%20%28OceanVeil%29.nzb" type="application/x-nzb" length="0"/>
```

mnightingale:
If the entry has any enclosure it will only use them, specifically because we don't want it to fall back to any link it can find.

If it has enclosure => only consider enclosure which might result in link = None
else => do as currently using link or links

```py
if "enclosures" in entry:
    # First enclosure with no type or application/x-nzb, else link remains None
else:
    link = entry.link
    if not link:
        link = entry.links[0].href

# link could be None here which will show a warning
```

In your example/snippet, it has enclosures so it will only consider them. If neither were present it would take the <link> element.

I tested it with a simple http server `npx http-server -p 9191`, with a modified version of animetosho' feed with a single item and adding/removing enclosure elements.

I suppose I could/should add tests, but the existing framework relies on loading them from sabnzbd.org maybe it could be made to load local test files instead.

- entry with only link element
- entry with enclosure but no `application/x-nzb`
- entry with enclosure `application/x-bittorrent` followed by`application/x-nzb`

Safihre:
Ah I see. Indeed!

mnightingale:
Thanks for the httpserver hint, I've ended up adding a handler that can be reused for serving the test files.
I suppose the sabnzbd.org files could be moved to the repo?

Any thoughts on these questions?

> Although type is a required field, I did allow it to be missing, maybe there are bad implementations out there. But happy to make it require the expected type if you would prefer.
>
>The only issue I can see is it may fall through to the Empty RSS entry found warning more often which could be annoying, maybe add a special option to ignore it?

Safihre:
Info level I would then say.

Safihre:
Still, new filters for minimal and maximum age should be nice.
I should ask Copilot 😄

Issue #2979, opened 2024-11-26
RSS "Next scan at:" time is not recalculated after manual "Read All Feeds Now" click


OneCDOnly:
### SABnzbd version

4.3.3

### Operating system

Debian 12.8

### Using Docker image

None

### Description

The following is an extract from here: https://forums.sabnzbd.org/viewtopic.php?p=132818

My RSS checking interval is 20 minutes.

I jumped to the SABnzbd RSS feed page, saw there were 2 minutes to-go until the next scan time. But, I couldn't wait, so I clicked "Read All Feeds Now". This was done and completed. I jumped back to the main queue and confirmed my items were now in the queue.

I jumped back to check the "Next scan time:" and it hadn't changed - I was guessing it should be the time 20 minutes from now. But it wasn't. Refreshed a few times and no-change.

Then the original scan time rolled around shortly afterward and (presumably) the RSS feed scan was activated again. Eventually, the scan time updated to 20 minutes from the original scan time.

I'm thinking the next scan time should always be calculated from the last scan time - whether it was an automated feed scan or a manual feed scan.

Thank you. 🤓

OneCDOnly:
Good work guys! Cheers! 🤓

PR #3102, opened 2025-06-25
Fix RSS 'Next scan at' time not updating after manual scan

mrshaun13:
## Problem
When a user manually triggers an RSS feed scan using the 'Read All Feeds Now' button, the 'Next scan at' time in the UI does not update to reflect the new scheduled time. This can cause confusion as users expect the next scan to happen after the configured RSS interval from the manual trigger time.

Original bug report: https://github.com/sabnzbd/sabnzbd/issues/2979

## Root Cause
The [force_rss()] method in scheduler.py was only scheduling a one-time scan without updating the  time. This meant that while a manual scan would run immediately, the next scheduled scan would still use the old [next_run] time.

## Solution
Modified the [force_rss()] method to update the  time to be the current time plus the configured RSS interval (in minutes). This ensures that after a manual scan, the next automatic scan will happen after the expected interval.

## Testing
1. Added a new test case [test_force_rss_updates_next_run] that verifies the [next_run] time is correctly updated when [force_rss()] is called.
2. The test mocks the necessary components to isolate the functionality:
   - Mocks the RSSReader class to track the [next_run] time
   - Mocks  to provide consistent test timing
   - Mocks the scheduler to avoid starting background threads
3. Verified that all existing tests pass with the changes.

### DIFF
--- a/sabnzbd/scheduler.py
+++ b/sabnzbd/scheduler.py
@@ -450,6 +450,9 @@ class Scheduler:
 
     def force_rss(self):
         """Add a one-time RSS scan, one second from now"""
+        # Update the next_run time to be current time + rss_rate minutes
+        # This ensures the next scheduled scan is rss_rate minutes after the manual scan
+        sabnzbd.RSSReader.next_run = time.time() + cfg.rss_rate() * 60
         self.scheduler.add_single_task(sabnzbd.RSSReader.run, "RSS", 1)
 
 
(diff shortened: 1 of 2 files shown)


Safihre:
Indeed this fixes the label.
However, does it actually move the scheduled task?
Because we have in the `scheduler.py`:
````python
            self.scheduler.add_interval_task(sabnzbd.RSSReader.run, "RSS", delay * 60, interval * 60)
```

Safihre:
Closing due to lack of response..

PR #3127, opened 2025-08-08
Use newer unrar versions in snaps

mnightingale:
The source tarball is found on https://www.rarlab.com/rar_add.htm and provides 7.13 but we'll have to manually update it.

CXXFLAGS are based on what linuxserver does.


Safihre:
I'll close this for now :)

Issue #2577, opened 2023-05-31
Files unrelated to job get removed when post-processing Cleanup List runs


albino1:
I wasn't sure the best way to title this, but essentially I noticed that if you have any files in your final download folder or any subfolder  that are on your Cleanup List then they get removed when the job finishes running, regardless of whether or not they're related to the actual job.  This might be a bug, or it might be just be a consequence of having a Cleanup List at all, but it seems to me that it shouldn't be removing files unrelated to the download.

Quick reproducible example:

* I set my Cleanup List in SABnzbd to `exe, png, nfo, nzb`
* I created a subfolder in my download directory called `MyApplication`
* Inside I placed some fake files named `fake.exe`, `fake.png`, `fake.nfo`, `fake.nzb`, `fake.ext` (.ext for control)
* I downloaded a random TV show NZB with a category set and sort turned on (this seems important, it doesn't happen with `Default` and no sorting)

Result: All of the cleanup list files are gone from the `MyApplication` subfolder.  In fact, if there's only cleanup files in the folder then the whole folder is removed.

Initial structure:

```
D:\downloads\sab\
D:\downloads\sab\MyApplication\
D:\downloads\sab\MyApplication\fake.exe
D:\downloads\sab\MyApplication\fake.png
D:\downloads\sab\MyApplication\fake.nfo
D:\downloads\sab\MyApplication\fake.nzb
D:\downloads\sab\MyApplication\fake.ext
```

After download and cleanup completes:

```
D:\downloads\sab\
D:\downloads\sab\Title.S01E01.720p.WEB-DL.AAC2.0.h264-Group.mkv
D:\downloads\sab\MyApplication\
D:\downloads\sab\MyApplication\fake.ext
```

Some logging:

```
2023-05-31 15:52:53,073::DEBUG::[filesystem:919] [sabnzbd\filesystem.py.cleanup_empty_directories] Removing dir \\?\D:\downloads\sab\Title.S01E01.720p.WEB-DL.AAC2.0.h264-Group
2023-05-31 15:52:53,074::DEBUG::[deobfuscate_filenames:88] No additional par2 files found to process
2023-05-31 15:52:53,076::INFO::[postproc:1080] Removing unwanted file \\?\D:\downloads\sab\MyApplication\fake.exe
[...]

PR #3162, opened 2025-10-10
Bugfix/cleanuplist

Safihre:
@jcfp me and my friendly AI tried to make a fix for #2577.
It made a mess of it, but after some friendly guidance it did a bit better.

It's stumbling a bit with the whole `move_to_parent_directory` function.
What do you think of the changes so far and maybe what is the best solution?


Safihre:
I had some success using AI, but on this one it really chokes.
Even with several prompting strategies, it just always goed off the rails.
Will have to fix it manually.

PR #3184, opened 2025-11-09
Optimize download performance for 10 Gbps+ connections

dimitribrooks:
This commit addresses critical bottlenecks preventing SABnzbd from achieving 10 Gbps (1250 MB/s) download speeds:

**Critical fixes:**
- Increase NNTP_BUFFER_SIZE from 800 KB to 4 MB (sabnzbd/constants.py:104)
  * 800 KB buffer fills in 0.64ms at 10 Gbps, causing excessive context switches
  * 4 MB provides ~3.2ms buffer time, reducing syscall overhead
- Increase NTTP_MAX_BUFFER_SIZE from 10 MB to 32 MB (sabnzbd/constants.py:105)
  * Allows larger bursts without buffer reallocation
- Increase MAX_ASSEMBLER_QUEUE from 12 to 100 (sabnzbd/constants.py:100)
  * Prevents article cache backpressure at high speeds
  * Original value caused throttling at ~600 MB/s

**Performance improvements:**
- Increase default receive_threads from 2 to 4 (sabnzbd/cfg.py:529)
  * Better parallelism for multi-server configurations
  * Reduces GIL contention in receive processing
- Reduce select() timeout from 1.0s to 0.1s (sabnzbd/downloader.py:657)
  * Improves article processing latency
  * Reduces idle time between download batches
- Reduce _SERVER_CHECK_DELAY from 0.5s to 0.1s (sabnzbd/downloader.py:53)
  * Faster article fetching from servers
  * Reduces pipeline stalls

**Expected improvements:**
- Eliminates ~600 MB/s bottleneck caused by assembler queue throttling
- Reduces network I/O overhead by 5x (fewer recv() calls)
- Improves multi-server throughput with better thread utilization
- Reduces latency in article processing pipeline

Tested for 10 Gbps connections with NVMe storage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Safihre:
1. The `NNTP_BUFFER_SIZE` is AI nonsense. This is not how NNTP works and what those constants do and how they affect performance
2. The Assembler queue is indeed something usefull, I will think about that.
3. Same for receive_threads, more are needed at high speed.
4. Changing select also seems a bit AI nonsense to me. The select call will always return right away if anything is available. In super-fast situations this should be always.
5. Changes to _SERVER_CHECK_DELAY only matter when switching servers if article is unavailable on 1 or the other server, or in timeout situations. Not relevant for performance.

Safihre:
In 699d97965c709896c334d93eb1594daa9cf43800 I now automatically increase `receive_threads` and `assembler_max_queue_size` in case a Bandwith Limit of 150MB/s or more is configured.
Below that value I think it will not benefit anyone.

PR #3185, opened 2025-11-09
STAT: tell on which server an article is present

sanderjo:
```
DEBUG: Article <article> is present on news.newshosting.com
DEBUG: Article <article> is present on news.eweka.nl

```
Useful to see if STAT on a certain says Present, but ARTICLE later on says otherwise.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -753,7 +753,7 @@ class Downloader(Thread):
 
             elif nw.status_code == 223:
                 article_done = True
-                logging.debug(f"Article {article.article} is present on {nw.server.host}")
+                logging.debug("Article <%s> is present on %s", article.article, nw.server.host)
 
             elif nw.status_code in (411, 423, 430, 451):
                 article_done = True
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -753,7 +753,7 @@ class Downloader(Thread):
 
             elif nw.status_code == 223:
                 article_done = True
-                logging.debug("Article <%s> is present", article.article)
+                logging.debug(f"Article {article.article} is present on {nw.server.host}")
 
             elif nw.status_code in (411, 423, 430, 451):
                 article_done = True


sabnzbd/downloader.py
```diff
@@ -753,7 +753,7 @@ def process_nw(self, nw: NewsWrapper):
 
             elif nw.status_code == 223:
                 article_done = True
-                logging.debug("Article <%s> is present", article.article)
+                logging.debug(f"Article {article.article} is present on {nw.server.host}")
```
Safihre: Please use the %s way of doing things. This way the string  substitute is only applied if Debug logging is set.

sanderjo:
Hmmm ... we have two types of formatting the article: without and with < resp >.  Do we want to streamline that? If so: with or without the < resp ?

```
2025-11-08 09:32:15,070::DEBUG::[downloader:761] Thread 9@news.newshosting.com: Article xxxxxxxxxdf1a7@56520167.6mQ missing (error=430)

2025-11-09 20:49:02,798::DEBUG::[downloader:758] Article <OfOkPwQyEwMnZhAuBtOnUjBv-1658672621044@nyuu> is present on news.newshosting.com
```

```
$ grep -rn Article * 2>&1 | grep "%s"
sabnzbd/decoder.py:138:                        logging.info("Article removed from server (%s)", art_id)
sabnzbd/downloader.py:756:                logging.debug("Article <%s> is present", article.article)
sabnzbd/downloader.py:761:                    "Thread %s@%s: Article %s missing (error=%s)",
sabnzbd/nzbstuff.py:268:        logging.info("Article %s unavailable on all servers, discarding", self.article)
sabnzbd/nzbstuff.py:293:        return "<Article: article=%s, bytes=%s, art_id=%s>" % (self.article, self.bytes, self.art_id)
```

So only in STAT we print with < resp >. Remove that ... ?

Safihre:
Sure. Use article object everywhere instead of the id.

PR #3197, opened 2025-11-17
Address low throughput reporting in diskspeed.py

L-Cie:
I noticed that the current version of `diskspeed.py` reports significantly lower throughput values than what is realistically achievable. Depending on the storage device and connection, the benchmark output shows only about half of the possible write speed.

To validate this, I used fio under Linux to measure sequential write performance. I ran `fio` with the following parameters:
```
fio --name=seqtest --rw=write --bs=128k --ioengine=libaio --size=10G --numjobs=1 --direct=1 --iodepth=1 --group_reporting --filename=/home/<USER>/10g.test
```
My `fio` results were:
SATA SSD: 508 MiB/s (533 MB/s)
ZFS HDD Pool: 507 MiB/s (531 MB/s)
NVMe SSD: 1002 MiB/s (1051 MB/s)

After noticing the discrepancy, I looked into the script and also read through the discussion at https://github.com/sabnzbd/sabnzbd/discussions/2846, where this issue had already been mentioned.

At first, I suspected synced writes to be the main cause, but after experimenting with changes to the `urandom` buffer and testing several other parameters, I arrived at the current version included in this PR. I also incorporated some code improvements from @mnightingale from the discussion. Additionally, the benchmark duration has been increased from 0.5 seconds to 3 seconds, which can potentially be reduced again, but it currently provides much more consistent results for me.

The new script now produces the following values:
SATA SSD: 461.95 MB/s
ZFS HDD Pool: 483.65 MB/s
NVMe SSD: 1034.10 MB/s

These numbers are significantly closer to the `fio` results compared to the previous version, which reported:
SATA SSD: 247.40 MB/s
ZFS HDD Pool: 195.40 MB/s
NVMe SSD: 917.30 MB/s

### DIFF
--- a/sabnzbd/utils/diskspeed.py
+++ b/sabnzbd/utils/diskspeed.py
@@ -7,14 +7,7 @@ import sys
 import logging
 import time
 
-BUFFERSIZE = [
-    16 * 1024 * 1024,
-    32 * 1024 * 1024,
-    128 * 1024 * 1024
-]
-
-# Prepare the whole buffer now for better write performance later
-buffer = os.urandom(BUFFERSIZE[-1]); # Dump 128 MB of trash in RAM
+BUFFERSIZE = 16 * 1024 * 1024
 
 def diskspeedmeasure(dirname: str) -> float:
     """Returns writing speed to my_dirname in MB/s
@@ -22,6 +15,9 @@ def diskspeedmeasure(dirname: str) -> float:
     Then divide bytes written by time passed
     In case of problems (ie non-writable dir or file), return 0.0
     """
+    # Prepare the whole buffer now for better write performance later
+    buffer = os.urandom(BUFFERSIZE); # Dump 16 MB of trash in RAM
+    
     start = time.time()
     maxtime = 1 # sec
     total_written = 0
@@ -35,11 +31,12 @@ def diskspeedmeasure(dirname: str) -> float:
         maxtime += start
         
         # Start looping
-        for i in range(len(BUFFERSIZE)):
+        for i in range(1,5):
             # Stop writing next buffer block, if time exceeds limit
             if time.perf_counter() >= maxtime:
                 break
-            total_written += os.write(fp_testfile, buffer[0:BUFFERSIZE[i]])
+            # Increase chunk size after each iteration
+            total_written += os.write(fp_testfile, buffer * (i**2))
             os.fsync(fp_testfile)
             
         total_time = time.perf_counter() - start
--- a/sabnzbd/utils/diskspeed.py
+++ b/sabnzbd/utils/diskspeed.py
@@ -17,9 +17,9 @@ def diskspeedmeasure(dirname: str) -> float:
     In case of problems (ie non-writable dir or file), return 0.0
     """
     # Prepare the whole buffer now for better write performance later
-    buffer = os.urandom(BUFFERSIZE); 
+    buffer = os.urandom(BUFFERSIZE) 
     # Dump 16 MB of trash in RAM
-    
+
     start = time.time()
     maxtime = 1  # sec
     total_written = 0
@@ -30,7 +30,7 @@ def diskspeedmeasure(dirname: str) -> float:
         # Use low-level I/O
         fp_testfile = os.open(
             filename, os.O_CREAT | os.O_WRONLY | getattr(os, "O_BINARY", 0) | getattr(os, "O_SYNC", 0), 0o777
-            )
+        )
         start = time.perf_counter()
         maxtime += start
 
@@ -42,7 +42,7 @@ def diskspeedmeasure(dirname: str) -> float:
             # Increase chunk size after each iteration
             total_written += os.write(fp_testfile, buffer * (i**2))
             os.fsync(fp_testfile)
-            
+
         total_time = time.perf_counter() - start
         # Have to use low-level close
         os.close(fp_testfile)
--- a/sabnzbd/utils/diskspeed.py
+++ b/sabnzbd/utils/diskspeed.py
@@ -7,8 +7,14 @@ import sys
 import logging
 import time
 
-_DUMP_DATA_SIZE = 256 * 1024 * 1024 # Increased buffer for better write performance
+BUFFERSIZE = [
+    16 * 1024 * 1024,
+    32 * 1024 * 1024,
+    128 * 1024 * 1024
+]
 
+# Prepare the whole buffer now for better write performance later
+buffer = os.urandom(BUFFERSIZE[-1]); # Dump 128 MB of trash in RAM
 
 def diskspeedmeasure(dirname: str) -> float:
     """Returns writing speed to my_dirname in MB/s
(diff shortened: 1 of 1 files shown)


sabnzbd/utils/diskspeed.py
```diff
@@ -18,30 +18,28 @@ def diskspeedmeasure(dirname: str) -> float:
     """
     dump_data = os.urandom(_DUMP_DATA_SIZE)
     start = time.time()
-    maxtime = 0.5  # sec
+    maxtime = 3  # sec
```
thezoggy: 3s is a bit much as you would do two tests (incomplete + complete). 
so that 6 seconds delay with the other tests would be more than most would probably be okay with. 1s would probably be fine.

Safihre:
Thanks for looking into this and sharing your feedback!

While we are aware that the current test is not accurate, as you saw from the discussion having an exact test is also not the main purpose of this information. It is really to identify the terrible performers (e.g. 10MB/s or lower, for example to a mapped NAS drive) from "normal" speeds.

Because this is part of the general tests-suite, we would need to keep this test fast so the total test-suite time doesn't exceed 10 seconds. 

We should also consider those terrible performers, because if the speed is indeed 10MB/s using 256MB would take 25 seconds. That's not acceptable.

What I think might work is trying a staggered approach: 
First try 10MB, if we have time left in our test window, try to write an additional 25MB, and if we have still time left after that, try writing an additional 50MB.
What do you think about that?

L-Cie:
Hey @Safihre 

I really appreciate your time and feedback on my PR. I have considered your suggestions and implemented the proposed improvements. The script now writes smaller chunks to disk first before utilizing the full `urandom` buffer, ensuring that the benchmark no longer exceeds the time limit when writing on slower storage. I have chosen chunk sizes of 16, 32, and 128 MB.

While the results for my NVMe are now slightly better and more consistent now, there is greater variability on my SATA SSD and ZFS HDD pool. However, the performance "scores" have overall still improved compared to before:

SATA SSD: 340 - 408 MB/s
ZFS HDD Pool: 360-410 MB/s
NVMe SSD: 1240 MB/s

sabnzbd/utils/diskspeed.py
```diff
@@ -7,41 +7,47 @@
...
 
-_DUMP_DATA_SIZE = 10 * 1024 * 1024
+BUFFERSIZE = [
+    16 * 1024 * 1024,
+    32 * 1024 * 1024,
+    128 * 1024 * 1024
+]
 
+# Prepare the whole buffer now for better write performance later
+buffer = os.urandom(BUFFERSIZE[-1]); # Dump 128 MB of trash in RAM
```
Safihre: This should be inside `diskspeedmeasure` otherwise we keep always 128MB of junk in memory.
We also run on small NAS devices that only have 512MB of memory for the whole device.
I would maybe even opt to only initialize the random data if the data is needed. So 128MB only if we were already able to write 16 and 32.

L-Cie: I moved the urandom dump to `diskspeedmeasure` now.

I also rewrote how we handle the `urandom` dump. The note about low-end devices was very helpful. Now, I use the same buffer for the writes and simply recycle it. With each iteration, more data is written to disk to ensure meaningful measurements can be obtained.

In total, there are now 4 writes, as long as time permits: 16 MB, 64 MB, 144 MB, and 256 MB. The buffer is simply multiplied, so we don’t unnecessarily fill up RAM with trash data.

Safihre: How does this version reported speed compare on your system to the fio one?

L-Cie: My SATA SSD does struggle more with this version, whereas the HDD Pool and my NVMe do not. I get inconistent results for the SSD, the rest is fine. I did about 10-15 runs.

SATA SSD:  291 - 419 MB/s 
ZFS HDD Pool: 418 - 467 MB/s
NVMe SSD: 1194 - 1231 MB/s

Safihre:
Could you fix the formatting issues?
Then I can merge👍
https://github.com/sabnzbd/sabnzbd/actions/runs/19512628948/job/55856050830?pr=3197

L-Cie:
Sorry for the formatting mess. Should be fine now.

Safihre:
Thanks for your contribution 👍🎉

mnightingale:
It's a pretty good improvement, but it's kind of a CPU and memory benchmark now.

`buffer * (i**2)` allocates bytes every iteration, that'll cause issues on low memory devices.
`print(psutil.Process(os.getpid()).memory_info().rss / 1024**2)` reports > 500 MB

`range(1, 5)` I'd switch back to purely time based, but also needs a `total_written` limit.

I'll send some improvements this evening, but I suggest something like:

```py
# Start looping
while time.perf_counter() < endtime:
    for _ in range(chunk_count):
        total_written += f.write(buffer)
    if chunk_count < max_chunk_count:
        chunk_count += 1
```

| i | range(chunk_count) | range(chunk_count**2) |
| - | - | - |
| 1 | 16 MB | 16 MB |
| 2 | 32 MB | 64 MB |
| 3 | 48 MB | 144 MB |
| 4 | 64 MB | 256 MB |
| 5 | 64 MB | 256 MB |
| 6 | 64 MB | 256 MB |

For the current scaling do `range(chunk_count**2)`, or make that the assignment instead of `chunk_count += 1`.

I think time could be reduced back to 0.5 seconds this way.

Safihre:
I just pushed some changes to remove the buffer-allocation from the timing.
But writing the same chunk every time isn't really the same though, writing a bigger buffer does have value I think.

mnightingale:
The current changes are better than they were, but it allocates a lot more memory than before. I just wanted to make sure were aware given the earlier 512 MB NAS device concerns.

I do like that we know the maximum amount of disk space required for the test.

It's just in my opinion it might as well have used BUFFERSIZE of 256 MB and a memoryview if it's going to allocate the initial 16 MB then copy it for passes of 16 + 64 + 144 + 256 = 16+480 MB per run, they are quickly released but at the end it has 16+256 allocated, which is more than just allocating 256 MB in the first place.

The larger buffers do help at least on COW filesystem (ZFS).

thezoggy:
> The current changes are better than they were, but it allocates a lot more memory than before. I just wanted to make sure were aware given the earlier 512 MB NAS device concerns.
> 
> I do like that we know the maximum amount of disk space required for the test.
> 
> It's just in my opinion it might as well have used BUFFERSIZE of 256 MB and a memoryview if it's going to allocate the initial 16 MB then copy it for passes of 16 + 64 + 144 + 256 = 16+480 MB per run, they are quickly released but at the end it has 16+256 allocated, which is more than just allocating 256 MB in the first place.
> 
> The larger buffers do help at least on COW filesystem (ZFS).

pretty sure sab wouldnt really work well on a 512MB device as that ram starved alone would prob puke trying to do nas+sab and then about anything of the process when it comes to article cache/paring/unraring/etc. but deff dont want to make it where someone in the middle of doing a download comes along and tries to do disk speed test for some reason and cause themselves an oom issue. 

could prob guard from letting someone do folder test if sab queue isnt empty or paused? (or we forcibly pause sab to run the test then unpause)

PR #3199, opened 2025-11-18
Pipelining and performance optimisations

### LINKED ISSUE #3171: NNTP pipelining
@mnightingale our friends at Newshosting said they also have experimented with pipelining recently and that especially in higher latency connections they see quite good improvements. So they (and their resellers) would be happy to see us implement it.
I would also be interested to look at implementing it, or do you already have something in mind? 😊

_Originally posted by @Safihre in https://github.com/sabnzbd/sabnzbd/discussions/2352#discussioncomment-14743621_

mnightingale:
I've had pipelining in the back of my mind for a couple of years; I just hadn't found the motivation to properly investigate and find a way to implement it.

There is an excellent explanation of the theory on https://github.com/sabnzbd/sabnzbd/issues/3171#issuecomment-3538705659 the general idea is that the current request => response loop is bottlenecked by latency, a lot of the time no data is being transmitted. 

[RFC 3977](https://datatracker.ietf.org/doc/html/rfc3977) describes NNTP server support for pipelining. The general idea is that by having a constant stream of requests going out we better saturate the incoming bandwidth.

There are a few aspects to the pull request:
- Pipelining, which is intended to improve speeds for >1 Gbps connections or at high latency, use fewer connections, and therefore fewer resources
- Streaming/incremental decoding, rather than reading a whole response into a buffer before decoding it all at once, process small parts as they are received, this can use less memory, process data quicker because the decoded data to be crc'd is still in cpu caches
- Performance optimisations, primarily use of more appropriate data structures for their use-case, [TimeComplexity](https://wiki.python.org/moin/TimeComplexity) - motivation is to challenge that SAB is "slow" because it's in Python, I don't think it is there are just some non optimal choices from 10+ years ago before fast connections were more common place
- Thread safety, fixed a few but identified issues

Closes #3171 🚀 

## What has changed and why

This is kind of a lot, but I expected you might want to know why I've made many of these changes, especially as it touches 15+ year old code in places.

<details>
  <summary>Implementation specifics</summary>

- Config/Interface: added pipelining_requests to specials (may need a better name), default 5, in the future (maybe not this PR) to be replaced with an ever-changing calculation based on average request -> complete time
- Server:
    - **article_queue:** a sequence of prefetched articles assigned to a server
        - Changed to a deque, efficient O(1) pop and append operations; with lists pop(0) is O(n) it memmoves all the other elements [ref](https://github.com/python/cpython/blob/b87613f21474ea848fec435cbfe63d8cb1c7c44c/Objects/listobject.c#L1558-L1561) so a prefetch of 20 items makes 20 memmoves for 190 elements in total; a deque is a doubly-linked list, appends and pops to either end are just setting a couple of pointers, in such a hot path this matters
        - Dequeue is thread safe (at least for append, pop, and len) - but locks may be required on occasion such as checking length before pop/peek
        - A single object instance is now used for the lifetime of the server, less constant creation, growing, and destruction of lists
    - **get_article(..., peek:bool = False):** I needed a way to peek at the next item without consuming it, only used to kickstart idle servers, if the socket isn't in a state to accept article requests yet it can't be added to command queue because it could interrupt auth.
    - **selectors:** replaces socket.select
        - Uses efficient select implementation for the platform
        - Lets us subscribe to availability for read and write events, without maintaining separate lists and combining results as with socket.select
- Decoder - one of two may need renaming:
    - Decoding of yEnc and UU are now handled by `sabctools.Decoder`
    - Decoder just copies required values out of the decoder metadata
    - May need to look at what happens for unknown formats, if that can happen they won't get to the decoder anyway.
- Downloader:
    - **__reset_nw => reset_nw:** yeah... I don't like this, it's mostly operations on a nw so I feel it should be a method on a nw but it needs the DOWNLOADER_LOCK, I needed access from NewsWrappper since it now writes requests.
    - **process_nw_queue.put_multiple:** no longer used, I profiled and it's now faster this way because we need `(NewsWrapper, event)`, use of MultiAddQueue could be replaced with queue.Queue but I wanted to keep the changes lean
    - **process_nw:** now reads from the socket in `process_nw_read` like before but in a loop to consume `sock.pending()`, second read only takes `sock.pending()` to avoid potentially reading from a single socket for too long. Much of the logic for dealing with the responses has moved to the NewsWrapper
    - **__finish_connect_nw => finish_connect_nw:** similar to the other one, it's now only used in NewsWrapper so could/should move but didn't want to complicate the PR
    - **__request_article:** removed, NewsWrapper needs to coordinate reads and writes
- NewsWrapper: The real meat of things
    - **command_queue and concurrent_requests:** the general idea is there is a queue of commands we want the server to send, and in-flight limits use a threading.Semaphore that is released once we have a whole response.
    - **send_buffer:** remainder of partially sent data, rare but allows us to make non-blocking writes and avoids potential deadlock when we can't write more until we read; sendall is basically a blocking while loop that doesn't terminate until it has all sent
    - **queue(command, article):** queues a command, blocks until space in the command queue
    - **command_complete:** not hugely in favour of the name, maybe completed_command?
        - Once a decoder has read a whole response, figure out what to do with it.
    - **recv_chunk => read(..., nbytes: int = 0):**
        - Per NewsWrapper buffer doesn't need to resize anymore, the decoder stores its streamed data internally, still enforce the limit against how many bytes a decoder has consumed to check something strange has not happened
        - nbytes is used when we want to limit to reading just sock.pending() and always uses the python default recv_into because we need to read <16 KiB anyway.
        - Tells sabctools how much data it needs to process, which handles all decoding and multiple responses per read
        - Reads back any number of completed responses from the decoder
    - **write:**
        - Writes command_queue and send_buffer to the socket, enqueues more requests to command_queue (I might make it limited loop, need to check if necessary - I doubt it), operates in a non-blocking manner that is if writes would block it stops so we can move on to the next socket
        - Handles idle connection detection - can write but nothing to write
    - **soft_reset:** removed, the NewsWrapper need to manage its state
- NzbQueue:
    - **get_articles:** including the chain from NzbQueue > NzbObject > NzbFile
        - Populates the server deque, instead of returning a new list instance
    - **register_article:** locks the nzf instead of copying the articles (I'll get onto this later)
- TryList
    - Adds a per object lock and can use it in `__enter__` and `__exit__`, e.g. `with nzf:`
    - **Why?** because multiple threads can manipulate per object state, affects other places, primarily an issue since `process_nw_worker` was added with multiple connections in different threads working on the same nzf, remove items from a list/dict, or iterate, and weird things can happen.
    - I think `TRYLIST_LOCK` is likely overused, contention on a single lock when the changes it's protecting don't conflict, but again I didn't want to fix it elsewhere in this PR.
    - n.b. I'm not sure I'm convinced by `try ... except` being faster than using `__enter__` and `__exit__` except in microbenchmarks at exceptional rates, the decorator is nice but can only use global locks and not all methods that modify data are using the TRYLIST_LOCK, so it can be changed by other threads anyway.
- NzbFile
    - **articles:** changed list to a dict, O(1) removals, test_download_10GB.nzb is a good demonstration, first file is 1498 articles, excluding the first, list removals are O(n) that’s worst case 1,496 memmove calls for 1,119,756 elements, we generally download in order so worse case is common for us.
    - **remove_article**
        - locks articles
        - removed the `article in self.articles` check and take advantage of dict pop method, single operation to remove if present, rather than searching and searching again for a removal (although cheaper now with dicts)
    - **get_articles**, **reset_all_try_lists** (not used?)
        - locks articles
    - **completed:**
        - locks articles when necessary
- General:
    - **self:** removed a few hot path uses of self by using a local variable
</details>

## Threading and Locks

Outlined in a few places above, but there are several existing not thread safe operations. I expect I'm finding them because it's trying to operate so fast.

I've had failures at `sabnzbd.ArticleCache.purge_articles` `Set changed size during iteration` `@synchronized(NZO_LOCK)` is not sufficient, an API request can remove an item while something else saves to the cache. There are many similar issues throughout the codebase, I intend to address them in a separate PR.

A common "fix" appears to be shallow copying objects instead of fixing the underlying threading issues, this is poor for performance and not safe. "Not locked for performance" doesn't pass the smell test, it's not safe if something else can modify the object while it's being used. When not handled correctly it can cause undefined state changes.

I expect there are ways test for thread safety issues, it's more likely an issue with many receive threads or the free-threaded model in the future.

## Does it work?

You tell me; testing on a local 10 Gbps network using `tc` with a 20ms delay, 16 connections, 5 pipelining, on a now 5-year-old desktop (5800x) has gone from 600 MBs to 1000 MBs, but it needs some real-world testing to see how real servers respond.

An **RPI4** shows a **100%** speed increase with the optimisations and pipelining makes little to no difference, it is bottlenecked by lack of hardware acceleration for TLS decoding.

An **i5 14500** shows minor **5-10%** increase with no pipelining, purely the CPU usage optimisations. Pipelining at even low levels 2-5 is extremely beneficial at low and high latency **100-300%** but too high is detrimental.


sanderjo:
To verify I understood the requirements, I did the below in a ubuntu docker, and ... it works. Is the below correct?

So now? Play with different settings, and see if with pipelining I need less connections? Most of my systems are CPU / disk bound compared to my 2500 Mbps connection, and I have short RTT to newshosting and eweka (10 ms), so limited in what I can test IRL.

```
apt update && apt install python3-pip git unrar par2 7zip -y

cd
git clone https://github.com/mnightingale/sabctools.git
cd sabctools/
git checkout origin/feature/streaming_decoder
python3 setup.py install

cd
git clone https://github.com/mnightingale/sabnzbd.git
cd sabnzbd/
git checkout origin/feature/pipelining
sed -i '/^sabctools/d' requirements.txt # to be sure: we don't want the generic sabctools
python3 -m pip install -r requirements.txt -U

./SABnzbd.py  -b0 -l2 -s 0.0.0.0
```


How can I see the sabctools is the one with streaming_decoder built in? Brute-force it works like this:




```
root@606d23133f32:~/sabnzbd# strings /usr/local/lib/python3.10/dist-packages/sabctools-8.2.6-py3.10-linux-x86_64.egg/sabctools/sabctools.cpython-310-x86_64-linux-gnu.so  | grep -i stream
no_stream_resources
stream_timeout
not_a_stream
iostream
root@606d23133f32:~/sabnzbd#
```
versus

```
sander@brixit:~$ strings /usr/lib/python3/dist-packages/sabctools/sabctools.cpython-312-x86_64-linux-gnu.so | grep -i stream
sander@brixit:~$
```

mnightingale:
It looks like you've built it correctly, it would soon tell you in SAB though because the versions are not compatible with each other.

I should remove iostream, it was useful while developing, but no need to bloat it.

You can check the exports with something like `nm sabctools.cpython-313-x86_64-linux-gnu.so | grep STATUS_SUCCESS`

I would keep the pipelining level low (1-10), I intend to reduce the allowed values, high connection count and high pipelining leads to worse performance. This will be address by managing it based on response time.

What I'm primarily looking for people to find is:

- Is CPU usage lower for the same connection count at pipelining 1 (compared to develop branch; this is just my non-pipelining performance improvements)
- Can the same speed by achieved with fewer connections and relatively low pipelining levels.
- Any discernible impact on speed, CPU, or memory usage by lowering NNTP_BUFFER_SIZE – optimal is likely 64-128 KB
- Pipelining effect on speed at high latency, compare very low connection counts
- What happens with 10 vs 50 connections - my testing shows too many has a big impact at high latency and too much pipelining

sanderjo:
To make testing easier for myself and others: I've created a Dockerfile: https://github.com/sanderjo/sabnzbd-with-pipelining


Within that docker, on my i3 laptop, wired to 2500 Mbps line, mean speed for 1GB-testfile (with 0.x second ramp up => impact on mean speed, but also: everything into RAM, so no disk limitation):

5 connections to newshosting: mean speed 104 MB/s
10 connections to newshosting: mean speed 210 MB/s # wow, I've never seen this speed with so little connections
30 connections to newshosting: mean speed 262 MB/s

sanderjo:
an NZB with all articles missing is not great: SAB shows, after a few minutes of trying:
image

There is no loop: all tried articles are unique. 

Ah, the trying took 6 minutes. But then: nothing, still in the queue. 

Weird.

End of sabnzbd.log:

```
2025-11-20 16:31:13,192::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.part11.rar in job all_missing_1000MB
2025-11-20 16:31:13,193::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol000+001.par2 in job all_missing_1000MB
2025-11-20 16:31:13,193::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol001+002.par2 in job all_missing_1000MB
2025-11-20 16:31:13,193::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol003+004.par2 in job all_missing_1000MB
2025-11-20 16:31:13,193::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol007+008.par2 in job all_missing_1000MB
2025-11-20 16:31:13,193::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol015+016.par2 in job all_missing_1000MB
2025-11-20 16:31:13,194::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol031+032.par2 in job all_missing_1000MB
2025-11-20 16:31:13,194::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol063+064.par2 in job all_missing_1000MB
2025-11-20 16:31:13,194::INFO::[nzbqueue:907] Resetting bad trylist for file sometestfile-1GB.vol127+072.par2 in job all_missing_1000MB
2025-11-20 16:31:13,194::INFO::[nzbqueue:911] Resetting bad trylist for job all_missing_1000MB

```

... and then nothing more.

mnightingale:
> an NZB with all articles missing is not great: SAB shows, after a few minutes of trying

Thanks, added it to my list of things to look into.

When it's working correctly the pipelining should mean it's able to pre-check or cope with missing responses much faster than currently, even more so if it dynamically scales and knows stat/head are short responses.

mnightingale:
> > an NZB with all articles missing is not great: SAB shows, after a few minutes of trying
> 
> Thanks, added it to my list of things to look into.

That should be fixed now

sanderjo:
> That should be fixed now

Yes, confirmed. Both with and without Check-Before-Download.

Question: if I find more, do you want it in this thread, or a separate issue (on your github, or here on sab's)? A separate issue might be cleaner?

mnightingale:
Separate issues would probably be better, you can report them on https://github.com/mnightingale/sabnzbd/issues to avoid cluttering this repo and so I can manage them, and I'll track their progress on https://github.com/users/mnightingale/projects/1

mnightingale:
I've switched the requirements.txt to my branch, so GitHub should be building working binaries.

There are not a whole lot of test failures, just test_decoder.py which is UU tests, they'd possibly be more appropriate in sabctools now?

I think I need to make a few changes in sabctools to allow creation of NNTPResponse instances from Python for where it wants to call `decoder.decode_uu` and modify a few tests so that they end `.\r\n` and start with a status line.

It looks like there are a bunch I'm not sure what to do with e.g. `test_missing_uu_begin` with ` b"222 0 <foo@bar>\r\n\r\n",  # Missing altogether` sabctools will decode that as `format=None` and `bytes_decoded=0` so it would never call `decoder.decode_uu`, in decode_uu I have maintaned the `raise BadUu` behaviour if binary data is empty, but as it stands this would follow the decode_yenc path so it probably needs a check for format=None and raise BadData or just make the yenc path raise if bytes_decoded is 0.

PR wise I "just" need to resolve the tests and make the pipelining smarter, I think best I can do is measure `full response time - request time` and derive how many it could have inflight but I'll need to only measure body requests and only sample them.

Safihre:
Your test changes sound good to me.
But I would strongly encourage not to make it smart from the start, but first focus on the concept and get it out in the world to test.
We can have many more releases to make it smart.

You're probably also in software development for your work life (assumption 😃), so let's keep things agile and iterate every version!

sanderjo:
> get it out in the world to test.

Indeed.

So on my main download system, I've now switched to the pipelining SABnzbd. It's running in a docker container, based on https://github.com/sanderjo/sabnzbd-with-pipelining

So far so good: downloads are working without problems.

The only proof I have I'm running it:

image

EDIT:

Ah, sabctools 9.0.0 is available on the repo's, so I can remove the sabctools special in my Dockerfile.

```
2025-12-02 20:31:59,809::INFO::[SABnzbd:422] SABCTools module (v9.0.0)... found!
2025-12-02 20:31:59,809::INFO::[SABnzbd:423] SABCTools module is using SIMD set: AVX512VL+VBMI2
```

mnightingale:
@safihre please can you release 9.0.1 when you have a moment.

A few non-critical areas you might want to change:

decoder.py is a little repetitive (decode_yenc and decode_uu), it also doesn't do any decoding anymore so there are paths it can't take such as except MemoryError.

The command queue per newswrapper may be buffering too many requests (pipelining level * 3 but with a limit) but I don't think it really matters. This is commands we have ready to send but are waiting until existing ones complete first. It's like the pre-fetch get_articles does but per connection. The purpose is that read and write are now disconnected from each other, for fastest speeds if a socket is available for writing we want to do so immediately and not have to spend CPU cycles deciding which article to request and holding up all sockets. Even with the prefetch get_articles still has a delay every 20 items. I think it could be much shorter because the buffer is populated in write whenever it’s not already full, which will be called a lot more often than it actually makes requests.

I decided against unsubscribing from EVENT_WRITE when we can't send anymore requests until we get a response, it would make it more complicated to keep track of and constantly be changing the events to listen for so likely slower than just listening and not doing anything.

I made a special rather than configuring per server for now, eventual intention is to have it as a simple on/off and dynamically adjust, or if there are no problems it could just be always on and transparent to the user. As it is currently per server would be nicer to adjust based on server latency and bandwidth, but I didn't want to if it is only going to be replaced anyway.

Due to above limitation, I was conservative with the default of 2 requests as I wanted to ensure it’s not too high for low bandwidth connections but still have it enabled by default, so that people try it. With low latency I find 2-5 is plenty. A challenge currently may be getting users to try fewer connections and/or greater pipelining.

The lock I added in TryList, could/should be used in a lot more places to lock at NZO/NZF level rather than a global lock and only lock for the specific calls that require it rather than whole methods.

sabnzbd/nzbstuff.py
```diff
@@ -118,11 +118,12 @@ class TryList:
     """TryList keeps track of which servers have been tried for a specific article"""
 
     # Pre-define attributes to save memory
-    __slots__ = ("try_list",)
```
Safihre, line 121: Is this really a good idea to have a lock for each article? That's a lotttttttt of locks :|

Safihre:
I released 9.1.0 as it's a required update while a patch could be interpreted as "optional".

sabnzbd/nzbstuff.py
mnightingale: I'm currently only using it at the NzbFile level so I could move it to that class.
If I make changes in the future I'll likely only additionally need it at NzbObject level.

However, let me check the paths to everything I used it on because I might be able to remove some or all of it since the global locks are still being used, I may have decided against changing it in this PR.

The time to create them will be micro seconds and maybe a hundred bytes but if I keep it I'll move it out of TryList.

mnightingale: I've moved it, but let me think about it a little longer.

mnightingale: I think the nzf locks need to remain, there are paths that want to iterate articles while others threads remove them.

mnightingale:
An interesting failure, I'm pretty sure the issue that sabnews is not a safe pipelining server implementation.
I think it can interleave responses to requests and just needs a lock on serve_article.

edit: yeah putting a `asyncio.sleep(random.random())` in the middle of a sabnews response makes it easy to reproduce, let me see how to make it work in asyncio.
edit2: I've fixed it locally but need to fix a different issue @sanderjo found

mnightingale:
I think I've resolved all the issues I know about.

sabnews I've switched to start_server, it creates a handler per connection with a nicer reader/writer interface especially when there could be multiple requests in the buffer. The implementation finishes each request before moving on to the next so it won't interleave responses anymore. The only functionality change I made was adding a `500 Unknown command` response, just as a safe-guard because if it was used for anything except quit/article/body you would get no response. It might be nice to add STAT/HEAD support sometime, that would have picked up my 'Check Before Download' failures earlier.

thezoggy:
> I think I've resolved all the issues I know about.
> 
> sabnews I've switched to start_server, it creates a handler per connection with a nicer reader/writer interface especially when there could be multiple requests in the buffer. The implementation finishes each request before moving on to the next so it won't interleave responses anymore. The only functionality change I made was adding a `500 Unknown command` response, just as a safe-guard because if it was used for anything except quit/article/body you would get no response. It might be nice to add STAT/HEAD support sometime, that would have picked up my 'Check Before Download' failures earlier.

does all your changes handle when people change their usenet server setup on sab (how many servers are enabled or remove+adding/priority/connections/pipeline requests)  in the middle of a job for example

mnightingale:
> does all your changes handle when people change their usenet server setup on sab (how many servers are enabled or remove+adding/priority/connections/pipeline requests) in the middle of a job for example

Perhaps not quite as well as it should.

- **Removing an active server** currently it won't queue any more work but will finish anything already queued (pipelining level * 3, capped at 20) but that's per connection.
  - I can and probably should at least make it abandon anything queued but not yet sent
  - Abandoning anything already requested should be may be possible, we'd be sending a quit message while the server is still sending data.
- **priority / connections** are similar, it won't restart until it has completed currently queued requests.
- **pipelining level** requires a client restart
- edit: on pause **reset_nw** does a hard_reset which discard send and unsent, so is immediate

@Safihre any input on which you'd prefer or think we need?

1. Clear unsent requests - delay to restart is time for already sent requests
2. (1) + discard sent requests

I'm also going to reduce the max queued to 1 or 2 requests, because 20 * 750_000 * #connections is a lot.

edit2: pipelining will effect responsiveness when changing queue order, let me think about it.

Safihre:
What about when you shutdown the application in the middle of downloading?

mnightingale:
I've removed the command_queue, and just prepare one request like the current implementation with nw.article.
It discards the next article request when the server is not active, or wants restart, etc.
Before sending the next request it checks if it is PAUSED and not FORCE_PRIORITY.
So it should behave as currently.

I haven't changed it but server reconfigs wait until all connections are idle, then recreate and replace. Far simpler than live changes.

The pipelining effect means even at the default (2) twice the requests have already been sent, so reordering the queue or pausing individual items will appear more sluggish, but that's unavoidable those requests have already been sent. For example, with 40 connections at 2 pipelining, that's approximately 80 * 0.75MB to receieve before you start getting responses after a queue order change. There is no way around that, those request are already on there way to you.

Shutdown or global pause are instant or the same as currently, hard_reset discards everything and just closes the socket.

edit: think I've found an issue related to resuming idle connections

thezoggy:
how does it handle retries / exceeding connection limits.. does it continue to hammer or adjust pipeline requests?

mnightingale:
> how does it handle retries / exceeding connection limits.. does it continue to hammer or adjust pipeline requests?

There is no pipelining until after authentication, so invalid auth or exceeding connection limits still follows the existing maybe block server behaviour.

By retries do you mean missing articles or failure to connect? the queue and connection logic is unchanged, so it'll continue trying lower priority servers or pause if it's a required server.
With pipelining precheck and responding to missing articles can be much faster.

sabnzbd/newswrapper.py
```diff
@@ -151,98 +163,254 @@ def finish_connect(self, code: int):
...
+                        nzo = article.nzf.nzo
+                        if nzo.removed_from_queue or nzo.status is Status.PAUSED and nzo.priority is not FORCE_PRIORITY:
+                            self.discard(article, count_article_try=False, retry_article=True)
+                            self.concurrent_requests.release()
+                            return
+                    self._response_queue.append(article)
+                    if sabnzbd.LOG_ALL:
+                        logging.debug("Thread %s@%s: %s", self.thrdnum, server.host, command)
+                    try:
+                        sent = self.nntp.sock.send(command)
```
Safihre, line 366: Why not use sendall for the writes?

sabnzbd/newswrapper.py
```diff
@@ -151,98 +163,254 @@ def finish_connect(self, code: int):
...
+                        if sent < len(command):
+                            # Partial send, store remainder
+                            self.send_buffer = command[sent:]
+                    except (BlockingIOError, ssl.SSLWantWriteError):
```
Safihre, line 370: This happens in practice? Or only theory?

sabnzbd/nzbstuff.py
```diff
@@ -469,6 +475,12 @@ def remove_admin(self):
         except Exception:
             pass
 
+    def __enter__(self):
+        self.lock.acquire()
```
Safihre, line 479: Isn't this a bit weird? I get why you need it later, but to see it as `with file:`  to indicate that it's using the lock?

sabnzbd/nzbstuff.py
```diff
@@ -347,7 +348,7 @@ def __init__(self, date, subject, raw_article_db, file_bytes, nzo):
         self.setname: Optional[str] = None
 
         # Articles are removed from "articles" after being fetched
-        self.articles: list[Article] = []
+        self.articles: dict[Article, Article] = {}
```
Safihre, line 351: Why did we move to a dict?

sabnzbd/nzbstuff.py
```diff
@@ -328,11 +328,12 @@ class NzbFile(TryList):
     """Representation of one file consisting of multiple articles"""
 
     # Pre-define attributes to save memory
-    __slots__ = NzbFileSaver
+    __slots__ = NzbFileSaver + ("lock",)
 
     def __init__(self, date, subject, raw_article_db, file_bytes, nzo):
         """Setup object"""
         super().__init__()
+        self.lock = threading.RLock()
```
Safihre, line 336: I'm a bit confused why the lock was introduced?

sabnzbd/newswrapper.py
mnightingale: I did cover this in the 'Implementation specifics' section above, there is a tiny risk that sendall could block.
However given pipelining is limitted to only a few requests, I think we coudl get away with removing send_buffer.
When working with non-blocking sockets, this is how it should work though else you're using non-blocking sockets but enforcing your own blocking anyway.

sabnzbd/newswrapper.py
mnightingale: As above I'm comfortable with removing send_buffer

sabnzbd/nzbstuff.py
mnightingale: It's just syntactic sugar around `with nzf.lock`, is the problem you don't think it's clear to the caller that it's using an nzf level lock?

sabnzbd/newswrapper.py
```diff
@@ -151,98 +163,254 @@ def finish_connect(self, code: int):
...
+                    and not server.restart
+                    and not (
+                        sabnzbd.Downloader.paused
+                        or sabnzbd.Downloader.shutdown
+                        or sabnzbd.Downloader.paused_for_postproc
+                    )
+                ):
+                    # Prepare the next request
+                    if not self.next_request and (article := server.get_article()):
+                        self.next_request = self.body(article)
```
Safihre, line 345: Why do we "buffer" the next request and not only prepare when needed?

sabnzbd/nzbstuff.py
mnightingale: This is covered in detail in the implementation specifics; it’s about optimising how many operations list removal and search do. The time complexity related changes made quite a large difference even before I made any pipelining changes. Popping from the head of a list means a lot of memmoves and we often pop from or near the head.

sabnzbd/newswrapper.py
mnightingale: For clarity the risk of sendall blocking, it is the risk of deadlock because we are waiting for a write that can't proceed until we read. However given pipelining is low I guess it shouldn't happen.

sabnzbd/newswrapper.py
mnightingale: This is to avoid a slow lookup for the next request, part of reducing the latency impact was also reducing our own delay for knowing what request to send next.

server.get_article() prefetch is low (20) so it’s constantly doing a more expensive lookup/loop for more articles to assign.

My aim was to make the read/write loop as quick as possible and the best way I found to achieve that was to have them prepare work before needed.

At one point I had it preparing several requests, but I reduced it to just one as I found that adequate. I did try removing it, but it made it slower. I expect because of DOWNLOADER_LOCK so you can have multiple download threads, but they are all synchronising with each other waiting for DOWNLOADER_LOCK and by buffering it there is no waiting when they actually need it. Without it 20 hit the prefetch so probably fast (except the lock) but when you are dealing with fast (1-10G) connections that could be 100 to 1000+ requests per second.

A potential improvement could be to separate populating server.article_queue and server.get_article(), then each connection wouldn’t need to buffer them at all. It was outside of scope, but I do think it’s a potential area for improvement. The nzbqueue iterates per server over articles potentially already assigned to pull out up to 20 to assign to ‘this’ server. It’s complicated because they need to be able to reorder, but I think it may be an area to look at in the future. I'm wondering if there is a way to have one orchestrator that assigns work to whoever needs it, rather than each server iterating the queue, but I'm not sure.

sabnzbd/newswrapper.py
Safihre: We could test that with a crazy high pipeline setting?

sabnzbd/nzbstuff.py
Safihre: Yeah don't think it's clear. It doesn't make sense to me to use a file-ish object in a with and then it turns out to be a lock and not for example a write to that file.

sabnzbd/nzbstuff.py
Safihre: We should add a comment about that.
I know it's in the PR now, but that will get lost over time.

sabnzbd/nzbstuff.py
Safihre: I see that the explanation is in the PR. But what actually operations could collide and what would be the problem if they do collide? 
Sometimes it's faster to just not lock and deal gracefully with any problems, if they only rarely happen.

sabnzbd/newswrapper.py
Safihre: Sounds indeed like we could prepare the commands in advance? Since we already get all the articles anyway for the server.

I thought you also made it a generator at some point, but it doesn't do that anymore?

sabnzbd/nzbstuff.py
mnightingale: Primarily thread safety and reliable state when manipulating NzbFile.articles, but even those that are sort of safe due to the GIL should be improved to not rely on that for a future without the GIL – I figure best to just iterate and fix such issues before it becomes an issue.

I believe it was primarily the following access to articles:
- NzbFile.get_articles – from NzbObject.get_articles => NzbQueue.get_articles => Downloader.get_article with DOWNLOADER_LOCK
- NzbFile.remove_article – from NzbObject.remove_article with NZO_LOCK
- NzbFile.reset_all_try_lists – with TRYLIST_LOCK
- NzbQueue.stop_idle_jobs – this was unclear to me, is the existence of this method likely due to threading issues? I figured ‘normal’ jobs would know when they are finished by the downloader/assembler logic, rather than a check every 10 seconds? The lock does prevent having to allocate a shallow copy of the articles, but it’s very rare anyway.

---

NzbFile.get_articles and NzbFile.remove_article are the primary ones I was interested in, they were already locking, but different manipulations or access uses different locks which renders the existing locking mostly useless. 

Shifting the locking to the owning object (NZF) makes this safer and locks at an appropriate level. Longterm I may look at the use of those global shared locks, if they can be improved or for example lock for a shorter amount of time rather than for whole calls.

I’m not 100% if we could remove the lock changes. I’d encourage not to do so, the locks are cheap to create, only per file, have low contention, and also make the methods thread safe at a lower level i.e. they don't rely on a decorator/lock much higher in the chain.

sabnzbd/newswrapper.py
mnightingale: Preparing them in advance is what I'm doing no? but if you mean it should be at the server or downloader.run level then sure.

Currently each Server prefetches 20 requests into server.article_queue but access has DOWNLOADER_LOCK, so there is contention of even getting them when they are prefetched. This is why at one point I had a deque of several unsent requests at the newswrapper level prefetching but immediately available to send without locks or delay, but I compromised on a single request because it seemed to perform well enough.

If there was no per-newswrapper buffering of the next request the process is something like:

1. Acquire DOWNLOADER_LOCK
2. Try to pop from server.article_queue and return
3. If not do a comparatively slower iteration of each nzo, nzf, until 20 articles are found - not as expensive as it was previously since it populates an existing deque
    1. Also does some filtering of those already selected articles if the are out of retention - not ideal, but anyway the point is there is relatively a large amount of work involved in getting the next article(s)
5. Pop and return from server.article_queue

As an alternative and remove the newswrapper buffering it may be possible to remove the lock from get_article, try...expect from article_queue and only lock the `sabnzbd.NzbQueue.get_articles` part `with DOWNLOADER_LOCK` when required. The problem is it has to try...expect because we couldn't check the length of article_queue without the lock.

My preferred approach is probably to change the prefetch a little, have it part of the downloader loop or a per server thread that constantly loads article_queue with requests.

I also suspect the prefetch of articles assigned to a server is quite low, it could be related to number of connections/pipelining or something i.e always have connections*2 next requests available. But it's tricky because you dont wan't two 0 priority servers were one is slow and the other fast, and the fast one end up with nothing to do. Anyway certainly an area for some thought, I have experimented with making it a generator and yielding requests but I don't think it was a noticable improvement.

PR #3201, opened 2025-11-20
Diskspeed improvements

mnightingale:
Following discussion in #3197

- Allocating 480 MB anyway so slice a 256 MB buffer instead
- Memory use: 282 MB vs 298 MB, or 154 MB with 8 MB base size
- Explicitly clears buffer when function finishes, I’m not sure it does much without gc.collect() though
- Can pass maxtime, passes, and chunksize arguments to the function

Results wise I'm seeing little difference looping the 16MB and 8MB except on a ZFS HDD pool.

The actual measurement is unchanged, just made more adaptable.

As a further improvement to mitigate the memory usage increase I suggest getting the total physical memory of the system and adjust chunksize as appropriate.

### DIFF
--- a/sabnzbd/utils/diskspeed.py
+++ b/sabnzbd/utils/diskspeed.py
@@ -7,22 +7,20 @@ import sys
 import logging
 import time
 
-BUFFERSIZE = 16 * 1024 * 1024
 
-
-def diskspeedmeasure(dirname: str) -> float:
+def diskspeedmeasure(dirname: str, maxtime: float = 1.0, passes: int = 4, chunksize: int = 16 * 1024 * 1024) -> float:
     """Returns writing speed to my_dirname in MB/s
     method: keep writing a file, until certain time is passed.
     Then divide bytes written by time passed
     In case of problems (ie non-writable dir or file), return 0.0
     """
-    maxtime = 1  # sec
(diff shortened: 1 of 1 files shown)


Safihre:
Hmm, this allocates the memory all at once?
The current version the buffer is re-created every loop and because of Python's reference counting the previous buffer is removed before new one is created.
The idea being that on a slow system we'll never get to that 500MB because it's slow anyway.
Allocating 500MB right away seems a bit too risky on low-mem devices?

Again, this test is mostly to separate <10MB/s from 100+ MB/s, not to be an actual accurate testing suite.

mnightingale:
I see you're expecting low memory devices to also have low disk speed, that does alleviate my concerns.

I few other suggestions:

```py
# No os.close needed
with os.fdopen(fp_testfile, 'wb', buffering=0) as f:
    # ...
    f.write(...)
    os.fsync(f.fileno())
```

os.fsync() could be called once after the while loop, still measure its time. That's closer to what assembler does, non-sync writes but flush/sync on close and will increase reported speeds quite a bit. edit: don't think it'll make any difference on platforms supporting O_SYNC.

Also I haven't tested, but what happens if disk space runs out, I think an exception will be thrown and the os.remove skipped?

PR #3214, opened 2025-12-04
update sfv help text

### LINKED ISSUE #3213: update explain-sfv_check text
for:
config > switches - Enable SFV-based checks

had a discord user complained about the in gui text:
> Do an extra verification based on SFV files

while wiki help is way more accurate of what it is:
> If no par2 files are available, .sfv files (if present) will be used to check files

i know this is been a long 15+ year old text string:
https://github.com/sabnzbd/sabnzbd/blob/master/sabnzbd/skintext.py#L445

any qualms about us just updating gui text to the help string

thezoggy:
closed #3213

### DIFF
--- a/po/main/SABnzbd.pot
+++ b/po/main/SABnzbd.pot
@@ -3127,7 +3127,7 @@ msgid "Enable SFV-based checks"
 msgstr ""
 
 #: sabnzbd/skintext.py
-msgid "Do an extra verification based on SFV files."
+msgid "If no par2 files are available, .sfv files (if present) will be used to check files."
 msgstr ""
 
 #: sabnzbd/skintext.py


po/main/SABnzbd.pot
```diff
@@ -3127,7 +3127,7 @@ msgid "Enable SFV-based checks"
 msgstr ""
 
 #: sabnzbd/skintext.py
-msgid "Do an extra verification based on SFV files."
+msgid "If no par2 files are available, .sfv files (if present) will be used to check files."
```
Safihre, line 3130: Nitpicking, but shouldn't we use "verified" instead of check? To keep terminology the samd

thezoggy: I just used what we have on our wiki. Feel free to change it (in both) places if that makes the most sense

Safihre: 👍

PR #3215, opened 2025-12-04
Prevent filepath creation before first article is processed

### LINKED ISSUE #3212: Files written to disk before par2-dobfuscate result in failed unpack
### SABnzbd version

4.6

### Operating system

All

### Using Docker image

None

### Description

See https://forums.sabnzbd.org/viewtopic.php?t=27128

Seems the problem is:

- The NZB has the "right" filenames, but the par2 has the obfuscated names just like yEnc
- part01 downloaded before the par2 is parsed, but the name from yEnc is ignored because its obfuscated
- part01 is sent to Direct Unpack, but then it gets stuck because part02 now has a different name than par02
- During post processing the stuck Direct Unpack is cancelled correctly
- Results in deleting the whole `__UNPACK__` folder _and_ clears the `unpack_dir_info`
- Seems unrar then in the next runs doesn't unpack anything because the destination `__UNPACK__` folder doesn't exist

Not sure of best solution here...

Safihre:
@mnightingale I think this fixed #3212.
But I worry about edge-cases I haven't considered?
I asked AI but it started hallucinating edge cases 😂

### DIFF
--- a/sabnzbd/nzbstuff.py
+++ b/sabnzbd/nzbstuff.py
@@ -447,9 +447,24 @@ class NzbFile(TryList):
             art.reset_try_list()
         self.reset_try_list()
 
+    def first_article_processed(self) -> bool:
+        """Check if the first article has been processed.
+        This ensures we have attempted to extract md5of16k and filename information
+        before creating the filepath.
+        """
+        # The first article of decodetable is always the lowest
+        first_article = self.decodetable[0]
+        # If it's still in nzo.first_articles, it hasn't been processed yet
+        return first_article not in self.nzo.first_articles
+
     def prepare_filepath(self):
         """Do all checks before making the final path"""
         if not self.filepath:
+            # Wait for the first article to be processed so we can get md5of16k
+            # and proper filename before creating the filepath
+            if not self.first_article_processed():
+                return None
+
             self.nzo.verify_nzf_filename(self)
             filename = sanitize_filename(self.filename)
             self.filepath = get_unique_filename(os.path.join(self.nzo.download_path, filename))


mnightingale:
For some edge-cases how about:

- Small files and queued in the assembler with file_done but haven’t processed the par2.
With these changes won’t it just get missed for assembly or will something else notice and requeue it?
  - Maybe needs some logic elsewhere, something like ‘when par2 gets processed => are there any completed files in the nzo that need to be sent to the assembler’
- ~~Doesn’t happen often but what happens if the NZB has no par2 files or they all fail to download, it’ll never be`nzf.filename_checked`.~~ oops if there is no par2 there is nothing to wait for

mnightingale:
Had a look with a debugger, the logic seems OK - don't assemble until the first article per file is done.

But I don't think it resovles the underlying issue, the logic doesn't prevent starting assembly until par2 has been processed?

https://github.com/sabnzbd/sabnzbd/blob/1b04e07d4086132b0bf3eaabbb535f64626bf44d/sabnzbd/nzbqueue.py#L735-L739

Here, single article files still go to the assembler, would pass this new check and be assembled before the par2 is processed.

The order the first pieces or any for that matter come back is undetermined, it's possible to get pieces of the .rar, or all of a .rar before the par2 has been processed.

---

I have tested this theory, if I modify one of the test downlaods, changing the NZB filenames and made an NNTP server that delays requests for the par2 articles.

Assembly is still happening and it logs "Detected filename from yenc or uu" not "Detected filename based on par2".

So although this helps, I still think it's a race to complete enough articles and start assembling.

Safihre:
> Small files and queued in the assembler with file_done but haven’t processed the par2.
With these changes won’t it just get missed for assembly or will something else notice and requeue it?

But this can't happen right? Because if it's small file it still always has a first article, and that will always be decoded at some point.

Yes, this isn't perfect at all. But dealing with "keep things from assembly before par2 is parsed" is just too much of a pain.

mnightingale:
I was thinking when an NZB has:
- something.par2: 1 article
- misnamed.rar: 100 articles
- misnamed.r00: 1 article

The `.r00` or any other small files could be file_done and have the first article before the par2 is processed, so they may be written without checking the filename.
Potentially par2 could be missing from the first server and have to be requeued on another.

I agree keeping track of things to assemble once par2 is proceseed is a pain, the PR does greatly reduce the chance of it happening.
It's why on the issue I suggested just trying to handle the direct unpack failues, checking if anything needs renaming now that's safe to do so:
- Have any of the completed files not been filename checked, do they need renames, restart direct unpack.
- Or just do the same thing in postproc, direct unpack failed, check for missed renames, recreate \_\_UNPACK\_\_ and unpack in pp, loses the direct unpack but it'll be rare.

Safihre:
This fix will cover a few more cases :)

PR #3217, opened 2025-12-05
Add option to keep NZB name prefix when processing multiple NZBs

Yaknar:
# Problem

When multiple NZBs are post processed via nzb_redirect, SABnzbd discards the previous nzbname. As a result, related jobs lose their shared name prefix, making it harder to identify or group downloads originating from the same batch.

# Solution

This change introduces the `multi_nzb_keep_prefix` Special Setting. When enabled, SABnzbd will preserve the supplied nzbname and append the individual NZB filename to it

This preserves backward compatibility while enabling improved job naming
for users working with multi-NZB redirects.

It helps me a lot to manage my downloads via the API.

Description for Wiki Page:

        <tr>
            <td><code>multi_nzb_keep_prefix</code></td>
            <td><span class="glyphicon glyphicon-check"></span></td>
            <td>During post-processing, during processing of nzb files new generated download jobs will well keep name of previous download job as prefix</td>
        </tr>


thezoggy:
can you share a real world example of what you are trying to fix?

Yaknar:
I would know which packages belong together and my software populates SABnzbd, and I would know that the new download jobs belong to the previous download:

image

Some users might find it quite interesting to be able to better understand the connection between the downloads.

thezoggy:
but if your software is queuing stuff to sab, why not just prepend it there when you add it to sab>?

Yaknar:
Yes, I do that, but with NZBs that only link to multiple NZBs or RAR archives containing multiple NZBs, where sabnzbd then creates new download jobs out of the new NZBs, these IDs than no longer exist in the new downloads jobs. 
Instead, they consist of the filename of the NZB linked in the first NZB.

And the pull request fixes this by adding (if enabled) a prefix from the initial job when it creates the new download jobs.

thezoggy:
feel free to add a test case with your change and we can review.

Yaknar:
>so from what vague info you have shared i assume your trying to upload a season pack but dont want sab to actually have that name of the season pack as the job but instead use each individual episode within the season pack? 

Yes, it occurs with Season Packs, but it's slightly different.

If I for example add Season1.nzb as Season1 which contains these Files:
Episode1.nzb
Episode2.nzb

Than in the postprocessing the Method nzb_redirect takes each individual NZB File an create a new Download Job for each:
Episode1.nzb
Episode2.nzb

These jobs download the corresponding files. 

Season1.nzb is than empty. 

To Track which Job belongs to the inital Job my change would than add these embedded NZBs as:
Season1_Episode1.nzb
Season1_Episode2.nzb

If the Setting multi_nzb_keep_prefix is Enabled, in default this behaviour won't change. 

>as i am just concern that this would break apps that expected the name to be what was sent. 

The name of the embedded NZBs was never set bei any Applications, and mine needs the prefex to keep track of the downloads. 
As default multi_nzb_keep_prefix is disabled, so the User has to explicity change this to break something, but I don't think it does. 

>but anyways feel free to add a test case with your change and we can review
Okay, I'll add test cases on Monday.

Safihre:
I think this can be default behaviour. No need for special setting.
In case of just single nzb inside it just keeps the name of the original job.

Yaknar:
@thezoggy Tests have been added.

Safihre:
I implemented the changes a bit different, but with the same effect.

Yaknar:
@Safihre Cool, but there's one thing I'd like to point out.
The `get_filename` function needs to be called inside the `for` loop, otherwise all newly created NZBs will have the same filename, because the `nzb_file` variable hasn't been populated.

Safihre:
Damn. You are right. My IDE even was trying to tell me that!

PR #3221, opened 2025-12-08
Split nzbstuff into separate files for Article, NzbFile and NzbObject

Safihre:
@mnightingale What do you think?


mnightingale:
How do you feel about grouping them together in a module?

I don't have a problem with nzbstuff as a module, just how hard the 2000 lines were to mentally keep track of.

- sabnzbd
  - nzb (or nzbstuff)
    - \_\_init\_\_.py
    - nzbarticle.py
    - nzbfile.py
    - ...
 - ...

\_\_init\_\_.py:

```py
from nzbarticle import Article
from nzbobject import NzbObject
# ...
```

Then the imports everywhere are just `from sabnzbd.nzb import Article` the same as when it was nzbstuff.

This seems a common approach looking at few large Python projects, keeps them organised and grouped by related domain while making them more manageable.

Safihre:
Sure. I'll do that.
But I don't really want to do more refactoring of other files because we also loose 20 years of easy git blame history with changes like this.
I agree for nzbstuff it was time. Just like filesystem split from misc a few years ago.

PR #3229, opened 2025-12-13
Make behaviour after reset more robust

### LINKED ISSUE #3226: downloader error - pop on empty _response_queue
### SABnzbd version

4.6a2

### Operating system

Debian 12

### Using Docker image

none

### Description

from discord user, requested debug logs but what they have shared so far
```
ERROR a few seconds ago Fatal error in Downloader
  File "/opt/sabnzbd/sabnzbd/downloader.py", line 734, in process_nw_worker
    self.process_nw(*nw_queue.get())
  File "/opt/sabnzbd/sabnzbd/downloader.py", line 744, in process_nw
    self.process_nw_read(nw)
  File "/opt/sabnzbd/sabnzbd/downloader.py", line 754, in process_nw_read
    n, bytes_pending = nw.read(nbytes=bytes_pending)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/sabnzbd/sabnzbd/newswrapper.py", line 312, in read
    article = self._response_queue.popleft()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: pop from an empty deque
```

mnightingale:
Fixes #3226

This adds checks in several places that the connection has not been reset since the select operation was enqueued. It also checks before writing and checks if _response_queue is empty before popleft.
I called it `reset_gen` rather than `resets` or similar because I wanted to avoid indicating that it's allowed to reset it to 0.

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -74,13 +74,14 @@ class NewsWrapper:
         "_response_queue",
         "selector_events",
         "lock",
-        "reset_gen",
+        "generation",
     )
 
-    def __init__(self, server, thrdnum, block=False):
+    def __init__(self, server: sabnzbd.downloader.Server, thrdnum: int, block: bool = False, generation: int = 0):
         self.server: sabnzbd.downloader.Server = server
         self.thrdnum: int = thrdnum
         self.blocking: bool = block
+        self.generation: int = generation
 
         self.timeout: Optional[float] = None
 
@@ -105,9 +106,6 @@ class NewsWrapper:
         self._response_queue: deque[Optional[sabnzbd.nzbarticle.Article]] = deque()
         self.selector_events = 0
         self.lock: threading.Lock = threading.Lock()
-        # Preserve reset generation across resets
-        if not hasattr(self, "reset_gen"):
-            self.reset_gen: int = 0
 
     @property
     def article(self) -> Optional["sabnzbd.nzbarticle.Article"]:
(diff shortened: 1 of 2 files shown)


sabnzbd/newswrapper.py
```diff
@@ -104,6 +105,9 @@ def __init__(self, server, thrdnum, block=False):
         self._response_queue: deque[Optional[sabnzbd.nzbarticle.Article]] = deque()
         self.selector_events = 0
         self.lock: threading.Lock = threading.Lock()
+        # Preserve reset generation across resets
```
Safihre: Why not in the init?

Safihre:
Hmm why an int and not a flag?
We don't need to preserve news wrappers of course, if we should revise that, we should.

mnightingale:
reset_gen could be passed via an optional argument in the \_\_init\_\_ signature if you’d prefer, the main reason I didn’t is to avoid callers thinking it’s something they should be controlling.

I’m OK with the reuse mechanics, replacing it with completely new instances/clones would likely introduce new bugs and it would still need to track resets in the same way or at least have nw.dead = True, idle/busy queue logic, avoid registering if dead, I think it ends up just makes it more complicated.

An int lets me pair a nw/fd registered for read/write and what has been selected for process_nw_queue and if the same state exists now as when it was enqueued. I think this is the most reliable way to avoid read/writes after reset and I additionally guarded before the popleft although because of the generation tracking I don't think it can happen anymore.

It guards against nw_resets before work starts (I don’t think that happens currently), resets during the read operations which may try to read the socket more than once, and against writing if the read operations caused a reset. I don’t think a simple “was reset” flag is sufficient, we need to reliably track that the select event was produced for the same connection state the newswrapper is currently in.

Possibly instead of reinitialising with \_\_init\_\_ there could be a reset_state method, which resets as init but increments reset_gen, that might be a little clearer than reusing init for multiple paths? But maybe confusing having hard_reset and reset_state.

Safihre:
Thanks for explaining.
Just want to avoid the hasattr pattern, that seems like bad object usage since the attribute should at least be specified when object is created.

mnightingale:
Because it's now passed via init there is a tiny window in hard_reset where generation could be stale, I don't think it matters because I don't think there are concurrent resets but it could have locked while reinitialising:

```py
with self.lock:
    # Reset all variables (including the NNTP connection)
    self.__init__(self.server, self.thrdnum, generation=self.generation + 1)
```

PR #3234, opened 2025-12-18
Increase max pipelining

mnightingale:
Since there haven't been any issues reported so far can we increase a little, from testing it's only useful for quite high latency.

Someone on Discord noticed the wiki already says up to 20 is allowed so that would be correct now.


Safihre:
Yeah AI wrote that part of the wiki 😂

Safihre:
I changed the wording and limits a bit. No longer suggesting 20 but up to 10.
They can always try 20 if they want of course.

PR #3236, opened 2025-12-20
Implement direct write

mnightingale:
I am not expecting this PR to be merged for the upcoming release, happy for you to ignore it until January or whenever.

## What

This PR consists of a few improvements to the assembler after analysing the behaviour and testing out a range of options.

I primarily noticed the performance problems at around 10-20 Gbps, however they are easily reproducible when download speed exceeds disk speed, which could be anything from a poor configuration, slow disk or 10 Gbps connection into a ~500MB/s SATA SSD.

1. **Direct write and sparse files** - it revisits a past attempt to bring a feature heavily influenced by "the other client" to sabnzbd, but in a more robust way than previously.
   - Efficiently write data at the correct offsets without wasting disk space and opens possibilities for retrying missing pieces
   - Attempts to allocate the expected file size, but it does not matter if this is wrong (some yenc data is), it can help avoid creating fragmented files.
2. **Assembler performance improvements**
   - When direct write is enabled prefer writing whole files
   - When multiple writes to a file are required keep a record of progress
   - Avoid enqueuing the same file multiple times
3. **Article Cache**
   - When the cache is almost full and direct write is enabled force a flush to the destination files at correct offsets, this has the outcome of almost never exceeding the cache limit or writing to temporary files.

### Direct Write and Sparse Files

Writing data at the correct offsets has the advantage that writes no longer strictly need to be in order like append mode requires, however it is preferred but probably matters less with SSDs.

- par2 verify/repair can be a little faster because the rolling CRC option is no longer required (I have not tested the difference, I doubt it is much these days)
- missing articles can be retried
- exotic repairs outside of sabnzbd can be more successful, p2p, or combining multiple failed downloads
- 0 article cache size benefits from usually writing directly to the destination rather than **write** (temporary) => **read** (assembler) => **write** (destination)

On startup the incomplete directory is tested for direct write support, and if the attempt to make a file sparse fails during assembly it switches back to append mode.

Changing the download_dir should probably also do this, but I wasn't sure if it should warn the user in the same way.

#### Responsibility

Previously nzbqueue would decide whether to tell the assembler about a completed article or not based on if the file is done, how many articles remain and if it can be written to disk.

The logic for this has moved to the assembler because it's the more appropriate place and the decision makes use of more assembler properties. For non-directwrites the logic remains as is but adds a several other checks:

- file_done => always enqueue (as before)
- or not direct_write and "not already queued" and "the existing logic"
- or direct_write and force and "not already force queued (no lock)"

`force` is used if the cache is full and instructs the assembler to write everything it can for each file, ignoring gaps.

#### Problems with `assembler_write_trigger`

The existing `assembler_write_trigger` is equivalent to the number of articles expected to use 5% of the cache size. Then trigger a write every number of trigger articles.

| Cache Size | Trigger (Articles) | Approximate Trigger (MB) |
| ---------- | ------------------ | ------------------------ |
| 512 M      | 36                 | 27 MB                    |
| 1 G        | 72                 | 54 MB                    |
| 2 G        | 144                | 108 MB                   |
| 4 G        | 287                | 215.3 MB                 |

When modern systems with fast SSDs are used this is quite inefficient due to the constant context switching, closing, and reopening of files and results it low utilisation of the chosen cache size.  
The only time most users would see cache usage much above this low level is if they are disk IO limited, hitting the trigger amount frequently, or downloading so fast they are paging to temporary files.

Another aspect is there is currently no logic to avoid enqueuing an nzf that is already queued, so especially at low cache sizes a 500 MB file might be queued 10 times, that's 10 file opens (relatively slow), 10 iterations of the whole decodetable (many of which won't require any writes).

A much better approach is to attempt to cache large chunks of data (ideally whole files) in memory and write them all at once in large sequential operations. Put simply asking an SSD to write ~200 MB at a time is slower than asking it to write 1GB once and the constant switching from downloader to assembler threads is also more work and slower.

The solution I have implemented is that when direct write is enabled, it prefers writing whole files but can be triggered by cache utilisation (90%) to force write to file and when the cache is full or not used write directly to the destination file.

What about HDD users you ask, hopefully rare these days but they should probably either not use direct write mode and get the previous behaviour or use a smaller cache size to force flushing more often than just whole files. It's also possibly HDD performance is perfectly fine with direct write but I have not tested it.

#### queue_level / check_assembler_levels

I do not think this ever really measured what it was hoped it did, and the delay was often self-inflicted by repeatably queuing the same item. I haven't made any changes to this but I think a more appropriate place to `delayed_assembler += 1` would be if the cache is in use but an article bypassed it and wrote directly to disk (or temporary files when not using direct write) but it also needs to account for first articles going to temporary files. So many just count how many times that happens per second i.e. record timestamp, and do not increment again until timestamp+1second.

#### Low level os.open and write_at_offset/pwrite

To cut out as much Python overhead as possible, I have switched to lower level os.open calls, this is similar to the approach taken in the diskspeed test.

For linux/macOS uses pwrite to write data at an offset in a single syscall, on Windows it must lock and call os.lseek and os.write.
It's possible to use Windows APIs to get the same behaviour but it is not worth the effort.

#### nzf.assembler_next_index

Keep track of which was the last article written, which can then be resumed from on future assemble attempts.  
This has advantages for both write modes.  
This state is not important to preserve so is not saved.

#### AssemblerTask

Added an AssemblerTask(typing.NamedTuple), the tuple was unwieldy plus I needed a more fields.

### Article Cache

The existing behaviour is to fill the cache, then write each article to a temporary file which only exacerbates the problem because the assembler must read from disk before writing to the destination, all the while even more articles were being temporarily written to disk.

I have made two improvements to this behaviour when direct write is used:

- If the cache is full, if possible write directly to the destination file and give the cache a chance to recover.
- The article cache is now a thread which monitors its utilisation, if it reaches 90% it triggers a forced assemble, attempting to write all of its contents to the destination files. **- it doesn't need to be a thread I suppose, reserve_space could trigger it** 
- If the downloader is paused, that also triggers a flush to disk rather than leaving articles in memory indefinitely.

#### Improve performance, locking, and thread safety

- **reserve_space** - return a bool for whether the space could be reserved, the previous behaviour would always reserve space even if it exceeded the limit, then check if limit was exceeded (without a lock), and free the space before flushing to a temporary file. The logic around this is now much simpler.
- **NZO_LOCK** - This was the only lock I added because I was seeing concurrent modifications to nzo.saved_articles while purging, ideally this would be a nzo level lock. Alternatively purge_articles could shallow copy the set :disappointed:
- **\_\_flush_article_to_disk** - throttles download if cache is exhausted since it is called from the downloader thread, if direct write is available it writes to the correct offset in file with a fallback to the existing temporary files when required.

I note the cache is purposely using try/except instead of locks because most will be successful. I doubt locks would be meaningfully slower, but I didn't bother changing it, I expect there are a few threading issues in there due to the lack of locks but I have tried to address only the one I was seeing. It does mean the article cache thread had to shallow copy the table, but it should happen infrequently anyway.

### Miscellaneous other changes:

- **assembler_write_trigger** - moved to the assembler because it makes more sense there, it is not used with direct write mode
- **cache_info** - this namedtuple was mostly unused except logging but the cache_size property was I think incorrectly mapped to the original cache_size, maybe it needs both instead?
- **cache_upper_limit** - considered increasing, but decided against it for now. However, I did add `misc.get_memory` and fixed the linux calculation within LXD (it reported the host amount)
- **update_crc32** - locks because multiple threads could write and update the value - if cache is full, uses nzf.lock so low/no contention since we only write one file at a time usually
- **filesystem.load_data** - adds `mutable: bool = False` because pwrite requires a writable buffer so it reads into a bytearray rather than bytes then requiring a copy.

---

I believe a previous problem with sparse files was unrar picking up the next file before it was finished because the size matched what it expected. I do not seem to be having a problem with that I think there have been changes in sabnzbd to make it always prompt?


sabnzbd/utils/sparse.py
```diff
@@ -0,0 +1,155 @@
...
+        raise ValueError("length must be non-negative")
+
+    # Get file handle from a file descriptor
+    handle = msvcrt.get_osfhandle(fd)
+    if handle == -1 or handle is None:
+        raise OSError("Invalid file descriptor / handle")
+
+    # Try to make sparse
+    bytes_returned = wintypes.DWORD(0)
+    if not kernel32.DeviceIoControl(
```
Safihre: Why not use win32 for this? We won't be getting rid of it for a long time.
https://timgolden.me.uk/pywin32-docs/win32file__DeviceIoControl_meth.html

sabnzbd/utils/sparse.py
```diff
@@ -0,0 +1,155 @@
...
+        return False
+
+    # Save current file pointer
+    cur_pos = ctypes.c_longlong(0)
+    if not kernel32.SetFilePointerEx(wintypes.HANDLE(handle), 0, ctypes.byref(cur_pos), FILE_CURRENT):
+        _raise_last_win_error("SetFilePointerEx (get current position) failed")
+
+    # Move to (END + length)
+    li_size = ctypes.c_longlong(length)
+    if not kernel32.SetFilePointerEx(wintypes.HANDLE(handle), li_size, None, FILE_END):
```
Safihre: Same here https://timgolden.me.uk/pywin32-docs/win32file__SetFilePointer_meth.html

sabnzbd/utils/sparse.py
Safihre: Or what is against using our sabctools version on Windows?

mnightingale: It didn’t occur to me that pywin32 could be used, I didn’t use sabctools because the low-level changes to os.open mean it needs to receive a file descriptor and it's only a few Windows api calls.

I don’t really mind which it’s in but I suppose since sabctools is 90% of the way there I can adapt it, I'll probably just make it accept files or fds.

mnightingale: So it wasn't too difficult to adapt the sabctools code.

What do you suggest doing with utils/sparse.py, I still need the is_sparse_supported and is_sparse methods somewhere.
I could make them part of sabctools, just keeping them in SAB I'm not sure where they would go maybe filesystem?

Safihre: Great!
Filesystem.

sabnzbd/interface.py
mnightingale, line 1: What do you think about this, could it go in advanced settings on the folders tab it does seem the type of option a user might want to access.

I made it default to enabled because that's what the other client does, most users I'd expect to be using SSDs so fragmentation doesn't matter.

mnightingale:
Further to our sabctools discussion, I would quite like to use `os.posix_fallocate(fd, 0, article.file_size)` (Linux only) for filesystems that support it efficiently - there is some discussion on https://github.com/sabnzbd/sabnzbd/issues/2169#issuecomment-1119111265

How do you feel about me adding a separate special for that? disabled by default.

sabnzbd/articlecache.py
```diff
@@ -22,76 +22,142 @@
...
+            if self.shutdown:
+                break
+
+            # Timeout so is reach when paused and no further articles arrive
+            if not self.__needs_flush():
+                continue
+
+            self.__next_flush = time.monotonic() + _SECONDS_BETWEEN_FLUSHES
+            for article in self.__article_table.copy():
+                if article.nzf.type == "yenc" and article.nzf not in nzfs and article.nzf.prepare_filepath():
```
Safihre: Is yenc safe enough? Isn't it only if the offset is defined that we can trigger this? Or am I maybe misunderstood what it's checking here?

sabnzbd/interface.py
mnightingale: I've moved it to the Tuning section, since it's related to the cache limit.

I struggled to come up with a description that explained it sufficiently but wasn't too long, I expect you'll trim it down and/or link to a wiki page.
Maybe it should have a different name, it's not really direct write since the cache is required.

I think the config page needs to make it clear that the cache is required, I wasn't sure how to disable the input if the cache is not set since it can be several values (-1, 0, "", 512M, etc.) or if you'd prefer to communicate that differently.

sabnzbd/articlecache.py
mnightingale: The decoder currently normalise data_begin to 0-indexed, and doesn't include data_end but I could change that.

Currently from `=ypart begin=15360001 end=15744000` the response gives us `part_begin = 15360000` and `part_size = 384000` but it's also provided in =yend anyway.
I could make it return `part_begin=15360001` `part_end=15744000` and `part_size = 384000`

`=ypart begin=1 end=100000` would give us `part_begin=0` so no way of determining if yenc was missing the offset or if it was non-multipart.

Then we can validate the offsets `not part_begin or not part_end` and revert to append mode.

Probably also needs some guards that offset is less than article.file_size and before any sparse call.

Safihre: If begin and size are defined, that would be same right? So er already have the check

Safihre:
> Further to our sabctools discussion, I would quite like to use `os.posix_fallocate(fd, 0, article.file_size)` (Linux only) for filesystems that support it efficiently - there is some discussion on [#2169 (comment)](https://github.com/sabnzbd/sabnzbd/issues/2169#issuecomment-1119111265)
> 
> How do you feel about me adding a separate special for that? disabled by default.

Why disabled by default? Doesn't nzbget also enable it by default?

sabnzbd/articlecache.py
mnightingale: Currently missing things like the size > 10 MB or huge part_begin which could cause 1 TB files to be created, although on most systems they would be sparse.

With the sabctools changes, it just needs to check `article.nzf.type == "yenc" and article.data_size` because data_size would be reset to 0 with invalid values.

Safihre: Indeed, so wouldn't be data size check be enough? Not needed yenc check.

mnightingale:
> > How do you feel about me adding a separate special for that? disabled by default.
> 
> Why disabled by default? Doesn't nzbget also enable it by default?

I would make direct write enabled by default.

This was about a seperate switch for `os.posix_fallocate` call it was one of the changes in the other sabctools PR but Python also has a method to access it (Unix only), for filesystems which support it efficiently it would be preferred rather than the `ftruncate` sabctools uses (basically creates a sparse file).

ftruncate only sets the logical file size
posix_fallocate sets both logical and reserves the disk space, it may allow the filesystem to make better choices for writes.

edit: nzbget doesn't use posix_fallocate, it just uses truncate with manual fallback to writing zeros. currently sabctools just does the truncate call. I wouldn't make posix_fallocate the default because some filesystems may behave poorly when using it.

sabnzbd/articlecache.py
mnightingale: With the extra validation and cleanup we can just use data_size because it's calculated from `end - begin - 1`, I think it still needs the changes and extra validation.

If you mean the `type == "yenc"` part that's just because I can only force flush articles to file if we know the offsets, UU ones can't be skipped because we don't have the offset metadata. So just forcing the yenc ones is sufficient.

sabnzbd/filesystem.py
```diff
@@ -1369,6 +1384,50 @@ def pathbrowser(path: str, show_hidden: bool = False, show_files: bool = False)
     return file_list
 
 
+if sys.platform == "linux":
+
+    def write_at_offset(fd: int, data: memoryview, offset: int) -> int:
+        return os.pwrite(fd, data, offset)
+
+elif sys.platform == "darwin":
```
Safihre: This whole block of code seems very non-SAB-style.
I would have expected to use 1 function with `if sabnzbd.WIN32` inside.
Or if it's really performance drive, not using `sys.platform` but still using `sabnzbd.WIN32` etc.

sabnzbd/interface.py
```diff
@@ -965,6 +966,7 @@ def saveSpecial(self, **kwargs):
     "username",
     "language",
     "cache_limit",
+    "direct_write",
```
Safihre: It's now in 2 spots :)

sabnzbd/filesystem.py
```diff
@@ -1369,6 +1384,50 @@ def pathbrowser(path: str, show_hidden: bool = False, show_files: bool = False)
...
+        # Make contiguous copy if needed
+        if not mv.contiguous:
+            mv = memoryview(bytearray(mv))
+
+        # Make writable copy if needed
+        if mv.readonly:
+            mv = memoryview(bytearray(mv))
+
+        buf = (ctypes.c_char * len(mv)).from_buffer(mv)
+        n = pwrite(fd, buf, len(mv), offset)
```
Safihre: Doesn't this also work on macOS? 
Since it mentions "Availability: Linux", and [the docs say](https://docs.python.org/3/library/intro.html#availability): "If not separately noted, all functions that claim “Availability: Unix” are supported on macOS, iOS and Android, all of which build on a Unix core."

sabnzbd/filesystem.py
```diff
@@ -1418,3 +1477,35 @@ def nzf_cmp_name(nzf1, nzf2):
     if m2 and m2.group(1) == ".rar":
         nzf2_name = nzf2_name.replace(".rar", ".r//")
     return sabnzbd.misc.cmp(nzf1_name, nzf2_name)
+
+
+def is_sparse(path: str) -> bool:
+    """Check if a path is a sparse file"""
+    info = os.stat(path)
+    if sys.platform == "win32":
```
Safihre: Use `sabnzbd.WINDOWS`

sabnzbd/filesystem.py
```diff
@@ -1418,3 +1477,35 @@ def nzf_cmp_name(nzf1, nzf2):
...
+
+def is_sparse_supported(check_dir: str) -> bool:
+    """Check if a directory supports sparse files"""
+    sparse_file = tempfile.NamedTemporaryFile(dir=check_dir, delete=False)
+    try:
+        sabctools.sparse(sparse_file.fileno(), 64)
+        sparse_file.close()
+        return is_sparse(sparse_file.name)
+    finally:
+        os.unlink(sparse_file.name)
```
Safihre: For code style: use `os.remove`, because we don't use `unlink` anywhere and they are the same.

sabnzbd/filesystem.py
mnightingale: I've no idea why I was having a problem calling that on macOS, seems to be working fine now it must have been caused by some other issue at the time.

sabnzbd/assembler.py
```diff
@@ -162,52 +226,113 @@ def diskspace_check(nzo: NzbObject, nzf: NzbFile):
...
 
         # Final steps
         if file_done:
             set_permissions(nzf.filepath)
             nzf.assembled = True
 
+    @staticmethod
+    def assemble_article(article: Article, data: bytearray) -> bool:
+        """Write a single article to disk"""
+        if not sabnzbd.cfg.direct_write.get() and article.can_direct_write:
```
Safihre: Make it an Assembler property? To prevent lookup every time.

sabnzbd/nzb/file.py
```diff
@@ -75,12 +75,13 @@ class NzbFile(TryList):
...
     # Pre-define attributes to save memory
-    __slots__ = NzbFileSaver + ("lock",)
+    __slots__ = NzbFileSaver + ("lock", "file_lock", "assembler_next_index")
 
     def __init__(self, date, subject, raw_article_db, file_bytes, nzo):
         """Setup object"""
         super().__init__()
-        self.lock = threading.RLock()
+        self.lock: threading.RLock = threading.RLock()
+        self.file_lock: threading.RLock = threading.RLock()
```
Safihre, line 84: Looking at the code, where can it happen that we write to the file twice? Only assembler can write and it's single-threaded?

mnightingale: Assembler and Downloader via ArticleCache.save_article if the cache is not used (disabled or full)
There's one assembler, but can be many downloader threads.

sabnzbd/assembler.py
mnightingale: I wasn't sure you'd go for that, so callbacks from the config like cache limit uses?

It's also used in articlecache, but less frequently so I think I'll just reorder it a little in there.

sabnzbd/articlecache.py
```diff
@@ -161,10 +221,27 @@ def purge_articles(self, articles: Collection[Article]):
...
+                    lambda: self.shutdown or self.__cache_size <= self.__cache_size_lower, timeout=2.0
+                )
+
+        nzf = article.nzf
+        # Direct write to destination
+        if (
+            sabnzbd.cfg.direct_write.get()
+            and self.__cache_limit
+            and article.can_direct_write
+            and sabnzbd.Assembler.assemble_article(article, data)
```
Safihre: I see now, missed it because it was part of the `if`.
Feels a bit "wrong" that data can be written from 2 spots, shouldn't always _just_ be the assembler?

sabnzbd/assembler.py
Safihre: I would say that we can also just make it require to restart if you change that setting.

sabnzbd/articlecache.py
mnightingale: I think it’s rare but unavoidable, this one-off direct write only occurs if the cache is being used and is full.

We can’t just store it in the cache anyway or queue article and bytearray in the assembler and go back to downloading, both would allow us to continue consuming memory and probably run out.

This approach has the advantage that the downloader thread is delayed while it writes so is acting as a hard limit and it’s not so different to what it does when not in direct write mode (sabnzbd.filesystem.save_data) both are synchronous operations.

thezoggy:
How resilient is direct write when users are doing things like putting their incomplete on a remote share (local or actually outside lan) and the file system or networking between them is flakey. Does it retry or just fallback to normal method?

mnightingale:
At startup it tests if sparse files are supported by the incomplete directory, if not direct_write is disabled and a helpful warning shown.

I’ve tried mounting a smb share on macOS, incomplete set to /Volumes/… and a rclone mount using FUSE over ssh, both are detected at startup.

I intend to make it also check the folder when config changed (if direct write is enabled), but I haven’t looked where to do that yet.

Support on obscure file systems probably varies, but with the startup and during assembly checks I think it’s a robust implementation that should will revert to the old method.

thezoggy:
> At startup it tests if sparse files are supported by the incomplete directory, if not direct_write is disabled and a helpful warning shown.
> 
> I’ve tried mounting a smb share on macOS, incomplete set to /Volumes/… and a rclone mount using FUSE over ssh, both are detected at startup.
> 
> I intend to make it also check the folder when config changed (if direct write is enabled), but I haven’t looked where to do that yet.
> 
> Support on obscure file systems probably varies, but with the startup and during assembly checks I think it’s a robust implementation that should will revert to the old method.

yeah but my concern is more of when it in theory should work.. but network connection drops / disk sleeps / power nap happens / whatever the reason and the low level write basically fails/gets discarded.  does it stay in the direct write logic because of the startup/inital test or does it fall back to non direct write logic ?

sabnzbd/assembler.py
mnightingale: I'm aware since changing this I've missed that it can get changed at runtime (statup actions), will likely add a callback for it.

mnightingale:
> yeah but my concern is more of when it in theory should work.. but network connection drops / disk sleeps / power nap happens / whatever the reason and the low level write basically fails/gets discarded. does it stay in the direct write logic because of the startup/inital test or does it fall back to non direct write logic ?

I don’t see any reason connection drops, sleep, etc would be any different to currently.

There are several paths which write to file:

Assember:
- Usually sequentially as currently, but designed to reduce disk access
- Forced when cached is full, may leave gaps and produce more fragmented files

ArticleCache:
- When cache is full

When either opens a file and it’s empty it attempts to make it sparse, if that fails it falls back to the current append mode and I’m going to make it change the config option to not try again – likely needs some callbacks so assembler/cache know.

If it was forced and couldn’t I’m going to simplify the logic around that.
file_done is always queued, so even with an abort it won’t miss the completed file notification.

If the cache was full and it attempted to write directly, it will return False and fallback to the current temporary storage mechanism.

The behaviour when it unexpectedly isn’t supported has been a key focus, it was the main area missing when I last attempted this a few years ago.

Safihre:
Should we include it for 4.6 next beta?

Or maybe with all this, 4.6 should be 5.0? 🤯

mnightingale:
I’d like to have another look over the Assembler.assemble logic and the fallback to append, which I’ll try and do tomorrow.

Perhaps 5.0 could make sense, pipelining and direct write are significant new features that represent a substantial difference between 4.x releases and does leave a stable release for people to revert to.

4.0 was unlocked SSL, if 5.0 is pipelining+direct write, I’m going to have to come up with something for 6.0.

Safihre:
Maybe we can then also make some breaking API changes if we call it 5.0.
Not sure which, but at least it's allowed 😂

> 4.0 was unlocked SSL, if 5.0 is pipelining+direct write, I’m going to have to come up with something for 6.0.

Hahah nice!

mnightingale:
> Maybe we can then also make some breaking API changes if we call it 5.0.

Any you had in mind? I'm not sure if it's an API change but the `add_active_history` behaviour surprised me in #3254 `status=Status.COMPLETED, storage=""` I would expect `Status.QUEUED` because it hadn't finished post-process at the time.

---

I'm going to look over this PR changes shortly.
It's also going to need a wiki entry, what do you think about the name?
"Direct Write" but we have ended up not doing the direct bit (no cache) we require that a cache is used, so it's never direct.

Safihre:
> It's also going to need a wiki entry, what do you think about the name?

With this in mind. I would suggest this becomes a Special again. 
It should only be needed to disable in very special cases, right?
And then it also doesn't need callbacks to deal with changes, user is just responsible to restart.

Safihre:
And API changes: removing all those fields were we pre-calculate from bytes to `XX MB`.
Like `"size": "9.5 GB",` "total_size": "255.9 G",` 
That should al be a front-end thing.

thezoggy:
> And API changes: removing all those fields were we pre-calculate from bytes to `XX MB`.
> Like `"size": "9.5 GB",` "total_size": "255.9 G",` 
> That should al be a front-end thing.

But is it quicker to have it precalculated or expect the front-end to do it each time for example on showing that field on the queue/history

mnightingale:
> With this in mind. I would suggest this becomes a Special again.
> It should only be needed to disable in very special cases, right?
> And then it also doesn't need callbacks to deal with changes, user is just responsible to restart.

That's correct, typically users should use it; it's just those odd people who insist on poor setups with remote download_dirs that might need to disable it, which it currently handles transparently.

I currently have it checking sparse is supported at start-up and if the download_dir changes.
But as an extra guard I also have it disabling if it fails to make a file sparse at assembly time.

If I revert it to requiring a restart, I'd probably have to remove that guard, but it would still work.

The only users who I think could see a worse experience are:
- Using slow storage - say 100MB/s
- and large cache size configured - >= 2 GB
- and large NzbFile(s), e.g. not split into rar archives, so cache force writes at 90% of 2GB and the slow disk has to write 1.8 GB

So, like direct unpack, I think you could make a good argument that it should be disabled by default and enabled at runtime if disk speed is sufficient.

Additionally, or alternatively I could include the triggered assembles but I don't want to compromise those systems with SSDs that can write significantly faster than they download. So, I would not want triggering at current 5% of cache size, it wastes too much time on IO.
So, I'll have to test the impact but maybe instead of 90% of the cache limit calculate `min(1GB, 0.9*cache limit)` so still ignore the configured cache limit but less pressure on slow disks.

The correct solution is for users to configure their cache size appropriately; the other client only recommends up to 1GB.

---

I have made quite a few changes recently I think it's looking very close, fixed bugs in fallback mode, a bunch of tests, simplified assembler logic.
~~I may take another look at does ArticleCache need to be a thread, or at least the use of `threading.Condition`, I quite like that it allows it to flush when paused though but probably other ways to achieve that.~~ edit: happy with this, just need to sort the special and behaviour

There is also deciding what to do about general vs specials, please let me know how you think the above effects that.

---

Maybe take the api discussion to discussions, but I agree on API returning bytes, I recall there were historic issues with max values allowed in json or by JavaScript engines but maybe that's not a concern anymore.

Safihre:
So far you kept it in General even in the last commits, not moving to Specials?

sabnzbd/articlecache.py
```diff
@@ -22,76 +22,147 @@
...
+    def save_article(self, article: Article, data: bytearray):
         """Save article in cache, either memory or disk"""
         nzo = article.nzf.nzo
         # Skip if already post-processing or fully finished
         if nzo.pp_or_finished:
             return
 
         # Register article for bookkeeping in case the job is deleted
-        nzo.saved_articles.add(article)
+        with NZO_LOCK:
```
Safihre: Let's add new locked methods to nzbobject instead of using the lock directly here.
It's cleaner to me, unless you disagree?

sabnzbd/filesystem.py
```diff
@@ -1081,7 +1082,9 @@ def save_data(data: Any, _id: str, path: str, do_pickle: bool = True, silent: bo
                 time.sleep(0.1)
 
 
-def load_data(data_id: str, path: str, remove: bool = True, do_pickle: bool = True, silent: bool = False) -> Any:
+def load_data(
```
Safihre, line 1085: Could probably change this to accept a constant `type=StoredData.PICKLE` or something like that instead of the 3 bools.
But fine for now of course.

mnightingale:
> So far you kept it in General even in the last commits, not moving to Specials?

I got distracted making the tests faster, shaved 30-60 seconds off so far [actions log](https://github.com/mnightingale/sabnzbd/actions/runs/20682582450)

I think it's correctly back is specials and I'll take a look at the other feedback now.

sabnzbd/assembler.py
```diff
@@ -245,6 +369,64 @@ def check_encrypted_and_unwanted(nzo: NzbObject, nzf: NzbFile):
...
+        # https://docs.python.org/3/library/io.html?highlight=write#io.RawIOBase.write
+        if written < len(data):
+            mv = memoryview(data)
+            while written < len(data):
+                if sabnzbd.WINDOWS:
+                    with nzf.file_lock:
+                        os.lseek(fd, pos + written, os.SEEK_SET)
+                        written += os.write(fd, mv[written:])
+                else:
+                    written += os.pwrite(fd, mv[written:], pos + written)
```
Safihre: Why not all in 1?
```python
mv = memoryview(data)
written = 0
while written < len(data):
    if sabnzbd.WINDOWS:
        # pwrite is not implemented on Windows so fallback to os.lseek and os.write
        # Must lock since it is possible to write from multiple threads (assembler + downloader)
        with nzf.file_lock:
            os.lseek(fd, pos + written, os.SEEK_SET)
            written += os.write(fd, mv[written:])
    else:
        written += os.pwrite(fd, mv[written:], pos + written)
```

sabnzbd/assembler.py
```diff
@@ -245,6 +369,64 @@ def check_encrypted_and_unwanted(nzo: NzbObject, nzf: NzbFile):
...
+
+        nzf.update_crc32(article.crc32, len(data))
+        article.on_disk = True
+        with nzf.lock:
+            if nzf.assembler_next_index == nzf_index:
+                nzf.assembler_next_index += 1
```
Safihre: Maybe add a comment why we don't always update the value?
I guess due to partial writes?

sabnzbd/articlecache.py
mnightingale: I'll do that and remove `__enter/exit__` from nzf.

How come NzbObject doesn't use `__slots__` ?

sabnzbd/assembler.py
```diff
@@ -245,6 +369,64 @@ def check_encrypted_and_unwanted(nzo: NzbObject, nzf: NzbFile):
...
+
+    @staticmethod
+    def open(nzf: NzbFile, direct_write: bool, file_size: int) -> tuple[int, int, bool]:
+        """Open file for nzf
+
+         Use direct_write if requested, with a fallback to setting the current file position for append mode
+        :returns (file_descriptor, current_offset, can_direct_write)
+        """
+        with nzf.file_lock:
+            fd = os.open(nzf.filepath, os.O_CREAT | os.O_WRONLY | getattr(os, "O_BINARY", 0), 0o644)
```
Safihre: Should we already do something with the permissions here that they match the user-configured ones?

sabnzbd/cfg.py
```diff
@@ -743,6 +745,19 @@ def new_limit():
     if sabnzbd.__INITIALIZED__:
         # Only update after full startup
         sabnzbd.ArticleCache.new_limit(cache_limit.get_int())
+        sabnzbd.Assembler.new_limit(sabnzbd.ArticleCache.cache_info().cache_limit)
+
+
+def new_direct_write():
+    """Callback for direct write changes"""
+    sabnzbd.Assembler.change_direct_write(direct_write.get())
```
Safihre: I think we should change all the `cfg.xx.get()`'s to just using `cfg.xx()` like we do everywhere else.

sabnzbd/assembler.py
mnightingale: 99.9% of the time the first write attempt is successful, so it saves constantly creating unnecessary memoryviews and delays getting access via the buffer protocol.

Those partial writes can happen on any system, but I've only seen it myself with mergerfs, because it can decide it needs to move files.

It could just do it like it currently does `data[written:]` is not using memory views and is copying data.

sabnzbd/cfg.py
```diff
@@ -743,6 +745,19 @@ def new_limit():
...
+    sabnzbd.ArticleCache.change_direct_write(direct_write.get())
+
+
+def new_download_dir():
+    """Callback for download dir changes"""
+    if direct_write.get():
+        threading.Thread(target=sabnzbd.filesystem.check_sparse_and_disable, args=(download_dir.get_path(),)).start()
```
Safihre: Still not fully in favour of this. If we do this, why not the whole `check_filesystem_capabilities`?

sabnzbd/misc.py
```diff
@@ -774,10 +766,35 @@ def __init__(self):
...
+        with open("/proc/meminfo") as f:
+            for line in f:
+                if line.startswith("MemTotal:"):
+                    return int(line.split()[1]) * 1024
+    except Exception:
+        pass
+    return os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES")
+
+
+def get_memory() -> int:
```
Safihre, line 787: Pushed a change to simplify `get_windows_memory`. 
Maybe this can then just all be merged into 1 function instead of 1 for each platform.
Not important for finishing of this PR, can be done later.

sabnzbd/assembler.py
```diff
@@ -162,52 +228,110 @@ def diskspace_check(nzo: NzbObject, nzf: NzbFile):
...
+            # Resume assembly from where we got to previously
+            for idx in range(nzf.assembler_next_index, len(decodetable)):
+                article = decodetable[idx]
 
-        # We write large article-sized chunks, so we can safely skip the buffering of Python
-        with open(nzf.filepath, "ab", buffering=0) as fout:
-            for article in nzf.decodetable:
                 # Break if deleted during writing
-                if nzo.status is Status.DELETED:
+                if nzo.status is status_deleted:
```
Safihre: I think we can spare this 1 extra attribute-lookup and just use `Status.DELETED` for clarity.

sabnzbd/articlecache.py
mnightingale: Hmm actually, I think I avoided that because ideally I would go through everywhere NZO_LOCK is used and see if only the current instance needs locking.

I was also planning on adding something like `@synchronized("lock")` or a callable `@synchronized(lambda self: self.lock)` so it can use the instance locks instead of only globals, because I do think it's neater than `with lock:` everywhere at least when the whole function requires the lock.

Shall I just use nzo.lock in articlecache for now?

sabnzbd/assembler.py
```diff
@@ -162,52 +228,110 @@ def diskspace_check(nzo: NzbObject, nzf: NzbFile):
...
+                    break
+
+                # When forced stop once reached an untried article unless paused
```
Safihre: Forced is when it's forced by the cache, right? Why do we stop at untried article and unless paused?
I looked through the explanations but couldn't find this one?

sabnzbd/assembler.py
```diff
@@ -162,52 +228,110 @@ def diskspace_check(nzo: NzbObject, nzf: NzbFile):
...
+            if fd is not None:
+                os.close(fd)
 
         # Final steps
         if file_done:
             set_permissions(nzf.filepath)
             nzf.assembled = True
 
+    @staticmethod
+    def assemble_article(article: Article, data: bytearray) -> bool:
```
Safihre, line 437: What if we just called `Assembler.assemble` for all `nzf` in the cache when we need to flush?
Just a thought.

sabnzbd/assembler.py
Safihre: Lets do that indeed, instead of code duplication.

tests/test_assembler.py
Safihre, line 1: I will admit I have not read the tests, I trust you 🎉

sabnzbd/assembler.py
mnightingale: That's correct, I'll add a comment about why.

Basically if a NzbFile has 1000 articles, they are assigned to servers and requested in order.
So if we reach one that has never been assigned to a server we can exit early, there is no need to loop through and check the same condition for the other articles.

I guess there are rare cases where articles are reset to 0 tries but doesn't seem worth accounting for, I could use saved_articles instead I suppose.

sabnzbd/assembler.py
mnightingale: I've extracted it to a seperate method, it's one extra line but hopefully that's ok.

sabnzbd/articlecache.py
mnightingale: I've just done it for usage in articlecache for now

sabnzbd/assembler.py
mnightingale: I've commented this, is it clear enough?

It's the next article index we want to write to remain sequential, it allows the below to skip checking those we have already written. Particularly relevant to append mode, the best way to make the assembler faster was to just make it do less.

```py
# Resume assembly from where we got to previously
for idx in range(nzf.assembler_next_index, len(decodetable)):
```

sabnzbd/assembler.py
mnightingale: I've changed this a little so it creates files with the same permissions as `with open(...)` in Python it's 666 and apply the users umask (typically 022 or 002 so resulting in 644 or 664)

But I don't think we should set the requested permission here, leave that in one place after assembly has finished. I don't want to risk accidently allowing execute permissions.

sabnzbd/assembler.py
mnightingale: The problem I see with that is you are probably suffering from poor IO, if the cache is full each article is going to be assembled directly from downloader threads, the assembler will be trying to write to probably the same file, if we then add the ArticleCache thread trying to also write the same and different files. We would have given already struggling IO even more work to do.

Best to just queue them in the assembler and let it try to work one at a time.
The downloader threads also try to back off to get under 100% cache usage.

I would have preferred set comprehension, but I don’t know if it’s too magic.

```py
nzfs: set[NzbFile] = {article.nzf for article in self.__article_table.copy() if article.can_direct_write}
for nzf in nzfs:
    logging.debug("Forcing write of %s", nzf.filepath)
    sabnzbd.Assembler.process(nzf.nzo, nzf, force=True)

self.__next_flush = time.monotonic() + _SECONDS_BETWEEN_FLUSHES
```

puzzledsab:
The fairly frequent writes with the current system is intentional from my side. It's a write cache so I think keeping as little unwritten data as possible is safer. I haven't experimented with it but I've been worried about what might happen if the OS shuts down and it doesn't give SAB enough time to write all the cached data. If there's a loss of power then it's all lost.

Writing less frequently will use somewhat less resources but as you pointed out, it will mostly fix itself if it's lagging behind because it will write all available data. Also, most posts use RAR files that are 50-500 MB so with max cache there won't be a lot of opening and closing for each file.

I also kept the multiple queuing of nzfs intentionally because otherwise we'd need another way to detect that the writing is lagging behind. It's not perfect but it works implicitly and I figured that it uses very little extra resources compared to all the other processing of each single article so changing it wouldn't have a real impact on performance.

sabnzbd/cfg.py
mnightingale: I've changed this, any idea why OptionBool returns an int `def __call__(self) -> int:` it seems intentional.

mnightingale:
I may have have introduced some regressions, particularly around handing IO-limited situations where force writes are required. Along with some handing around items being removed from the queue while the cache is flushing directly. I've switched to draft while I figure that out.

@puzzledsab, I intend to respond tomorrow after considering some of the points you raised.

Direct write has been NZBGet’s default approach for around 18 years. While it doesn’t enable a cache by default, it does recommend one albeit significantly smaller than SAB. As a result, I’ll likely compromise with some aspects of the trigger system, especially around detecting and handing situations where IO can’t keep up. Some of the other changes should provide benefits to both write methods anyway.

My primary goal is to write at the correct offsets. I’ve used simpler patches to achieve this, and over the years it has allowed me to repair many files that wouldn’t otherwise and not just via par2. Extreme performance is a secondary concern; I’m unlikely to have a connection exceeding 1 Gbps anytime soon. 

That said, I don’t see a technical reason why the go to for achieving the highest speeds needs be a different client. I’m not convinced this is inherently due to Python being slower; rather, it’s my opinion that there are several areas where the code is simply not optimal. That’s not a criticism of the original authors, much of it was written over 15 years ago for connections speeds far lower than what are becoming increasingly common today.

With that in mind, I’ll take care to analyse performance and impact under various disk and file size bottlenecks to ensure the PR works well across a range of user setups.

mnightingale:
@puzzledsab

The main problem I have with the throttling on the strict append method, is that I don’t believe it always achieves the intended goal of throttling download speed to allow disk IO to catch up because ultimately the disk can’t catch up unless the next article it needs is downloaded which is exacerbated at low cache levels or fast connections.

On the develop branch with no pipelining and testing on my desktop with a disk and cache bottleneck:
Cache: 256M (Assembler trigger = 18)
Connection: 10 Gbps / 32
Disk: "slow" SATA SSD - an old Crucial MX500 1TB it can sustain 400-450 MB/s

The 256M cache size exacerbates it, but it’s not so different to any connection speed where the cache is not sufficient to hold several seconds worth of articles and maybe more at higher speeds. The impact is just greater at high speeds because it can build up so rapidly and fill the cache.

The issue is that users use many connections, so the order of requests and responses is undetermined.

`articles_left` doesn’t account for whether the “next” article can be written yet, so queuing an assemble every 18 articles downloaded often doesn’t do anything except open and close the file, this is where I have issue with the currently queue_length throttling mechanism, in many cases it’s not throttling anything and even if it was I don’t think it would be achieving what was desired.

To explain what I mean, If I add an extra line logging if the assembler managed to write anything:

```
[assembler:81] Decoding part of …
[assembler:210] Opened but did not write anything
```

It is logged 60+ times per second, I am never triggering the check_assembler_levels sleeping because opening and closing files is instantaneous. But even if it was, I don’t believe it would achieve the goal of allowing the disk to catch up, the assembler doesn’t have the next article it needs so can’t write anything and sleeping wouldn’t have achieved anything and the cache remains full. Note: this is not to say I don’t think there is value in the queue_level mechanism, if the cache is large enough for the connection speed it will usually have the intended result however, I don’t believe it scales well to fast connections at least not without digging into the `assembler_max_queue_size` setting, which users are not going to do.

Tons of `[filesystem:931] [sabnzbd.filesystem.load_data] Deleting file …` are logged because it couldn’t store further articles in the cache and had to write them to temporary files later to be read again by the assembler and appended to the destination. It results in the job taking ages to finish and stuck in the queue at 100% progress because there are tons to tiny files to read and delete.

The result is an unstable speed of 250-350 MB/s that has actually written far more data to disk than necessary, in contrast my direct_write method although it prefers writing sequentially it does not have the append mode limitations, has a relatively flat speed and reports an average of 515 MB/s i.e. what I expect since it is limited by the disk speed.

The solution should not just be to give it a bigger cache so it has the “next” article when it needs it, again I want something that works for all users and does not waste resources.

n.b. If you want to see it have a really bad time, disable the article cache entirely, it triggers on every completed article and most of the time must write to temporary files.

---

I am going to give some thought to if even in direct write mode it should trigger writes every so often, I think it probably should at least if the file is large, maybe relative to the cache limit, non-rar’d posts are more common than they used to be so I might trigger every 500 MB or something - but not 5% of cache limit because it's too much.

I will also consider if there are different approaches for the queue_level mechanism when using append mode, for example seeing if the article we are being notified is the next one we need. My favoured approach is something at least incrementing the delayed counter if the article cache had to write to temporary files.

The shutdown behaviour a valid concern, I think SAB only gives a brief time for threads to finish but like with power cuts SAB should just recover and redownload them. I will see what it does currently, but I think article.decoded and tries should be reset on startup, with a few edge cases like the temporary files.

I do appreciate your thoughts on the PR; it has given me things to be conscious of. I am trying to give users the best experience, and they always have the append mode to fallback to although I intend for them never to need to.

---

I believe that I have resolved the performance problems I recently introduced, but I am going to 
do a little more testing and consider the above points before marking it ready to review again.

sabnzbd/cfg.py
mnightingale: Removed, unnecessary since I made runtime changes of the config possible.

Although I think it should check_filesystem_capabilities when it changes anyway, but I'll leave that out of this PR.
A user start SAB for first time, defaults to local disk, checks all pass, configures a non-optimal directory or mount, everyhing looks good and they don't get any warnings or errors until they next start the program.

sabnzbd/misc.py
mnightingale: Yes, I agree with combining but will leave it out of this PR.

sabnzbd/assembler.py
mnightingale: I've added an explanation in the comments.

sabnzbd/assembler.py
mnightingale: I've made this loop a little simpler, and only logs which files are forced once per second.

sabnzbd/nzb/file.py
```diff
@@ -265,7 +265,7 @@ def bytes_written_sequentially(self) -> int:
                     offset += article.decoded_size
                 elif os.path.exists(self.filepath):
                     # old queues were always files opened in append mode, so use the file size
-                    return os.stat(self.filepath).st_size
```
Safihre: I noticed that sometimes you use getsize and sometimes st_size?

Safihre:
I have no further pressing comments.
To prevent the PR discussion becoming too long, should we merge and then have follow-up PR's for improvements? 🎉
Only very last comment would be to open the PR to add `direct_write` to the wiki.

sabnzbd/nzb/object.py
```diff
@@ -210,6 +210,7 @@ def __init__(
         dup_check: bool = True,
     ):
         super().__init__()
+        self.lock: threading.RLock = threading.RLock()
```
Safihre, line 213: After this I will create a new decorator that uses the object's `lock` attribute.
I find the decorator approach cleaner. But could just be my personal preference.
Then we no longer need `NZO_LOCK`

sabnzbd/nzb/file.py
mnightingale: I think the only places left I use os.stat/os.fstat is when I have an already open file

sabnzbd/cfg.py
mnightingale: fyi my guess was it's used in an api somewhere that expects a return of 0 or 1 for bools?

mnightingale:
I kept the changes I was working on in a separate branch, to avoid spamming and allow me to rebase or drop commits.

I've addressed the points raised by @puzzledsab and my concerns with the append throttling - which Warp/GPT-5.1 agreed with me about – however much that’s worth.

tl;dr; `articles_left` does not tell you if there is anything append mode assemble can write, so measuring how many times a non-actionable task has been queued is not a meaningful measure.

I didn’t intend to make any changes to the append method, but Warp pointed out I’d already done most of it:

`assembler_next_index` tells us what article to write next
`article.decoded_size` had already been added recording the actual size of data to write to disk

I have added `contiguous_ready_bytes`which counts how many contiguous bytes from `assembler_next_index` onwards are ready to be written to disk and uses that as the trigger therefore only queuing when there is processible work.

I have also added a write trigger to direct write mode; however, it is at a larger size (75% of cache, capped at 500M) and works in a more naive way of scheduling a assemble without computing if it can do so contiguously - at 75% of cache it should be except at very small sizes.

I have kept the `check_assembler_levels` sleep in the downloader, it’s mostly the same calculation but it is adapted to use the actual bytes waiting to be assembled (cache and temporary). Along with some extra logic to avoid making the calculation when below the 0.5 trigger. It also has some deadlines and exits if the assembler has nothing it’s working on.

Separately I have renamed `force` to `allow_non_contiguous` to try and make it clearer.

I think it’s ready now, I’ll work on a wiki page.

edit: also given the trigger changes I could probably remove the ArticleCache force/allow_non_contiguous writing because it'll be rare to hit, but I do like it for flushing when paused.

Safihre:
Nice. We can do another release some time next week.

PR #3242, opened 2025-12-24
Refactor preparing the rss config

mnightingale:
An initial cleanup of the config parsing, is this the kind of thing you had in mind?

I wanted it in the config.py, with a method on ConfigRSS but circular references prevented it.

There are a few other changes I'd like to add to this PR:
- Stop checking for string values everywhere add a class with the constants
- Maybe pull out some other constants
- Perhaps myCat => my_category but I might leave those for another PR

Would you prefer I split not listed above other changes over many smaller PRs?

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -25,8 +25,8 @@ import time
 import datetime
 import threading
 import urllib.parse
-from dataclasses import dataclass
-from typing import Union, Optional
+from dataclasses import dataclass, field
+from typing import Union, Optional, Iterator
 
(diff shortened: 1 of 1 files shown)


Safihre:
Great. And changing those variable names would definitely help 😁

Smaller PR's is better for me indeed. Even these changes already look good?

mnightingale:
I don't think the failures are related to any of these changes, probably retry?

I'll leave the changes at this for now and make some more PRs in a few days.

Safihre:
Nice 🎄

PR #3243, opened 2025-12-28
Simplify RSS rule evaluation

mnightingale:
I think this completes my refactoring of the rules side of rss processing, there was a minor bug in the last PR; it didn’t record which rule was used correctly.

I’ve basically split the reading of the feed and testing it for matches as much as possible. In some places I’ve reduced duplication and in others actually added some (matched_rule is None) because it greatly reduces the complexity, I'll likely add some helpers in another PR.

The logic around category/priority/pp/script is greatly simplified by normalising values early rather than checking everywhere with `notdefault` - which is removed because its not needed anymore and it's name is confusing `not notdefault(...)`, the normalisation means ‘no override’ becomes None. This way the priority of values has much simpler logic with some special handling of pp and priority where ‘0’ is a valid value.

```py
my_pp = coalesce(matched_rule.pp, category_pp, self.default_pp) # coalesce(0, None, None) = 0
my_script = matched_rule.script or category_script or self.default_script # 'None' or None or 'blah.py' = 'None'
```

There is a stray `logging.debug("Size %s", size)` that I've kept, do we need that or can it be combined with `logging.debug("Trying title %s", title)` something like `logging.debug("Trying title=%r, size=%d", title, size)`

Focus for any further changes will probably be on cleaning up the rest of the run_feed method and sqlite for storage.

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -26,7 +26,7 @@ import datetime
 import threading
 import urllib.parse
 from dataclasses import dataclass, field
-from typing import Union, Optional, Iterator
+from typing import Union, Optional
 
 import sabnzbd
 from sabnzbd.constants import RSS_FILE_NAME, DEFAULT_PRIORITY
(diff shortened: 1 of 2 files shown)


sabnzbd/rss.py
Safihre, line 712: Maybe we can also "fix" this part? It has always bothered me.

mnightingale: The _HandleLink part? I have another ~500 lines of local changes which focus on splitting run_feed up into separate steps (configure, parse, evaluate, and other side effects). I've just tried to keep each PR limitted in scope.

So _HandleLink would likely be renamed, typed, and made part of the RSSReader class, likewise with all/most other methods. I've just kept it seperate else it would be a 1000+ line PR changing almost all of it at once.

Safihre: Ah perfect 👍

PR #3247, opened 2025-12-29
Refactor RSS flow

mnightingale:
Apologies, this one is kind of a lot but was unavoidable.

The goal here was to make the logic flow easier to follow.

All the new dataclasses from previous PRs have moved to the top of the file to allow easier typing without quotes everywhere.

Generally, I’ve broken up the logic into separate methods so that each is usually only responsible for a single task.

So run_feed basically follows:
- configure_rss
- fetch_rss or use jobs dictionary directly
- normalise entry to a NormalisedEntry
- evaluate the entry against filters/rules
- process the evaluation result, persist and maybe download
- email
- remove obsolete

Next will be adding persistence in sqlite, likely with a separate storage class.

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -48,72 +48,441 @@ import sabnzbd.emailer as emailer
 import feedparser
 
 
-##############################################################################
-# Wrapper functions
-##############################################################################
-
-
-def _normalise_default(value: Optional[str]) -> Optional[str]:
-    """Normalise default values to None"""
-    if not value:
-        return None
-    v = str(value).strip()
-    if v.lower() in ("", "*", "default"):
-        return None
-    return v
-
-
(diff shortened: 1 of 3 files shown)


sabnzbd/rss.py
```diff
@@ -48,392 +48,173 @@
...
+    rule: int
 
-    @synchronized(RSS_LOCK)
-    def clear_downloaded(self, feed):
-        # Mark downloaded jobs, so that they won't be displayed any more.
-        if feed in self.jobs:
-            for item in self.jobs[feed]:
-                if self.jobs[feed][item]["status"] == "D":
-                    self.jobs[feed][item]["status"] = "D-"
+    status: str  # "G", "B", "G*", "D"
```
Safihre, line 212: Only minor comment: should we make this an Enum?

PR #3248, opened 2025-12-30
Use blocking writes instead of buffering

Safihre:
I made the changes, even with high pipelining setting I didn't experience any problems.
@mnightingale thoughts?

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -60,7 +60,6 @@ class NewsWrapper:
         "blocking",
         "timeout",
         "decoder",
-        "send_buffer",
         "nntp",
         "connected",
         "user_sent",
@@ -86,7 +85,6 @@ class NewsWrapper:
         self.timeout: Optional[float] = None
 
         self.decoder: Optional[sabctools.Decoder] = None
-        self.send_buffer = b""
 
         self.nntp: Optional[NNTP] = None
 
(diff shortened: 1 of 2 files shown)


sabnzbd/newswrapper.py
```diff
@@ -375,25 +366,24 @@ def write(self):
...
-                            self.send_buffer = command[sent:]
+                        self.nntp.sock.sendall(command)
+                        self._response_queue.append(article)
                     except (BlockingIOError, ssl.SSLWantWriteError):
-                        # Can't send now, store full command
-                        self.send_buffer = command
+                        # Couldn't send now, try again later
+                        self.concurrent_requests.release()
+                        self.next_request = (command, article)
+                        return
```
mnightingale: This part is the only worry, if ssl.SSLWantWriteError occurs we don't know how much of the command was written leaving the state undefined.

We could have written `BODY <a` then the next write we are going to try again and the entire command can end up as `BODY <aBODY <article-id>\r\n`

I'm not 100% but I think BlockingIOError would be no bytes sent.
But the ssl.SSLWantWriteError could lead to subtle difficult to diagnose bugs, could be unknown commands or missing articles.

If the code noise and complexity is your concern, perhaps we could instead split the writing and command queue into a different wrapper class, so it's hidden from newswrapper and seperated from finding the next request/article?

Safihre: Indeed it feels like it's doing too many things in 1 function, might be better to split it indeed.
Although, we have used `sendall` for ages and never had any problems..
Yes, of course sending 20 requests in 1 write will be much larger than the 1 request we had so far.

Safihre: Hmm. Thinking about this: 1 ssl frame is 16k. Which is such a huge amount of data compared to what we are sending.

I am just going to try a test release with this in it, and see what happens in the field.

PR #3249, opened 2025-12-30
Convert RSS Status to Enum

Safihre:
@mnightingale

### DIFF
--- a/sabnzbd/interface.py
+++ b/sabnzbd/interface.py
@@ -38,6 +38,7 @@ from typing import Optional, Callable, Union, Any
 from guessit.api import properties as guessit_properties
 
 import sabnzbd
+from sabnzbd.rss import RSSStatus
 from sabnzbd.misc import (
     to_units,
     from_units,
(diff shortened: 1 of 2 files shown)


mnightingale:
The star stuff makes it rather complex, I know they are strings currently but maybe look at Flag or IntFlag and a helper for converting from strings.

So, we could do something like `status = RSSStatus.DOWNLOADED | RSSStatus.HIDDEN`

I'm also consious of storing these in sqlite, so IntFlags might be a good fit.

---

But at least the strings stop you having impossible states, so im not sure.

mnightingale:
Strings backed are OK with me, sqlite has check constraints.

```sql
CREATE TABLE rss (
    "id" INTEGER PRIMARY KEY,
    "feed" TEXT NOT NULL,
    "status" TEXT NOT NULL
             CHECK (status IN (
             'G',  -- Good
             'B',  -- Bad
             'D',  -- Downloaded
             'X',  -- Expired
             'G*', -- Good Initial
             'D-' -- Downloaded Hidden
             )),
```

mnightingale:
There is also a `NEW = "N"` status

PR #3253, opened 2025-12-31
SQLite storage for RSS

mnightingale:
Moves RSS to SQLite for more efficient access.

Always stores dates and times in UTC and adjusts for the systems current timezone

Also allows for some future enhancements:

- Pagination, search and filtering in the UI like history
- Cleanup of archived (downloaded) jobs


thezoggy:
i would say we shouldnt be putting legit indexer urls in the tests... replace with dummy domains

Safihre:
Basic question: why it's a seperate database file and not part of the main databae file?

mnightingale:
> Basic question: why it's a seperate database file and not part of the main databae file?

It can be, for some reason I thought you preferred them separate.

I based my class on the history one, so it shouldn't be very difficult to combine them. There are several changes and improvements I wanted to make to the history one anyway, to start with use WAL mode and ensure closing, improve indexes (were not auto), and use connections per thread as `get_db_connection` does but make it not part of CherryPy so `with HistoryDB() as history_db:` doesn't need to be used elsewhere.

edit: circular import problems though :(

mnightingale:
> i would say we shouldnt be putting legit indexer urls in the tests... replace with dummy domains

I did notice those, there are a few tests which load feeds from sabnzbd.org which could just be local files. Then I can change the URLs.

---

I've moved everything to database.py, I had to split the models into a separate package.
I'll leave WAL, connection per query, and history index changes for another time.

thezoggy:
i'm sure you saw but recently someone in discord was asking for us to do pagination with our rss stuff. I'd imagine the move to sqlite backend should make this more attainable for a future update. Just mentioning it now if that plays a part in the refactor ideas

mnightingale:
Yeah, that’s what got me looking at rss.py and I couldn’t leave it be. But yes, using sqlite makes such pagination much simpler. Currently it does as Python was doing with dictionaries and groups all the entries into good, bad, or downloaded but sqlite greatly simplifies the sorting.

I think pagination in the future would probably mean splitting it into 3 queries, and paginating each tab independently, but again far simpler with sqlite along with things like search. I'd also like a way to delete single entries so they can be downloaded again, especially since the current history can go back years.

mnightingale:
I think I’m happy with this, but I’m going to review the changes before marking it ready. 

I’ve checked the handling of dates and times; the old way used three different formats (time.localtime, datetime.datetime – without tz, and float timestamp) so no timezone infomration, everything was just assumed to be localtime.

I switched to the develop branch initially, changed my device’ timezone, read a rss feed and added a nzb to the queue. Then switch to this PR, it imported the old storage, times and relative ages are still correct, switched to a different timezone while running and dates/times are still correct - 4.5.5/develop don’t cope with that.

How would you like me to handle the existing `rss_data.sab` after it has been imported, just delete it?
Currently if it does a migration to a newer DB version it imports it, but on a fresh install it is ignored.

As mentioned above there are a few new features that are worthwhile implementing (pagination especially), but for this PR I’m just moving the storage - maybe someone else will look at the interface/pagination changes and I can advise on any query or index changes.

thezoggy:
Does file get backed up during backup process?

About keeping old file, if we do it allows to downgrade but IMHO I'd say treat it. Like anything where warn you can only move forward and downgrading requires restoring backup.

mnightingale:
> Does file get backed up during backup process?

Yes, the rss table is added to history1.db so part of the normal backup archive.

> About keeping old file, if we do it allows to downgrade but IMHO I'd say treat it. Like anything where warn you can only move forward and downgrading requires restoring backup.

I don't think there is much benefit to keeping it because if a user upgrades and changes their mind they've still lost the recent history which is only really there to keep track of what has already been grabbed.

For example, upgrade to 4.6.0 or 5.0.0 whatever it becomes, run it for a while, RSS grabs some releases, downgrade to 4.5.5 which has no knowledge of those downloads and queues them again. In which case the user would be better off just restoring a backup from before the upgrade.

Safihre:
> How would you like me to handle the existing rss_data.sab after it has been imported, just delete it?

Just keep the file or rename it to `rss_data.sab.old`.
New installs will not get it, so seems fine to me?

mnightingale:
I've completed my own look over the changes.

I copied the remote test rss feeds to the repository, but removed references to indexers.

I didn't really want to split the models into rssmodels.py, but I couldn't see any other way to resolve a circular dependency since database and rss both need them.

As far as I can remember the only real change I made to the rss parsing is if it falls back to the description field try to strip out html tags and only take before \<br\/\> - some feeds dump 1KB+ html content in there.

I decided to leave pp (0-3) and priority as integers in the database to avoid needless conversions, I did consider something like [gist: PostProcessOpts](https://gist.github.com/mnightingale/3b898015a17fec09681c9178e6fe82c0) but decided against it for now - I didn't like the \_\_int\_\_ method for compatability elsewhere instead of using the flags i.e. 7 = REPAIR|UNPACK|DELETE.

It should open some doors for further enhancements, pagination in the UI, per item management (delete), maybe removal of downloaded items.

sabnzbd/constants.py
```diff
@@ -87,7 +87,6 @@
 
 CONFIG_BACKUP_FILES = [
     BYTES_FILE_NAME,
-    RSS_FILE_NAME,
```
jcfp: this has the side effect of rss info not getting restored when a user restores from a backup created prior to this changeset. Maybe fixup `config.restore_config_backup()` to restore the old rss file if present in the backup?

mnightingale: Well spotted, I've added CONFIG_RESTORE_FILES with additional files to restore from backups. If the restart after includes it and the database requires a migration it'll import the old database.

mnightingale:
Rebased and refactored, just a single ResolvedEntry dataclass now which could perhaps be better named.

I should probably check what does and should happen when the server is one timezone and client another.

`HistoryDB.rss_import_done` is a little odd but the only way I could think to achieve it since the database cannot import rss (circular dependency), without extracting the models to a seperate file as I'd done previously. Maybe some thought needs to go into how to structure history/database/rss.

Save this for 5.1.0 (hopefully) but also need to think about how it should fit within other changes, some I'm interested in but didn't want to bloat this PR with:

- Download an rss entry again, perhaps use history cleanup behaviour on rss history, or have the UI show already downloaded with a "Download (again)" button, note inserting again will only keep the latest download. Especially useful for user specific feeds e.g. carts.
- Improve display of past downloads (pagination), maybe a global view of history with multiple feeds.

sabnzbd/rss.py
```diff
@@ -51,10 +56,22 @@
...
+    GOOD = "G"  # Matched by filter rules (should be grabbed)
+    BAD = "B"  # Rejected by filter rules
+    DOWNLOADED = "D"  # Successfully downloaded to queue
+    EXPIRED = "X"  # No longer in feed (marked for cleanup)
+
+
+@dataclass
+class ResolvedEntry:
+    feed: str
     link: Optional[str]
```
Safihre: Can it really be optional?

sabnzbd/rss.py
```diff
@@ -486,47 +545,39 @@ def _resolve_options(
...
         # Patch feedparser
         self.patch_feedparser()
 
     def stop(self):
         self.shutdown = True
+        for store in list(self._active_db_connections):
+            store.close()
+
+    @property
+    def db(self) -> HistoryDB:
```
Safihre: Bit of context maybe why we need this? 
For other things we let the caller deal with this?

sabnzbd/rss.py
```diff
@@ -680,97 +710,146 @@ def _start_newznab_attr(self, attrsD):
...
+        Expired links older than 3 days are removed
         """
-        all_entries = []
-        msg = ""
+        now = int(datetime.datetime.now(datetime.timezone.utc).timestamp())
+        limit = now - 3 * 24 * 3600  # 3 days in seconds
+
+        if new_urls:
+            # Create temporary table for all new URLs
+            self.db.execute("CREATE TEMP TABLE temp_urls(url TEXT PRIMARY KEY)")
```
Safihre, line 542: Why do we need this? Just add new ones and remove old ones?

sabnzbd/rss.py
```diff
@@ -680,97 +710,146 @@ def _start_newznab_attr(self, attrsD):
...
+
+                for entry in entries:
+                    normalised = ResolvedEntry.from_feed_entry(feed, entry)
+                    if not normalised:
+                        continue
+                    # Merge the existing state
+                    existing = self.find_job_by_url(feed, normalised.link)
+                    if existing:
+                        normalised.merge(existing)
+                    yield normalised
```
Safihre: Why do we need the yield?

sabnzbd/rss.py
Safihre: Pycharm is flagging this a lot throughout the file, because we abuse the variable a bit. It means something else for different type of rules. Not sure what we can do about it.

sabnzbd/rss.py
mnightingale: The goal was for the caller to not need to worry about it by just giving each thread a connection which can be reused.

Lots of existing usage of HistoryDB does `with HistoryDB() as history_db: …` so opens a new connection for each query which has some overhead.

> For other things we let the caller deal with this?

Sometimes the caller does deal with it (cherrypy via `get_db_connection`), others it takes the easy less efficient approach, by abstracting the connection per thread away it reduced complexity for the caller.

sabnzbd/rss.py
mnightingale: I've split this into separate properties and resolved most other linter warnings

sabnzbd/rss.py
mnightingale: The temporary table which are deleted on close is much more efficient at performing this cleanup.

Without it we would have to perform a query for each URL individually or build a giant `NOT IN (…)` query taking care of the param limits.

Handling it individually in Python:

- Select all GOOD or BAD (could be thousands)
- Foreach not in new_urls
  - Update state = RSSState.EXPIRED

The set-based query is much more efficient, 4 queries to update up to 500 records and it offloads the work to SQLite which can perform the query very efficiently.

sabnzbd/rss.py
mnightingale: Primarily reduced memory usage, when reading a feed of 200+ entries we just need to handle each one by adding it to the database.

There are a couple of places I’ve used yield, I’d also say it reduces complexity because there is no huge list to manage.

The key part is we need each entry, but we don't need them all at once.

sabnzbd/rss.py
mnightingale: This has been addressed, the caller can choose to pass a database instance to the rss_repository helper otherwise it will create a HistoryDB instance.

mnightingale:
I think I’ve simplified and tidied up the changes if you wouldn’t mind reviewing sometime, no rush I expect there will be a 5.0.4 first anyway.

sabnzbd/rss.py
```diff
@@ -466,43 +524,305 @@ def evaluate(
         )
 
 
-class RSSReader:
-    def __init__(self):
-        self.jobs = {}
-        self.next_run = time.time()
-        self.shutdown = False
+class RSSRepository:
```
Safihre, line 527: Why the name repository/repo? 
Maybe because I'm not a native-speaker, but it just doesn't seem to fit?

Safihre:
Gonna trust you again on this one 🥳 If I don't like some of the stuff, I can always change it later.
Only that small comment above.

sabnzbd/rss.py
mnightingale: Repository because it follows the repository pattern, it’s abstracting the database access.

RSSReader => is only responsible for reading feeds and applying the config and rules.
RSSRepository => everything SQLite

The primary reason for it is it allows me to remove some awkward workarounds I had to use previously for `import_rss_records` and creating database connections per thread.

It’s actually very similar to how HistoryDB itself works and all the data access methods could be part of that class. However, it means several RSS related classes would also need to move and probably be renamed (RSSState, ResolvedEntry, normalise_*, etc.) otherwise there would be circular dependencies.

I'm not against moving things around, ideally, I think it should all be part of the HistoryDB class to be consistent or it would be split up further with Database, HistoryRepository, and RSSRepository.

PR #3254, opened 2026-01-02
Fix failing downloader tests and make tests less fragile

### LINKED ISSUE #3250: Occasional download test failures
### SABnzbd version

develop

### Operating system

N/A

### Using Docker image

None

### Description

I've been seeing these quite a lot recently.

```
FAILED tests/test_functional_sorting.py::TestDownloadSorting::test_download_season_sorting[sea_sort_s01_4k_uhd-SABnzbd-result0] - Failed: Download did not complete
FAILED tests/test_functional_sorting.py::TestDownloadSorting::test_download_season_sorting[sea_sort_s02_4k_uhd-SABnzbd-result1] - Failed: Download did not complete
```

https://github.com/sabnzbd/sabnzbd/actions/runs/20593138976/job/59142134416?pr=3249

I'm not sure what to make of them, it looks like sabnews must have crashed or otherwise exitted?

It tried to connect, but timed out after 60 seconds.

```
2025-12-30 09:26:26,685::INFO::[downloader:659] 5@127.0.0.1: Initiating connection
2025-12-30 09:26:26,686::INFO::[downloader:659] 2@127.0.0.1: Initiating connection
2025-12-30 09:26:26,686::INFO::[downloader:659] 3@127.0.0.1: Initiating connection
2025-12-30 09:26:26,686::INFO::[downloader:659] 6@127.0.0.1: Initiating connection
2025-12-30 09:26:26,687::INFO::[downloader:659] 8@127.0.0.1: Initiating connection
2025-12-30 09:26:26,687::INFO::[downloader:659] 4@127.0.0.1: Initiating connection
2025-12-30 09:26:26,687::INFO::[downloader:659] 7@127.0.0.1: Initiating connection

...

2025-12-30 09:27:26,825::INFO::[downloader:924] Thread 5@127.0.0.1: Timed out
2025-12-30 09:27:26,826::INFO::[downloader:924] Thread 2@127.0.0.1: Timed out
2025-12-30 09:27:26,826::INFO::[downloader:924] Thread 3@127.0.0.1: Timed out
2025-12-30 09:27:26,826::INFO::[downloader:924] Thread 6@127.0.0.1: Timed out
2025-12-30 09:27:26,826::INFO::[downloader:924] Thread 8@127.0.0.1: Timed out
2025-12-30 09:27:26,826::INFO::[downloader:924] Thread 4@127.0.0.1: Timed out
2025-12-30 09:27:26,827::INFO::[downloader:924] Thread 7@127.0.0.1: Timed out
2025-12-30 09:27:26,827::INFO::[downloader:924] Thread 1@127.0.0.1: Timed out
```

How can we make sabnews logging more visible? - at least when it starts and shuts down

I'm not familiar with the threads, but could ` 2025-12-30 09:26:17,074::INFO::[SABnzbd:1447] Performing triggered restart` be the cause - if SAB restarts does sabnews get shutdown?

mnightingale:
Fixes #3250

There were a few issues in the downloader logic:

- "Timed Out" in logs, was a race where idle_threads/busy_threads collided so a socket could be removed from listening but in busy_threads
- `nw.timeout = None` was missing which prevented kick starting idle threads, it used to be in soft_reset, that still kind of happens but it's remove_socket now.
- The biggest issue was if 8 connections were initially made, but they do not all immediately have article requests queued, they return to idle_threads. Then there were bugs related to those idle connections having not completed the authentication flow so not `connected` and not getting more work assigned.
- Split the logic for queuing the next request for a server and allow downloader to call it to kick start requests.

Test improvements:

- Made `download_nzb` helper less fragile
  - Clear queue/history between tests
  - When random job names are used use `"TestDownload_%s" % str(uuid.uuid4()).replace("-", "")` so deobfuscator won't get involved and they are easier to see in logs instead of many timestamps within 5 minutes of each other
  - Use API instead of selenium for reliability, it was matching name, but the name requested isn't always what it is created as
  - Sleep less between checks = quicker tests
  - Log more meaningful information about failure, queue, history, and completed files.
- Because download_nzb is less fragile I have removed flaky from download tests, it hid the failures
- Fixed assertions in `test_functional_adding_nzbs_clean.py` - it's where I copied the api tests from and found its assertions failed after 10 seconds and it carried on anyway, it does it twice so that's ~20 seconds saved

Shaves a few minutes off the test times, removing time.sleep(1) helps a lot.

`test_download_unicode_made_on_windows` fails on all platforms because it tries to deobfuscate, which used to waste 20 seconds due to flaky, so only ~10 seconds now. Will recreate it as a separate PR.

---

I ran test_functional_downloads.py 6000 times overnight without a failure.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -28,7 +28,7 @@ import sys
 import ssl
 import time
 from datetime import date
-from typing import Optional, Union, Deque
+from typing import Optional, Union, Deque, Callable
 
 import sabctools
 
@@ -173,7 +173,6 @@ class Server:
     def stop(self):
         """Remove all connections and cached articles from server"""
         for nw in self.idle_threads:
-            sabnzbd.Downloader.remove_socket(nw)
             nw.hard_reset()
         self.idle_threads = set()
         self.reset_article_queue()
@@ -375,6 +374,9 @@ class Downloader(Thread):
     def add_socket(self, nw: NewsWrapper):
         """Add a socket to be watched for read or write availability"""
         if nw.nntp:
+            server = nw.server
+            server.idle_threads.remove(nw)
+            server.busy_threads.add(nw)
             try:
                 self.selector.register(nw.nntp.fileno, selectors.EVENT_READ | selectors.EVENT_WRITE, nw)
                 nw.selector_events = selectors.EVENT_READ | selectors.EVENT_WRITE
@@ -395,6 +397,10 @@ class Downloader(Thread):
     def remove_socket(self, nw: NewsWrapper):
         """Remove a socket to be watched"""
         if nw.nntp:
+            server = nw.server
+            server.busy_threads.discard(nw)
+            server.idle_threads.add(nw)
+            nw.timeout = None
             try:
                 self.selector.unregister(nw.nntp.fileno)
                 nw.selector_events = 0
@@ -649,11 +655,9 @@ class Downloader(Thread):
                         if not server.get_article(peek=True):
                             break
 
-                        server.idle_threads.remove(nw)
-                        server.busy_threads.add(nw)
-
-                        if nw.connected:
-                            self.add_socket(nw)
+                        if nw.connected or nw.nntp:
+                            if nw.prepare_request():
+                                self.add_socket(nw)
                         else:
                             try:
                                 logging.info("%s@%s: Initiating connection", nw.thrdnum, server.host)
@@ -924,13 +928,6 @@ class Downloader(Thread):
             logging.info("Thread %s@%s: %s", nw.thrdnum, nw.server.host, reset_msg)
         elif reset_msg:
             logging.debug("Thread %s@%s: %s", nw.thrdnum, nw.server.host, reset_msg)
-
-        # Make sure this NewsWrapper is in the idle threads
-        nw.server.busy_threads.discard(nw)
-        nw.server.idle_threads.add(nw)
-
-        # Make sure it is not in the readable sockets
-        self.remove_socket(nw)
 
         # Discard the article request which failed
         nw.discard(article, count_article_try=count_article_try, retry_article=retry_article)
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -374,9 +374,8 @@ class Downloader(Thread):
     def add_socket(self, nw: NewsWrapper):
         """Add a socket to be watched for read or write availability"""
         if nw.nntp:
-            server = nw.server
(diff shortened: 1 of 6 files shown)


mnightingale:
6000 test runs and I was still nervous waiting for those checks... 4m 🚀

sabnzbd/newswrapper.py
```diff
@@ -201,7 +201,8 @@ def body(self, article: "sabnzbd.nzb.Article") -> tuple[bytes, "sabnzbd.nzb.Arti
     def on_response(self, response: sabctools.NNTPResponse, article: Optional["sabnzbd.nzb.Article"]) -> None:
         """A response to a NNTP request is received"""
         self.concurrent_requests.release()
-        sabnzbd.Downloader.modify_socket(self, EVENT_READ | EVENT_WRITE)
+        if not self.blocking:
```
Safihre: Can you add a comment why we do this?

sabnzbd/downloader.py
```diff
@@ -395,6 +397,10 @@ def modify_socket(self, nw: NewsWrapper, events: int):
     def remove_socket(self, nw: NewsWrapper):
         """Remove a socket to be watched"""
         if nw.nntp:
+            server = nw.server
```
Safihre: Do we need this optimization really?

sabnzbd/downloader.py
```diff
@@ -375,6 +374,9 @@ def init_server(self, oldserver: Optional[str], newserver: str):
     def add_socket(self, nw: NewsWrapper):
         """Add a socket to be watched for read or write availability"""
         if nw.nntp:
+            server = nw.server
+            server.idle_threads.remove(nw)
+            server.busy_threads.add(nw)
```
Safihre: It's added directly to busy threads because we send the first command?

mnightingale: Actually because the server sends a welcome message on connect, so we need to listen and consume that.

Only if nw.write finds articles to request do we send further requests and trigger the auth flow.
Otherwise it'll be removed after timeout, however to have been started we had requests to send.
But there might not be enough requests to need all the connections configured.

init_connect could do `self.timeout = time.time() + 5` so we recreate those never used connections?

sabnzbd/newswrapper.py
mnightingale: I've commented it in `on_response`, `blocking` is just because they don't use the read/write mechanism that's only for non-blocking sockets. It was always raising KeyError because it was not registered.

mnightingale:
I think I've resolved the issues, I'm going to improve the time.sleep in all tests after.

Issue #3209, opened 2025-11-30
post processing - multi threaded


thezoggy:
### Description

As peoples connections get faster we have started to get people to come in asking about how to make sab faster at post processing. 
With modern utilities no longer taking ages to do tasks and fast hdd able to keep up with demands there is a new bottle neck that shows up and that is sab actual post processing queue being a FIFO single thread.

I shared a recent example (in dev channel) where each job only takes ~2s to do but you can see the jobs build up in history just 'waiting' which really shows up after all the downloads finished. Thoughts on making this multi threaded where we can have multiple jobs running..  id envision it would be like direct unpackers and we have a config value to let users adjust to what makes sense (or flexible to do x per category?)

Safihre:
Nzbget also has the more precise tweaking to configure how many par2 and unpack can be active at same time.

Not even sure why pp takes 4s, maybe also worth a look.

thezoggy:
and just to call out, the post-processing time isnt the issue its just the lag between it going from the queue to being processed.
```
2025-11-29 17:27:17,326::INFO::[nzbqueue:767] [sabnzbd.nzbqueue.register_article] Ending job Bobs.Burgers.S09E07.I.Bob.Your.Pardon.1080p.DSNP.WEB-DL.DDP5.1.H.264-POWER
2025-11-29 17:27:17,759::INFO::[assembler:88] Decoding finished /scratch/incomplete/Bobs.Burgers.S09E07.I.Bob.Your.Pardon.1080p.DSNP.WEB-DL.DDP5.1.H.264-POWER/Bobs.Burgers.S09E07.I.Bob.Your.Pardon.1080p.DSNP.WEB-DL.DDP5.1.H.264-POWER.mkv
2025-11-29 17:27:17,760::DEBUG::[nzbstuff:467] Removing article database for SABnzbd_nzf_j2hli6n6
2025-11-29 17:27:17,760::DEBUG::[filesystem:931] [sabnzbd.nzbstuff.remove_admin] Deleting file /scratch/incomplete/Bobs.Burgers.S09E07.I.Bob.Your.Pardon.1080p.DSNP.WEB-DL.DDP5.1.H.264-POWER/__ADMIN__/SABnzbd_nzf_j2hli6n6
[...]

PR #3264, opened 2026-01-06
One Lock

Safihre:
@mnightingale what do you think?

### DIFF
--- a/sabnzbd/decorators.py
+++ b/sabnzbd/decorators.py
@@ -20,7 +20,7 @@
 ##############################################################################
 import time
 import functools
-from typing import Union, Callable
+from typing import Union, Callable, Any
 from threading import Lock, RLock, Condition
 
 
@@ -35,15 +35,21 @@ DOWNLOADER_CV = Condition(NZBQUEUE_LOCK)
 DOWNLOADER_LOCK = RLock()
 
 
-def synchronized(lock: Union[Lock, RLock]):
+def synchronized(lock: Union[Lock, RLock, Condition, None] = None):
     def wrap(func: Callable):
         def call_func(*args, **kw):
-            # Using the try/finally approach is 25% faster compared to using "with lock"
+            # Either use the supplied lock or the object-specific one
+            # Because it's a variable in the upper function, we cannot use it directly
+            lock_obj = lock
+            if not lock_obj:
+                lock_obj = getattr(args[0], "lock")
+
+            # Using try/finally is ~25% faster than "with lock"
             try:
-                lock.acquire()
+                lock_obj.acquire()
                 return func(*args, **kw)
             finally:
-                lock.release()
+                lock_obj.release()
 
         return call_func
 
--- a/sabnzbd/nzb/__init__.py
+++ b/sabnzbd/nzb/__init__.py
@@ -20,7 +20,7 @@ sabnzbd.nzb - NZB-related classes and functionality
 """
 
 # Article-related classes
-from sabnzbd.nzb.article import Article, ArticleSaver, TryList, TRYLIST_LOCK
+from sabnzbd.nzb.article import Article, ArticleSaver, TryList
 
 # File-related classes
 from sabnzbd.nzb.file import NzbFile, NzbFileSaver, SkippedNzbFile
@@ -30,7 +30,6 @@ from sabnzbd.nzb.object import (
     NzbObject,
     NzbObjectSaver,
     NzoAttributeSaver,
-    NZO_LOCK,
     NzbEmpty,
     NzbRejected,
     NzbPreQueueRejected,
@@ -42,7 +41,6 @@ __all__ = [
     "Article",
     "ArticleSaver",
     "TryList",
-    "TRYLIST_LOCK",
     # File
     "NzbFile",
     "NzbFileSaver",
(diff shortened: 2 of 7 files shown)


sabnzbd/decorators.py
```diff
@@ -35,15 +35,21 @@
 DOWNLOADER_LOCK = RLock()
```
mnightingale, line 36:
```suggestion

LockType = Union[Lock, RLock, Condition]
LockGetter = Callable[[object], LockType]

```

sabnzbd/decorators.py
```diff
@@ -20,7 +20,7 @@
 ##############################################################################
 import time
 import functools
-from typing import Union, Callable
+from typing import Union, Callable, Any
```
mnightingale, line 23:
```suggestion
from typing import Union, Callable, Optional
```

Not sure why Any was added, I don't think I see it anywhere

sabnzbd/decorators.py
```diff
@@ -35,15 +35,21 @@
 DOWNLOADER_LOCK = RLock()
 
 
-def synchronized(lock: Union[Lock, RLock]):
+def synchronized(lock: Union[Lock, RLock, Condition, None] = None):
```
mnightingale, line 38:
```suggestion
def synchronized(lock: Optional[Union[LockType, LockGetter]] = None):
```

sabnzbd/decorators.py
```diff
@@ -35,15 +35,21 @@
...
     def wrap(func: Callable):
         def call_func(*args, **kw):
```
mnightingale, line 40:
```suggestion
        @functools.wraps(func)
        def call_func(*args, **kw):
```

> It copies metadata from func to wrapper, such as:
__name__ → the original function’s name
__doc__ → the original function’s docstring
__annotations__ → type hints
__module__ → the module it’s defined in
__qualname__ → qualified name
…and it also sets __wrapped__ so tools like inspect and debuggers can see the original function.

> Why it matters in practice
**Debugging**
Stack traces will show the real function name, not wrapper.
**Documentation**
Tools like help(), Sphinx, or IDE tooltips show correct docstrings.
**Introspection**
inspect.signature() and other reflection tools will work as expected.
**Nested decorators**
When stacking multiple decorators, `@wraps` preserves metadata so each layer knows the original function.

sabnzbd/decorators.py
```diff
@@ -35,15 +35,21 @@
...
-            # Using the try/finally approach is 25% faster compared to using "with lock"
+            # Either use the supplied lock or the object-specific one
+            # Because it's a variable in the upper function, we cannot use it directly
+            lock_obj = lock
+            if not lock_obj:
+                lock_obj = getattr(args[0], "lock")
```
mnightingale, line 45:
```suggestion
            if lock is None:
                lock_obj = getattr(args[0], "lock")
            elif isinstance(lock, str):
                lock_obj = getattr(args[0], lock)
            elif callable(lock):
                lock_obj = lock(args[0])
            else:
                lock_obj = lock
```

mnightingale (commented):
I'm not sure I got those suggestions quite right, string one I can do without but it and/or callable may we wanted when objects have multiple locks protecting different attributes.

Here is the full version:

```py
LockType = Union[Lock, RLock, Condition]
LockGetter = Callable[[object], LockType]


def synchronized(lock: Optional[Union[LockType, LockGetter]] = None):
    def wrap(func: Callable):
        @functools.wraps(func)
        def call_func(*args, **kw):
            # Either use the supplied lock or the object-specific one
            # Because it's a variable in the upper function, we cannot use it directly
            if lock is None:
                lock_obj = getattr(args[0], "lock")
            elif isinstance(lock, str):
                lock_obj = getattr(args[0], lock)
            elif callable(lock):
                lock_obj = lock(args[0])
            else:
                lock_obj = lock

            # Using try/finally is ~25% faster than "with lock"
            try:
                lock_obj.acquire()
                return func(*args, **kw)
            finally:
                lock_obj.release()
```

`@functools.wraps(func)` so the debugger knows where it is.
May want to check what happens when used on a classmethod, it won't have args.

sabnzbd/decorators.py
Safihre: Why? This is also what Warp came up with but I don't see why we would need it.

mnightingale: I'd have liked to use it in assembler:

```py
def total_ready_bytes(self) -> int:
    with self.ready_bytes_lock:
        ...

def update_ready_bytes(self, nzf: NzbFile, delta: int) -> int:
    with self.ready_bytes_lock:
        ...

def clear_ready_bytes(self, *nzfs: NzbFile) -> None:
    with self.ready_bytes_lock:
        ...
```

Since those whole method calls need the lock, I'd like to do `@synchronized("ready_bytes_lock")` it's probably more useful than the callable way; I can't think of anywhere I would want that currently.

mnightingale: Also thinking about it the isinstance path should be above `if lock is None` in case the object has both.

sabnzbd/decorators.py
mnightingale: I suppose it's extra overhead that's only useful to developers so can do without this.

sabnzbd/decorators.py
Safihre: It's an improvement we can make later indeed :)

mnightingale:
I expect you're aware but this could help with multiple postproc concurrently (#3209), and postproc and downloader/queue fighting to update state.

PR #3265, opened 2026-01-06
Do not throttle flushing cache during shutdown

mnightingale:
You may want to adjust the timeout, I just used what the other threads have.

As discussed:

Append triggers at 5 of cache size, but it's going to vary a little, so at 4GB call that 200-400MB
Direct write is 75% of cache size capped at 500MB, so 1GB = 500M, 4GB = 500M, so allow 500-750MB

3 seconds seems reasonable to let it try to flush to disk, even modest SATA ssds should be able to do that.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -450,7 +450,7 @@ def trigger_restart(timeout=None):
 def save_state():
     """Save all internal bookkeeping to disk"""
     config.save_config()
-    sabnzbd.ArticleCache.flush_articles()
+    sabnzbd.ArticleCache.flush_articles(timelimit=3)
     sabnzbd.NzbQueue.save()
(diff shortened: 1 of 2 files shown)


Safihre:
But what happens after the timeout?

mnightingale:
Oh yeah... Ideally this would have the article level lock you're introducing and also use it elsewhere that on_disk is updated. Although I think the assembler should have stopped by this point.

Does it need anything or would you just use `allow_new_fetcher` instead?

```py
# After the deadline based loop

while self.__article_table:
    try:
        article, _ = self.__article_table.popitem()
        article.decoded = False
        article.remove_from_try_list(article.fetcher)
    except KeyError:
        # Could fail if already deleted by purge_articles or load_data
        logging.debug("Failed to flush item from cache, probably already deleted or written to disk")
```

Issue #3266, opened 2026-01-08
Queue stops downloading and items at 100% won't complete


Exophobias:
### SABnzbd version

4.6.0Beta2

### Operating system

Linux (TrueNAS Scale 25.10.1 - Goldeye)

### Using Docker image

linuxserver

### Description

I am using the **nightly** version, and I have noticed for the past couple of updates that items in my queue will stop downloading o,r even at 100% will not complete until I restart SABnzbd,

<img width="1720" height="165" alt="Image" src="https://github.com/user-attachments/assets/3899e123-a9f0-4038-b09f-ce6da495dfce" />

That image shows what it looks like, and it can look like that with 8 or more things in the queue.
I changed my logging to Debug a day or so ago. Here is the log: [sabnzbd.log](https://github.com/user-attachments/files/24488844/sabnzbd.log)

I restarted SABnzbd and grabbed the log again after the two items finished to see if it might show anything else that may help. 

[sabnzbd2.log](https://github.com/user-attachments/files/24488919/sabnzbd2.log)

mnightingale:
Thanks, for reporting, I saw the same when testing and reported as #3260

I'll send the PR to fix it shortly.

Safihre:
@mnightingale Also see a lot of this:
``` 
2026-01-07 22:13:12,224::INFO::[newswrapper:654] 78@news.giganews.com: Failed to close socket (error=EOF occurred in violation of protocol (_ssl.c:2427))
2026-01-07 22:13:12,225::INFO::[newswrapper:404] Looks like server closed connection: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:2427)
2026-01-07 22:13:12,226::INFO::[newswrapper:404] Looks like server closed connection: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:2427)
2026-01-07 22:13:12,226::INFO::[newswrapper:404] Looks like server closed connection: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:2427)
2026-01-07 22:13:12,226::INFO::[downloader:930] Thread 57@news.giganews.com: Server broke off connection
```

mnightingale:
That one is probably #3216, should be able to verify by not using unlocked_ssl because normal Python ignores certain EOF errors.

mnightingale:
[...]

PR #3267, opened 2026-01-08
Adjust monitored socket events as required to prevent hot looping

### LINKED ISSUE #3260: Downloader thread hot loops when no more requests but not force disconnected yet
### SABnzbd version

develop

### Operating system

linux, maybe macOS, does not appear to affect Windows but not sure why

### Using Docker image

None

### Description

When queue has finished, but items are still being assembled the sockets are still listening for read/write events because the item is still in the queue, so it hot loops and delays assembly.

Relevent parts:

https://github.com/sabnzbd/sabnzbd/blob/ecb36442d3a44d77e2dccc0a36d7b21fcdc0ab8e/sabnzbd/downloader.py#L702-L705

https://github.com/sabnzbd/sabnzbd/blob/ecb36442d3a44d77e2dccc0a36d7b21fcdc0ab8e/sabnzbd/newswrapper.py#L395-L401

I'll resolve it, just making note of it for now.

At least I think that's what the problem is, only reproduced it on Linux so far and at high speeds.

mnightingale:
Fixes #3260 #3266

`prepare_request` returns True if there is a request to send buffered or there is a request already in-flight we are waiting for a response for. So if it's False, the connection can be idled.

The problem was once all articles are downloaded, the connection was still active because the force disconnection isn't until the queue is empty. So select was returning the sockets as available for writing hundreds of times per second, starving the assembler of cpu resources.

The main culprit was this logic was incorrect, it waited until timeout.

```py
not self.next_request
and not self._response_queue
and (not server.active or server.restart or not self.timeout or time.time() > self.timeout)
```


Safihre:
The PR is marked as Draft, are there any concerns?

mnightingale:
I'm going to remove the modify_socket calls to make the logic simpler.

mnightingale:
Actually that would have made it worse because it's basically always available to write so would loop.

I am going to change it a little. I'm just testing the changes by repeating functional download tests for a while.

mnightingale:
OK I’m happy now, it has run through thousands of repeated tests.

Also, I added quite a few detailed comments about why it does certain things because it’s difficult to visualise that the socket could be monitoring for READ or READ | WRITE and if you are monitoring for WRITE but don’t have anything to write you are basically doing `if True: pass` hundreds if not thousands of times per second.

PR #3271, opened 2026-01-09
Fix queue cannot be loaded

### LINKED ISSUE #3269: Queue not restored on startup
### SABnzbd version

develop

### Operating system

macos

### Using Docker image

None

### Description

Appears to have been broken by https://github.com/sabnzbd/sabnzbd/commit/17dcff49b25aaf3988bbc28f46175981908f9211

```
  File "/sabnzbd/sabnzbd/filesystem.py", line 1107, in load_data
    data = pickle.load(data_file, encoding=sabnzbd.encoding.CODEPAGE)
  File "/sabnzbd/sabnzbd/nzb/article.py", line 221, in __setstate__
    self.lock = self.nzf.nzo.lock
                ^^^^^^^^^^^^
AttributeError: 'sabnzbd.nzb.file.NzbFile' object has no attribute 'nzo'
```

mnightingale:
Fixes #3269

I've used nzf level locks, it makes setting them simpler.

Part I missed initially is that `super().__setstate__(dict_.get("try_list", []))` is called from file and needs `self.lock` on article, so it needs to come after the locks have been restored.

I think I could make an argument for at least the lock initialisation lines being executed before anything else in `__setstate__` methods.

I've tried this with:
- Items added to the queue with no downloading
- Items added and partially downloaded, so trylist is restored

### DIFF
--- a/sabnzbd/nzb/article.py
+++ b/sabnzbd/nzb/article.py
@@ -19,6 +19,7 @@
 sabnzbd.article - Article and TryList classes for NZB downloading
 """
 import logging
+import threading
 from typing import Optional
 
 import sabnzbd
@@ -214,6 +215,7 @@ class Article(TryList):
             except KeyError:
                 # Handle new attributes
                 setattr(self, item, None)
+        self.lock = threading.RLock()
         super().__setstate__(dict_.get("try_list", []))
         self.fetcher = None
         self.fetcher_priority = 0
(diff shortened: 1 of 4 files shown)


mnightingale:
I should probably write a test

sabnzbd/nzb/article.py
Safihre, line 126: Also needs changing of course.

mnightingale:
Testing this is a pain because I needed to test try_list behaviour.
I used bits of other changes I was already working on, I couldn't see a way to test such state via Selenium.

But it has revealed that since article is always recreated first, it needs a temporary lock to restore it's try_list.

Safihre:
Thanks!

Issue #3216, opened 2025-12-05
Server closed connection with SSL


mnightingale:
### SABnzbd version

develop

### Operating system

N/A

### Using Docker image

None

### Description

If I use non-ssl connections this does not happen.

But with SSL I've noticed this getting raised occasionally, which causes it to recreate the connection - which may cause connection limit issues.

https://github.com/sabnzbd/sabnzbd/blob/1cde76433650c638c0807a6b8120ffba6db4c4da/sabnzbd/newswrapper.py#L299-L300

edit appears to be a unlocked_ssl_recv_into issue, and not new or related to pipelining

mnightingale:
edit: actually coming from ConnectionAbortedError('Failed to read data') at: 

https://github.com/sabnzbd/sabnzbd/blob/1cde76433650c638c0807a6b8120ffba6db4c4da/sabnzbd/downloader.py#L758-L761

Which is from unlocked_ssl_recv_into

mnightingale:
I can't find the cause, first SSL_read_ex returns 0 and SSL_get_error is an unexpected EOF.

But it's not a new error so 🤷

mnightingale:
I'm going to look into this a bit more, there are some things we don't do on the Python side.

https://github.com/python/cpython/blob/46295677a13f141b8d52cc36202384554adba65d/Lib/ssl.py#L1339-L1356

Notice this ignores the EOF error so I need to try passing that out of unlocked_ssl_recv_into.

https://github.com/python/cpython/blob/46295677a13f141b8d52cc36202384554adba65d/Lib/ssl.py#L1144-L1163

mnightingale:
Not stale, initial version required a lot of changes and openssl headers.
I'm going to see if I can trim it down at all and just handle the case we need to ignore the EOF.

mnightingale:
Branch I want to slim down to just the EOF errors https://github.com/sabnzbd/sabctools/compare/master...mnightingale:sabctools:feature/handle_ssl_errors not expected you to review it or anything, just showing it seems too fragile.

Unfortunately we have to deal with different Python version having different data structures, and it's probably not just minor version there could be changes in.

It needs openssl headers, of which we won't always get the same version as the system but constants between 1.1 and 3.0 seemed the same. It's only for translating error codes to reasons, it doesn't link against OpenSSL; just ERR_GET_LIB and ERR_GET_REASON.

[...]

PR #3272, opened 2026-01-10
Fix pipelining connection read/write logic errors

mnightingale:
This fixes several issues I've found while investigating #3266

- When write failed, self.next_request was already set to None so the request was lost and never sent, preventing progress
- macOS used KqueueSelector which creates separate selector events for READ and WRITE, which has logic issues because the rest of the code wasn’t expecting or written to handle that – things like read and write trying to reset the connection from different threads, or write being executed before read. I resolved by making macOS use selectors.SelectSelector.
- Newswrapper lock was recreated when \_\_init\_\_ was called; I didn’t find it causing an issue but best to prevent that
- A connection error (closed connection) during a write would timeout for the server default, on reads this is only 5 seconds, so I’ve made it consistent.
- Close might not call sock.close if sending QUIT failed; not 100% sure it matters but probasbly best to ensure the socket is closed regardless
- Removed a redundant generation check

The main other one is #3216 I'm not sure if these changes help or resolve that but it will take a bit longer to figure out.
edit: I don't see it occuring anything like as often, not at all in the last 24 hours, but it wasn't downloading that whole time. I suspect it was due to KqueueSelector, will maybe leave the underlying issue unresolved since it's been like this since 4.0

I think you mentioned it before, but shall I split reinitialising into a separate method, rather than checking if lock is already set? `def reset()` or something?

The best way I've found to test this failure was to restart my NNTP server.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -300,7 +300,7 @@ class Downloader(Thread):
         # macOS/BSD will default to KqueueSelector, it's very efficient but produces separate events for READ and WRITE.
         # Which causes problems when two receive threads are both trying to use the connection while it is resetting.
         if selectors.DefaultSelector is getattr(selectors, "KqueueSelector", None):
-            self.selector: selectors.BaseSelector = selectors.SelectSelector()
+            self.selector: selectors.BaseSelector = selectors.PollSelector()
         else:
             self.selector: selectors.BaseSelector = selectors.DefaultSelector()
 
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -299,11 +299,10 @@ class Downloader(Thread):
 
         # macOS/BSD will default to KqueueSelector, it's very efficient but produces separate events for READ and WRITE.
         # Which causes problems when two receive threads are both trying to use the connection while it is resetting.
-        self.selector: selectors.BaseSelector = (
-            selectors.SelectSelector()
-            if isinstance(selectors.DefaultSelector(), selectors.KqueueSelector)
-            else selectors.DefaultSelector()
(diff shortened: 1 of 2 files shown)


mnightingale:
This appears to have greatly improved the situation; In the last 36 hours I’ve done 50 TB from my local testing server and 1 TB from a real server both with test_download_1000MB.nzb.

I’m also not seeing the EOF errors either, so perhaps that was related to the full-duplex communication KqueueSelector was causing.

sabnzbd/downloader.py
```diff
@@ -297,7 +297,12 @@ def __init__(self, paused=False):
 
         self.force_disconnect: bool = False
 
-        self.selector: selectors.DefaultSelector = selectors.DefaultSelector()
+        # macOS/BSD will default to KqueueSelector, it's very efficient but produces separate events for READ and WRITE.
+        # Which causes problems when two receive threads are both trying to use the connection while it is resetting.
+        if selectors.DefaultSelector is getattr(selectors, "KqueueSelector", None):
```
Safihre, line 302: Does macOS not support any of the "better" ones like poll or epoll?

mnightingale: I’ve tested and macOS and TrueNAS Core, on Python 3.9 and both can use poll.

It’s important that we never use KqueueSelector which is why I did not just use sabnzbd.MACOS, at least not without rewriting the rest of the logic and adding locks in several places.

I also considered manually merging events to behave like the other selectors, but it didn’t seem worth the effort and probably slower than just using a different selector.

I’m not sure why the Windows binary build got stuck, not sure it’s related to these changes but will investigate this evening.

PR #3273, opened 2026-01-11
Fix bug when bandwidth limit is removed

mnightingale:
While testing I had restricted speeds by setting ‘Maximum line speed’ then while downloading I removed the restriction (empty field) and observed the speed collapse to 0 or almost because it’s a float and never actually hits 0 exactly, it takes quite a while to decay to displaying zero in the UI.

It remained stuck reading 0 for a while until I added a new limit.

Initially the check is True and it starts enforcing the speed limit, however the value of ` self.bandwidth_limit` then changes to 0 so it becomes `while sabnzbd.BPSMeter.bps > 0` and loops indefinitely.


Safihre:
Nice catch.

PR #3280, opened 2026-01-16
Handle non-fatal errors during read

mnightingale:
This needs the sabctools PR, as explained there it's possible to get write errors while reading and currently we treat that as a fatal error.


thezoggy:
prob time to cut a new alpha to bump/catch everything up to this as well?

Safihre:
Thought I already merged this one, oops!

PR #3281, opened 2026-01-16
Add assemble interval trigger

mnightingale:
This is based on feedback/logs I’ve seen where with many connections the 500MB direct write trigger was being hit, but because of the number of connections the next article wasn’t available, so although an assemble task was repeatably being enqueued to actual assembling was taking place.

I purposely haven't added the `contiguous_ready_bytes()` call to the direct write path because the trigger is much higher than append, so it would be calculating over a large number of articles on each call.

This changes the logic in `should_queue_nzf` in several related ways:
- Trigger a write (direct or append) every 5 seconds after the first part if the next article is available - helps slow connections configured with large cache sizes
- Similar to append, direct write only enqueues if the next article is decoded or on disk - avoids the issue described above, it might end up only writing one article but it's better than enqueuing a no-op
- Max direct write trigger raised to 1GB - the 5 second interval guards against the reason this was low previously, this allows fast connections to write less frequently

I also fixed a bug I noticed with the `finally`- if `nzf.prepare_filepath()`returns None the nzf was not removed from queued_nzf_non_contiguous or queued_nzf, so they wouldn’t assemble again until file_done. I flipped the `if filepath := nzf.prepare_filepath()` logic around to avoid having to indent and change many lines.


Safihre:
I'm gonna trust you on this one because the explanation seems logical 😊
But I remember years ago when I just started digging through the Sab internals I also made complex logic like this.
Over time, I ended up removing quite some of those again because there's always some setup somewhere that just doesn't fit the well thought out original plans.
We will have to see in the wild!

PR #3283, opened 2026-01-17
Add database indexes

mnightingale:
Currently the majority if not all of the history database queries require full scans of the database, adding a few indexes gives the database much less work to do for queries.

There are other queries that could benefit from indexes, but they are rarely used so I just focused on those updating by nzo_id, paginating history, and checking for duplicates.

This is probably the most important, the main fetch_history with no filters does something like

> EXPLAIN QUERY PLAN SELECT * FROM history WHERE archive is null ORDER BY completed desc LIMIT 10

SCAN history
USE TEMP B-TREE FOR ORDER BY

Without any indexes this involves creating temporary structures, scanning and copying ALL records where archive is null, then sorting by completed (the b-tree handles it) and returning the first 10, or skipping and returning 10.

SEARCH history USING INDEX idx_history_archive_completed (archive=?)

With indexes this doesn't use temporary structures, they are already sorted and filtered and it just needs to take records until it has 10.

SABs itself has some optimisations to avoid full api calls if nothing has changed, other API consumers do not.

> EXPLAIN QUERY PLAN SELECT EXISTS(
                SELECT 1
                FROM history
                WHERE duplicate_key = ? AND status != "Failed"
            ) as found

SCAN CONSTANT ROW
SCALAR SUBQUERY 1
SEARCH history USING INDEX idx_history_duplicate_key_active (duplicate_key=?)

> EXPLAIN QUERY PLAN SELECT EXISTS(
                SELECT 1
                FROM history
                WHERE md5sum = ? AND status != "Failed"            
                UNION ALL
                SELECT 1
                FROM history
                WHERE name = ? COLLATE NOCASE AND status != "Failed"
            ) as found

SCAN CONSTANT ROW
SCALAR SUBQUERY 2
COMPOUND QUERY
LEFT-MOST SUBQUERY
SEARCH history USING INDEX idx_history_md5sum_active (md5sum=?)
UNION ALL
SEARCH history USING INDEX idx_history_name_nocase_active (name=?)

SQLite does not like OR queries, it would have to pick one of the indexes for the query. By splitting this up both subqueries hit the indexes.
There is also a NOCASE index for the name matching, avoids the on demand LOWER(name) which would require table scans.

> EXPLAIN QUERY PLAN SELECT sum(bytes) FROM history WHERE completed > ?

SEARCH history USING COVERING INDEX idx_history_completed_bytes (completed>?)

The index contains all the data required for sum(bytes) queries, no need to lookup the record in the main table.

### DIFF
--- a/sabnzbd/database.py
+++ b/sabnzbd/database.py
@@ -413,13 +413,13 @@ class HistoryDB:
             SELECT EXISTS(
                 SELECT 1
                 FROM history
-                WHERE md5sum = ? AND status != ?
-            
-                UNION ALL
-            
+                WHERE md5sum = ?
+                  AND status != ?
+            ) OR EXISTS(
                 SELECT 1
                 FROM history
-                WHERE name = ? COLLATE NOCASE AND status != ?
+                WHERE name = ? COLLATE NOCASE
+                  AND status != ?
             ) as found
             """,
             (md5sum, Status.FAILED, name, Status.FAILED),
--- a/sabnzbd/database.py
+++ b/sabnzbd/database.py
@@ -395,16 +395,11 @@ class HistoryDB:
             SELECT EXISTS(
                 SELECT 1
                 FROM history
-                WHERE md5sum = ?
-                  AND status != ?
-            ) OR EXISTS(
-                SELECT 1
-                FROM history
-                WHERE name = ? COLLATE NOCASE
+                WHERE (name = ? COLLATE NOCASE OR md5sum = ?)
                   AND status != ?
             ) as found
             """,
-            (md5sum, Status.FAILED, name, Status.FAILED),
+            (name, md5sum, Status.FAILED),
         ):
             return bool(self.cursor.fetchone()["found"])
         return False
--- a/sabnzbd/database.py
+++ b/sabnzbd/database.py
@@ -119,7 +119,6 @@ class HistoryDB:
                     self.execute("PRAGMA user_version = 6;")
                     and self.execute("CREATE UNIQUE INDEX idx_history_nzo_id ON history(nzo_id);")
                     and self.execute("CREATE INDEX idx_history_archive_completed ON history(archive, completed DESC);")
-                    and self.execute("CREATE INDEX idx_history_completed_bytes ON history(completed, bytes);")
                     and self.execute(
                         "CREATE INDEX idx_history_duplicate_key_active ON history(duplicate_key) WHERE status != 'Failed';"
                     )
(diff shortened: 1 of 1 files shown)


sabnzbd/database.py
```diff
@@ -114,6 +114,22 @@ def connect(self):
                 _ = self.execute("PRAGMA user_version = 5;") and self.execute(
                     "ALTER TABLE history ADD COLUMN time_added INTEGER;"
                 )
+            if version < 6:
+                _ = (
+                    self.execute("PRAGMA user_version = 6;")
+                    and self.execute("CREATE UNIQUE INDEX idx_history_nzo_id ON history(nzo_id);")
+                    and self.execute("CREATE INDEX idx_history_archive_completed ON history(archive, completed DESC);")
+                    and self.execute("CREATE INDEX idx_history_completed_bytes ON history(completed, bytes);")
+                    and self.execute(
```
Safihre: Should we apply the KISS-method and only do the first 3?
The rest is nice, but how often are those actually triggered?
Even the completed_bytes, when is that triggered?

mnightingale: For new tables, the `CREATE UNIQUE INDEX` could be a table constraint, so no index and smaller.
But I avoided it to keep new and upgraded databases the same schema.

Completed bytes can go, I hadn’t realised where it was used - I thought it was the totals at the bottom of the UI but it’s not.

The 3 duplicate indexes are for `have_duplicate_key` and `have_name_or_md5sum`, so they’re only used when an item is added so only really a problem when users have very large histories.

The COLLATE NOCASE will be the largest because it basically duplicates the name for each record. I would suggest that one could be removed but keep the other two (duplicate_key and md5sum), with a slight restructing of the have_name_or_md5sum queryit can at least benefit from filtering out the failed items and short circuit on an md5sum match.

850 items are 1024 KB
Without NOCASE index 1122 KB (+10%)
With NOCASE index 1171 KB (+4%)

I don't think the size or impercetable increased insert/update times should be a concern, but certainly there is a balance between indexing nothing and everything.

Safihre: But why the duplicate check related ones? They are executed so rarely. Then we also don't need the query changes.

I only want to keep the top ones.
And why not index the name of the Failed ones?

mnightingale: The duplicate check indexes with `WHERE status != "Failed"` would only index rows that matter to the queries, and the duplicate checks are only interested in non-failure duplicates.

I’ve removed the duplicate indexes, but I’ve kept a minor change to two queries because SELECT EXISTS can stop scanning the whole table once it finds a match and even though there isn’t an index COLLATE NOCASE is cheaper than calling a function for ever row - it’s the same comparison only A-Z compared as a-z

Safihre:
Nice!

PR #3289, opened 2026-01-22
Handle SSLWantWriteError exceptions and buffer writes

mnightingale:
I think this handles the ssl.SSLWantWriteError.
I also added the send buffer back just a little neater than before.

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -75,6 +75,7 @@ class NewsWrapper:
         "selector_events",
         "lock",
         "generation",
+        "tls_wants_write",
     )
 
     def __init__(self, server: "sabnzbd.downloader.Server", thrdnum: int, block: bool = False, generation: int = 0):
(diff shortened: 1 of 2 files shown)


sabnzbd/newswrapper.py
```diff
@@ -375,6 +377,14 @@ def write(self):
         server = self.server
 
         try:
+            # Flush any buffered data
+            if self.nntp.write_buffer:
+                sent = self.nntp.sock.send(self.nntp.write_buffer)
+                self.nntp.write_buffer = self.nntp.write_buffer[sent:]
+                # If buffer still has data, wait for next write opportunity
+                if self.nntp.write_buffer:
```
Safihre, line 385: Can you add a Debug line here to see if we ever need this?

Safihre: Oh wait that's below. When is this send?

Safihre: Got it.

PR #3291, opened 2026-01-22
Fix adding sockets mid-connect

### LINKED ISSUE #3288: Traceback: TypeError: unlocked_ssl_recv_into() argument 1 must be SSLSocket, not socket
### SABnzbd version

5.0.0Beta1 [d215d4b]

### Operating system

docker

### Using Docker image

linuxserver

### Description

Traceback, and downloads stop. Pausing and unpasuing does not help. I'm going to restart the docker.

```
2026-01-22 18:50 Fatal error in Downloader
  File "/app/sabnzbd/sabnzbd/downloader.py", line 745, in process_nw_worker
    self.process_nw(*nw_queue.get())
  File "/app/sabnzbd/sabnzbd/downloader.py", line 758, in process_nw
    self.process_nw_read(nw, generation)
  File "/app/sabnzbd/sabnzbd/downloader.py", line 774, in process_nw_read
    n, bytes_pending = nw.read(nbytes=bytes_pending, generation=generation)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/sabnzbd/sabnzbd/newswrapper.py", line 298, in read
    bytes_recv = sabctools.unlocked_ssl_recv_into(self.nntp.sock, self.decoder)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unlocked_ssl_recv_into() argument 1 must be SSLSocket, not socket
ERROR 2026-01-22 18:50 Fatal error in Downloader
  File "/app/sabnzbd/sabnzbd/downloader.py", line 745, in process_nw_worker
    self.process_nw(*nw_queue.get())
  File "/app/sabnzbd/sabnzbd/downloader.py", line 758, in process_nw
    self.process_nw_read(nw, generation)
  File "/app/sabnzbd/sabnzbd/downloader.py", line 774, in process_nw_read
    n, bytes_pending = nw.read(nbytes=bytes_pending, generation=generation)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/sabnzbd/sabnzbd/newswrapper.py", line 298, in read
    bytes_recv = sabctools.unlocked_ssl_recv_into(self.nntp.sock, self.decoder)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unlocked_ssl_recv_into() argument 1 must be SSLSocket, not socket

```

Image

mnightingale:
Fixes #3288 mostly reverts to 4.5.5 behaviour: https://github.com/sabnzbd/sabnzbd/blob/11ba9ae12ade8c8f2abb42d44ea35efdd361fae5/sabnzbd/downloader.py#L609-L617

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1415,7 +1415,7 @@ def test_nntp_server_dict(kwargs: dict[str, Union[str, list[str]]]) -> tuple[boo
         while test_server.active:
             nw.write()
             nw.read(on_response=on_response)
-            if nw.connected:
+            if nw.ready:
                 break
 
     except socket.timeout:
@@ -1521,10 +1521,10 @@ def build_status(calculate_performance: bool = False, skip_dashboard: bool = Fal
     info["servers"] = []
     # Servers-list could be modified during iteration, so we need a copy
     for server in sabnzbd.Downloader.servers[:]:
-        activeconn = sum(nw.connected for nw in server.idle_threads.copy())
+        activeconn = sum(nw.ready for nw in server.idle_threads.copy())
         serverconnections = []
         for nw in server.busy_threads.copy():
-            if nw.connected:
+            if nw.ready:
                 activeconn += 1
             if article := nw.article:
                 serverconnections.append(
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -659,7 +659,7 @@ class Downloader(Thread):
                         if nw.connected:
                             if nw.prepare_request():
                                 self.add_socket(nw)
-                        else:
+                        elif not nw.nntp:
                             try:
                                 logging.info("%s@%s: Initiating connection", nw.thrdnum, server.host)
                                 nw.init_connect()
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -656,7 +656,7 @@ class Downloader(Thread):
                         if not server.get_article(peek=True):
                             break
 
-                        if nw.connected or nw.socket_connected:
+                        if nw.connected:
                             self.add_socket(nw)
                         elif not nw.nntp:
                             try:
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -656,10 +656,12 @@ class Downloader(Thread):
                         if not server.get_article(peek=True):
                             break
 
-                        if nw.connected or nw.nntp:
+                        if nw.connected:
                             if nw.prepare_request():
                                 self.add_socket(nw)
                         else:
+                            nw.server.idle_threads.remove(nw)
+                            nw.server.busy_threads.add(nw)
                             try:
                                 logging.info("%s@%s: Initiating connection", nw.thrdnum, server.host)
                                 nw.init_connect()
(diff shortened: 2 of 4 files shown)


Safihre:
Wonder if there's some way we can unit test this, does AI have some suggestions?
They are all such difficult issues to reproduce.

mnightingale:
I’ll have a look at how it could be tested; I can’t see an obvious way.

The issue I came across is the local side of the socket can be created (socket.socket), connect could succeed, but TLS negotiation could fail, SAB warns about it but leaves the socket connected or with just the local side which I haven't checked but presumably responds to selectors/select.select.

`nw.connected` is a bit misleading, it’s connected in the sense it has finished connecting the socket but also that it has finished the auth flow and is ready to serve requests.

I realised that the edge case I was previously trying to fix is that it is possible for `nw.nntp` and the underlying socket/tls to be connected but not have ever had a request to trigger the auth flow. 

For example, if you have 100 connections, but few requests it's possible in the time connections are established the requests have been exhausted leaving connections in limbo connected to the server but never starting the auth flow and `nw.connected` is False. With the PR currently they remain in limbo until timeout.

So, although I think it's sensible to check if `nw.nntp` is not None to avoid races and clearing it when a `nntp.connect()` fails. I think I need to either rename `nw.connected` in 15 places to `nw.ready` or similar and repurpose `nw.connected` to just mean the socket is connected and if `not nw.ready` then it is ready for the auth flow and can accept requests. Maybe just a `nw.socket_connected` will be sufficient.

mnightingale:
The test failure isn't related to the changes, failed to connect to sabnzbd.org or something.

With some help I’ve added tests for the expected state transitions so at least I’ll know when I next break it: 😄

There are 3 states:

- nw.connected or nw.socket_connected => ready for requests or connected but not completed auth flow (needs add_socket to progress)
- not nw.nntp => no connection in progress, safe to start one
- nw.nntp and not socket_connected => connection in progress

Then `hard_reset` clears nw.nntp instance on connect failure.

Safihre:
Shouldn't it be on the nw.nntp.connected?
Or that makes it more complicated?

mnightingale:
It's a little problematic to store the state at nntp level, because then everywhere has to check is nw.nntp is None first so to avoid a race it'd have to do `if nw.connected or (nntp := nw.nntp) and nntp.connected:`

I think it should probably be exposed as a NewsWrapper level concern rather than NNTP, the rest of the application is only interested in the NewsWrapper not how it internally splits the responsibilities up i.e. treat it more like nw._nntp

I'd perhaps suggest renaming nw.connected to nw.ready and nw.socket_connected to nw.connected to better represent what they mean, but that means changing it in ~15 places.

Safihre:
Let's do that, change it to ready and connected.
Cleaner 😊

mnightingale:
Latest commit changed it to ready and connected, and documented the two properties.

I also made the check in downloader a little simpler.

`nw.ready` => tcp/tls socket established
`not nw.nntp` => no connection in progress

PR #3294, opened 2026-01-28
Add more logging to stop_idle_jobs

mnightingale:
I only anticipate the `logging.debug` lines being temporary, I'm trying to figure out what the state it is stuck in is.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -221,12 +221,7 @@ class Assembler(Thread):
         # Always write
         if (override_trigger or article_has_first_part) and filename_checked and not import_finished:
             return True
-        next_ready = (
-            (next_index := nzf.assembler_next_index) >= 0
-            and next_index < len(nzf.decodetable)
-            and (next_article := nzf.decodetable[next_index])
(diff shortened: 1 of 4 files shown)


Safihre:
That macOS job is failing quite often, maybe worth a look? It succeeded after retry.

mnightingale:
I saw the other comment, the timeout issue was resolved.

I've also noticed arm64 snaps taking ages or timing out, same happens locally even on installing an Ubuntu VM - terribly slow apt ports.ubuntu.com - not sure if we could setup caching of the lxc image/container, it could be too big.

Probably hold off on this PR for now, I have possibly found some deadlocks in nzb/* to fix, the locks are correct but `@synchronized()` can mean they are locked in different orders.

For example allow_new_fetcher locks nzf.lock then nzf.nzo.lock but NzbObject.remove_article locks nzf.lock then nzo.lock.

It needs to be consistent, probably nzo > nzf > article (shares nzf) so which I can't currently do with `@synchronized()` so it will probably need with context - which from my benchmarking of some future nzbqueue changes is is much faster than the decorator/try-catch approach.

Will try and send a new PR in a few hours.

PR #3296, opened 2026-01-29
Bind socket throughout test but don't listen and configure a timeout

mnightingale:
Fixes recent test failure, timeout was too high the connect thread finishes after the test.

It was logging the expected connect failure, but the test had already failed.

```
WARNING  root:newswrapper.py:673 Failed to connect: [Errno 61] Connection refused 1@127.0.0.1:49216 ()
```


Safihre:
Ah you already saw 😂

PR #3298, opened 2026-01-29
Fix nzb deadlocks

mnightingale:
Fixes several lock ordering issues, and simplifies some to avoid any deadlock risk.

The order locks are acquired must always be NzbObject > NzbFile
This means `@synchronized()` cannot always be used.

**allow_new_fetcher** nzf.lock => nzo.lock
**remove_article** nzo.lock => nzf.lock

This is the one most likely to occur, missing articles on multiple receive threads.

**contiguous_offset** nzf.lock => file_lock
**Assember.open** file_lock => nzf.lock (contiguous_offset)
**Assember.assemble_article** file_lock => nzf.lock

A few other similar ordering problems.

### DIFF
--- a/sabnzbd/nzb/file.py
+++ b/sabnzbd/nzb/file.py
@@ -247,26 +247,25 @@ class NzbFile(TryList):
 
         Note: there could be non-sequential direct writes already beyond this point
         """
-        with self.file_lock:
-            # If last written article has valid yenc headers
-            if self.assembler_next_index:
-                article = self.decodetable[self.assembler_next_index - 1]
-                if article.on_disk and article.data_size:
-                    return article.data_begin + article.data_size
+        # If last written article has valid yenc headers
+        if self.assembler_next_index:
+            article = self.decodetable[self.assembler_next_index - 1]
+            if article.on_disk and article.data_size:
+                return article.data_begin + article.data_size
 
-            # Fallback to summing decoded size
-            offset = 0
-            for article in self.decodetable[: self.assembler_next_index]:
-                if not article.on_disk:
-                    break
-                if article.data_size:
-                    offset = article.data_begin + article.decoded_size
-                elif article.decoded_size is not None:
-                    # queues from <= 4.5.5 do not have this attribute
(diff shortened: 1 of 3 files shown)


sabnzbd/nzb/file.py
```diff
@@ -247,26 +247,25 @@ def contiguous_offset(self) -> int:
...
-                    break
-                if article.data_size:
-                    offset = article.data_begin + article.decoded_size
-                elif article.decoded_size is not None:
-                    # queues from <= 4.5.5 do not have this attribute
-                    offset += article.decoded_size
-                elif os.path.exists(self.filepath):
-                    # fallback for <= 4.5.5 because files were always opened in append mode, so use the file size
-                    return os.path.getsize(self.filepath)
+        # If last written article has valid yenc headers
```
Safihre, line 250: Why does it no longer need a `file_lock`?

mnightingale: contiguous_offset itself never actually needed it:
nzf.lock protects article state (assembler_next_index, on_disk, etc.)
file_lock protects IO operations on the file (open, lseek+write on Windows)

contiguous_offset reads article state, so it needs nzf.lock
It doesn’t do any IO so doesn’t need file_lock

It is only called from Assember.open which has file_lock - but again this method doesn’t need that lock itself.
Assember.open does need the stable state of a critical section for file open + lseek, but that's the caller’s decision and adding file_lock to contiguous_offset wouldn't necessarily mean it was safe. When correctly used the caller is always going to hold file_lock already, else they risk a race/stale state and adding file_lock to contiguous_offset wouldn't prevent that.

For example, this is wrong:

```py
# if contiguous_offset acquired file_lock; not safe
offset = contiguous_offset()

with nzf.file_lock:
    fd = os.open(...)
    os.lseek(fd, offset, os.SEEK_SET)  # Unsafe because something could have opened and written to the file
```

It was removed because it’s unnecessary, but keeping it added a deadlock risk i.e. some future change using it elsewhere and causing a lock order deadlock.

PR #3300, opened 2026-01-30
Remove sleep when cache is full

mnightingale:
This sleep can cause problems and doesn't really provide any benefit.

For example I've seen a slow to respond backup server with few connections cause the 0 priority to get way ahead and cause the whole download thread to sleep, providing no benefit.

KISS; the thread already monitors cache usage and performs a allow_non_contiguous assemble when required.

### DIFF
--- a/sabnzbd/articlecache.py
+++ b/sabnzbd/articlecache.py
@@ -230,9 +230,6 @@ class ArticleCache(threading.Thread):
         if self.__cache_limit and self.__direct_write and sabnzbd.Assembler.assemble_article(article, data):
             with article.nzf.nzo.lock:
                 article.nzf.nzo.saved_articles.discard(article)
-            if not self.shutdown:
-                sabnzbd.Assembler.process(article.nzf.nzo, article.nzf, article=article, override_trigger=True)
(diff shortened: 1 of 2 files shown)


sabnzbd/articlecache.py
```diff
@@ -230,9 +230,6 @@ def __flush_article_to_disk(self, article: Article, data: bytearray):
         if self.__cache_limit and self.__direct_write and sabnzbd.Assembler.assemble_article(article, data):
             with article.nzf.nzo.lock:
                 article.nzf.nzo.saved_articles.discard(article)
-            if not self.shutdown:
-                sabnzbd.Assembler.process(article.nzf.nzo, article.nzf, article=article, override_trigger=True)
```
Safihre, line 234: `override_trigger` can be removed then?

mnightingale: Yes done, it didn’t really help either because it prevented the main thread from queuing allow_non_contiguous assembles by queuing a normal one that was probably past the trigger value anyway but hadn’t been making assemble progress so far.

PR #3304, opened 2026-02-03
Fix server retention not always applied

### LINKED ISSUE #3302: server retention time not honored
### SABnzbd version

5.0.0b1

### Operating system

linux

### Using Docker image

None

### Description

Grabbed the 100MB test download (currently 73 days old according to the history entry), and noticed a bunch of 430 responses in the debug log. Turns out there are connections being made to 2 servers despite those having a retention time limit of 30 days. The servers in question are at the priority 5 - the highest in my current config - and share that level with 3 other servers without a limit on retention time.

Image

mnightingale:
Fixes #3302

The peek option sees if there is a request available without consuming it from `article_queue` but peek was not applying the retention preference which will result subsequent calls taking articles from the prefetch bypassing the retention check entirely and potentially going to an ineligible server.

### DIFF
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -191,6 +191,9 @@ class Server:
                 if self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
                     if not peek:
                         sabnzbd.Downloader.decode(article)
+                    # sabnzbd.NzbQueue.get_articles stops after each nzo with articles.
+                    # As a result, if one article is out of retention, all remaining
+                    # entries in article_queue will also be out of retention.
                     while self.article_queue:
                         sabnzbd.Downloader.decode(self.article_queue.pop())
                 else:
--- a/sabnzbd/downloader.py
+++ b/sabnzbd/downloader.py
@@ -188,8 +188,9 @@ class Server:
             if self.article_queue:
                 article = self.article_queue[0] if peek else self.article_queue.popleft()
                 # Mark expired articles as tried on this server
-                if not peek and self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
-                    sabnzbd.Downloader.decode(article)
+                if self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
+                    if not peek:
+                        sabnzbd.Downloader.decode(article)
                     while self.article_queue:
                         sabnzbd.Downloader.decode(self.article_queue.pop())
                 else:


sabnzbd/downloader.py
```diff
@@ -188,8 +188,9 @@ def get_article(self, peek: bool = False):
             if self.article_queue:
                 article = self.article_queue[0] if peek else self.article_queue.popleft()
                 # Mark expired articles as tried on this server
-                if not peek and self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
-                    sabnzbd.Downloader.decode(article)
+                if self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
+                    if not peek:
+                        sabnzbd.Downloader.decode(article)
```
Safihre, line 193: I feel stupid, but I'm not understanding why the while shouldn't also be in the if?

mnightingale: I didn’t write that part originally, but it also confused me a while ago - it just looks wrong.

I believe it’s because each call to populate the prefetch only selects from at most one NzbObject so if one article is beyond the retention, all of them must be.

mnightingale: Verified the above (line 710). I believe the reason for this is otherwise it would load first articles for the whole queue, triggering loading each NzbFile into memory.

https://github.com/sabnzbd/sabnzbd/blob/e1ea4f1e7eb6fb47576eb0c97e7c9b99421fb796/sabnzbd/nzbqueue.py#L699-L713

I can add a brief comment explaining that if you'd like?

Safihre: Actually that wasn't even my problem with it 😂
I did a small stylistic change, now it looks more clear to me. If you're ok with it?
Although now it will only do the retention check after the connect..

mnightingale: Your commit changes it back to how it's currently broken, not filtering on the peek with `sabnzbd.NzbQueue.get_articles` path.

The second `if peek` check needs to come after the retention check, which if you want to break the ternary up would be:

```py
        if self.article_queue:
            if peek:
                return True
            return self.article_queue.popleft()

        if self.next_article_search < time.time():
            # Pre-fetch new articles
            sabnzbd.NzbQueue.get_articles(self, sabnzbd.Downloader.servers, _ARTICLE_PREFETCH)
            if self.article_queue:
                if peek:
                    article = self.article_queue[0]
                else:
                    article = self.article_queue.popleft()

                # Mark expired articles as tried on this server
                if self.retention and article.nzf.nzo.avg_stamp < time.time() - self.retention:
                    if not peek:
                        sabnzbd.Downloader.decode(article)
                    while self.article_queue:
                        sabnzbd.Downloader.decode(self.article_queue.pop())
                else:
                    if peek:
                        return True
                    return article
            else:
                # No available articles, skip this server for a short time
                self.next_article_search = time.time() + _SERVER_CHECK_DELAY
        return None
```

I don't really like the Union return, gives more to reason about and be aware of the behaviour.
I'd prefer Optional["sabnzbd.nzb.Article"]

mnightingale: Are you happy for me to change it as above but return Optional["sabnzbd.nzb.Article"]

Maybe in the future I'll suggest shifting the responsibilities around, make `sabnzbd.NzbQueue.get_articles` a generator and let Server.get_article populate article_queue after applying the retention filter.

Safihre: I removed my changes. This code is indeed unclear, maybe a comment here or there would help :)

mnightingale: Added a comment, but the logic is probably something I'll address in the future because I've always found it odd that `sabnzbd.NzbQueue.get_articles(server, servers, ...)` queues articles that are not eligible for the given server.

PR #3313, opened 2026-02-08
Add socket for read/write if next request or partial write

mnightingale:
SSLWantReadError could modify for EVENT_READ.
But if it's the last request next_request will be None, so need to also check nntp.write_buffer


Safihre:
Add that to the comment so we won't forget in many years :)

PR #3321, opened 2026-02-16
bump 7zip and par2

thezoggy:
7zip 26.0
```
26.00          2026-02-12
-------------------------
- improved code for ZIP, CPIO, RAR, UFD, QCOW, Compound.
- 7-Zip File Manager: improved sorting order of the file list. It uses file name as secondary sorting key.
- 7-Zip File Manager: improved Benchmark to support systems with more than 64 CPU threads.
- the bug was fixed: 7-Zip could not correctly extract TAR archives containing sparse files.
- some bugs were fixed.
```

par2cmdline-turbo 1.4.0
```
   Update base par2cmdline to v1.1.1
```
which is mainly to resolve unicode/long path issues on windows.


Safihre:
Thanks!

PR #3322, opened 2026-02-16
Reduce assembler trigger

mnightingale:
After much more testing, increasing this so much puts too much pressure on low to mid range consumer SSDs.
Along with the other changes to make the trigger point more meaningful i.e. a measure of contiguous articles, smaller sizes are not having the impact seen previously.

This basically returns the trigger to the 4.5.5, 5% of cache size and the same value is used for both assemble modes to keep it simpler.

The other change is just to only do the contiguous_ready_bytes() check if the total bytes ready is past the trigger.


sabnzbd/assembler.py
```diff
@@ -91,27 +87,26 @@ def stop(self):
...
+            to_units(self.assembler_trigger),
             to_units(self.delay_trigger),
         )
 
     def change_direct_write(self, direct_write: bool) -> None:
-        self.direct_write = direct_write and self.direct_write_trigger > 1
+        self.direct_write = direct_write and self.assembler_trigger > 0
         self.calculate_delay_trigger()
 
+    def calculate_assembler_trigger(self):
```
Safihre: Do we need a function since we only call it once?

PR #3326, opened 2026-02-19
Allow direct write usage without the cache

mnightingale:
I was not intending implementing this for 5.0.0 because I thought it would need to keep files open which would make it overly complicated. However, it appears to work perfectly fine opening the file for each article - which is what it would be doing previously to write temporary files anyway.

This is not usually a recommended setup because it had a temporary file penalty (write, read, write) and constant IO activity. However, when using SSD storage, it seems to work well[^1] almost indistinguishable from using the cache (even at 2GB/s+)

I will need to update the wiki a little bit (cache not required, but not recommended for HDD users) - will do so in a day or two.

[^1]: Windows suffers a little (no pwrite) but still managed 700+MB/s on my system. Opening and closing files so much may trigger antivirus but that is no different to when using temporary files.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -88,7 +88,6 @@ class Assembler(Thread):
         """Called when cache limit changes"""
         self.cache_limit = limit
         self.assembler_trigger = max(1, int(self.cache_limit * ASSEMBLER_TRIGGER_PERCENTAGE))
-        self.calculate_delay_trigger()
         self.change_direct_write(cfg.direct_write())
         logging.debug(
             "Assembler trigger=%s, delay=%s",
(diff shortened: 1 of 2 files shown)


sabnzbd/assembler.py
```diff
@@ -88,7 +88,6 @@ def new_limit(self, limit: int):
         """Called when cache limit changes"""
         self.cache_limit = limit
         self.assembler_trigger = max(1, int(self.cache_limit * ASSEMBLER_TRIGGER_PERCENTAGE))
-        self.calculate_delay_trigger()
```
Safihre, line 91: Why should we not calculate it when the limit changes?

mnightingale: The next line calls `change_direct_write` which as a side effect does the trigger calculation

Issue #3287, opened 2026-01-22
Downloads not Completing Again


Exophobias:
### SABnzbd version

5.0.0Beta1 (nightly - Commit: <HASH>6966a37a6f6a1fc)

### Operating system

Linux (TrueNAS 25.10.1)

### Using Docker image

linuxserver

### Description

I had a previous issue on this that was resolved, but it has come back with recent updates on the **nightly** version.
Some media is close to completion or is already completed, but has stopped downloading.

<img width="1709" height="355" alt="Image" src="https://github.com/user-attachments/assets/4c2a65ea-3528-487d-b36a-6a7f9c3659f6" />

The image shows that it is still downloading at 44 MB/s, but my TrueNAS dashboard is showing that it is not downloading at all anymore.

<img width="537" height="202" alt="Image" src="https://github.com/user-attachments/assets/2fa6114e-19df-449c-a604-e9f2d0bd71c9" />

I deleted all of my sab logs to ensure they were on the latest version/commit available to me in TrueNAS and restarted sab and verified the issue was still happening. So the log attached to this report is all from the commit noted in the log.

[sabnzbd.log](https://github.com/user-attachments/files/24786794/sabnzbd.log)

mnightingale:
It may have been caused by https://github.com/sabnzbd/sabnzbd/commit/a19b3750e37491029159757a7d6bfbbe20da3520 previously you were triggering this a lot, it caused connections to be closed and reconnected unnecessarily. But I'm also suspicious of https://github.com/sabnzbd/sabnzbd/commit/bf412371358ed3773c5f0fff65400c9d1547bdc9 given then 430/451 errors.

I'll take a look this evening, but I think I see how I've caused a TLS deadlock and should register for EVENT_WRITE and change the read/write logic.

Exophobias:
Let me know when you've pushed an update for it. I have had to move my sabnzbd back to `latest` branch so that my queue could complete. I will go back to nightly to test and help correct this issue once you have a potential fix in place.

mnightingale:
linuxserver nightly should include what I hope resolves the problem, let me know how you get on.

If it does get stuck, please could you see if the connections tab from the status/wrench/"s" hotkey screen shows any active connections that think they're working on an article.
If so be sure to wait until the configured server timeout (60 seconds in your config) and see if it's still stuck.

[...]

PR #3331, opened 2026-02-21
Fix queuing requests on reset connections

### LINKED ISSUE #3327: Received unknown status code 0 for article
### SABnzbd version

5.0.0Beta1 [507edc3]

### Operating system

Debian 12

### Using Docker image

None

### Description

```
WARNING 8 hours ago 1@news-eu.newshosting.com: Received unknown status code 0 for article cokOvJCsqamNVTgLhmeBhkOLHacljIL@xsjAbrbRENJz-r-cb.DmNji
--
WARNING 8 hours ago 2@news-eu.newshosting.com: Received unknown status code 0 for article IRNBqNOFwzUTZlYhtCxTTUjIQTZKfCZekWjhIUgUoU@6DfYB.SHW.W$LSzz.BhIK
```
Queue frozen ever since

mnightingale:
Fixes #3327 and #3287

`Received unknown status code 0 for article` means the decoder saw a response line but couldn't extract the status code from the first 3 characters which doesn’t match the nntp protocol. The status_code remains 0 and passed back to Python and should be handled as the unknown state and reset.

I did actually see this error from a real server, an nzb failed a few times then a few minutes later worked fine. I'm not sure if I was hitting different server nodes or it had self healed itself. Unfortunately I didn't get a chance to dump the responses.

I've tested with a server which always responds with in invalid response for a specific article.

The problem is this trigger the `allow_new_fetcher` path, resets `article.tries` and retries the same server, which could have the same issue again.
I've changed it to `retry_article=response.status_code != 0` so that it moves on to a different server, it probably shouldn't but I'm not sure what the behaviour should be for required servers.

https://github.com/sabnzbd/sabnzbd/blob/9b1b503ce6f2b70df39d2296aed4b7e296f2f18f/sabnzbd/newswrapper.py#L499-L507

`not self.server.required` is part of the issue, even if it remembered tries if it's a required server it would still retry indefinitely.
`article.allow_new_fetcher()` resets `article.tries`

So, this works but it doesn't seem ideal because if a server has some temporary internal error and is returning invalid responses but is required we shouldn't just move on to lower priority servers but I'm not sure what the behaviour should be?

Perhaps if status is 0 it should increment `server.bad_cons` and trigger `maybe_block_server`? But if there is only one required server it could still be stuck.

---

The read part is the main bug, the most common path is `on_response` may reset the connection, then `prepare_request` sets `next_request` on a NewsWrapper with no nntp instance and the request gets overwritten during the authentication phase.

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -262,7 +262,11 @@ class NewsWrapper:
 
                 # Ditch this thread, we don't know what data we got now so the buffer can be bad
                 sabnzbd.Downloader.reset_nw(
-                    self, f"Server error or unknown status code: {response.status_code}", wait=False, article=article
+                    self,
+                    f"Server error or unknown status code: {response.status_code}",
+                    wait=False,
+                    article=article,
+                    retry_article=response.status_code != 0,  # could be 0 if response violates nntp protocol see #3327
                 )
                 return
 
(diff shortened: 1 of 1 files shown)


mnightingale:
I'm going to split this in two

PR #3334, opened 2026-03-02
Improve direct write helpful warning message and add link to the wiki

mnightingale:
I think it makes the 'what and why' a little clearer and hopefully including the wiki link will reduce users needing to seek support.


Safihre:
👍

PR #3339, opened 2026-03-08
Ignore missing data and consume OSError

mnightingale:
First is an incorrect logic change from [4.5.5](https://github.com/sabnzbd/sabnzbd/blob/77f7490aeae60cce9de0afb7ecd3f573bae33946/sabnzbd/assembler.py#L181-L195)

`sabnzbd.ArticleCache.load_article` - load and remove data from the cache
The write raises `errno.ENOSPC` which is caught in assemble, but that article data is now lost, `article.on_disk` is not set and will be attempted again on the next assemble for the file.
Hence, the previous `break` behaviour was incorrect, if the data which is expected because of `article.decoded` is not present it will never get any further.
The `file_done` and `allow_non_contiguous` checks are redundant.

A possibly better solution I can implement but different to 4.5.5 would be to catch the IOError and put the data back in the cache, but it would still have to be allowed to silently fail - if the cache is full it'll do `__flush_article_to_disk => assemble_article (fail) => sabnzbd.filesystem.save_data (silent fail)`. Or skip the known failure and go straight to sabnzbd.filesystem.save_data. I think that's likely to introduce bugs or be complicated though, it might work ok for normal priority because it'll pause, but force will soldier on and complete the job without writing everything.

---

The second is if the cache is full `__flush_article_to_disk` will call `assemble_article` which could fail with errno.ENOSPC, so it needs to catch that and report a failure to revert to `sabnzbd.filesystem.save_data` which will fail silently.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -455,11 +455,8 @@ class Assembler(Thread):
                     cfg.direct_write.set(False)
                     return False
                 Assembler.write(fd, None, nzf, article, data)
-            except FileNotFoundError:
-                # nzo has probably been deleted, ArticleCache tries the fallback and handles it
-                return False
-            except IOError:
-                # Probably not enough disk space
+            except OSError:
+                # nzo has probably been deleted or not enough disk space, ArticleCache tries the fallback and handles it
                 return False
             finally:
                 os.close(fd)
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -408,8 +408,8 @@ class Assembler(Thread):
                         continue
                     break
 
-                # Could be empty in case nzo was deleted or a previous write attempt encountered errno.ENOSPC and data
-                # was removed from the cache but could not be written to disk.
+                # Could be empty in case nzo was deleted or a previous write attempt failed and the data was removed
+                # from the cache but could not be written to disk.
                 data = load_article(article)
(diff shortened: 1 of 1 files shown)


sabnzbd/assembler.py
```diff
@@ -408,17 +408,12 @@ def assemble(nzo: NzbObject, nzf: NzbFile, file_done: bool, allow_non_contiguous
                         continue
                     break
 
-                # Could be empty in case nzo was deleted
+                # Could be empty in case nzo was deleted or a previous write attempt encountered errno.ENOSPC and data
```
Safihre: Is it specific to `errno.ENOSPC`? Seems it's just any write error?

mnightingale: Updated, yes any OSError because it could be at least `errno.ENOSPC` or `errno.EDQUOT` but there are likely others that are possible.

mnightingale:
There is another issue around this code, but I'm not sure if it should go in this PR or not.

In `assemble` the `if file_done:` part needs to be in the finally block, but it also needs to consume the OSError when `file_done == True` or the code in `run` needs to be able to run the code `# Clean-up admin data` rather than exit and handle the exception. Or just move the `file_done` hanlding out of `assemble`.

Otherwise you can be `file_done`, run out of space but never remove_admin, tell the directunpacker or handle par2 files.

PR #3343, opened 2026-03-10
Abort DirectUnpack on retryable errors

mnightingale:
This handles the `linebuf.endswith(b"[R]etry, [A]bort ")` case, I have only implemented this in the direct unpacker for now since post processing needs more changes - it needs to read lines without `\n` like the direct unpacker. It may be worth a refactor in the future so there is only one class handling unpacking, with a `non-interactive` mode for post-processing.

Some systems will already handle this because they see a `not enough space on the disk`, but that's a system error message rather than from unrar itself so the message will vary by platform and perhaps filesystem. I did see it not handle disk quota errors. For this reason I think it's better to handle errors/messages from unrar itself.

The other changes I included are to fix a couple of race conditions:

```py
while not self.have_next_volume() and not self.killed and not self.nzo.pp_active:
    with self.next_file_lock:
        self.next_file_lock.wait()
```

There is a window were after the while condition is checked, somewhere calls notify before we call wait. Causing a lost wakeup.

### DIFF
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -258,10 +258,11 @@ class DirectUnpacker(threading.Thread):
                     extracted = []
 
                     # Are there more files left?
-                    while not self.nzo.removed_from_queue and not self.next_sets:
+                    with self.next_file_lock:
                         logging.debug("Direct Unpack for %s waiting for more sets", self.nzo.final_name)
-                        with self.next_file_lock:
-                            self.next_file_lock.wait()
+                        self.next_file_lock.wait_for(
+                            lambda: self.nzo.removed_from_queue or self.next_sets or self.killed
+                        )
 
                     # Is there another set to do?
                     logging.debug(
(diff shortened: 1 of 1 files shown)


sabnzbd/directunpacker.py
```diff
@@ -373,9 +378,8 @@ def wait_for_next_volume(self):
         """Wait for the correct volume to appear but stop if it was killed
         or the NZB is in post-processing and no new files will be downloaded.
         """
-        while not self.have_next_volume() and not self.killed and not self.nzo.pp_active:
-            with self.next_file_lock:
-                self.next_file_lock.wait()
+        with self.next_file_lock:
+            self.next_file_lock.wait_for(lambda: self.have_next_volume() or self.killed or self.nzo.pp_active)
```
Safihre, line 382: It's the same as the while loop right? 
https://github.com/python/cpython/blob/3.14/Lib/threading.py#L385
Guess it's just style preference. I don't mind.

Safihre: Ah yeah I get it now. Nevermind

PR #3344, opened 2026-03-10
Execute file_done actions even if assemble failed

mnightingale:
Resolves a problem were if `assemble` raises IO/OSError due to running out of disk space, the final step actions are not performed potentially leaving a job stuck.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -291,8 +291,7 @@ class Assembler(Thread):
 
                     # Final steps
                     if file_done:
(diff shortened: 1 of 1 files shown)


mnightingale:
I think that’ll have to do, I could make the finally steps catch IOError like they would previously. However, most calls already try/except internally and I think only check_encrypted_and_unwanted might try to write to disk but not in the common path.

I don’t really want to complicate it even more with another level of try/except, I just wanted to fix the state where a file_done is missed.

PR #3345, opened 2026-03-11
Check next_request before changing the value

mnightingale:
It's a possible place requests/articles could be lost, which was the cause of a previous bug.
I have no evidence that it still happens but best to gauard against it

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -186,10 +186,14 @@ class NewsWrapper:
         article: Optional["sabnzbd.nzb.Article"] = None,
     ) -> None:
         """Add a command to the command queue"""
+        if self.next_request:
+            # This should never happen, but guard in case it does
+            _, article = self.next_request
+            self.discard(article, count_article_try=False)
         self.next_request = command, article
 
-    def body(self, article: "sabnzbd.nzb.Article") -> tuple[bytes, "sabnzbd.nzb.Article"]:
-        """Request the body of the article"""
+    def queue_article(self, article: "sabnzbd.nzb.Article"):
+        """Add an article request to the command queue"""
         self.timeout = time.time() + self.server.timeout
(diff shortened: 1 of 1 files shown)


sabnzbd/newswrapper.py
```diff
@@ -186,10 +186,14 @@ def queue_command(
         article: Optional["sabnzbd.nzb.Article"] = None,
     ) -> None:
         """Add a command to the command queue"""
+        if self.next_request:
+            # This should never happen, but guard in case it does
+            _, article = self.next_request
+            self.discard(article, count_article_try=False)
```
Safihre, line 192: If it can never happen, why do we discard if it does happen?

mnightingale: I can’t see where in the code it can happen, but that’s not to say it doesn't.

The change is more I want to protect against if it does happen because it’s one of the few places that a request could be assigned to a connection, then lost and causing the job to be stuck.

It’s related to #3287 because such situation did happen in the past and from the logs, it’s what I think happens.

```
2026-03-07 03:59:42,333::DEBUG::[nzbqueue:905] Next article to assemble for file <FILE> is <Article: article=li5hdAJXy17tYmY6UH7fZ@vSp7C.5Wb, bytes=768000, art_id=None>, decoded: False, on_disk: False, decoded_size: None, has_fetcher: True, tries: 1
```

` has_fetcher: True ` => The article was assigned to the server, so it went into `server. article_queue` or was returned by `get_article`

I think from there it went into a connection’ `next_request` and was somehow lost during a reset, but another angle I need to look at is what happens when 10s or 100s of server connections go bad maybe it's something to do with a server going bad - of note that user has 100s of connections being use and appear to have network issues, they also don't have autodisconnect active.

It’s not so much that the code doesn’t allow it to happen, just I’m not 100% sure that it does and if it ever does happen it will always result in the reported issue.

Issue #3329, opened 2026-02-20
test nzb (force) priority bypassing min free space checks


thezoggy:
### SABnzbd version

5.0 beta1

### Operating system

unraid

### Using Docker image

linuxserver

### Description

In sab I have my incomplete and complete min free space set to 40G each.

While testing some code I set my incomplete + complete to `/ramdrive` which is a 32G (half host ram) mount in my container.
Loaded up the 10G test nzb. 

![Image](https://github.com/user-attachments/assets/c9ac42e1-0d4d-4d54-b832-e9728ff20e5a)

Job gets added, paused.. notification sent because well no min free space. However, the job still continue which I assume is because of the 'force' priority. I would think that we would not want that to happen. 

and from quick glance this was complete folder check not incomplete? also I noticed we arent logging which location we are checking/the sizes.

snippet of logs for the job:
```
2026-02-20 04:15:33,130::INFO::[notifier:165] Sending notification: SABnzbd - SABnzbd 5.0.0Beta1 started (type=startup, job_cat=None)
2026-02-20 04:15:33,131::INFO::[__init__:473] Commit = <HASH>ed4b7e296f2f18f
2026-02-20 04:15:33,131::INFO::[__init__:474] Python-version = 3.12.12 (main, Oct 11 2025, 01:16:26) [GCC 15.2.0]
```
https://gist.githubusercontent.com/thezoggy/da1b70e3ad9da6878624dc23ce0e92cb/raw/d1d450a9089fedecf1e0c6cdff832ba7a32cd6f9/gistfile1.txt

nobrainletsallowed:
Force priority ignoring free space checks is good and should be kept. isn't that the purpose of the priority?

Safihre:
I agree.
Why didn't you expect that @thezoggy?
Force is force.

mnightingale:
I understand force should usually be actioned when paused that's the whole point, but I'm not sure it should continue even when out of disk space.

I'll probably check what happens when the cache and disk is full, given the direct write changes `assemble_article` probably needs to catch more general exceptions.

There is `paused_for_postproc` which it doesn't look like force bypasses, so as a suggestion perhaps there could be a `paused_for_diskspace` cleared by the regular diskspace check or manually unpausing.

nobrainletsallowed:
> I understand force should usually be actioned when paused that's the whole point, but I'm not sure it should continue even when out of disk space.

[...]

PR #3346, opened 2026-03-11
Perform diskspace check on complete directory for nzo

mnightingale:
Related to discussion in #3329 the assembler diskspace_check currently doesn't check the per-category override of the complete directory.

I've split the logic in `prepare_extraction_path` so we can get a return of the base path for the complete directory, without creating anything.

Thereby allowing the diskspace requirements to be applied to the directories the job is using.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -26,7 +26,7 @@ import re
 import threading
 from threading import Thread
 import ctypes
-from typing import Optional, NamedTuple, Union, Literal
+from typing import Optional, NamedTuple, Union
 import rarfile
 import time
 
@@ -325,7 +325,7 @@ class Assembler(Thread):
         """Check diskspace requirements.
         If not enough space left, pause downloader and send email"""
         diskspace_info = diskspace(force=True, complete_dir=get_complete_directory(nzo)[0])
-        full_dir: Optional[Literal["download_dir", "complete_dir"]] = None
+        full_dir: Optional[str] = None
         required_space = (cfg.download_free.get_float() + nzf.bytes) / GIGI
         if diskspace_info.download_dir.free < required_space:
             full_dir = "download_dir"
@@ -344,7 +344,10 @@ class Assembler(Thread):
                 required_space = (complete_free + nzo.bytes) / GIGI
 
             if required_space and diskspace_info.complete_dir.free < required_space:
-                full_dir = "complete_dir"
+                if diskspace_info.complete_dir.path == sabnzbd.cfg.complete_dir.get_path():
+                    full_dir = "complete_dir"
+                else:
+                    full_dir = diskspace_info.complete_dir.path
 
         if full_dir:
             logging.warning(T("Too little diskspace forcing PAUSE"))
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -26,7 +26,7 @@ import re
 import threading
 from threading import Thread
 import ctypes
-from typing import Optional, NamedTuple, Union
+from typing import Optional, NamedTuple, Union, Literal
 import rarfile
 import time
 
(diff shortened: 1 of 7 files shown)


Safihre:
Should we then not always check the free space on the complete folder?
Currently we only do it if configured.

mnightingale:
I hadn't noticed that yes, `diskspace_check` checks if `download_free + nzf.bytes`.
So for complete it would need to be `nzo.bytes` the sum of nzf.

So it needs to use `cfg.complete_free.get_float()` and `freespace["complete_dir"][1]` for two different checks.

mnightingale:
I think I'll have to add some tests to this.

Unpacking tuples and the dict hurt my head so I've changed it to dataclasses.

I also need to figure something out for the scheduler since it may need to check a different path.

mnightingale:
Before I fix the 3.9 failure, are we keeping 3.9 support?

Safihre:
Yes for 5.0.

Safihre:
Hahha that last commit really helped, I did not like the 2 dataclasses 😂

mnightingale:
Same 😄 I had to way up what to replace it with, keeping the helper wrapper or just using diskspace_base everywhere but I decided it would make it more complicated for most callers.

I think the functionality is all there now, the required_space calculations are unchanged just always enforced even if unset or 0.

sabnzbd/assembler.py
```diff
@@ -342,8 +343,11 @@ def diskspace_check(nzo: NzbObject, nzf: NzbFile):
                 # downloading until 95% complete before checking
```
Safihre, line 343: Should we maybe do it earlier?

mnightingale: I think an argument could be made for the directunpack to be at least `required_space = (complete_free + nzf.bytes) / GIGI` like the download_dir calculation. So “is there enough free space to extract the just downloaded file” – I did have this at one point but decided against changing the behaviour in this PR.

The unpacker path with the 95% downloaded calculation, is tricky because in the time to pause that remaining 5% could be inflight. The main benefit I can think of checking at 95% is it gives time for previously completed jobs to be moved, etc. An optimal setup has both on the same filesystem anyway so most would hit this via the download_dir check instead.

It’s probably safer to do at a lower percentage but running out of space due to complete_dir is less of an issue because the direct unpacker can be aborted and post-proc retried. Whereas download_dir running out results in lost article data. But it’s also important to think of how this fits in with concurrent downloading and other jobs extracting.

Where I think there is an issue but it’s a more fundamental behaviour change is that the defaults allow download_dir and complete_dir to reach 0 space in the first place, it would be safer to have a minimum disk space of 1GB or something. Users will usually have both paths on the same disk, and often the same filesystem as their operating system which can cause all sorts of problems.

Therefore, I feel that the direct unpack calculation should change, the 95% check is probably fine but could be reduced a bit to lower the risk of completing the job while paused, there could be a minimum disk space requirement to protect SAB state/persistence and the OS.

Safihre: We could just set default for the minium free size. So just for new users.

sabnzbd/assembler.py
```diff
@@ -342,8 +343,11 @@ def diskspace_check(nzo: NzbObject, nzf: NzbFile):
...
                 required_space = (complete_free + nzo.bytes) / GIGI
 
-            if required_space and freespace["complete_dir"][1] < required_space:
-                full_dir = "complete_dir"
+            if required_space and complete_dir.free < required_space:
+                if complete_dir.path == sabnzbd.cfg.complete_dir.get_path():
+                    full_dir = "complete_dir"
```
Safihre, line 348: Can we do a nicer way then pass these strings around? I know it's for the planned resume, but it feels ugly. I asked AI but it came up with something _meh_.

mnightingale: I think the reason it passes the strings instead of paths is to allow resuming when the user "fixes" it by changing their config. I could change it to always use the absolute path which will get rid of some of the noise but then it may not perform a planned resume, but I suppose if the user has reconfigured paths they could be expected to resume manually.

We could pull the constants out `DISKSPACE_DOWNLOAD_DIR = "download_dir"`and add typing to the resume methods to make it clear that those values have special meaning without referring to the documentation.

Do you have any preference or suggestion?

sabnzbd/assembler.py
Safihre: I set it to be 500M, equal to the default for the Article Cache

PR #3349, opened 2026-03-12
Fix race when direct unpacker is killed

### LINKED ISSUE #3347: Fatal error in Assembler: threads can only be started once
### SABnzbd version

4.5.5

### Operating system

macOS

### Using Docker image

None

### Description

I need to check this againt develop, just adding here so there is a record of it.

Basically I checked out the 4.5.5 branch, so the unrar binaries changed and macOS didn't trust the publisher or where it is. It seemed to result it in failing to launch the process, as expected.

I'm pretty sure it's a race involving https://github.com/sabnzbd/sabnzbd/blob/d80d4f57b5e247a7558b0fa3d33b0880c9ddbfbf/sabnzbd/directunpacker.py#L354-L361 and https://github.com/sabnzbd/sabnzbd/blob/d80d4f57b5e247a7558b0fa3d33b0880c9ddbfbf/sabnzbd/directunpacker.py#L135-L137 and or https://github.com/sabnzbd/sabnzbd/blob/d80d4f57b5e247a7558b0fa3d33b0880c9ddbfbf/sabnzbd/directunpacker.py#L151-L159

Perhaps build_and_run_command will have hung until I did something with the macOS prompt, or it was just a race because the thread launched then immediately was killed.

Result is `Fatal error in Assembler - threads can only be started once` which causes a restart then stuck `2026-03-12 19:04:34,000::DEBUG::[downloader:841] Delayed - 12 seconds - Assembler queue: 12`

edit: I think my recent changes to the assembler should have made it recoverable, but I still want to fix the race and trying to restart threads.

Perhaps build_and_run_command will have hung until I did something with the macOS prompt, or it was just a race because the thread launched then immediately was killed.

[455assembler.log](https://github.com/user-attachments/files/25947056/455assembler.log)

mnightingale:
Fixes #3347

AI mostly found it after feeding it the symptoms and log.

> The race is between run()'s final cleanup and add() checking state. Here's the sequence:
>
>1. run() calls self.reset_active() (line 355) → sets active_instance = None
>2. add() acquires START_STOP_LOCK, enters
>3. check_requirements() checks self.killed → still False → passes
>4. Line 151: not self.active_instance → True (reset in step 1)
>5. Meanwhile run() finishes, sets self.killed = True (line 361), returns → thread dies
>6. Back in add(): not self.is_alive() → True (thread just died)
>7. self.start() → crash — thread already ran once
>
> The root cause: run()'s final cleanup (lines 355–361) is not protected by START_STOP_LOCK, so add() can observe a half-updated state — active_instance is None and killed is still False, and is_alive() flips to False between the two checks.

The order doesn't matter but it is required that both `self.reset_active()` and `self.killed = True` are called under the lock.

### DIFF
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -351,14 +351,14 @@ class DirectUnpacker(threading.Thread):
         if unrar_log:
             logging.debug("DirectUnpack Unrar output: \n%s", "\n".join(unrar_log))
 
-        # Make more space
-        self.reset_active()
-        if self in ACTIVE_UNPACKERS:
-            ACTIVE_UNPACKERS.remove(self)
-        logging.debug("Closing DirectUnpack for %s", self.nzo.final_name)
-
-        # Set the thread to killed so it never gets restarted by accident
-        self.killed = True
+        with START_STOP_LOCK:
+            # Set the thread to killed so it never gets restarted by accident
+            self.killed = True
+            # Make more space
+            self.reset_active()
+            if self in ACTIVE_UNPACKERS:
+                ACTIVE_UNPACKERS.remove(self)
+            logging.debug("Closing DirectUnpack for %s", self.nzo.final_name)
 
     def have_next_volume(self):
         """Check if next volume of set is available, start


Safihre:
Trusting you and Al 😊

PR #3355, opened 2026-03-17
Reduce chance of assembly starting before filenames are available

mnightingale:
Files could be created with obfuscated names because prepare_filepath only waited for each file's own first article - par2 may not have been assembled yet, so md5of16k wasn't populated when verify_nzf_filename ran.

- first_article_processed(): Non-par2 files now wait for all first_articles to complete, not just their own. Since mini-par2 files download first and are typically single-article, handle_par2 has almost always populated md5of16k by then. Par2 files still only wait for their own first article.
- prepare_filepath(force): Accepts force=True (passed by the assembler on file_done) to bypass the gate, preventing completed files from being dropped if they finish before all first articles are done.
- should_queue_nzf(): Early-write path now checks prepare_filepath() so it respects the new gating

This doesn't eliminate the race entirely but it reduces the window significantly because it encourages par2 to be assembled before other files.

**Before:** race starts when that file's first article is downloaded. Before the par2 has been assembled and handle_par2 is called - par2 might not have even downloaded yet.
**New:** race starts when the last first article completes, until handle_par2 finishes. Files are sorted so mini-par2 are first, so the chance that the assembler has ran for a par2 and populated md5of16k by the time all first articles are finished is high.

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -206,7 +206,7 @@ class Assembler(Thread):
             return False
         # Always write
         if article_has_first_part and filename_checked and not import_finished:
-            return True
+            return nzf.prepare_filepath() is not None
         next_ready = (next_article := nzf.assembler_next_article) and (next_article.decoded or next_article.on_disk)
         # Trigger every 5 seconds if next article is decoded or on_disk
         if next_ready and time.monotonic() > self.queued_next_time.get(nzf.nzf_id, 0):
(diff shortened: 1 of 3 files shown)


mnightingale:
Maybe reject, and I'll make a PR of just the latest commit?

---

For context, this is a problem I’ve seen happen where the NZB has normal looking file names but the par2 has different names and you sometimes end up with a mixture of filenames so direct unpack fails.

Adding a time.sleep here tends to trigger it.

https://github.com/sabnzbd/sabnzbd/blob/2c91e61d8ede7beda055deb1ec368295faee6846/sabnzbd/assembler.py#L305-L307

I think just the `filename_checked` change may prevent that sufficiently but I’ll test it first.

Safihre:
Sounds good!

PR #3363, opened 2026-03-24
Check for unwanted extensions after repair and unpack

### LINKED ISSUE #3361: Unwanted extension checks are not robust enough
### SABnzbd version

5.0.0b4

### Operating system

macOS

### Using Docker image

None

### Description

Split from https://github.com/sabnzbd/sabnzbd/issues/3358#issuecomment-4109194674 to track separately.

There appear to be several issues with the unwanted extensions checks that make it easy to circumvent.

Identifying the issues is still a work in-progress, for now I'm just throwing some ideas out there for ways I suspect bypass the checks and may need some improvement.

If anyone can provide other methods, would like to test the unverified methods, or can provide NZBs verifying the issues that would be appreciated.

# Issues

## Verified

- Repair bypasses check - can't read file list from a corrupt RAR, a crafted NZB can force repair to be required
- Nested archives (rar => rar => exe)
- Extra hidden files in par2 repair volumes - force a repair and add a new file
- 
## Unverified

- Appears to only check rar files - what about 7z, zip, and any others that can be extracted - I'm not 100% which others SAB does
- par2 rename - NzbObject checks original filenames but I don't think renames from par2 filenames are checked

# Proposal

## Detection

Keep a set of `nzo.unchecked_files`, add any file that fails the assembler check, including those which couldn't be identified as rar files.

In postproc, after `parring` compare the directory listing before and after and combine with `nzo.unchecked_files` for a list of files to check.

After `unpacker` check `newfiles` for unwanted extensions.

Retry job should skip the two unwanted extensions checks.

## Action

Open to ideas, I suggest after the repair step they would have to fail the job which can be retried if the user wants to keep the unwanted files.
After extraction may be a little trickier to deal with, we could delete the unpacked content?

mnightingale:
Fixes #3361

Still a work in progress but it doesn't appear there would be much to it.

- [x] Needs tests
- [ ] Remove local import of `check_encrypted_and_unwanted_files` - move it somewhere else to avoid circular import, but probably refactor to avoid the repetition
- [x] Allow skipping the check on resumed jobs
- [x] Allow skipping the check on retried jobs
- [ ] May need adding to NzbObjectSaver but should probably use relative paths first, or just at runtime and list the directory to populate on restart i.e. treat all as unchecked on restart
- [ ] Possibly only needs to check after unpack when enable_recursive is enabled
- [ ] Check direct unpack is checked
- [ ] Behaviour on fail after unpack; probably needs to delete unpack result instead of leaving extracted files, else will be an orphaned job
- [ ] Probably exclude \_\_admin\_\_

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -306,6 +306,10 @@ class Assembler(Thread):
                             elif par2file.is_par2_file(filepath):
                                 # Parse par2 files, cloaked or not
                                 nzo.handle_par2(nzf, filepath)
+
+                            else:
+                                # May be RAR files which could not be identified due to missing first part
+                                nzo.unchecked_files.add(filepath)
                 except Exception:
                     logging.error(T("Fatal error in Assembler"), exc_info=True)
                     break
@@ -686,6 +690,8 @@ def check_encrypted_and_unwanted_files(nzo: NzbObject, filepath: str) -> tuple[b
                             zf.trigger_parse()
                         except Exception:
                             pass
+                    if not zf.namelist():
+                        nzo.unchecked_files.add(filepath)
                     for somefile in zf.namelist():
                         logging.debug("File contains: %s", somefile)
                         if has_unwanted_extension(somefile):
@@ -695,5 +701,6 @@ def check_encrypted_and_unwanted_files(nzo: NzbObject, filepath: str) -> tuple[b
                 del zf
         except rarfile.Error as e:
             logging.info("Error during inspection of RAR-file %s: %s", filepath, e)
+            nzo.unchecked_files.add(filepath)
 
     return encrypted, unwanted
--- a/sabnzbd/nzb/object.py
+++ b/sabnzbd/nzb/object.py
@@ -354,6 +354,8 @@ class NzbObject(TryList):
         # Reuse the existing directory
         if reuse and os.path.exists(reuse):
             self.download_path = long_path(reuse)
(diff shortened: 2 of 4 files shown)


mnightingale:
I'm not sure how to better handle [Skip unwanted check for retried jobs](https://github.com/sabnzbd/sabnzbd/pull/3363/commits/33db6de983ecd18b5271770bd739d3f5184e0021) this makes all jobs that are repaired skip the unwanted extension check.

Ideally I think we would want `retry_job` to check `unwanted_ext == 1` on the history item but we don't store it, we only really have `status = "Failed" and fail_message = "Aborted, unwanted extension detected"` T(...) which we could use but seems a little fragile.

Safihre:
While I appreciate the effort, I feel this is maybe solving a problem that isn't a huge problem? Like @thezoggy said, most people rely on Cleanup list anyway.

Why not only a final check after unpack and that's it?
It's best effort, it doesn't need 100% accuracy in all situations.

mnightingale:
I'm inclined to agree, there has been a recent wave of them being uploaded but can be negated by sonarr options.

I'll leave it for 5.0.0, maybe revisit it in the future only after unpacking.

I do think it is a risk for users, they think that having configured unwanted extensions and having never being warned means the resulting files are safe. Guides don't appear to recommend entering executables in both unwanted and cleanup, I understand they are supposed to serve different purposes but the unwanted part can be bypassed and cleanup doesn't mark jobs as failed.

Issue #2274, opened 2022-08-23
Test failure in TestPar2Repair.test_basic


jcfp:
Been seeing this one for a while on some CI environments ([debian ci](https://ci.debian.net/data/autopkgtest/testing/amd64/s/sabnzbdplus/25082821/log.gz), [ubuntu ci](https://autopkgtest.ubuntu.com/results/autopkgtest-kinetic/kinetic/amd64/s/sabnzbdplus/20220823_010337_9057a@/log.gz), salsa.debian.org) but not others (github). At first I assumed it was a fluke, but it keeps popping up so the time for a bug report has come. No clue what's causing this one.

Log excerpt:
```
__________________________ TestPar2Repair.test_basic ___________________________

self = <tests.test_newsunpack.TestPar2Repair object at 0x7ffb54f54af0>
caplog = <_pytest.logging.LogCaptureFixture object at 0x7ffb57e2d240>

    def test_basic(self, caplog):
        # Run code
        nzo, dir_contents = self._run_par2repair("tests/data/par2repair/basic", caplog)
    
>       assert dir_contents == [
            "__ADMIN__",
            "notarealfile.rar",
            "par2test.part1.rar",
            "par2test.part2.rar",
            "par2test.part3.rar",
            "par2test.part4.rar",
            "par2test.part5.rar",
            "par2test.part6.rar",
        ]
E       AssertionError: assert ['__ADMIN__',...rt4.rar', ...] == ['__ADMIN__',...rt4.rar', ...]
E         Left contains one more item: 'stillrarbutnotagoodname.txt'
E         Full diff:
E           [
E            '__ADMIN__',
E            'notarealfile.rar',
E            'par2test.part1.rar',
E            'par2test.part2.rar',
E            'par2test.part3.rar',
E            'par2test.part4.rar',
E            'par2test.part5.rar',
E            'par2test.part6.rar',
E         +  'stillrarbutnotagoodname.txt',
E           ]

tests/test_newsunpack.py:148: AssertionError
```

Safihre:
This is strange. 
'stillrarbutnotagoodname.txt' should be renamed to "par2test.part4.rar".
It seems it just made a copy.
Is there a way to see which par2cmdline is running on these runners?
Maybe from the CI output of the functional tests?

jcfp:
[...]

PR #3364, opened 2026-03-25
builder/package.py: stop mangling test data (#2274)

joecool1029:
Broke tests on Debian/Ubuntu, they reported it, never figured it out.

On Gentoo we skipped the failing test for years but 'just plain fails' didn't sit right with me. We found the tests only fail using release tarball, not git.

package.py script 'fixes' the line endings on .txt extensions but doesn't account for a stillrarbutnotagoodname.txt rar binary in the test data.

This breaks tests on anyone using the release tarballs. So we skip mangling the test data to allow for downstream CI to work.


Safihre:
Thanks 👍

PR #3370, opened 2026-04-03
Show what the entry is alternative to

### LINKED ISSUE #2793: Alternative feature enhancement: Show what the entry should be the alternative to
### Description

I really like the new "Alternative" label, unfortunately it's not always clear which other download it refers to. It would be great if we could somehow link the different entries. 

You could add an "original" label and add a number to each label. Or make it clickable and jump to the original. A tooltip with the title of the original entry could help, too.

Safihre:
Closes #2793

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1642,6 +1642,7 @@ def build_queue(
         slot["script"] = nzo.script if nzo.script else "None"
         slot["filename"] = nzo.final_name
         slot["labels"] = nzo.labels
+        slot["duplicate_info"] = nzo.duplicate_info
         slot["password"] = nzo.password if nzo.password else ""
         slot["cat"] = nzo.cat if nzo.cat else "None"
         slot["mbleft"] = "%.2f" % mbleft
--- a/sabnzbd/nzb/object.py
+++ b/sabnzbd/nzb/object.py
@@ -161,6 +161,7 @@ NzbObjectSaver = (
     "bad_articles",
     "duplicate",
     "duplicate_key",
+    "duplicate_of",
     "oversized",
     "precheck",
     "incomplete",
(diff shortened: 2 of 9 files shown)


sabnzbd/nzb/object.py
```diff
@@ -1562,6 +1575,7 @@ def duplicate_check(self, repeat: bool = False):
         if repeat:
             self.duplicate = None
             self.duplicate_key = None
+            self.duplicate_of = None
 
         duplicate_in_history = smart_duplicate_in_history = False
         duplicate_in_queue = smart_duplicate_in_queue = False
```
mnightingale, line 1581:
```suggestion
        duplicate_in_history = smart_duplicate_in_history = None
        duplicate_in_queue = smart_duplicate_in_queue = None
```

To resolve complaints about `Expected type 'str | None', got 'bool | str' instead` with the duplicate_of lines below.

sabnzbd/api.py
```diff
@@ -1642,6 +1642,7 @@ def build_queue(
         slot["script"] = nzo.script if nzo.script else "None"
         slot["filename"] = nzo.final_name
         slot["labels"] = nzo.labels
+        slot["duplicate_info"] = nzo.duplicate_info
```
mnightingale, line 1645: Could this be a little restrictive to only return a str?

I'm envisaging a future UI overhaul where we want the nzo_id of the duplicate, or the url it downloaded from for doing things like more hover information or clicking to filter the history by nzo id.

To avoid breaking API changes in the future I would suggest considering returning a dict, even if it just has "reason" or something that returns the str.

Did you decide against making click filter the history by nzo id?

mnightingale (commented):
Only minor suggestions but I'm not sure how useful just the name is, aren't you going to often just end up with many duplicates with the same name anyway so it may no help you identify them?

I did like the idea of clicking filtering the history, but I don't think the frontend currently supports filtering by id which would need to be separate from search.

Safihre:
Yeah I thought about filtering. Claude came up with this solution. But filtering will be better.
Will try again!

mnightingale:
It doesn't look like it would be too difficult to add filtering via `parseSearchQuery` so you can search for something like `id:SABnzbd_nzo_xyz` and proably allow multiple and map those to a list for the `nzo_ids` argument.

Issue #1136, opened 2018-04-28
[Feature Request] Please Add CORS Support


hl2guide:
Please consider adding SABNZBD server support for CORS.
Ref: https://enable-cors.org/

Currently:
![example](https://user-images.githubusercontent.com/17044982/39391931-ae9f6178-4aee-11e8-8757-b587b18435d0.png)

Normally you can just add the header:
`Access-Control-Allow-Origin: *`

_NB: This would allow for 3rd-party User Interfaces to be made, I'm making one._

Safihre:
What do you think @beardog108?
Would make sense for the `/api` endpoint.

Cool to hear you are making a new user interface, will it be full featured or subset? If full featured, we can also integrate it in SABnzbd directly. 
Can we see it somewhere? :+1:

hl2guide:
Most likely full featured, however it's too early on in development so far.
The main features I'm aiming for are:
- Responsive layout and font size for small and large displays
- Support for custom fonts
- Light and Dark modes
- Charts and proper progress indicators
- Colour blind modes
- High contrast mode

Safihre:
Cool! This is also how I started with SABnzbd a few years back: I hated the Plush interface so I made a new one, Glitter.
And now I do everything 😁
I could use some help.. Too little time to take up big projects (like interface rework).

Safihre:
If you find and API endpoint is missing, please open an issue and I can add it :+1:.
There's quite a few actions that don't have proper API implementation and are only available through "template" URLs.
Or if the API endpoint doesn't make sense or has too little information, we can change this too.

beardog108:
If the API key is securely generated, there should be no security issue that I can think of for enabling it on /api. It would be cool to have user created interfaces.

Safihre:
Added it 👍 
Looking forward to see what you create for new interface! ❤️

PR #3373, opened 2026-04-06
Replace cherrypy by uvicorn/starlette

Safihre:
Work in progress.
Warp failed me, but Claude Code is really making good progress.
Because it can also control the browser, it can actually visually check all changes if they really work 🌟🌟

Maybe @mnightingale already sees high-level problems to address?

[ckip ci] for now.

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -39,16 +39,14 @@ import time
 import re
 import gc
 import threading
-import http.cookies
 import uvicorn
-from typing import List, Dict, Any
+from typing import Any
 
 try:
     import sabctools
     import Cheetah
     import feedparser
     import configobj
-    import portend
     import cryptography
     import charset_normalizer
     import guessit
@@ -85,6 +83,8 @@ import sabnzbd.newsunpack
 from sabnzbd.misc import (
     exit_sab,
     split_host,
+    port_is_free,
+    find_free_port,
     create_https_certificates,
     ip_extract,
     set_serv_parms,
@@ -656,18 +656,6 @@ def is_sabnzbd_running(url):
         return False
 
 
-def find_free_port(host, currentport):
-    """Return a free port, 0 when nothing is free"""
-    n = 0
-    while n < 10 and currentport <= 49151:
-        try:
-            portend.free(host, currentport, timeout=0.025)
-            return currentport
-        except Exception:
-            currentport += 5
-            n += 1
-    return 0
-
 
 def check_for_sabnzbd(url, upload_nzbs, allow_browser=True):
     """Check for a running instance of sabnzbd on this port
@@ -980,18 +968,10 @@ def main():
     # When this is a daemon, just check and bail out if port in use
     if sabnzbd.DAEMON:
         if enable_https and https_port:
-            try:
-                portend.free(web_host, https_port, timeout=0.05)
-            except IOError:
+            if not port_is_free(web_host, https_port):
                 abort_and_show_error(browserhost, web_port)
-            except Exception:
-                abort_and_show_error(browserhost, web_port, "49")
-        try:
-            portend.free(web_host, web_port, timeout=0.05)
-        except IOError:
+        if not port_is_free(web_host, web_port):
             abort_and_show_error(browserhost, web_port)
-        except Exception:
-            abort_and_show_error(browserhost, web_port, "49")
 
     # Windows instance is reachable through registry
     url = None
@@ -1003,55 +983,39 @@ def main():
     # SSL
     if enable_https:
         port = https_port or web_port
-        try:
-            portend.free(browserhost, port, timeout=0.05)
-        except IOError as error:
-            if str(error) == "Port not bound.":
-                pass
-            else:
-                if not url:
-                    url = "https://%s:%s%s/api?" % (browserhost, port, sabnzbd.cfg.url_base())
-                if new_instance or not check_for_sabnzbd(url, upload_nzbs, autobrowser):
-                    # Bail out if we have fixed our ports after first start-up
-                    if sabnzbd.cfg.fixed_ports():
-                        abort_and_show_error(browserhost, web_port)
-                    # Find free port to bind
-                    newport = find_free_port(browserhost, port)
-                    if newport > 0:
-                        # Save the new port
-                        if https_port:
-                            https_port = newport
-                            sabnzbd.cfg.https_port.set(newport)
-                        else:
-                            # In case HTTPS == HTTP port
-                            web_port = newport
-                            sabnzbd.cfg.web_port.set(newport)
-        except Exception:
-            # Something else wrong, probably badly specified host
-            abort_and_show_error(browserhost, web_port, "49")
+        if not port_is_free(browserhost, port):
+            if not url:
+                url = "https://%s:%s%s/api?" % (browserhost, port, sabnzbd.cfg.url_base())
+            if new_instance or not check_for_sabnzbd(url, upload_nzbs, autobrowser):
+                # Bail out if we have fixed our ports after first start-up
+                if sabnzbd.cfg.fixed_ports():
+                    abort_and_show_error(browserhost, web_port)
+                # Find free port to bind
+                newport = find_free_port(browserhost, port)
+                if newport > 0:
+                    # Save the new port
+                    if https_port:
+                        https_port = newport
+                        sabnzbd.cfg.https_port.set(newport)
+                    else:
+                        # In case HTTPS == HTTP port
+                        web_port = newport
+                        sabnzbd.cfg.web_port.set(newport)
 
     # NonSSL check if there's no HTTPS or we only use 1 port
     if not (enable_https and not https_port):
-        try:
-            portend.free(browserhost, web_port, timeout=0.05)
-        except IOError as error:
-            if str(error) == "Port not bound.":
-                pass
-            else:
-                if not url:
-                    url = "http://%s:%s%s/api?" % (browserhost, web_port, sabnzbd.cfg.url_base())
-                if new_instance or not check_for_sabnzbd(url, upload_nzbs, autobrowser):
-                    # Bail out if we have fixed our ports after first start-up
-                    if sabnzbd.cfg.fixed_ports():
-                        abort_and_show_error(browserhost, web_port)
-                    # Find free port to bind
-                    port = find_free_port(browserhost, web_port)
-                    if port > 0:
-                        sabnzbd.cfg.web_port.set(port)
-                        web_port = port
-        except Exception:
-            # Something else wrong, probably badly specified host
-            abort_and_show_error(browserhost, web_port, "49")
+        if not port_is_free(browserhost, web_port):
+            if not url:
+                url = "http://%s:%s%s/api?" % (browserhost, web_port, sabnzbd.cfg.url_base())
+            if new_instance or not check_for_sabnzbd(url, upload_nzbs, autobrowser):
+                # Bail out if we have fixed our ports after first start-up
+                if sabnzbd.cfg.fixed_ports():
+                    abort_and_show_error(browserhost, web_port)
+                # Find free port to bind
+                port = find_free_port(browserhost, web_port)
+                if port > 0:
+                    sabnzbd.cfg.web_port.set(port)
+                    web_port = port
 
     # We found a port, now we never check again
     sabnzbd.cfg.fixed_ports.set(True)
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1299,48 +1299,6 @@ def main():
         sabnzbd.WEBLOGFILE = os.path.join(logdir, DEF_LOG_CHERRY)
         cherrypy.log.access_file = str(sabnzbd.WEBLOGFILE)
 
-    # Force mimetypes (OS might overwrite them)
-    forced_mime_types = {"css": "text/css", "js": "application/javascript"}
-
-    static = {
-        "tools.staticdir.on": True,
-        "tools.staticdir.dir": os.path.join(sabnzbd.WEB_DIR, "static"),
-        "tools.staticdir.content_types": forced_mime_types,
-    }
-    staticcfg = {
-        "tools.staticdir.on": True,
-        "tools.staticdir.dir": os.path.join(sabnzbd.WEB_DIR_CONFIG, "staticcfg"),
-        "tools.staticdir.content_types": forced_mime_types,
-    }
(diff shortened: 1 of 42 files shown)


mnightingale:
Just keeping this high level for now.

This aspect probably requires a whole other discussion as it isn’t strictly related to these changes, but exposing the long-lived api key to the frontend by todays standards is poor security. It should probably be session based using a cookie for authentication – would also allow us to fix some of the cookie issues. I’m only mentioning this because uvicorn/starlette make it quite straightforward to support.

Default Gzip compression level is 9, I would drop it way down to 2 for faster responses.
`Middleware(GZipMiddleware, minimum_size=1000, compresslevel=2)`

Several forms have changed to `method="get"` this isn’t good for security and breaks HTTP semantics. Things like possible ways of leaking sensitive data; preserved in browser history, logged by proxies, access logs, SAB logs.

You have a method merging body and query arguments, with body winning if present. Maybe it has made it easier although I can’t think where you might want to mix get and post so maybe it was just a convenience thing, so you don’t have to worry if it’s a get/post. It could lead to security issues due to the larger attack surface. It also changes all non-files to str, not sure if that could cause problems anywhere. It massively expands the API route (/api) since you have committed to supporting requests via GET or POST or a combination of the query and body.

Safihre:
The get/post part and merging of arguments is actually a non-AI part that I had to make myself due to all the weird ways the Config submits requests. Sometimes it posts, sometimes it ajaxs, sometimes get... All on the same page.
I will have to generalise that indeed, but I just wanted to get something working first.

Safihre:
What do you think of the new parameter merging?

sabnzbd/api.py
```diff
@@ -1021,15 +1034,15 @@ def _api_server_stats(name: str, kwargs: dict[str, Union[str, list[str]]]) -> by
...
 
 
-def _api_gc_stats(name: str, kwargs: dict[str, Union[str, list[str]]]) -> bytes:
+def _api_gc_stats(name: str, kwargs: QueryParams) -> Response:
     """Function only intended for internal testing of the memory handling"""
     # Collect before we check
     gc.collect()
     # We cannot create any lists/dicts, as they would create a reference
-    return report(data=[str(obj) for obj in gc.get_objects() if isinstance(obj, TryList)])
+    return report(kwargs, data=[str(obj) for obj in gc.get_objects() if isinstance(obj, sabnzbd.nzbstuff.TryList)])
```
mnightingale:
```suggestion
    return report(kwargs, data=[str(obj) for obj in gc.get_objects() if isinstance(obj, TryList)])
```

Also need to revert the import line for TryList, `from sabnzbd.nzb import TryList, NzbObject`.
Should fix 14/24 test failures.

interfaces/Config/templates/config_cat.tmpl
```diff
@@ -129,7 +129,7 @@
                         jQuery(elm).extractFormDataTo(data);
                         jQuery.ajax({
                             type: "GET",
-                            url: window.location.pathname + 'save',
+                            url: window.location.pathname + '/save',
```
mnightingale: Should be POST (get is not supported by route) and url could just be "categories/save"

interfaces/Config/templates/config_sorting.tmpl
```diff
@@ -499,7 +499,7 @@
                         jQuery(elm).extractFormDataTo(data);
                         jQuery.ajax({
                             type: "GET",
-                            url: window.location.pathname + 'save_sorter',
+                            url: window.location.pathname + '/save_sorter',
```
mnightingale: POST + `url: "sorting/save_sorter"`

interfaces/Glitter/templates/include_menu.tmpl
```diff
@@ -91,7 +91,7 @@
                     <li><a href="#modal-help" data-toggle="modal"><span class="glyphicon glyphicon-question-sign"></span> $T('menu-help')</a></li>
                     <li><a href="https://sabnzbd.org/donate" target="_blank"><span class="glyphicon glyphicon-heart"></span> $T('menu-donate')</a></li>
                     <!--#if $have_logout or $have_quota or $have_rss_defined or $have_watched_dir or $pp_pause_event#--><li class="divider"></li><!--#end if#-->
-                    <!--#if $have_logout#--><li><a href="./login/?logout=1"><span class="glyphicon glyphicon-log-out"></span> $T('logout')</a></li><!--#end if#-->
+                    <!--#if $have_logout#--><li><a href="./login?logout=1"><span class="glyphicon glyphicon-log-out"></span> $T('logout')</a></li><!--#end if#-->
```
mnightingale: I'd understand leaving this behaviour unchanged for now.

Logout should be a separate route and use a post request, get is semantically the wrong kind of request but the main issue is susceptibility to CSRF.

sabnzbd/config.py
```diff
@@ -117,6 +117,10 @@ def set_dict(self, values: dict[str, Any]):
             except KeyError:
                 pass
 
+    def get_from_dict(self, values: dict[str, Any], kw: str) -> Any:
+        """Extract this option's value from a dict by key, raising KeyError if absent"""
+        return values[kw]
```
mnightingale, line 122: I'd suggest extracting `get_from_dict` to a helper method, its implementation does not depend on the Option or OptionList.

Typing could be better for the values argument, checking for "getlist" attr on what is supposed to be a dict is odd. Perhaps typingMapping would be more appropriate for "an object supporting \_\_gettiem\_\_"

sabnzbd/interface.py
```diff
@@ -406,205 +402,229 @@ def template_filtered_response(file: str, search_list: dict[str, Any]):
...
+    threading.Thread(target=sabnzbd.shutdown_program).start()
+    return PlainTextResponse(T("SABnzbd shutdown finished"))
+
+
+@secured_expose(route="/api", check_api_key=True, access_type=1)
+async def api(request: Request):
+    """Redirect to API-handler, we check the access_type in the API-handler"""
+    # Pass a fresh mutable copy so api_handler can assign kwargs["keyword"] etc.
+    # without mutating the request-cached _query_params.
+    return api_handler(MultiDict(request.query_params.multi_items()))
```
mnightingale: `Expected type 'QueryParams', got 'MultiDict' instead`

sabnzbd/api.py
```diff
@@ -1175,9 +1190,14 @@ def report(error: Optional[str] = None, keyword: str = "value", data: Optional[A
...
 
-    cherrypy.response.headers["Content-Type"] = content
-    cherrypy.response.headers["Pragma"] = "no-cache"
-    return response
+    return Response(
+        response,
+        media_type=content,
+        headers={
+            "Pragma": "no-cache",
+            "Access-Control-Allow-Origin": "*",
```
mnightingale, line 1198: `Access-Control-Allow-Origin: *` should not be a hardcoded default CORS policy, at least it's only on API routes though.

[CWE-942: Permissive Cross-domain Security Policy with Untrusted Domains](https://cwe.mitre.org/data/definitions/942.html)
[GitHub Advisory Database: CWE-942](https://github.com/advisories?query=cwe%3A942)

SABnzbd.py
```diff
@@ -19,8 +19,8 @@
 
 # Trick to show a better message on older Python
 # releases that don't support walrus operator
-if Python_39_is_required_to_run_SABnzbd := sys.hexversion < 0x03090000:
-    print("Sorry, requires Python 3.9 or above")
+if Python_39_is_required_to_run_SABnzbd := sys.hexversion < 0x030A00F0:
```
mnightingale, line 22:
```suggestion
if Python_310_is_required_to_run_SABnzbd := sys.hexversion < 0x030A00F0:
```

sabnzbd/nzbparser.py
```diff
@@ -57,7 +56,7 @@
 
 
 def add_nzbfile(
-    nzbfile: Union[str, cherrypy._cpreqbody.Part],
+    nzbfile,
```
mnightingale, line 59: Can we type this? There are a few places where `cherrypy._cpreqbody.Part` is removed and there is now no typing at all.

mnightingale (commented):
Minor nitpick: there are several forms with `name="fullform"` that doesn't appear to serve any purpose, `class="fullform"` is being used instead.

sabnzbd/api.py
Safihre: I know, but we're not really a webserver so this would cause more problems than it's worth for that tiny tiny number of users that really run SABnzbd on a publicly exposed domain. Right?

mnightingale: The permissive value was added in response to https://github.com/sabnzbd/sabnzbd/issues/1136 it allows requests from a browser context to access SAB on a different hostname.

Default browser protections would not allow requests to sab.domain.xyz from the ui.domain.xyz page.
A more sensible approach to a custom UI would probably be to proxy the API requests so it’s not cross origin.

In normal usage, the policy is not required because it’s not cross origin.
If a user accesses via a domain pointing at a public or private instance, API requests go to the same origin as the current page which is allowed.

The only time I think a less restrictive setting might be required are indexers that let you configure SAB within them to send NZBs to the queue but run client side. Thankfully in my experience this kind of integration is rarely provided these days; it is a bad idea to give indexers credentials to your SAB instance. Note there may already be issues with this due to Access-Control-Allow-Private-Network if you wanted to send cross origin to a private instance.

The browser extension approach such as SABconnect++ should also not require a permissive CORs policy since it won’t operate in the page context.

The risk is it allows any site to interact with SAB; this would be more important if unauthenticated requests could be made but it's only on the /api route so I don’t think it’s exposing anything, but it’s still a bad security practise to outright disable the protection rather than allow configuration to just the hosts that require it.

How it would ideally work is:
- Default no CORs policy - no Access-Control-Allow-Origin header
- Let the user specify the value themselves (specials) and it should be a comma separated allowlist of origins (https://ui.domain.xyz,https://someindexer.foo) although they could still specify (*)

Safihre: Let's make that a separate issue: https://github.com/sabnzbd/sabnzbd/issues/3377

Safihre:
Pushed more changes. I feel it's ready now? I hope 😂
Build failure I will resolve later when I have access to my ancient Macbook again.

sabnzbd/interface.py
```diff
@@ -657,75 +686,84 @@ def get_access_info():
...
+
+@secured_expose(route="/config", check_configlock=True, methods=["GET"])
+async def config_general_index(request: Request):
+    conf = build_header(sabnzbd.WEB_DIR_CONFIG)
+    conf["configfn"] = clip_path(config.get_filename())
+    conf["cmdline"] = sabnzbd.CMDLINE
+    conf["build"] = sabnzbd.__baseline__[:7]
+
+    conf["have_7zip"] = bool(sabnzbd.newsunpack.SEVENZIP_COMMAND)
+    conf["have_sabctools"] = sabnzbd.decoder.SABCTOOLS_ENABLED
```
mnightingale, line 755:
```suggestion
    conf["have_7zip"] = bool(sabnzbd.newsunpack.SEVENZIP_COMMAND)
    conf["have_par2_turbo"] = sabnzbd.newsunpack.PAR2_TURBO
    conf["have_sabctools"] = sabnzbd.decoder.SABCTOOLS_ENABLED
```

#3379

sabnzbd/interface.py
```diff
@@ -1060,7 +1055,8 @@ async def config_general_save(request: Request):
     cfg.password.set(request.query_params.get("password"))
 
     if web_dir := request.query_params.get("web_dir"):
-        change_web_dir(web_dir)
+        if msg := change_web_dir(web_dir):
+            return report(request.query_params, error=msg)
```
mnightingale, line 1059: `change_web_dir` doesn't return anything?
I think you wanted it to return the error instead of just logging it, else return None?

edit: will be because of the removal of badParameterResponse - should return the error string

mnightingale:
Just a couple of further suggestions then I think it’s good to go, presumably you are saving it for after the 5.0 release anyway so it can be tested more and for 3.9 compatibility?

I think it's sensible to make it close to a drop-in replacement for now without changing how it works. For a future discussion and it depends if there is motivation to tackle it but there are several areas of improvements I could suggest; api (OpenAPI), frontend (React/Vue) + use api (instead of mix of rpc/public/private) + tooling (Vite), sessions/cookie, etc.

> Build failure I will resolve later when I have access to my ancient Macbook again.

You might be able to setup a trusted phone number on https://account.apple.com for 2FA rather than needing a trusted device (Mac).

Safihre:
> Just a couple of further suggestions then I think it’s good to go, presumably you are saving it for after the 5.0 release anyway so it can be tested more and for 3.9 compatibility?

Yes!

> You might be able to setup a trusted phone number on https://account.apple.com/ for 2FA rather than needing a trusted device (Mac).

Never noticed, thanks!

sabnzbd/interface.py
```diff
@@ -826,41 +858,37 @@ def saveDirectories(self, **kwargs):
...
 
-    @secured_expose(check_api_key=True, check_configlock=True)
-    def saveSwitches(self, **kwargs):
-        for kw in SWITCH_LIST:
-            if msg := config.get_config("misc", kw).set(kwargs.get(kw)):
-                return badParameterResponse(msg, kwargs.get("ajax"))
+@secured_expose(route="/config/switches/save", check_api_key=True, check_configlock=True, methods=["POST"])
+async def config_switches_save(request: Request):
+    for kw in SWITCH_LIST:
+        if msg := config.get_config("misc", kw).set(request.query_params.get(kw)):
```
Kcchouette, line 883: `request.query_params.get(kw)` -> `request.state.params.get(kw)`
`return report(request.query_params, error=msg)` -> `return report(request.state.params, error=msg)`

sabnzbd/interface.py
```diff
@@ -104,98 +112,75 @@
...
-                cherrypy.request.remote_label,
-                kwargs,
-            )
-
-        # Add X-Frame-Headers headers to page-requests
-        if cfg.x_frame_options():
-            cherrypy.response.headers["X-Frame-Options"] = "SameOrigin"
+    async def internal_wrap(request: Request, *args, **kwargs):
+        # Populate request._query_params from the correct source for this method
+        request._query_params = await get_request_params(request)
```
Kcchouette, line 142: Use a [BaseHTTPMiddleware](https://starlette.dev/middleware/#basehttpmiddleware) instead

Safihre:
Closing this but leaving the branch.

Issue #3381, opened 2026-04-24
running with "uv run python3" on Ubuntu: SABnzbd says "Platform Snap" because uv is a snap


sanderjo:
### SABnzbd version

5.0.0rc3

### Operating system

Ubuntu

### Using Docker image

None

### Description

... so we need a better, narrower detection for snap. EDIT: I propose `SNAP_NAME=sabnzbd` ENV variable.

<img width="723" height="272" alt="Image" src="https://github.com/user-attachments/assets/65afb272-78a8-44cc-b7a2-b2854ca88a8a" />

Very probably because `uv` on my Ubuntu is a snap install:

```
$ which uv
/snap/bin/uv
```

Proof

```
sander@haring1:~/git/sabnzbd$ python3 print_environ_items.py | grep -i snap
XDG_DATA_DIRS: /usr/share/gnome:/usr/local/share:/usr/share:/var/lib/snapd/desktop
PATH: /home/sander/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin

```

```
sander@haring1:~/git/sabnzbd$ uv run python3 print_environ_items.py | grep -i snap
PATH: /home/sander/git/sabnzbd/.venv/bin:/snap/astral-uv/1448/bin:/snap/astral-uv/1448/usr/bin:/home/sander/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
SNAP: /snap/astral-uv/1448
SNAP_ARCH: amd64
SNAP_COMMON: /var/snap/astral-uv/common
SNAP_CONTEXT: HYOAu96brDIbL0it-auhvHSlOgj0YdNkCHpy7dz032pXJ_yKp2Jy
SNAP_COOKIE: HYOAu96brDIbL0it-auhvHSlOgj0YdNkCHpy7dz032pXJ_yKp2Jy
SNAP_DATA: /var/snap/astral-uv/1448
SNAP_EUID: 1000
SNAP_INSTANCE_KEY:
SNAP_INSTANCE_NAME: astral-uv
SNAP_LIBRARY_PATH: /var/lib/snapd/lib/gl:/var/lib/snapd/lib/gl32
SNAP_NAME: astral-uv
SNAP_REAL_HOME: /home/sander
SNAP_REEXEC:
[...]

PR #3382, opened 2026-04-24
Use SNAP_NAME env var for more precise snap detection

sanderjo:
Correct check for sabnzbd running as snap. No more false info due to running sabnzbd from source with uv-with-snap.


sanderjo:
Solves https://github.com/sabnzbd/sabnzbd/issues/3381

Issue #3380, opened 2026-04-20
uvicorn: "host = ::" only listens on ipv6, not ipv4. Problem for dual-stack systems


sanderjo:
### SABnzbd version

d1578ce520c01a037843b7a3403f650aa29ad998

### Operating system

Ubuntu Linux

### Using Docker image

None

### Description

Setting:
```
$ cat ~/.sabnzbd/sabnzbd.ini | grep ::
host = ::
```

then SABnzbd with uvicorn only listens on ipv6:
```
$ sudo netstat -tulpen | grep 8080
tcp6       0      0 :::8080                 :::*                    LISTEN      1000       1820099    93547/venv/bin/pyth
```
With cherrypy, SAB will listen on both ipv4 and ipv6. Which is best with dual stack systems. Or at least: a user should have a way to specify dual stack.

And it now goes wrong: 
```
2026-04-20 10:40:32,558::INFO::[SABnzbd:1180] Starting web-interface on :::8080
2026-04-20 10:40:32,572::INFO::[server:224] Uvicorn running on http://[::]:8080 (Press CTRL+C to quit)
2026-04-20 10:40:32,579::INFO::[panic:229] Launching browser with http://127.0.0.1:8080
```
... and that last line will result in

<img width="1089" height="590" alt="Image" src="https://github.com/user-attachments/assets/c86e60e5-945c-4e9c-892b-f81069b06de1" />

Might be related to https://github.com/Kludex/uvicorn/discussions/1529 and https://www.tomechangosubanana.com/2025/uvicorn-dual-stack/#:~:text=Run%20uvicorn%20on%20ipv4%20and%20ipv6%20dual%2Dstack%20(empty%20string%20passed%20to%20host)%3A

sanderjo:
Hacky, but works: empty string.

in SABnzbd.py:

```
    if web_host=="::":
        web_host=""

    server_config = uvicorn.Config(
        sabnzbd.interface.app,
        host=web_host,
        port=web_port,
        log_config=uvicorn_logging_config,
        ssl_keyfile=https_key if enable_https else None,
        ssl_certfile=https_cert if enable_https else None,
        ssl_ca_certs=https_chain if enable_https else None,
    )
```

And then it works:

```
sander@zwarte:~/git/uvicorn-sabnzbd$ sudo netstat -tulpen | grep 8080
[...]

PR #3383, opened 2026-04-24
Allow empty host value in validate_host() and add test

sanderjo:
solves https://github.com/sabnzbd/sabnzbd/issues/3380

@Safihre correct PR method ... ?

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -217,6 +217,10 @@ def validate_host(value: str) -> ValidateResult:
         logging.debug("No valid host name")
         pass
 
+    # Allow empty value
+    if not value:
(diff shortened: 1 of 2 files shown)


Safihre:
No this not enough. AI is only providing a point solution. It also needs to fix get_webhost and access_info.
It's not that easy.

sanderjo:
Ah, yes, I should've seen: not good ... empty host is allowed, but still listening only on ipv4:

```
$ sudo netstat -tulpen | grep 8080
tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN      1000       7656470    330536/python3     
```

> It's not that easy.

Shall I delete this PR?

PR #3390, opened 2026-05-02
Fix PostProcessor handling of empty queues

Safihre:
User reported that end-of-queue was never called anymore in 5.0.
This seems to fix it but wanted a double check.

### DIFF
--- a/sabnzbd/postproc.py
+++ b/sabnzbd/postproc.py
@@ -299,11 +299,6 @@ class PostProcessor(Thread):
                 self.work_available.clear()
                 continue
 
-            # If queues are empty (spurious wake or race condition), clear and loop back
-            if self.slow_queue.empty() and self.fast_queue.empty():
-                self.work_available.clear()
-                continue
-
             # Something in the fast queue?
             try:
                 # Every few fast-jobs we should allow a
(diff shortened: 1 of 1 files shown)


mnightingale (commented):
I'm not 100% sure on this logic because the removed part and what it called 'spurious wake' I think is because `resume()` sets the event.

I'm not sure how to trigger the pause/resume of pp to test it.

But with these changes if the event is set (resume), fast and slow queue are empty, it'll hit the `handle_empty_queue()` which at first glance seems ok because it checks if nzbqueue has items, but I suspect it may race when items are being removed.

Safihre:
This should catch it?

mnightingale:
I think that's better, `empty()` and `qsize()` are not guaranteed to be accurate.

Before it was possible although unlikely to skip fast and try slow, and perform the empty_queue operation without ever checking fast. The comment change "No fast or slow jobs, better luck next loop!" is a little misleading. The use of `empty()` is probably good enough.

An alternative logic flow and avoiding empty and qsize:

```py
queues = (
    (self.fast_queue, self.slow_queue)
    if self.__fast_job_count < MAX_FAST_JOB_COUNT
    else (self.slow_queue, self.fast_queue)
)
# get_nowait, break
# if nzo is None:  # after checking both handle the empty queue
# counter stuff, if queues[0] is self.fast_queue
```

PR #3395, opened 2026-05-03
Use UUID4 for nzo_id and rekey existing duplicates

### LINKED ISSUE #3391: History nzo_id could not be unique
### SABnzbd version

5.0

### Operating system

Not relevant

### Using Docker image

None

### Description

https://www.reddit.com/r/SABnzbd/comments/1t1lraj/sql_error_unique_constraint_failed_historynzo_id/?utm_source=share&utm_medium=mweb3x&utm_name=mweb3xcss&utm_term=1&utm_content=share_button

Wonder how that happened and if more users experience this. Could have a pre-check before we do the migration.

mnightingale:
Fixes #3391

The NZO id wasn't always unique, I suspect there was a bug at some point causing history items to be inserted multiple times.

This switches nzo_id to UUID4 going forward so we don't need to worry about collisions, and rekeys existing duplicates.

- `with self.connection:` runs the migration in a transaction; should get in the habit of doing this, better to fail entirely rather than make some of the changes
- `CREATE [UNIQUE] INDEX IF NOT EXISTS` successful migrations to 6 make no changes to the schema
- `NZO_FILE` because repair job cleans up `SABnzbd_nz?` and there can only be one NZO per admin folder[^1] avoiding path length issues, similar to the SABnzbd_attrib file.

future nzo files are prefixed "SABnzbd_nzo_" but the per queue item file is called "SABnzbd_nzo", most of the changes are dealing with nzo_id no longer having the prefix but there being several cleanups looking for SABnzbd_nz?_* files.

[^1]: existing queue items will have the old naming for \_\_admin\_\_ (SABnzbd_nzo_IDIDIDID), but subsequent saves will be to SABnzbd_nzo, I don't think it's a problem for repairs because it tries to loa SABnzbd_nzo first.

### DIFF
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -49,7 +49,7 @@ JOB_ADMIN = "__ADMIN__"
 VERIFIED_FILE = "__verified__"
 RENAMES_FILE = "__renames__"
 ATTRIB_FILE = "SABnzbd_attrib"
-NZO_FILE = "SABnzbd_nzo"
+NZO_FILE = "SABnzbd_nzo_data"
 REPAIR_REQUEST = "repair-all.sab"
 
 SABCTOOLS_VERSION_REQUIRED = "9.4.0"
--- a/sabnzbd/constants.py
+++ b/sabnzbd/constants.py
@@ -49,6 +49,7 @@ JOB_ADMIN = "__ADMIN__"
 VERIFIED_FILE = "__verified__"
 RENAMES_FILE = "__renames__"
 ATTRIB_FILE = "SABnzbd_attrib"
+NZO_FILE = "SABnzbd_nzo"
 REPAIR_REQUEST = "repair-all.sab"
 
 SABCTOOLS_VERSION_REQUIRED = "9.4.0"
--- a/sabnzbd/database.py
+++ b/sabnzbd/database.py
@@ -116,22 +116,38 @@ class HistoryDB:
                     "ALTER TABLE history ADD COLUMN time_added INTEGER;"
                 )
             if version < 7:
-                with self.connection:
-                    self.execute("PRAGMA user_version = 7;")
-                    if self.execute("""
+                try:
+                    self.connection.execute("BEGIN")
+                    self.execute("PRAGMA user_version = 7", raise_on_error=True)
+                    if self.execute(
+                        """
                         WITH duplicates AS (SELECT id, ROW_NUMBER() OVER (PARTITION BY nzo_id ORDER BY id) AS rn FROM history)
                         SELECT id FROM duplicates WHERE rn > 1
-                        """):
+                        """,
+                        raise_on_error=True,
+                    ):
                         for (row_id,) in self.cursor.fetchall():
-                            self.execute("UPDATE history SET nzo_id = ? WHERE id = ?", (str(uuid.uuid4()), row_id))
-                    self.execute("CREATE UNIQUE INDEX IF NOT EXISTS idx_history_nzo_id ON history(nzo_id);")
+                            self.execute(
+                                "UPDATE history SET nzo_id = ? WHERE id = ?",
+                                (str(uuid.uuid4()), row_id),
+                                raise_on_error=True,
+                            )
                     self.execute(
-                        "CREATE INDEX IF NOT EXISTS idx_history_archive_completed ON history(archive, completed DESC);"
+                        "CREATE UNIQUE INDEX IF NOT EXISTS idx_history_nzo_id ON history(nzo_id)",
+                        raise_on_error=True,
                     )
+                    self.execute(
+                        "CREATE INDEX IF NOT EXISTS idx_history_archive_completed ON history(archive, completed DESC)",
+                        raise_on_error=True,
+                    )
+                    self.connection.commit()
+                except:
+                    self.connection.rollback()
+                    raise
 
             HistoryDB.startup_done = True
 
-    def execute(self, command: str, args: Sequence = ()) -> bool:
+    def execute(self, command: str, args: Sequence = (), raise_on_error: bool = False) -> bool:
         """Wrapper for executing SQL commands"""
         for tries in (4, 3, 2, 1, 0):
             try:
@@ -145,6 +161,8 @@ class HistoryDB:
                     continue
                 elif "readonly" in error:
                     logging.error(T("Cannot write to History database, check access rights!"))
+                    if raise_on_error:
+                        raise
                     # Report back success, because there's no recovery possible
                     return True
                 elif match_str(error, ("not a database", "malformed", "no such table", "duplicate column name")):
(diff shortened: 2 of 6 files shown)


sabnzbd/nzb/object.py
```diff
@@ -1428,11 +1430,12 @@ def purge_data(self, delete_all_data=True):
             # If duplicate is discarded during URL-fetches, no nzo_id is known yet
             if self.nzo_id:
                 # Remove temporary file left from URL-fetches
-                remove_data(self.nzo_id, self.admin_path)
+                remove_data(f"SABnzbd_nzo_{self.nzo_id}", self.admin_path)
```
mnightingale: Could there be existing future types with old ids? 

If so it should do

```
remove_data(self.nzo_id, self.admin_path)
remove_data(f"SABnzbd_nzo_{self.nzo_id}", self.admin_path)
```

Similar at nzbqueue.save maybe check if startswith SABnzbd_nzo_

tests/data/tavern/api_queue_format.yaml
```diff
@@ -56,7 +56,7 @@ stages:
           finish: !anyint
           slots: !anylist
             index: !re_match "[0-9]+"
-            nzo_id: !re_match "SABnzbd_nzo_.*"
+            nzo_id: "!anystr"
```
mnightingale, line 59: Even without this change the test passes and I'm not sure why, the xml response test is already checking for !anystr

sabnzbd/database.py
```diff
@@ -114,12 +115,19 @@ def connect(self):
...
                 )
-            if version < 6:
-                _ = (
-                    self.execute("PRAGMA user_version = 6;")
-                    and self.execute("CREATE UNIQUE INDEX idx_history_nzo_id ON history(nzo_id);")
-                    and self.execute("CREATE INDEX idx_history_archive_completed ON history(archive, completed DESC);")
-                )
+            if version < 7:
+                with self.connection:
+                    self.execute("PRAGMA user_version = 7;")
```
mnightingale: Need to rethink this because execute catches the exception that the context probably relies on, maybe use self.cursor.execute directly or add a throws=True arg

sabnzbd/nzb/object.py
Safihre: Yes that's possible. If somebody has just read an RSS feed before updating and dumped 100 URLs in the queue.

sabnzbd/constants.py
```diff
@@ -49,6 +49,7 @@
 VERIFIED_FILE = "__verified__"
 RENAMES_FILE = "__renames__"
 ATTRIB_FILE = "SABnzbd_attrib"
+NZO_FILE = "SABnzbd_nzo"
```
Safihre: Maybe SABnzbd_nzo_data?

thezoggy:
so we are going from `SABnzbd_nzo_12345678` which is 20 characters to a uuid.uuid4 which is 36 characters..  which get shoved in admin folder of a job.. so id be worried about if this causes stuff to be longer than path allowances. as the increase in length will make it much more likely

Safihre:
Why? This is the incomplete folder, we only have the job folder and the rars. Unpacking is only in the Complete so there we get path problems.

mnightingale:
> so we are going from `SABnzbd_nzo_12345678` which is 20 characters to a uuid.uuid4 which is 36 characters..

No, for that reason it will store `__admin__/SABnzbd_nzo_data` since each admin directory only needs to store one. Only future in the config directory will temporarily store `SABnzbd_nzo_11eef886-ae99-4a9a-93aa-0c7ee4f5d5b5`.

sabnzbd/nzbqueue.py
```diff
@@ -312,7 +321,7 @@ def get_nzo(self, nzo_id) -> Optional[NzbObject]:
     def add(self, nzo: NzbObject, save: bool = True, quiet: bool = False) -> str:
         # Can already be set for future jobs
         if not nzo.nzo_id:
-            nzo.nzo_id = sabnzbd.filesystem.get_new_id("nzo", nzo.admin_path, self.__nzo_table)
+            nzo.nzo_id = str(uuid.uuid4())
```
Safihre, line 326: Why not use `uuid.uuid4().hex`?

mnightingale: ~~Although it would be 16 instead of 36 bytes.~~
It's not human readable compared to dashes and alphanumeric and would require conversion everywhere for use in logs, the api would probably have to accept the 36 byte version then convert it to query the database.
The storage saving, in memory or database are just not worth the tradeoffs.

It would just add a load of complexity for a negligible store saving.

edit: mis read for bytes

Safihre:
What if we truncate it a bit?
I asked AI:

If your total pool is max 100,000 IDs, then truncation can work—but pick enough chars.

Approx collision chance for generating 100k random IDs from `uuid4().hex[:k]`:

•  k=8 → ~69% (too high)
•  k=10 → ~0.45%
•  k=11 → ~0.028%
•  k=12 → ~0.0018% (about 1 in 56k such runs)

sabnzbd/nzbqueue.py
mnightingale: Oops misread hex for bytes, hex would be ok it's the same as string just without the dashes so 32 characters.
Going from 20 to 32 seems fine?
There are already fields taking up way more space in the database, multiple names, paths, logs.

mnightingale: But is saving 4 characters worth losing the readability of having dashes?

Safihre: I guess it doesn't really matter indeed. We can keep the dashes.

sabnzbd/database.py
mnightingale: Along with execute consuming exception `isolation_level = None` prevents the default context/transaction behaviour

sabnzbd/database.py
```diff
@@ -114,16 +115,39 @@ def connect(self):
...
+                        raise_on_error=True,
+                    )
+                    self.execute(
+                        "CREATE INDEX IF NOT EXISTS idx_history_archive_completed ON history(archive, completed DESC)",
+                        raise_on_error=True,
+                    )
+                    self.connection.commit()
+                except:
+                    self.connection.rollback()
+                    raise
```
mnightingale, line 146: Not sure what to do here instead of `raise`, I don't think the migration can fail now but there should be handling now and in the future so if they do we shutdown or something to prevent attemping to carry on regardless.

sabnzbd/nzbqueue.py
```diff
@@ -238,8 +247,10 @@ def save(self, save_nzo: Union[NzbObject, None, bool] = None):
                     if not nzo.futuretype:
                         # Also includes save_data for NZO
                         nzo.save_to_disk()
-                    else:
+                    elif nzo.nzo_id.startswith("SABnzbd_nzo_"):
```
Safihre, line 250: Why do we need this? Worst case we just save it in a new file?

mnightingale: It is for saving futures that are already in the queue so have old IDs, if the nzo already exists on disk (SABnzbd_nzo_xyz) then without this it would save a duplicate (SABnzbd_nzo_SABnzbd_nzo_xyz) which I think would leave one available for queue repair/orphans.

PR #3397, opened 2026-05-03
Fix spurious FileExistsError logs when download directories already e…

mordax7:
create_all_dirs checks os.path.exists() before calling os.mkdir(), the directory can appear between the check and the mkdir (concurrent threads, stale dentry cache, etc.), resulting in a FileExistsError that the outer except OSError treats as a real failure and logs repeatedly.

Catch FileExistsError directly on os.mkdir() instead. If the path is a directory, continue normally. If something else occupies the name, re-raise.

### DIFF
--- a/sabnzbd/filesystem.py
+++ b/sabnzbd/filesystem.py
@@ -740,19 +740,22 @@ def create_all_dirs(path: str, apply_permissions: bool = False) -> Union[str, bo
             if not os.path.exists(path):
                 os.makedirs(path)
         else:
-            # We need to build the directory recursively, so we can
-            # apply permissions to only the newly created folders
-            # We cannot use os.makedirs() as it could ignore the mode
+            # Build the directory tree one level at a time so we can
+            # apply permissions to only the newly created folders.
+            # We cannot use os.makedirs() as it could ignore the mode.
             path_part_combined = "/"
             for path_part in path.split("/"):
                 if path_part:
                     path_part_combined = os.path.join(path_part_combined, path_part)
-                    # Only create if it doesn't exist
-                    if not os.path.exists(path_part_combined):
+                    try:
+                        # Only create if it doesn't exist
                         os.mkdir(path_part_combined)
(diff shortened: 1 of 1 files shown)


mnightingale:
Likely caused by 1833173bac5ac49c7258c766d02e683550c6aac6 so could just add the global lock back to create_all_dirs, creating a directory structure should be quick? I recall the main issue was move_to_path and renamer because users could move between local an network shares.

Alternatively fix the cause, add nzo.lock to `create_all_dirs` calls, but also be suspicous of and of the other removed locks in that commit. There will be TOCTOU races, os.path.exists followed by create_all_dirs etc.

Do you have a stacktrace of the failure you experienced?

mordax7:
Hey hey @mnightingale :wave: 

Yeah, looks like 1833173 exposed this.

I went with catching `FileExistsError` instead of re-adding the lock because a lock only helps against other SABnzbd threads, on network filesystems `os.path.exists()` can return stale results regardless. The try/except makes `create_all_dirs` properly idempotent without needing any lock.

Stacktrace (SABnzbd in container, downloads on NFS):

```
  File "/app/sabnzbd/sabnzbd/filesystem.py", line 741, in create_all_dirs
    os.mkdir(path_part_combined)
FileExistsError: [Errno 17] File exists: '/data/downloads/complete'
```

Downloads worked fine, dirs existed, just log spam from the `except OSError` handler + `create_real_path` logging on top.

Safihre:
@mnightingale This solution doesn't seem too bad? Otherwise I almost can't think of a way that would not include re-introducing global lock.

mnightingale:
It looks fine, maybe keep the "# Try to set permissions if desired, ignore failures" comment.

As they say locks aren't really necessary since they don't protect against someone else deleting or creating directories outside of SAB.

Safihre:
@mordax7 Can you add back the removed comment?

mordax7:
@Safihre done.

Safihre:
Thanks 👍

PR #3399, opened 2026-05-03
Fix RSS default priority rules

### LINKED ISSUE #3398: RSS feeds not respecting default Priority setting
### SABnzbd version

5.0.1

### Operating system

Windows

### Using Docker image

None

### Description

If you change the default Priority setting to Paused while leaving the filter set as Default it doesn't add the nzb as paused. You have to change the individual filter to Paused to get it to work like it used to. 

Image

mnightingale:
Fixes #3398


Safihre:
How did that happen? 😅

iggles:
I don't think this was actually fixed. I just tried on 5.0.2RC1 and it's still adding things unpaused.

Issue #3389, opened 2026-05-02
Bad try list blocking downloads


Safihre:
> Resetting bad trylist still blocking downloads in 5.0.1
> 
> trylist.txt is the overnight log (info only), I re-added the download with debug logs as .38.log and it hangs on a different file
> 
> [Der.Schaum.der.Tage.Extended.2013.German.1080p.BluRay.x264-ENCOUNTERS.zip](https://github.com/user-attachments/files/27297647/Der.Schaum.der.Tage.Extended.2013.German.1080p.BluRay.x264-ENCOUNTERS.zip) 

 _Originally posted by @nobrainletsallowed in [#3287](https://github.com/sabnzbd/sabnzbd/issues/3287#issuecomment-4363084093)_

mnightingale:
I've not looked into it yet but @nobrainletsallowed you need to reset your indexers rss token - see sab-trylist.txt
You need to be really careful sharing the unsanitised log.

Safihre:
Indeed, we sanitize the log if you click "Show logging".

nobrainletsallowed:
whoopsie :^) updated token

@Safihre can you remove the [original comment](https://github.com/sabnzbd/sabnzbd/issues/3287#issuecomment-4363084093) + the link to trylist.txt in the OP? I can't edit the original comment as the issue is locked

mnightingale:
The primary cause is because initially all the NZF have the same name.

If I modify the NZB replacing the subject lines removing what's enclosed in the first set of quotes `&#34;` is downloads successfully.

I think the problem is with prepare_filepath and maybe nzf.file_lock when opening files, since they initially want the same name.

I've logged the filename prepare_filepath creates, before the par2 is parsed there are duplicates.
The main problem is prepare_filepath is just looking to see if the name exists on disk, it's not reserving it so before the file is created there can be duplicates.

It would need to use locks but get_unique_filename could take a set of existing names that have been reserved but may not yet exist on disk?

The simplest fix is just to make nzb add/import ensure all filenames are unique initially.

Safihre:
That could also result in problems if we do it at import.
Some way to only do it when we actually start writing only?

mnightingale:
I've tried adding locks to prepare_filepath and moving where it's called from but it hasn't resolved the problem of getting stuck and I haven't figured out why yet. So I think the duplicate filenames was only part of the problem.

mnightingale:
I'm ending up with:

[...]

PR #3400, opened 2026-05-04
Fix stuck queue when articles complete while removed from queue

mnightingale:
Fixes one issue causing #3389

The linked issue triggers a situation where the item moves back and forth between the queue and postproc, the use of `article.nzf.nzo.removed_from_queue` allowed discarded requests to be ignored leaving the request never sent but left assigned to a server resulting in a stuck queue if the item is return to the queue.

### DIFF
--- a/sabnzbd/newswrapper.py
+++ b/sabnzbd/newswrapper.py
@@ -500,7 +500,12 @@ class NewsWrapper:
         retry_article: bool = True,
     ) -> None:
         """Discard an article back to the queue"""
-        if article and not article.nzf.nzo.removed_from_queue:
+        if article:
+            # The job might return to the queue in the future so clear the current server
+            if article.nzf.nzo.removed_from_queue:
+                article.allow_new_fetcher()
+                return
+
             # Only some errors should count towards the total tries for each server
             if count_article_try:
                 article.tries += 1


mnightingale:
Instead of creating the files I'm open to making a `get_unique_filename` method on NzbObject.

Safihre:
> Instead of creating the files I'm open to making a `get_unique_filename` method on NzbObject.

That sounds like such a nice clean approach 👍

And thanks of course for the deep dive into this issue 🎉

mnightingale:
I'm not sure if restoring used filenames from `files_table` is failing, maybe it would be better to just include it in NzbObjectSaver

mnightingale:
> I'm not sure if restoring used filenames from `files_table` is failing, maybe it would be better to just include it in NzbObjectSaver

Seems to just be https://github.com/sabnzbd/sabnzbd/issues/3402#issuecomment-4377084204 not worth duplicating in NzbObjectSaver as it would have the same potential issue.

sabnzbd/newswrapper.py
```diff
@@ -500,7 +500,12 @@ def discard(
         retry_article: bool = True,
     ) -> None:
         """Discard an article back to the queue"""
-        if article and not article.nzf.nzo.removed_from_queue:
+        if article:
```
Safihre, line 503: Maybe combine in 1 if?

mnightingale: Do you mean combine it with https://github.com/sabnzbd/sabnzbd/pull/3400/changes#diff-6cf7245b84be1f361101ed6ab6ef1466ccd0ef9365facc3d7d65645be654a7b3L509 ?

It did consider it but a big if makes the logic more complicated, it can't do`if article and not article.nzf.nzo.removed_from_queue:` because it always needs to either `sabnzbd.Downloader.decode` or `article.allow_new_fetcher()` otherwise it can never be downloaded again.

Safihre: Aah like that. OK 👍

Safihre:
Isn't only the second problem really causing the stuck queue?

mnightingale:
> Isn't only the second problem really causing the stuck queue?

I can split it into two PRs?

The problem NZB in the issue hits both and each on their own wouldn't resolve its issues.

mnightingale:
I've split this PR in two

PR #3406, opened 2026-05-05
Simplify postproc queue storage and resolve serialization issues

### LINKED ISSUE #3396: RuntimeError: dictionary changed size during iteration when saving postproc2.sab
### SABnzbd version

Version: 5.0.1-ls251 (Docker) 

### Operating system

Alpine Linux v3.23.4, kernel 6.12.57+deb13-amd64

### Using Docker image

Other

### Description

When saving the postproc queue (postproc2.sab), SABnzbd throws a RuntimeError due to a thread-safety issue in filesystem.py. The error occurs consistently when multiple jobs finish post-processing simultaneously at high download
   speeds (100+ MB/s with 2 Usenet servers).                                                                                                                                                                                          
                                                                                                                                                                                                                                      
  Traceback:                                                                                                                                                                                                                          
    File "/app/sabnzbd/sabnzbd/filesystem.py", line 1079, in save_data
      pickle.dump(data, data_file, protocol=pickle.HIGHEST_PROTOCOL)                                                                                                                                                                  
                                                                                                                                                                                                                                      
  The postproc queue dictionary appears to be modified by one thread while another is serializing it via pickle.dump(). This happens repeatedly, not just once.                                                                       
   
  Setup: Docker (linuxserver/sabnzbd:5.0.1-ls251), Alpine Linux 3.23.4, 2 Usenet servers running in parallel, NNTP pipelining enabled.

mnightingale:
Fixes #3396

Pickling the whole list of NzbObject is problematic, to resolve the mutation problems in the issue we'd have to lock all the NzbObjects and their NzbFiles throughout the save process (bad).

But we don't actually need to pickle the objects, they should already be on disk so if we reuse the storage approach that nzbqueue uses and just store the work_name and nzo_id, we can reload the pickled files when required.

While testing I found and fixed a bug with `nzo.save_to_disk()`, it should be called before `nzo.removed_from_queue = True` otherwise it will not save, and I needed to ensure it's saved before postproc.


sabnzbd/postproc.py
```diff
@@ -150,11 +157,18 @@ def load(self):
             if POSTPROC_QUEUE_VERSION != version:
                 logging.warning(T("Old queue detected, use Status->Repair to convert the queue"))
             elif isinstance(history_queue, list):
-                self.history_queue = [nzo for nzo in history_queue if os.path.exists(nzo.download_path)]
+                for item in history_queue:
```
Safihre: Shouldn't it have compatibility for the previous format? If anyone updated with active pp queue?

mnightingale: I wrote that originally, then realised the POSTPROC_QUEUE_VERSION change means it'll never be reached.

Old queues will have to restore orphaned files via the status screen.

~~Changing POSTPROC_QUEUE_VERSION keeps it compatible with downgrades to 4.5.5, where queue repair would also be required.~~

edit: The version change means it'll load a different file anyway, postproc3.sab

Safihre:
Indeed cleaner, but wouldn't simply using `nzo.lock` not also resolve it?

Safihre:
Nevermind, we can't of course.

Safihre:
Wait, we can just `@synchronized` the `__getstate__` of `NzbObject`?

mnightingale:
> Wait, we can, we can just `@synchronized` the `__getstate__` of `NzbObject`?

I should have explained why that approach is problematic, it's what I tried to do first.

`__getstate__` can't use `@synchronized` it gets error `_pickle.PicklingError: Can't pickle local object <function synchronized.<locals>.wrap.<locals>.call_func` maybe that can be resolved but for now just switching to `with self.lock` inside also won’t allow consistent snapshots or resolve the list/dict mutation.

- Postproc.save
  - pickle.dump, for each NzbObject
    - (locks) NzbObject and creates a dict
    - (locks) NzbFile
    - returns a list of dicts
    - (unlocked) Pickles/serializes the dict, but it has copied mutable properties (files, articles)

At a glance it appears the locks would allow getting a consistent snapshot, but because the lock is released after return the copied mutable properties may be modified which is the same error the issue is showing.

Possible solutions that might help would be to create shallow copies within `__getstate__` of the mutable properties, alternatively lock all the NzbObject and NzbFile before pickle.dump is called, but both those approaches are expensive, complicated, or just poor design.

This data is only needed for restarts to know which jobs are in postproc, so I am avoiding all this complexity by just behaving like nzbqueue and saving paths and ids.

I suspect the main cause currently is there are no locks on postproc, so a few paths that could result in active items in `history_queue`.

nzbqueue.fail_to_history => PostProcessor.process => PostProcessor.save
Assembler.run => PostProcessor.process => PostProcessor.save
PostProcessor.run => might NzbQueue.add() + PostProcessor.remove() => PostProcessor.save

So multiple threads using history_queue and you might end up racing and saving something that has been added back to the queue, hence files and articles changes.

Flagging that PostProcessor needs to save and handling it in its thread could be one approach but that has other complexities. I understand that the persistence of state is more important hence why it wants to save everywhere, so I just added `@synchronized(POSTPROC_LOCK)` anywhere that touches `history_queue`. Jobs are not being added/removed for postproc frequently enough for the locks to matter.

---

Instead of the change I've made to end_job and avoid risk that the mutable properties do change, maybe I should look at why it wasn't beign saved...

https://github.com/sabnzbd/sabnzbd/blob/fc214a7ef9932246eebbc62575976a3d0e460473/sabnzbd/nzbqueue.py#L756-L768

`nzo.save_to_disk()` is only called if file_done AND it has been several seconds, `post_done => end_job` can be reached without saving.

Take the 100MB test file, it might only save after the first file completes because the rest downloads too quick.

So maybe just `if file_done and (post_done or nzo.next_save is None or time.time() > nzo.next_save):` then the post_done path doesn't need to save which also avoids possibly saving twice.

edit: maybe it's intentially not after each file and the post_done path should have a `nzo.save_to_disk(force=True)`

I'll look at it this evening, and probably make a separate PR.

mnightingale:
I'll fix the `nzo.save_to_disk()` issue separately, I noticed it because I had a breakpoint in save, copied the postproc3.sab and job directory, let a job finish, removed it from history and restarted with the restored postproc3.sab and job directory. It recorded the incorrect metadata in the history, because it was an out of date version of the NZO on disk.

So I think the main concern is just how POSTPROC_QUEUE_VERSION is handled, it's odd having it in the filename and contents because the warning (logging.warning(T("Old queue detected, use Status->Repair to...) is never going to be reached.

Safihre:
Yeah oke now it makes sense about the locks.

Indeed that version check is pretty bad... What suggestions are you thinking about?

sabnzbd/postproc.py
```diff
@@ -134,11 +138,14 @@ def __init__(self):
         self.__busy = False
         self.paused = False
 
+    @synchronized(POSTPROC_LOCK)
     def save(self):
         """Save postproc queue"""
         logging.info("Saving postproc queue")
-        sabnzbd.filesystem.save_admin((POSTPROC_QUEUE_VERSION, self.history_queue), POSTPROC_QUEUE_FILE_NAME)
+        snapshot = [os.path.join(nzo.work_name, nzo.nzo_id) for nzo in self.history_queue]
```
Safihre: Can't we store more of the path so we don't need old/new-style handling for the nzo-data file in the `load()`?

mnightingale: There are a few ways I can think of, my preferred I think is shifting it to save:

Using absolute path might be sensible if we ever do something like allowing different incomplete paths per job, worse that happens is the path doesn't exist and the user repairs the queue which isn't really unexpected.

```py
snapshot = []
for nzo in self.history_queue:
    if nzo.nzo_id.startswith("SABnzbd_nzo_"):
        snapshot.append(os.path.join(nzo.admin_path, nzo.nzo_id))
    else:
        snapshot.append(os.path.join(nzo.admin_path, NZO_FILE))
```

This could also be handed off to a property on NzbObject.

What you lose is you can't check the expected nzo_id was loaded, for instance if a path is reused, so I'd maybe store a tuple of (path, id), sensible to check in case the same path gets used again.

mnightingale:
> Indeed that version check is pretty bad... What suggestions are you thinking about?

I’d consider the history/database approach more sensible, it’s always called `history1.db` and uses `user_version` to store the version of the schema and migrate.

That’s close to what storing POSTPROC_QUEUE_VERSION within the data was probably for, so we could follow a similar pattern and facilitate a migration by keeping it always called `postproc2.sab`, looking at some of the constants it appears this kind of thing has come up before and decided regularly losing the users history/state is not a good idea.

Downgrading to 4.5.5 would load the same file but warn the versions don’t match so nothing broken.

Something like this, a little different/simpler if absolute paths are used.

```py
version, history_queue = data
if version == 2:
    self.history_queue = [nzo for nzo in history_queue if os.path.exists(nzo.download_path)]
elif version == 3:
    for item in history_queue:
        folder, _id = os.path.split(item)
        normal_path = get_admin_path(folder, future=False)
        if _id.startswith("SABnzbd_nzo_"):
            nzo = sabnzbd.filesystem.load_data(_id, normal_path, remove=False)
        else:
            nzo = sabnzbd.filesystem.load_data(NZO_FILE, normal_path, remove=False)
        if nzo and nzo.nzo_id == _id:
            self.history_queue.append(nzo)
else:
    logging.warning(T("Old queue detected, use Status->Repair to convert the queue"))

```

This storage hasn’t changed for ~15 years so hopefully wont again with how simple it's now.

mnightingale:
This seems much neater and keeps compatability with 4.5.5.

What do you think about the use of absolute paths, maybe remove for now and worry about it if required in the future?

edit: I've changed to relative to download_dir for now

mnightingale:
I think comment "Is this correct for featuretype?" was deleted, but you're correct for relative paths. Absolute is ok because `self.admin_path` gives the full path to admin_dir/download_dir.

It could do this, but the caller would have to know to use it against `cfg.admin_dir` or `cfg.download_dir`, or try both.

```py
if absolute:
    return os.path.join(self.admin_path, filename)
if self.futuretype:
    return os.path.join(FUTURE_Q_FOLDER, filename)
return os.path.join(self.work_name, JOB_ADMIN, filename)
```

It'll never be futuretype where postproc uses it, so leave it as it is? FUTURE_Q_FOLDER? or None?

PR #3412, opened 2026-05-13
Reduce time to test

mnightingale:
Locally on macOS and Linux `CI=1 pytest -q` I can run the tests in approximately 1m40s saving about 1m.
The benefit in GitHub runners is much more variable, some runs are at best maybe 30-45 seconds faster on Linux runners, macOS/Windows less so due to fewer resources, so the overall time is not usually significantly better but you do see the failures from Linux runners quicker. I run the tests a lot locally so see a pretty big difference.

Replaces some test decorators with fixtures, primarily to resolve an issue where you can't use both but also allows some things like dynamic config without needing nested functions.

The main change is removing a lot of long sleeping which really adds up, it also fixes a few cases where the tests do try/except, but don't xfail if the expected value is never seen and just carry on.

The remaining slow tests are mostly due to `run_sabnzbd` I'm not sure I can do anything with them, maybe in the future making it works with `pytest-xdist` with multiple instances running would give a benefit but it requires quite a few changes such as making the tests deterministic.

edit: overall runtime on GitHub runners is probably around 30m vs 35m

### DIFF
--- a/interfaces/Config/templates/staticcfg/js/script.js
+++ b/interfaces/Config/templates/staticcfg/js/script.js
@@ -423,11 +423,11 @@ $(document).ready(function () {
                     do_restart();
                 } else {
                     $('#config_err_msg').text(" ");
-                    setTimeout(config_success, 1000);
+                    setTimeout(config_success, config.disableDelays ? 0 : 1000);
                 }
             } else {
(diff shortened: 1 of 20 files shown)


Safihre:
Nice.

PR #3413, opened 2026-05-14
Extend Cleanup List with filename and path pattern matching

Safihre:
Previously limited to cleaning up files based only on their extensions, the 'Cleanup List' now supports more flexible matching. Users can specify:
- Exact filenames (e.g., `Thumbs.db`)
- Wildcard filename patterns (e.g., `*.tmp`, `cleanup.*`)
- Relative path patterns (e.g., `images/*`, `*/test.jpg`)

This enhancement provides greater control and flexibility for automated post-processing cleanup.

Based on request in https://forums.sabnzbd.org/viewtopic.php?t=27245

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -25,7 +25,7 @@ import re
 import argparse
 import socket
 import ipaddress
-from typing import Union, Optional
+from typing import Union
 
 import sabnzbd
 from sabnzbd.config import (
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -25,7 +25,7 @@ import re
 import argparse
 import socket
 import ipaddress
-from typing import Union
+from typing import Union, Optional
 
 import sabnzbd
 from sabnzbd.config import (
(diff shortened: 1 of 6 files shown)


sabnzbd/misc.py
```diff
@@ -853,17 +853,43 @@ def get_platform_description() -> str:
...
+
+            # Type 2: Filename/wildcard pattern (has dots, no slashes)
+            elif "." in entry and "/" not in entry and "\\" not in entry:
+                if safe_fnmatch(filename_lower, entry):
+                    return True
+
+            # Type 3: Path pattern (has slashes)
+            elif "/" in entry or "\\" in entry:
+                if relative_path:
+                    relative_path_lower = relative_path.lower().replace("\\", "/")
```
mnightingale: I suggest pulling this out of the loop so it's only calculated once, like `filename_lower`

tests/test_misc.py
```diff
@@ -237,6 +237,51 @@ def test_on_cleanup_list(self):
...
+
+    # cleanup_list is always lowercased by validator
+    @pytest.mark.config({"cleanup_list": ["images/*", "*/test.jpg", "cache/*.log"]})
+    def test_on_cleanup_list_path_patterns(self):
+        # Path patterns with relative paths
+        assert misc.on_cleanup_list("file.jpg", relative_path="images/file.jpg")
+        assert misc.on_cleanup_list("pic.png", relative_path="images/pic.png")
+        assert not misc.on_cleanup_list("file.jpg", relative_path="file.jpg")
+        assert misc.on_cleanup_list("test.jpg", relative_path="subfolder/test.jpg")
+        assert misc.on_cleanup_list("test.jpg", relative_path="deep/nested/test.jpg")
```
mnightingale, line 269: This works correctly, but I'd suggest adding a specific test for a nested path to check it doesn't break sometime i.e. `deep/*/pic.png` (needs different name so doesn't hit `*/test.jpg`)

I'm not sure if it's intentional for wildcards to match multiple levels, `deep/a/b/test.jpg` would be matched by `deep/*/test.jpg`

sabnzbd/cfg.py
```diff
@@ -138,6 +138,28 @@ def lower_case_ext(value: Union[str, list]) -> tuple[None, Union[str, list]]:
     return None, value.lower().strip(" .")
 
 
+def cleanup_list_validator(value: Union[str, list]) -> tuple[None, Union[str, list]]:
+    """Validate cleanup list entries
+    Accepts extensions (exe, nfo), filenames (Thumbs.db), patterns (*.tmp), and paths (images/*)
+    Rejects entries containing .. to prevent directory traversal
```
jcfp: the last comment looks off, with `validate_entry` returning paths unchanged except for lowercasing and stripping spaces from both ends.

Safihre: Oh yeah that's something I removed indeed. That's somewhere else.

albino1:
I wonder if this would also help fix this issue:

https://github.com/sabnzbd/sabnzbd/issues/2577

I know you [gave it a shot](https://github.com/sabnzbd/sabnzbd/pull/3162) once before, but the way SAB handles cleanup was messy.

Safihre:
@albino1 It won't fix that for sure :(

Safihre:
@mnightingale Since the changes to the test, I do get quite a few of intermitted failures.
Did you see?

PR #3415, opened 2026-05-15
Check SAB_CACHE_DIR exists first and allow to already exist

mnightingale:
This is a little closer to what it used to be.

If it continues to have errors I suggest repeated tries like teardown does.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,9 +35,10 @@ from tests.testhelper import *
 def clean_cache_dir(request):
     # Remove cache if already there
     try:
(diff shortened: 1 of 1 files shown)


mnightingale:
Will create a separate issue to discuss the test_failed_connect_allows_retry failures, if I can't find a fix.
I think I know why it's a bit unpredictable.

Safihre:
It seems oddly Python 3.12 specific, I've never seen it fail on any other than 3.12? 
Although I already merged the removal.

PR #3425, opened 2026-05-19
Match to previous RSS behaviour

### LINKED ISSUE #3398: RSS feeds not respecting default Priority setting

mnightingale:
Fixes #3398

The previous refactoring has changed RSS behaviour,because I missunderstood some of the stateful aspect of the rules.

The error is that the first decisive rule controls subsequent fallback behavior, rather than directly supplying the final values.
For example, where it was doing `first_not_none(rule.pp, feed.pp, category.pp)` the original is closer to:

```py
if rule.pp:
    use it
elif no rule category and no RSS category:
    use category pp
else:
    keep feed pp
```

So, the precedence/logic is:

- reject rules still matter
- RSS categories suppress category defaults
- feed defaults can beat more-specific metadata
- category defaults are not true defaults

There are several tests you added, their names now don't really reflect what the expected or actual outcome they test for is. They're very similar or the same as some of the parametrize tests, so I could just add them in there?

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -742,7 +742,7 @@ class RSSReader:
         download: bool,
         force: bool,
         readout: bool,
-    ) -> tuple[Optional[FeedEvaluation], Optional[bool], Optional[bool]]:
+    ) -> tuple[Optional[FeedEvaluation], bool, bool]:
         """Evaluate a normalised entry against filters
 
         Returns a tuple (evaluation, should_download, star) or None if the entry should be skipped.
@@ -752,7 +752,7 @@ class RSSReader:
         job_status = job.get("status", " ")[0] if job else "N"
 
         if job_status not in "NGB" and not (job_status == "X" and readout):
-            return None, None, None
+            return None, False, False
 
         # Match this title against all filters
         logging.debug("Trying title=%r, size=%d", feed_entry.title, feed_entry.size)
@@ -764,7 +764,7 @@ class RSSReader:
             episode=feed_entry.episode,
         )
 
-        is_starred = job and job.get("status", "").endswith("*")
+        is_starred = bool(job and job.get("status", "").endswith("*"))
         star = first or is_starred
         should_download = (download and not first and not is_starred) or force
 
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -415,10 +415,12 @@ class FeedConfig:
             last_rule = rule
             break
 
+        rule_has_category = bool(last_rule and last_rule.category)
+
         # Category resolution
         if not rule_matched and self.default_category:
             effective_category = self.default_category
-        elif rule_matched and last_rule and last_rule.category is not None:
+        elif rule_matched and rule_has_category:
             effective_category = last_rule.category
         elif entry_cat and not self.default_category:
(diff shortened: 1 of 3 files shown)


Safihre:
Definitely feel free to change the tests I added. They are AI (slop).

mnightingale:
I'll take another look tomorrow, I think I'll have to try porting some tests to 4.5.5 which is tricky.
4.5.5 I think had a bug where it did `-100 == str(-100)` and without the normalisation is hard to follow with `None, "", "0", 0` all meaning different things.

mnightingale:
I've changed the tests to pass a RSS feed and compare the job data rather than the internals directly, it allowed me to run the same data through 4.5.5. I removed duplicate tests and added a few to the parametrize table.

This almost matches now, the couple of differences are not my ones but they seem fine.

In job data, pp used to be str and is now int, and season/episode used to be int and are now str - although it was not consistent, 0 would be an int but nondefault would be str "1" etc. I can't see a reason all three could not just be ints, seems the most fitting.

https://github.com/sabnzbd/sabnzbd/blob/a8751c5e77cea4753d139bd785eef05ad99d22f0/sabnzbd/rss.py#L801

This behaviour is different; it weirdly used to do str(None) which normalisation refactoring was trying to get rid of. 
It'll eventually lead to NzbObject.set_priority("None") which would result in NORMAL_PRIORITY I think, now it's explicitly DEFAULT_PRIORITY which may be ok although different behaviour - I don't think passing 'None' was expected behaviour or explicitly handled.

So this could possible be `"prio": str(resolved_entry.priority) if resolved_entry.priority is not None else str(NORMAL_PRIORITY),`, the job data would be different to 4.5.5 but it should have the same outcome when added to queue, but as I said it may be fine leaving it at default and taking the category priority.

Safihre:
> In job data, pp used to be str and is now int, and season/episode used to be int and are now str - although it was not consistent, 0 would be an int but nondefault would be str "1" etc. I can't see a reason all three could not just be ints, seems the most fitting.

Fix this in this PR or separate one?

mnightingale:
I've fixed it here but I think I'll leave the different priority behaviour, because I don't think it was intentional.

```py
# 4.5.5 did this
nzo = NzbObject("test_basic")
nzo.set_priority("None")
assert nzo.priority == NORMAL_PRIORITY
```

Safihre:
Maybe from the API? Or pre queue script?

mnightingale:
The RSS entries are rendered via `config_rss.tmpl` so the change in types should not matter.
The add to queue request just sends the feed name and url which looks up the data in the RSS entries.
NzbObject is typed for pp: `Optional[int]` although it does an int_conv so will accept a str, priority is `Optional[Union[int, str]]` so accepts either, and season/episode do not get passed on.
Pre queue is part of NzbObject init, so it'll have already done it's own normalisation and they'll get passed as string args/envvars. 

Maybe after the sqlite changes we will add an API to allow search/pagination.

Safihre:
Didn't do a deep inspection in last changes, trusting you 😊

Issue #2735, opened 2023-11-24
Implement article-level retry


Safihre:
### Description

Since we now have the option to write sparse-files, we should implement really article-level retries.

mnightingale:
Just as a note #2628 resulting in disabling sparse files because of an unrar issue testing passwords for encrypted files, I think that part is fixable, but it would require more tarfile changes.

I think @thezoggy raised a good point about having a switch to disable sparse files; there could be some non-standard setups where a directory is a network mount or something and I'm not sure how it would cope with truncate/sparse calls.

Strictly speaking the sparse file part isn't required to fill in the gaps of missing articles since the more important change was writing articles at the correct offsets.

For SAB the main benefit of sparse files would be if there are lots of missing articles because keeping failed downloads wouldn't take up so much space, i.e. several 100MB files but we only had the last 1MB article, would mean disk allocation of ~1MB.

Rather than just being able to add more servers and retry failures I have a pipe dream of being able to take two .nzb files pointing to different uploads and SAB being able to merge them and fille the gaps with different articles because it can tell that some/all files are the same.
I'm not sure how effective it would be though if upload split sizes vary a lot.

mnightingale:
Have you given any thought to how we might achieve this?

Article currently has `on_disk` but it’s currently also set for failed/missing articles, so I think we need an extra `failed` property.

But how to handle that in `repair_job`, currently it deletes `SABnzbd_nzo_data` and reconstructs from the .gz eventually calling `check_existing_files` which finds files `Matched file %s to %s of %s` and doesn’t consider missing pieces.

Options I see are load the old `SABnzbd_nzo_data` but avoid using it directly, in `check_existing_files` lookup the matching nzf, finish_import and set on_disk/missing as appropriate, only call `remove_nzf` if none need to retry parts.

That seems achievable but more complicated than I’d like, and one purpose of repair is to be able to achieve a repair without relying on potentially broken state.

Other areas avoid reading the `SABnzbd_nzo_data` with simpler structures for state we might need such as `__renames__` and `__verified__` so we could do something like.

- if cfg.direct_write() and "any nzf missing articles"
- Store a dictionary of missing parts, could be article-id or for space efficiency with bitmap (101 = index 1 is missing)

For a bitmap without dependencies AI spat this out https://gist.github.com/mnightingale/a57598fe57eccac121218050adcea42d but I’d probably store it as dict[str, tuple[int, bytes]] or just a list of tuples for portability.

[...]

PR #3433, opened 2026-05-24
Repair with no par2 sets and missing data should have failed status

### LINKED ISSUE #3430: Download with missing articles has status completed
### SABnzbd version

5.0.3 / develop

### Operating system

macOS

### Using Docker image

None

### Description

I'm investigating article level retries, setting up a test failing job with "Abort jobs that cannot be completed" disabled and an NZB with no par2 and missing some articles.

Image

The job gets marked Completed and not Failed?
It means there is no retry button, granted it wouldn't do anything currently.

For now I'll test with "Abort jobs that cannot be completed" enabled, but should we change this behaviour?

mnightingale:
Fixes #3430

Post processing is "+Delete" so it should try to repair, test NZB has no par2, 10 files, 1 complete, 9 each missing one or more articles.

`fail_hopeless_jobs` is disabled

I think the message may need changing, but I'm not sure how to communicate that verification failed because there are no par2 files and some articles are not available.

Path to get here is:
- No par2 files so `verified` is empty
- No sfv for sfv check
- No RARs for RAR-check
- There are non-par2 files with missing data (f.bytes.left)

So we know data is missing and we couldn't repair, so this should be a failed job.
If there was par2 data this would already be handled correctly as a failed job.

If PP mode was just "Download" it will still be marked completed as expected.

### DIFF
--- a/sabnzbd/postproc.py
+++ b/sabnzbd/postproc.py
@@ -869,6 +869,14 @@ def parring(nzo: NzbObject) -> tuple[bool, bool]:
                 if rars:
                     par_error = not try_rar_check(nzo, rars)
 
+            # Any non-par2 files missing data
+            if not par_error and any(not f.is_par2 and f.bytes_left for f in nzo.finished_files):
+                emsg = T("Repair failed - Not on your server(s)") + " - https://sabnzbd.org/not-complete"
+                nzo.fail_msg = emsg
+                nzo.set_unpack_info("Repair", emsg)
+                nzo.status = Status.FAILED
+                par_error = True
+
             # Save that we already tried SFV/RAR-verification
             verified[""] = not par_error
 


mnightingale:
Briefly draft because "Retry" marks it as complete, but that's a separate issue related to #2735 because it recreates with no knowledge of which articles are missing.

Safihre:
I still don't think we need to change this. 
I think we had this in the past, but users were angry when there were just missing articles in for example the screenshot. Or when indexers had this thing where they put fake articles just to track their NZB's.

What does nzbget does in this case?

thezoggy:
my two cents on this, the verification should be assumed good if we dont run into an error/dont have a way to not make that assumption. even if verification is technically unknown because we have nothing..  now we could flag that sort of thing for a status or something to expose in case someone wanted to do something in a script to decide to 'fail' the job or something.. but we shouldnt just assume the job is bad because there isnt verification data.

par is king when it comes to verification due to obfuscation, if no par then we need some other way to validate and quick-check sfv is used if enabled. but if no par/sfv then just relying on rar to say its good. but par is pretty much always used if its there

normally those nzb with no parset are fallout of people getting nzb from lower tier nzb (where the indexer is not getting the nzb from the uploader but instead are generating their nzb from regex/logic to match said post automatically, which by design of the poster ends up getting yielding an partial upload in the nzb or no parset).

mnightingale:
Rare enough edge case for a job without par2 so I'm not going to push for it.

PR #3435, opened 2026-05-24
Implement article-level retry

### LINKED ISSUE #2735: Implement article-level retry

mnightingale:
Fixes #2735

# Why

- Retry missing articles after activating new servers, or maybe a server had a temporary problem
- May allow reduced usage of block servers, download only with unlimited and decide if it's worth enabling a block server just for the missing pieces.

# Implementation

During postproc (repair) If direct_write is enabled and there are files that completed with missing articles, a `__ondisk__` file is created which stores if each article is on disk - limitted to just files missing articles.

During `repair_job / check_existing_files` for each file on disk it checks the on_disk dictionary and if an entry exists restores the on_disk status for each article, allowing the missing ones to be retried.

Also improves handling of renamed files.

# Testing

I've added tests, but it can be tested locally with a specially crafted NZB for [NServ](https://nzbget.github.io/nserv-nntp-server) can do `testfile.dat?2=50000:50000!2` (missing on instance 1)


sabnzbd/nzb/article.py
```diff
@@ -121,6 +122,7 @@ def __init__(self, article, article_bytes, nzf):
         self.data_size: Optional[int] = None
         self.decoded_size: Optional[int] = None  # Size of the decoded article
         self.on_disk: bool = False
+        self.failed: bool = False  # Download failed, assembler should skip this article
```
mnightingale, line 125: Should `__setstate__` set this if None, it didn't look like we bother elsewhere.

```py
if self.failed is None:
    self.failed = False
```

sabnzbd/nzb/object.py
```diff
@@ -830,11 +832,6 @@ def check_existing_files(self, wdir: str):
         # Substitute renamed files
         if renames := load_data(RENAMES_FILE, self.admin_path, remove=True):
             for name, old_name in renames.items():
-                if name in existing_files or old_name in existing_files:
-                    if name in existing_files:
-                        existing_files.remove(name)
-                    if os.path.exists(os.path.join(wdir, old_name)):
-                        existing_files.append(old_name)
```
mnightingale: This was a bit of a nonsense; it didn’t really let the later code cope with renamed files.

`if name in existing_files` this makes it ignore the existing renamed file, so it’ll download again and is what could cause the xyz.1.rar files and doesn't work when we want to just retry the missing articles.

`os.path.exists(os.path.join(wdir, old_name))` the old_name should not exist, but if it does it would already be in existing_files anyway.

Resolved later by matching existing files by `nzf.filename` but also checking `self.renames.get(…` to lookup if the NZF filename has been renamed previously and is now a different file on disk.

mnightingale:
Storing list[bool] is only 8x less efficient so doesn't seem worth worrying about.
100 MB is around 1500 articles so only 1.5KB, there are other far larger items in the admin directory anyway and they only get created if files are missing articles.

PR #3437, opened 2026-05-25
Sanitise host_whitelist and login/request public IP addresses

### LINKED ISSUE #3426: sabnzbd - not sanitizing public ip of xff or host_whitelist
Right now in sab 5.0.3/dev we do not sanitized the login success/failures thus users end up leaking their public ip due to the xff.

in this example `112.193.118.100`
```
2026-05-19 18:35:18,271::INFO::[notifier:169] Sending notification: Warning - Unsuccessful login attempt from ::ffff:172.18.0.1 (X-Forwarded-For: 112.193.118.100, 112.193.118.100) [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0] (type=warning, job_cat=None)
2026-05-19 18:35:18,271::WARNING::[interface:689] Unsuccessful login attempt from ::ffff:172.18.0.1 (X-Forwarded-For: 112.193.118.100, 112.193.118.100) [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0]
2026-05-19 18:35:22,740::INFO::[interface:681] Successful login from ::ffff:172.18.0.1 (X-Forwarded-For: 112.193.118.100, 112.193.118.100) [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0]
```

and then from the ini users are leaking their reverse proxy with `host_whitelist` so we prob should sanitized that as well...

mnightingale:
Fixes #3426

- `host_whitelist` => \<REMOVED\>
- LOG_IP_RE removed IP and X-Forwarded-For but keeps the useragent
  - "Unsuccessful login attempt from ... ["
  - "Successful login from ... ["
  - "Request ... from ..." for example "Request GET /api from ... ["

With lines from issue report:

```
2026-05-19 18:35:18,271::INFO::[notifier:169] Sending notification: Warning - Unsuccessful login attempt from <REMOVED> [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0] (type=warning, job_cat=None)
2026-05-19 18:35:18,271::WARNING::[interface:689] Unsuccessful login attempt from <REMOVED> [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0]
2026-05-19 18:35:22,740::INFO::[interface:681] Successful login from <REMOVED> [Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0]
```

---

I also noticed `api_key` and `nzb_key` in the ini are picked up by `LOG_HASH_RE` but only partially masked, is this intentional? or should they be in `LOG_INI_HIDE_RE`

---

Could this one do `b"'\\1': '<REMOVED>'"` to keep the keys or is there a reason it doesn't?

https://github.com/sabnzbd/sabnzbd/blob/25d20f0f08ce2078402faf0e7b281db9ceb04643/sabnzbd/api.py#L724

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -728,7 +728,7 @@ def _api_showlog(name: str, kwargs: ApiParams) -> Generator[bytes, Any, None]:
 
     def sanitize_line(line: bytes) -> bytes:
         """Apply regex substitutions to a single line to remove sensitive data"""
-        line = LOG_JSON_RE.sub(b"'REMOVED': '<REMOVED>'", line)
+        line = LOG_JSON_RE.sub(b"'\\1': '<REMOVED>'", line)
         line = LOG_INI_HIDE_RE.sub(b"\\1 = <REMOVED>", line)
         line = LOG_NNTP_AUTH_RE.sub(b"\\1 <REMOVED>", line)
         line = LOG_HASH_RE.sub(b"<HASH>", line)
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -702,7 +702,6 @@ LOG_INI_HIDE_RE = re.compile(
 LOG_NNTP_AUTH_RE = re.compile(rb"(authinfo (?:user|pass)) [^\\'\'\r\n]+", re.I)
 LOG_HASH_RE = re.compile(rb"([a-zA-Z\d]{25})", re.I)
 LOG_REMOTE_LABEL_RE = re.compile(
-    rb"(?P<prefix>^.*?]\s*.*?)"
     rb"(?P<ip>(?:\d{1,3}\.){3}\d{1,3}|(?:[A-Fa-f0-9:]+:+)+[A-Fa-f0-9.]+)"
     rb"(?:\s+\(X-Forwarded-For:\s*(?P<xff>[^)]+)\))?"
     rb"\s+\[(?P<ua>[^]]+)]"
@@ -722,8 +721,8 @@ def remote_label_replacement(m: re.Match[bytes]) -> bytes:
                 xff.append(xff_ip.encode())
             else:
                 xff.append(b"<REMOVED>")
-        return b"%s%s (X-Forwarded-For: %s) [%s]" % (m.group("prefix"), ip, b", ".join(xff), m.group("ua"))
-    return b"%s%s [%s]" % (m.group("prefix"), ip, m.group("ua"))
+        return b"%s (X-Forwarded-For: %s) [%s]" % (ip, b", ".join(xff), m.group("ua"))
+    return b"%s [%s]" % (ip, m.group("ua"))
 
 
 def sanitize_line(line: bytes, cur_user_bytes: Optional[bytes] = None) -> bytes:
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -700,29 +700,12 @@ LOG_INI_HIDE_RE = re.compile(
 )
 LOG_NNTP_AUTH_RE = re.compile(rb"(authinfo (?:user|pass)) [^\\'\'\r\n]+", re.I)
 LOG_HASH_RE = re.compile(rb"([a-zA-Z\d]{25})", re.I)
-
-# https://gist.github.com/dfee/6ed3a4b05cfe7a6faf40a2102408d5d8
-# fmt: off
-IPV4SEG  = r'(?:25[0-5]|(?:2[0-4]|1{0,1}[0-9]){0,1}[0-9])'
-IPV4ADDR = r'(?:(?:' + IPV4SEG + r'\.){3,3}' + IPV4SEG + r')'
-IPV6SEG  = r'(?:(?:[0-9a-fA-F]){1,4})'
-IPV6GROUPS = (
-    r'(?:' + IPV6SEG + r':){7,7}' + IPV6SEG,                  # 1:2:3:4:5:6:7:8
-    r'(?:' + IPV6SEG + r':){1,7}:',                           # 1::                                 1:2:3:4:5:6:7::
-    r'(?:' + IPV6SEG + r':){1,6}:' + IPV6SEG,                 # 1::8               1:2:3:4:5:6::8   1:2:3:4:5:6::8
-    r'(?:' + IPV6SEG + r':){1,5}(?::' + IPV6SEG + r'){1,2}',  # 1::7:8             1:2:3:4:5::7:8   1:2:3:4:5::8
-    r'(?:' + IPV6SEG + r':){1,4}(?::' + IPV6SEG + r'){1,3}',  # 1::6:7:8           1:2:3:4::6:7:8   1:2:3:4::8
-    r'(?:' + IPV6SEG + r':){1,3}(?::' + IPV6SEG + r'){1,4}',  # 1::5:6:7:8         1:2:3::5:6:7:8   1:2:3::8
-    r'(?:' + IPV6SEG + r':){1,2}(?::' + IPV6SEG + r'){1,5}',  # 1::4:5:6:7:8       1:2::4:5:6:7:8   1:2::8
-    IPV6SEG + r':(?:(?::' + IPV6SEG + r'){1,6})',             # 1::3:4:5:6:7:8     1::3:4:5:6:7:8   1::8
-    r':(?:(?::' + IPV6SEG + r'){1,7}|:)',                     # ::2:3:4:5:6:7:8    ::2:3:4:5:6:7:8  ::8       ::
-    r'fe80:(?::' + IPV6SEG + r'){0,4}%[0-9a-zA-Z]{1,}',       # fe80::7:8%eth0     fe80::7:8%1  (link-local IPv6 addresses with zone index)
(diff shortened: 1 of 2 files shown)


thezoggy:
apikey+nzbkey  - yeah i think that was just a value add that it somewhat masked that, we should indeed prob just hide them from the ini.

Safihre:
It's nice to have, since they can also appear in the request logging so they also need to be captured by the hash one.

Indeed should be \\1.

Safihre:
Hmm, but those are translatable texts. What if the user is in non English it won't work..

mnightingale:
I think I can make it use the translatable text, `Successful login from %s` doesn't because it's info level, Unsuccessful is a warning to shown in UI.

Safihre:
Why not look for something IP-like?
Must be some regexes online for that.

mnightingale:
I came up with a too complicated way for translatable text with %s, but it doesn't handle all uses of `log_warning_and_ip`

```py
def request_ip_func(search: str) -> Callable[..., bytes]:
    """Apply substitutions to a line including IP addresses for the request"""
    replacement = rb"[\w:.]+(?: \(X-Forwarded-For: .*?\))?"
    pattern = (search % "%s").encode() % replacement
    repl = (search % "<REMOVED>").encode()
    rx = re.compile(pattern, re.I)
    return lambda l: rx.sub(repl, l)

request_successful_login = request_ip_func("Successful login from %s")
...
line = request_successful_login(line)
```

I did consider just searching for IP, the IPv6 ones are very long and over match the start of log lines around time/level/location.
Would we want to just remove all IPs for logs?

An easier way would be to just replace `request.remote_label` with something easier to replace.

mnightingale:
A crazy IPv6 regex to remove all IPv4/IPv6 from the logs is the only way I can see to do it, but I don't know if that's desirable.

For just "Unsuccessful login" and "Unsuccessful login" it could check it came from interface or notifier.

Safihre:
Seems a bit much? Let's keep it simple and try to catch 90%, best effort..

thezoggy:
yeah to do proper v4/v6 you do something like:
```
ipv4 = '''^(25[0-5]|2[0-4][0-9]|[0-1]?[0-9][0-9]?)\.( 
            25[0-5]|2[0-4][0-9]|[0-1]?[0-9][0-9]?)\.( 
            25[0-5]|2[0-4][0-9]|[0-1]?[0-9][0-9]?)\.( 
            25[0-5]|2[0-4][0-9]|[0-1]?[0-9][0-9]?)$'''

ipv6 = '''(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|
        ([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:)
        {1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1
        ,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}
        :){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{
        1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA
        -F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a
        -fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0
        -9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,
        4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}
        :){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9
        ])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0
        -9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]
        |1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]
        |1{0,1}[0-9]){0,1}[0-9]))'''
```
but since were not really trying to sanitized the data as input..  we can do something much simpler like:

```
IP_4_RE = r"""(?:\d{1,3}\.){3}\d{1,3}/\d{1,2}"""
IP_6_RE = r"""(?:[0-9a-fA-F]{1,4}::?|::){1,7}(?:[0-9a-fA-F]{1,4}|(?<=:):/\d{1,3})"""
``` 

but the argument we should only really sanitized public ip and leave private ip space alone since one would argue as a noob its harder for them to get the logs outside of the gui.. and them being able to do basic triage of when something hits their stuff and see the internal ip is useful

looking at solution for that.
https://regex101.com/library?orderBy=RELEVANCE&search=Matches+publicly+available+IPv4+addresses

> Matches publicly available IPv4 addresses (excludes 10., 172.16-32, and 192.168 blocks).
```
(\d+)(?<!10)(?<!127)\.(\d+)(?<!192\.168)(?<!172\.(1[6-9]|2[0-9]|3[0-1]))\.(\d+)\.(\d+)
```

v6, dunno if we need to cover zero expanded or not but anyways you get the idea

mnightingale:
This is a simpler approach, the main IP matching will need improving to be less likely of false positives. But I need to figure out what will be good enough matching.

Safihre:
That looks good. Want to tinker more with it or you also like it this way? Since PR is in draft.
Probably good to add a test for this whole function, but that can also be a new PR.

mnightingale:
I need to make the ip part stricter for `IP [USERAGENT]` based on what @thezoggy outlined above; currently `(?P<ip>\S+)` is very lax but it doesn’t need to be 100% validated just got enough not to flag false positives. Ideally just public IP addresses but I wont if it gets too complicated.

Currently it’s just showing the approach but would match `foo [anything in square brackets]` so I just need to make it look for something that looks like an IP address.

I'm not sure I'll do it in this PR but I have though about adding tests I've been testing with the key methods copied into a test.

mnightingale:
I think I'm done now, the main ip part should match "ip looking" content and it allow IPs from private address space to pass through unchanged.
It's moved around a bit to add tests, which could be expanded to cover the other replacements.

thezoggy:
i saw online one persons solution was just extract ip from string, toss it to function to just leverage python lib to see if its viewed as local or not, if true return back value otherwise return back the sanitized data. dunno if overall this ends up being faster ?

mnightingale:
That’s pretty much what I’ve done.
A simplified pattern to match IPv4/6 like strings without enough to validate.
The pattern extracts the relevant parts (ip, xff, and useragent).
A replacement function that for each IP in ip and xff determines if they are loopback/private addresses; if so keeps them, otherwise replace with `<REMOVED>`.
So only public IP addresses should be removed.

jcfp:
@mnightingale you might want to use `is_lan_addr` rather than `is_local_addr`, because the latter takes the user-configurable local_ranges setting into account. Some people insert trusted remote ip addresses into that to give them favourable treatment normally reserved for local area networks.

thezoggy:
> @mnightingale you might want to use `is_lan_addr` rather than `is_local_addr`, because the latter takes the user-configurable local_ranges setting into account. Some people insert trusted remote ip addresses into that to give them favourable treatment normally reserved for local area networks.

I've seen way to many people do dumb things with local_ranges where they end up making everything (even actual local stuff) be viewer as external and end up making it where they can't access sab. Being able to see what they put for that field makes it very easy to triage when they post logs.

mnightingale:
> @mnightingale you might want to use is_lan_addr rather than is_local_addr

Thanks

> I've seen way to many people do dumb things with local_ranges where they end up making everything (even actual local stuff) be viewer as external and end up making it where they can't access sab. Being able to see what they put for that field makes it very easy to triage when they post logs.

The only addition to INI and JSON filtering is `host_whitelist` however, it hides the value regardless of if it is empty or not so we wont be able to tell if it's configured or not.
We could tweak the patterns to ignore empty values.

`local_ranges` is left as is so may show public ranges, not sure what to do here. It could sanitise using the replacement method I've used for requests. Then if configured it'll show `<REMOVED>` for any configured public IPs - which should be all of them, but at least then you could still tell if it's configured while hiding public IPs.

thezoggy:
> > @mnightingale you might want to use is_lan_addr rather than is_local_addr
> 
> Thanks
> 
> > I've seen way to many people do dumb things with local_ranges where they end up making everything (even actual local stuff) be viewer as external and end up making it where they can't access sab. Being able to see what they put for that field makes it very easy to triage when they post logs.
> 
> The only addition to INI and JSON filtering is `host_whitelist` however, it hides the value regardless of if it is empty or not so we wont be able to tell if it's configured or not. We could tweak the patterns to ignore empty values.
> 
> `local_ranges` is left as is so may show public ranges, not sure what to do here. It could sanitise using the replacement method I've used for requests. Then if configured it'll show `<REMOVED>` for any configured public IPs - which should be all of them, but at least then you could still tell if it's configured while hiding public IPs.

filtering host_whitelist is good, what i asked for. local_ranges ideally should not be touched just cause that needs to be shared... and hopefully people arent putting in public ip space in here (but ideally i would want to see it so i could tell them dont do it). which sadly people do, or dont realize when they set this field its not additive and end up doing stuff where it makes their rfc1914 space now is 'external' because of it and then struggle on trying to get access to sab again.

PR #3438, opened 2026-05-25
Remove disableDelays because config is not defined

mnightingale:
I should have added a variable to the main template set by `bool(os.environ.get("CI", False))` [i did try a branch like this](https://github.com/sabnzbd/sabnzbd/compare/develop...mnightingale:sabnzbd:bugfix/fixdisableconfigdelays?expand=1)

But I think the impact is pretty low <5 seconds (at least locally) so we should just remove it.


Safihre:
MacOS failed on this branch and on develop, worth a look maybe?

PR #3456, opened 2026-06-09
Support extracting tar files

### LINKED ISSUE #3016: Support extracting .tar

mnightingale:
Fixes #3016 for Python >= 3.12

I'll need to send a wiki entry sometime.

I'm not sure about the UI showing the enable_tar option for older Python versions.
It could filter it out of `SPECIAL_BOOL_LIST` but I don't think it should, better to show the user it is supported and the reason it's not working is an old Python version.
Maybe a debug or warning if tar files are found but not supported?


Safihre:
I would suggest a message in the Unpack state that it was skipped due to Python version.

Safihre:
No not like that, just a Unpack info step that we couldn't do it for that set.
It's certainly not a failure.

sabnzbd/newsunpack.py
```diff
@@ -1012,6 +1026,103 @@ def seven_extract_core(
...
+    """
+
+    untar_failed = False
+    new_files = []
+
+    # Unpack each tar
+    for tar_path in tars:
+        setname = setname_from_path(tar_path)
+        if sys.version_info < (3, 12):
+            msg = T("Unpacking failed, TAR support requires Python 3.12 or later")
```
Safihre:
```suggestion
            msg = T("Unpacking skipped, TAR support requires Python 3.12 or later")
```

sabnzbd/newsunpack.py
```diff
@@ -1012,6 +1026,103 @@ def seven_extract_core(
...
+
+    return ret, new_files
+
+
+def tar_non_executable_data_filter(member, path) -> Optional[tarfile.TarInfo]:
+    """Applies tarfile.data_filter and removes executable permissions from files"""
+    member = tarfile.data_filter(member, path)
+
+    if member is not None and member.isreg():
+        member = member.replace(mode=member.mode & ~(stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH))
```
Safihre: Should we not use `UNWANTED_FILE_PERMISSIONS`?

tests/test_functional_downloads.py
```diff
@@ -34,6 +35,10 @@ def test_download_zip(self):
     def test_download_7zip(self):
         self.download_nzb("test_7zip", ["My_Test_Download.bin"])
 
+    @pytest.mark.skipif(sys.version_info < (3, 12), reason="tarfile extraction filter requires Python 3.12+")
+    def test_download_tar(self):
```
Safihre, line 39: Due to security issues, we should really add a test also that includes some executable file and show it's clean.

sabnzbd/newsunpack.py
mnightingale: Yes can do, tar_filter already applies the extra ones but it's cleaner to use it

tests/test_functional_downloads.py
mnightingale: I've added more tests based on TestRarUnpack, testfile.tar includes executable helloworld.sh

sabnzbd/newsunpack.py
```diff
@@ -1012,6 +1026,103 @@ def seven_extract_core(
...
+        for member in tar.getmembers():
+            if member.isfile():
+                if one_folder:
+                    member.name = os.path.basename(member.name)
+                try:
+                    if not cfg.overwrite_files():
+                        # Ensure a unique filepath by appending .N before extension if needed
+                        info = tar_non_executable_data_filter(member, extraction_path)
+                        if info is not None and os.path.exists(os.path.join(extraction_path, info.name)):
+                            base_name, ext = os.path.splitext(info.name)
```
Safihre: Why not use `get_unique_filename`?

tests/test_newsunpack.py
```diff
@@ -632,3 +633,134 @@ def test_rar_unpack_invalid_windows_filenames(self):
...
+            extracted_files_set
+        ), f"{complete_contents_set} should be in {extracted_files_set}"
+
+        # Verify the expected files are present using full paths
+        expected_full_paths = {os.path.join(temp_complete_dir, filename) for filename in expected_files}
+        assert expected_full_paths.issubset(
+            extracted_files_set
+        ), f"{expected_full_paths} should be in {extracted_files_set}"
+
+    def test_basic_tar_unpack(self):
```
Safihre, line 753: I would suggest to also add verification of permissions (on non-Windows).

mnightingale: Expanded tests to path traversal, executable, permissions, owner, and group.

sabnzbd/newsunpack.py
mnightingale: I forgot about that, I don't know how you remember so much of it 😄

Safihre: 12 years of manual coding this project 😂

And doing the Python 2 to 3 conversion required touching every bit of code.

Safihre:
Looks good to me.
It's still marked as Draft, do you want to add more?

sabnzbd/newsunpack.py
```diff
@@ -1012,6 +1027,90 @@ def seven_extract_core(
...
+        with tarfile.open(tar_path) as tar:
+            members = []
+            for member in tar.getmembers():
+                member = tar_non_executable_data_filter(member, extraction_path)
+                if not member or not member.isfile():
+                    continue
+                if one_folder:
+                    member = member.replace(name=os.path.basename(member.name))
+                members.append(member)
+                new_files.append(os.path.join(extraction_path, member.name))
```
mnightingale: Draft because I spotted a slight bug here, collapsing into a one folder might have duplicates.
The later `tar_non_executable_data_filter` call as part of `extractall` will fix them but the wrong path will be added to `new_files`, I think I just need to move it above the manual `tar_non_executable_data_filter` call but I'll add a test for it also.

Safihre:
Nice.

PR #3461, opened 2026-06-15
Replace getopt with argparse in SABnzbd CLI

Kcchouette:
## Summary

Replace the legacy `getopt`-based CLI argument parsing with `argparse.ArgumentParser`, while preserving full backward compatibility with existing options and usage patterns.

## Motivation

The current CLI parsing uses `getopt` with manual `print_help()` / `print_version()` functions, a hand-built `(opt, arg)` tuple list (`sab_opts`), and a long `if/elif` chain in `main()` to consume those tuples.

## Changes

- **`SABnzbd.py`**:
  - Replace `import getopt` with `import argparse`
  - Add `build_parser()`, single source of truth for all CLI options
  - Add `_logging_type()`, custom type that validates `--logging` range (-1..2) and handles negative numbers
  - `commandline_handler()` returns the argparse `Namespace` directly (no more tuple intermediary)
  - `main()` accesses options via `opts.attribute` instead of iterating `(opt, arg)` tuples
  - Remove dead code: `print_help()`, `print_version()`, `get_f_option()`
  - Backward-compat aliases via argparse: `--config`/`--config-file`, `--ipv6`/`--ipv6_hosting`, `--inet`/`--inet_exposure`, `--disable`/`--disable-file-log`
  - `-2` short option restored for `--wait` (Win32 service)
  - Win32 service options hidden from `--help` via `argparse.SUPPRESS`
  - Pre-processing of `sys.argv` to convert `-l -1` → `--logging=-1` (argparse limitation with negative numbers)

- **`tests/test_cli_argparse.py`** (new, 51 tests):
  - `TestBuildParser` — 30 tests: flags, options, aliases, validation, platform-specific behavior, help/version
  - `TestCommandlineHandler` — 21 tests: full integration, NZB collection, backward compat, complex flag combinations

## Backward Compatibility

All existing CLI options work unchanged:

| Before (getopt) | After (argparse) | Status |
|------------------|-------------------|--------|
| `-f`, `--config-file` | `-f`, `--config`, `--config-file` | ✅ Enhanced |
| `-l -1` | `-l -1`, `--logging=-1` | ✅ Works |
| `--ipv6_hosting 1` | `--ipv6 1`, `--ipv6_hosting 1` | ✅ Enhanced |
| `--inet_exposure 3` | `--inet 3`, `--inet_exposure 3` | ✅ Enhanced |
| `--disable-file-log` | `--disable`, `--disable-file-log` | ✅ Enhanced |
| `--help`, `--version` | `--help`, `--version` | ✅ Auto-generated |

### DIFF
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -29,7 +29,7 @@ import logging
 import logging.handlers
 import importlib.util
 import traceback
-import getopt
+import argparse
 import signal
 import socket
 import subprocess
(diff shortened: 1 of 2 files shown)


Safihre:
@Kcchouette Will you try to fix the failures?

mnightingale:
The failures appear to be because of setting `sabnzbd.WINDOWS` etc in the tests breaks later ones, I suggest using the platform fixture ando/or the monkleypatch approach it uses if you need others.

See testhelper `@pytest.mark.platform("win32")` and maybe `monkeypatch.setattr(sabnzbd, "__version__", "4.0.0")`

Kcchouette:
> See testhelper @pytest.mark.platform("win32") and maybe monkeypatch.setattr(sabnzbd, "__version__", "4.0.0")

~do I need to call it in each 51 tests?~ :ok:

mnightingale:
Instead of faking the WINDOWS=true which I don't think will actually work since it'll want to import Windows only packages.

You can just not change the platform and skip them based on the host platform, `@pytest.mark.skipif(not sys.platform.startswith("win"), reason="Windows tests")` the CI runs on all the platforms separately anyway.

`_build` could be a fixture and used to pass the parser to each test

```
@pytest.fixture()
def parser(self, monkeypatch):
    import sabnzbd

    monkeypatch.setattr(sabnzbd, "MY_NAME", "SABnzbd")
    monkeypatch.setattr(sabnzbd, "__version__", "4.0.0")
    from SABnzbd import build_parser

    return build_parser()

def test_parser_returns_argument_parser(self, parser):
    import argparse

    assert isinstance(parser, argparse.ArgumentParser)
```

Kcchouette:
do you want I use your way of doing @mnightingale or is https://github.com/sabnzbd/sabnzbd/pull/3461/commits/5576781765657caccf7d5d01c05289ef3dc27abd ok?

mnightingale:
That's not too bad now, I'd clean up the imports within tests and skip platform tests because unless you're on the tested platform they aren't really representative anyway.

https://gist.github.com/mnightingale/05d81449ac7401e926d32eb025eb6567

I haven't really looked through the changes to check the behaviour is the same but a couple of things I did notice:

- `python SABnzbd.py --version` on macOS (doesn't on Windows, not tested Linux) remains running instead of immediately exiting, this possible affects other options, not looked closely but probably missing a few `exit_sab` calls - [trace](https://gist.github.com/mnightingale/5a0e86586fbe225543044b52689556a6) - edit: unclear to me why cherrypy has even started making the exit calls necessary
- `python SABnzbd.py --version` output doesn't have copywrite/license part, probably needs to be somewhere if not version output?

Kcchouette:
I have taken remarks about copyright + the exit

Safihre:
We had a discussion among the team and device to reject this PR. While we appreciate the effort and tokens spent, it's not something that needed to be fixed. It improves things in some way, but barely on other ways. 
We would like to suggest to always check the open issues, or create a new issue, before opening PR's.

PR #3462, opened 2026-06-19
Track files during cleanup to prevent removing unrelated files

### LINKED ISSUE #2577: Files unrelated to job get removed when post-processing Cleanup List runs

Safihre:
Fixes #2577

@mnightingale wanted your quality check :)


sabnzbd/sorting.py
```diff
@@ -610,29 +619,47 @@ def ends_in_file(path: str) -> bool:
     return bool(RE_ENDEXT.search(path) or RE_ENDFN.search(path))
 
 
-def move_to_parent_directory(workdir: str) -> tuple[str, bool]:
-    """Move all files under 'workdir' into 'workdir/..'"""
+def update_tracked_file(tracked_files: list[str], old_path: str, new_path: str):
+    """Record the rename of a job file in the tracked files list"""
+    if new_path and old_path in tracked_files:
+        tracked_files[tracked_files.index(old_path)] = new_path
```
mnightingale, line 625:
```suggestion
    if new_path:
        try:
            tracked_files[tracked_files.index(old_path)] = new_path
        except ValueError:
            pass
```

To avoid searching for the item twice, alternatively if the order doesn't matter use a set and discard/add.

sabnzbd/sorting.py
```diff
@@ -610,29 +619,47 @@ def ends_in_file(path: str) -> bool:
...
-
-    cleanup_empty_directories(workdir)
-    return dest, True
+            return workdir, True, files
+
+    logging.debug("Moving all job files from %s to %s", workdir, dest)
+
+    moved_files = []
+    for nr, path in enumerate(files):
+        path = os.path.abspath(os.path.normpath(path))
```
mnightingale: More a question than me thinking anything is wrong, why is abspath not enough I thought it already normalises like normpath?

There are a handful of other places which do it, I'm just interested if there is a case when it is not sufficient?

thezoggy: it looks like that abspath wrapping normpath was something done forever ago, just a lot more surfaced when saf did the py2->py3 work

going back 19 years ago is where i really first see this
https://github.com/sabnzbd/sabnzbd/commit/453f2c91e9a867d518903028e52f2c885d877eaa#diff-f191357a735a0cf878f1c15693b964a8a255a1a00efae431916ab3258f0c60b5R150

from quick glance abspath is just normpath but handles symlinks better but normpath is more for multiple os support..  from what I assume is that it was just used as a way to take in `.` (and prob other user input paths) to do fixup of forward/backward slashes and stuff before having abspath generate what path to actually use.. 

snippet of first thing that used it:
image

surely there is a more efficient way to do this by now..

Safihre:
Claude seems to agree after burning a few thousand tokens 😁

---

Following the review question about why we call os.path.abspath(os.path.normpath(path)): the inner normpath is redundant — abspath already normalizes internally on both platforms.

From the CPython source:
```python
# ntpath.py (Windows)
def abspath(path):
    try:
        return _getfullpathname(normpath(path))   # normpath first, then GetFullPathNameW
    except (OSError, ValueError):
        ...
        return normpath(path)                       # fallback also ends in normpath

# posixpath.py (POSIX)
def abspath(path):
    if not path.startswith('/'):
        path = join(os.getcwd(), path)
    return normpath(path)                           # ends in normpath
```
Since normpath is idempotent, `abspath(normpath(x)) == abspath(x)` in all cases — including Windows `\\?\` extended-length paths with .. components (verified empirically). So the extra call was pure belt-and-suspenders.

sabnzbd/sorting.py
Safihre: I think the clarity is better with the double lookup, I think we can afford it :)

Issue #3106, opened 2025-07-02
Auto Resume is on, but SABnzbd is not resuming from disk full


sanderjo:
### SABnzbd version

4.5.1

### Operating system

Linux

### Using Docker image

None

### Description

See https://forums.sabnzbd.org/viewtopic.php?p=133615&sid=6ee29ede16eaf02d3066867ef3173927#p133615 ... which I can reproduce:

Auto Resume is on
Put a few 10GB testfiles into SABnzbd, it started downloading ... until disk full, and thus Paused. That is half-way a 10GB download.
I freed up space on my disk, and SABnzb does see it: " 10.3 GB left 50.1 GB Free Space". Good.
BUT: even after 20 minutes, SAB is still paused, and does not unpause.

And even manually Unpausing SABnzbd does not restart the downloading.

SABnzbd stdout says:

```
2025-06-19 14:57:00,041::DEBUG::[downloader:841] Delayed - 1712 seconds - Assembler queue: 12
2025-06-19 14:57:01,043::DEBUG::[downloader:841] Delayed - 1713 seconds - Assembler queue: 12
2025-06-19 14:57:02,044::DEBUG::[downloader:841] Delayed - 1714 seconds - Assembler queue: 12
2025-06-19 14:57:03,046::DEBUG::[downloader:841] Delayed - 1715 seconds - Assembler queue: 12
2025-06-19 14:57:04,048::DEBUG::[downloader:841] Delayed - 1716 seconds - Assembler queue: 12
2025-06-19 14:57:05,049::DEBUG::[downloader:841] Delayed - 1717 seconds - Assembler queue: 12
2025-06-19 14:57:06,051::DEBUG::[downloader:841] Delayed - 1718 seconds - Assembler queue: 12
2025-06-19 14:57:07,053::DEBUG::[downloader:841] Delayed - 1719 seconds - Assembler queue: 12
2025-06-19 14:57:08,054::DEBUG::[downloader:841] Delayed - 1720 seconds - Assembler queue: 12
2025-06-19 14:57:09,055::DEBUG::[downloader:841] Delayed - 1721 seconds - Assembler queue: 12
2025-06-19 14:57:10,057::DEBUG::[downloader:841] Delayed - 1722 seconds - Assembler queue: 12
```

... is SABnzbd doing something too internally?

Safihre:
Not sure what you mean? It should always work right?

Safihre:
The self.paused is inside the no active jobs.
[...]

PR #3463, opened 2026-06-19
Abort postproc unrar on prompt

mnightingale:
Related to #3106 but I haven’t setup an environment to test it aborts when out of space yet.

Direct Unpack passes `text_mode=False` and uses `platform_btou` which has a note about "We cannot use encoding-parameter because it's Python 3.7+" - can we simplify that?
Main point is we have two similar code paths that deal with slightly different encoding since postproc reads strings.
I'm not sure if would be useful to make them both the same?

`StringIO or BytesIO` could also be used on direct unpacker to also avoid creating new bytes after each character read.

Git diff makes the changes look more complicated than it is:

```py
# Read character at a time until " ", "\n", ""
if char == "\n":
  # The previous whole line handling, with skipping on empty lines
elif line.endswith("[C]ontinue, [Q]uit "):
  # ...
elif line.endswith("[R]etry, [A]bort "):
  # ...
```

I've not checked if the volume prompt can occur with whatever options are being passed, but I assume so.

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -880,11 +880,6 @@ def rar_extract_core(
                 if cfg.flat_unpack():
                     unpacked_file = os.path.basename(unpacked_file)
                 extracted.append(real_path(extraction_path, unpacked_file))
-        elif line.endswith("[C]ontinue, [Q]uit "):
-            fail = 1
-            with suppress(IOError):
-                p.stdin.write("Q\n")
-                time.sleep(0.2)
         elif line.endswith("[R]etry, [A]bort "):
             fail = 1
             msg = "%s %s" % (T("Unpacking failed, write error or disk is full?"), line)
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -887,7 +887,6 @@ def rar_extract_core(
             nzo.set_unpack_info("Unpack", msg, setname)
             with suppress(IOError):
                 p.stdin.write("A\n")
-                time.sleep(0.2)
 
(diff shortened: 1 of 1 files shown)


Safihre:
Still draft?

mnightingale:
I was going to see if I could avoid moving so many lines, but it makes the logic ugly.

~~The last part I need to change is the `[C]ontinue, [Q]uit ` path, that was never handled previously which made me think maybe it can't happen we must check all volumes exist before postproc unrar?~~

~~But a user could still delete a rar while it's downloading, so maybe it should still be handled.~~

mnightingale:
```
sabnzbd/macos/unrar/arm64/unrar e rar-files.part1.rar 

UNRAR 7.22 freeware      Copyright (c) 1993-2026 Alexander Roshal

Extracting from rar-files.part1.rar

Extracting  100MB-2025-11-22.bin                                      26%

Extracting from rar-files.part2.rar

...         100MB-2025-11-22.bin                                      47%
No space left on device
Write error in the file 100MB-2025-11-22.bin [R]etry, [A]bort a
Write error in the file 100MB-2025-11-22.bin
No space left on device
Program aborted
```

On macOS "No space left on device" is a OS/FS level error although it's a whole line there will many others to handle and it'll vary by platform.

The existing `Write error` check is not hit because the retry/abort prompt is before a whole line.

---

I suppose we could add a case for `No space left on device` which it appears 7z does, or better yet use `os.strerror(errno.ENOSPC)` but Unix only and there are probably many others. I think it's safer to just detect the abort prompt, and quit gracefully.

PR #3466, opened 2026-06-23
Improve rarfile performance with RAR3 archives using header encryption and long passwords

mnightingale:
@thezoggy reported an issue on Discord that some jobs cause the cache fill and cause extremely slow download speeds.

After analysing the logs, I found that the time to check a rar password was extremely long at over 10 seconds. The issue was only with RAR4 archives with long passwords and utilising header encryption.

If config options `unwanted_extensions` and `action_on_unwanted_extensions` are enabled, it causes the assembler to call `check_encrypted_and_unwanted_files` after each rar volume downloads. If that’s 30 volumes it could easily be 5 minutes total were the assembler thread is blocked decrypting the file list. If you have a fast connection or small cache that could result in a jagged download speed that looks like a disk IO limitation, indeed SABnzbd will be reporting`delayed_assembler` but it may actually be delayed by rar SHA1 calculations.

The actual impact could be greater because we don’t log subsequent header decryption times, and at high CPU load it could take longer. The blocked thread causes the cache to fill which slows downloader, the cache can’t be flushed because the assembler is busy.

After digging into what `rarfile` is doing the issue is with `Rar3Sha1` is doing in `update `, `_corrupt`, and the number of calls they receive from `rar3_s2k`. The result is millions of Python operations and calls performing bitwise operations on the password and salt, Python is just not good at these operations.

I have resolved the issue (with some Claude help) by moving the heavy computation part of `rar3_s2k` to a native extension, let it do the calculations and only make 16 calls from Python.

I’ll try and open an issue on the rarfile repository sometime, but I think it’s unlikely to go anywhere. For now, a patch to rarfile allows sabnzbd to use our implementation that is easily 100-200 times faster.

```py
rarfile.rar3_s2k = sabctools.rarfile_rar3_s2k
```

Tests will fail until sabctools release and version requirements updated.

---

I may investigate further changes to `zf.setpassword`, I think in the cases that header encryption is and it's the first password tested `setpassword` already calls parse, then we manually parse again. But it probably doesn't matter if the call takes 0.05 seconds now.


mnightingale:
Testing on a local server limited to 600Mb/s, 32 connections, 1 pipelining

## 4.5.5

Notice it speeds up once it has done the first password check

https://github.com/user-attachments/assets/3aad6475-4b16-49d7-9f16-229b853fd449

## 4.5.5 with unwanted_extensions checks

https://github.com/user-attachments/assets/e8bbb20b-db9c-4ad8-aeb8-aa7b8f5a90fe

## PR

https://github.com/user-attachments/assets/143e8dc4-af11-4ab8-940c-abf367b30552

## PR with unwanted_extensions checks 

A little slower, which is reason I think I'll look at the overall double parse or moving it out of the assembler path.

https://github.com/user-attachments/assets/21c7c767-0b38-41dd-9a66-fbe12588349e

---

# PR with unwanted_extensions checks and release GIL

https://github.com/user-attachments/assets/e751ae85-69ca-4b68-ab99-b31a23e2b1a2

PR #3471, opened 2026-06-27
Improve Glitter header menu labels

coopa11y:
This cleans up some of the Glitter header controls for screen reader use.

I noticed several icon-only buttons were announced as just "menu" or "button", and the speed limit menu was especially confusing because the visible caret was reachable as a separate control. This adds labels to the header actions and hides the duplicate caret from keyboard/screen reader navigation.

Tested locally with NVDA.

### DIFF
--- a/interfaces/Glitter/templates/include_menu.tmpl
+++ b/interfaces/Glitter/templates/include_menu.tmpl
@@ -1,6 +1,6 @@
 <nav class="navbar navbar-inverse main-navbar">
     <div class="navbar-header">
-        <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navbar-collapse" aria-label="Toggle SABnzbd navigation menu" aria-controls="navbar-collapse" aria-expanded="false">
+        <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navbar-collapse" aria-label="$T('Glitter-toggleNavigation')" aria-controls="navbar-collapse" aria-expanded="false">
             <span class="icon-bar"></span>
             <span class="icon-bar"></span>
             <span class="icon-bar"></span>
@@ -13,7 +13,7 @@
                 <button type="button" class="btn btn-default navbar-btn" data-bind="click: pauseToggle, attr: { 'title': downloadsPaused() ? '$T('link-resume')' : '$T('link-pause')', 'aria-label': downloadsPaused() ? '$T('link-resume')' : '$T('link-pause')' }">
                     <span class="glyphicon" data-bind="css: { 'glyphicon-play' : downloadsPaused(), 'glyphicon-pause' : !downloadsPaused() }"></span>
                 </button>
-                <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown" aria-label="$T('pauseFor')" aria-haspopup="true">
+                <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown" aria-label="$T('pauseFor')" aria-haspopup="true" aria-expanded="false">
                     <span class="caret"></span>
                 </button>
                 <ul class="dropdown-menu">
@@ -26,7 +26,7 @@
                     <li class="divider"></li>
                     <li><a href="#" data-bind="click: openCustomPauseTime">$T('Glitter-custom')</a></li>
                 </ul>
-                <span class="btn btn-default navbar-btn navbar-timeleft" role="status" aria-live="polite" data-bind="text: timeLeft, attr: { 'aria-label': timeLeftLabel }">0:00</span>
+                <span class="btn btn-default navbar-btn navbar-timeleft" data-bind="text: timeLeft" data-toggle="dropdown">0:00</span>
             </div>
         </div>
     </div>
@@ -37,13 +37,13 @@
             <li class="speedlimit-dropdown dropdown">
                 <div class="btn-group">
                     <div class="btn-group">
-                        <button type="button" class="btn btn-default navbar-btn navbar-speed" data-toggle="dropdown" aria-haspopup="true" data-bind="attr: { 'aria-label': speedMenuLabel }">
+                        <button type="button" class="btn btn-default navbar-btn navbar-speed" data-toggle="dropdown" aria-haspopup="true">
                             <span data-bind="text: speedText">0 KB/s</span>
                         </button>
                         <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown" onclick="keepOpen(this)" aria-hidden="true" tabindex="-1">
                             <span class="caret"></span>
                         </button>
-                        <a href="#" class="max-speed-input-clear hover-button" aria-label="Clear speed limit" data-bind="click: clearSpeedLimit, visible:(speedLimit() < 100 && speedLimit() > 0)" style="display: none;">
+                        <a href="#" class="max-speed-input-clear hover-button" aria-label="$T('button-clear')" data-bind="click: clearSpeedLimit, visible:(speedLimit() < 100 && speedLimit() > 0)" style="display: none;">
                             <span class="glyphicon glyphicon-link"></span>
                         </a>
                         <div class="dropdown-menu max-speed-input">
@@ -82,7 +82,7 @@
                 <a href="./config/" aria-label="SABnzbd $T('menu-config')"><span class="glyphicon glyphicon-cog"></span></a>
             </li>
             <li class="dropdown main-menu-link" data-bind="css: { 'active-on-queue-finish-menu': finishaction()}">
-                <a href="#" data-toggle="dropdown"  onclick="keepOpen(this)" aria-label="Open SABnzbd main menu" aria-haspopup="true">
+                <a href="#" data-toggle="dropdown"  onclick="keepOpen(this)" aria-label="$T('Glitter-mainMenu')" aria-haspopup="true" aria-expanded="false">
                     <span class="icon-bar"></span>
                     <span class="icon-bar"></span>
                     <span class="icon-bar"></span>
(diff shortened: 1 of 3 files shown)


thezoggy:
dunno if ai, but seems to have hit many repos all at the same time for this same thing

coopa11y:
Totally fair question. Yes, I do use AI tooling to help with the coding side of things, but I promise this is not a mysterious robot roadshow rolling through GitHub.

My main goal is accessibility. I use and care about these apps, and I'm trying to make them work better with screen readers like NVDA, JAWS, and VoiceOver, both for myself and for anyone else in the community who benefits from that work.

I'm trying to keep each PR small, focused, and easy to review. If there's a project preference for how these accessibility fixes should be shaped, I'm very happy to adjust.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -105,6 +105,14 @@ function ViewModel() {
         return self.speed() + ' ' + (self.speedMetrics[self.speedMetric()] ? self.speedMetrics[self.speedMetric()] : "B/s");
     });
 
+    self.timeLeftLabel = ko.pureComputed(function() {
```
Safihre: Our application is already quite heavy, so is it ok if we do this without another extra compute?
So maybe only the timeLeft would maybe also be clear enough?

interfaces/Glitter/templates/static/javascripts/glitter.main.js
```diff
@@ -105,6 +105,14 @@ function ViewModel() {
...
+        return glitterTranslate.timeRemaining + ': ' + self.timeLeft();
+    });
+
+    self.speedMenuLabel = ko.pureComputed(function() {
+        return glitterTranslate.downloadSpeed + ': ' + self.speedText() + '. ' + glitterTranslate.openSpeedLimitMenu;
```
Safihre: Same for this one.

Besides that, no comments.

interfaces/Glitter/templates/static/javascripts/glitter.main.js
coopa11y: Updated this to use the existing time-left value directly, so there is no extra computed label.

coopa11y: Same here. The speed button now uses the existing speed text directly, and I removed the unused label strings.

Safihre (commented):
Thanks for the accessibility cleanup — good direction overall, and nice reuse of existing translation keys where labels were needed. A few things worth addressing, noted inline.

**Should fix**
- The new `aria-live` on the time-left element will announce the ETA roughly every second during downloads (details inline) — likely counterproductive for the AT users this PR is meant to help.

**Minor**
- A couple of redundant `aria-label`s that just duplicate existing visible text.
- `aria-expanded` consistency between the navbar-toggle and the dropdown toggles.
- One avoidable new translation string (`Glitter-clearSpeedLimit`) — can reuse an existing key.
- `main.tmpl` only deletes a blank line; unrelated to the a11y work and could be dropped to keep the diff focused.

Verified against `glitter.main.js` for the refresh cadence and the Bootstrap 3 collapse/dropdown behavior. None of this blocks the overall approach.

interfaces/Glitter/templates/include_menu.tmpl
```diff
@@ -26,7 +26,7 @@
                     <li class="divider"></li>
                     <li><a href="#" data-bind="click: openCustomPauseTime">$T('Glitter-custom')</a></li>
                 </ul>
-                <span class="btn btn-default navbar-btn navbar-timeleft" data-bind="text: timeLeft"  data-toggle="dropdown">0:00</span>
+                <span class="btn btn-default navbar-btn navbar-timeleft" role="status" aria-live="polite" data-bind="text: timeLeft, attr: { 'aria-label': timeLeft }">0:00</span>
```
Safihre: **`aria-live` here will spam screen readers.** `timeLeft` is rewritten on every queue refresh — `setNextUpdate` re-fires at `refreshRate` (default ~1s) and calls `self.timeLeft(...)` in `glitter.main.js`. A polite live region on a value that changes every second means NVDA/JAWS announce the ETA continuously, drowning out everything else — worse than the original for AT users. Suggest dropping `role="status"` + `aria-live` entirely; the ETA is glanceable, not announce-worthy. (Note `role="status"` already implies `aria-live="polite"`, so the two are redundant together regardless.)

Two more on this line:
- `data-toggle="dropdown"` was removed here, so clicking the time-left no longer opens the pause menu. It's still reachable via the caret button, so nothing breaks — just flagging the behavior change.
- `'aria-label': timeLeft` duplicates the element's own text content and can be dropped.

interfaces/Glitter/templates/include_menu.tmpl
```diff
@@ -1,19 +1,19 @@
...
             #include $webdir + "/../../Config/templates/staticcfg/images/logo-small.svg"#
         </a>
         <div class="btn-group">
             <div class="btn-group">
-                <button type="button" class="btn btn-default navbar-btn" data-bind="click: pauseToggle, attr: { 'title': downloadsPaused() ? '$T('link-resume')' : '$T('link-pause')' }">
+                <button type="button" class="btn btn-default navbar-btn" data-bind="click: pauseToggle, attr: { 'title': downloadsPaused() ? '$T('link-resume')' : '$T('link-pause')', 'aria-label': downloadsPaused() ? '$T('link-resume')' : '$T('link-pause')' }">
                     <span class="glyphicon" data-bind="css: { 'glyphicon-play' : downloadsPaused(), 'glyphicon-pause' : !downloadsPaused() }"></span>
                 </button>
-                <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown">
+                <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown" aria-label="$T('pauseFor')" aria-haspopup="true">
```
Safihre: Minor consistency: this toggle and the main-menu toggle (line 85) get `aria-haspopup` but no initial `aria-expanded="false"`, whereas the navbar-toggle (line 3) did get one. Bootstrap 3's dropdown JS toggles `aria-expanded` at runtime so it works either way, but adding the static `aria-expanded="false"` here would match.

interfaces/Glitter/templates/include_menu.tmpl
```diff
@@ -37,13 +37,13 @@
             <li class="speedlimit-dropdown dropdown">
                 <div class="btn-group">
                     <div class="btn-group">
-                        <button type="button" class="btn btn-default navbar-btn navbar-speed" data-toggle="dropdown">
+                        <button type="button" class="btn btn-default navbar-btn navbar-speed" data-toggle="dropdown" aria-haspopup="true" data-bind="attr: { 'aria-label': speedText }">
```
Safihre: `'aria-label': speedText` is redundant — the button already contains `<span data-bind="text: speedText">` as visible text, so it already has an accessible name. The `aria-label` just duplicates it. (`aria-haspopup` is good to keep.)

interfaces/Glitter/templates/include_menu.tmpl
```diff
@@ -37,13 +37,13 @@
...
                             <span data-bind="text: speedText">0 KB/s</span>
                         </button>
-                        <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown" onclick="keepOpen(this)">
+                        <button type="button" class="btn btn-default navbar-btn dropdown-toggle" data-toggle="dropdown" onclick="keepOpen(this)" aria-hidden="true" tabindex="-1">
```
Safihre, line 43: 👍 `aria-hidden="true"` + `tabindex="-1"` is a valid way to hide this redundant caret — it passes axe's `aria-hidden-focusable` rule (the `tabindex="-1"` removes focusability) and the primary trigger above is now properly labeled.

interfaces/Glitter/templates/include_menu.tmpl
```diff
@@ -37,13 +37,13 @@
...
                             <span class="caret"></span>
                         </button>
-                        <a href="#" class="max-speed-input-clear hover-button" data-bind="click: clearSpeedLimit, visible:(speedLimit() < 100 && speedLimit() > 0)" style="display: none;">
+                        <a href="#" class="max-speed-input-clear hover-button" aria-label="$T('Glitter-clearSpeedLimit')" data-bind="click: clearSpeedLimit, visible:(speedLimit() < 100 && speedLimit() > 0)" style="display: none;">
```
Safihre: This label can reuse an existing, already-translated key instead of adding `Glitter-clearSpeedLimit`. `button-clear` ("Clear") is translated in all 19 languages, so `aria-label="$T('button-clear')"` reads fine for this clear button inside the speed dropdown — no new strings to translate. See the skintext.py comment.

sabnzbd/skintext.py
```diff
@@ -813,6 +813,9 @@
     "Glitter-pause6h": TT("Pause for 6 hours"),
     "Glitter-setMaxLinespeed": TT("You must set a maximum bandwidth before you can set a bandwidth limit"),
     "Glitter-left": TT("left"),
+    "Glitter-toggleNavigation": TT("Toggle SABnzbd navigation menu"),
+    "Glitter-clearSpeedLimit": TT("Clear speed limit"),
```
Safihre: `Glitter-clearSpeedLimit` is avoidable: reuse the existing `button-clear` ("Clear", line 723) on the template element instead — already translated in all 19 languages, saving ~19 fresh translations. The other two new keys (`Glitter-toggleNavigation`, `Glitter-mainMenu`) have no existing equivalent, so keeping those as new strings is justified.

Safihre:
Lol, Claude took over. Sorry for the spam.

coopa11y:
Thanks, these all make sense.

I removed the live region from the time-left value and restored the dropdown behavior there. I also dropped the redundant dynamic labels for time-left and speed, added the initial expanded state on the dropdown toggles, and switched the clear action to the existing Clear translation.

Also cleaned up the unrelated blank-line diff in main.tmpl.

Safihre:
Thank you 👍

Issue #1964, opened 2021-10-12
Sample-detection broken for `.par2` files


Safihre:
Some users reported it on the forums, but now I also saw it myself: the sample detection based on `guessit` is just not working well enough.
Files like `last.week.tonight.with.john.oliver.s08e26.1080p.web.h264-glhf-sample.par2` or not detected as a sample.
@jcfp Maybe we should just go back to the old method with the super-simple regex?
Despite theoretical cases in which it would fail (as shown in the unittests), did we even get real-life complaints? I don't remember any.
Seems `guessit` just can't handle this well.

thezoggy:
We have (I have). Why I ended up turning off sample detection and just let sonarr handle it. But I do agree the false positive is very rare

PR #3475, opened 2026-06-30
Use media duration for accurate sample detection

### LINKED ISSUE #2083: Shows/Movies with `sample` or `proof` in the name get ignored

Safihre:
Enhances the `is_sample()` logic to reduce false positives for files whose names contain "sample" or "proof" but are actually full-length content.

Integrates the `hachoir` library to parse media files and extract their duration. Files are only considered samples if their duration is less than or equal to `SAMPLE_MAX_DURATION` (default 2 minutes).

If media duration cannot be determined (e.g., not a media file or parsing fails), the detection falls back to the name-based approach.

Closes #2083

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -1085,19 +1085,23 @@ def get_all_passwords(nzo) -> list[str]:
     return unique_passwords
 
 
-def is_sample(filename: str, filepath: Optional[str] = None) -> bool:
-    """Try to determine if filename is (most likely) a sample.
-    When filepath points to the actual file on disk, the media duration is used
+def is_sample(filename_or_filepath: str) -> bool:
+    """Try to determine if the file is (most likely) a sample.
+    When given a path to an actual file on disk, the media duration is used
     to rule out false-positives on titles that merely contain "sample" or
     "proof" (e.g. "The.Moment.of.Proof.S01E01")."""
-    if not re.search(RE_SAMPLE, filename):
+    if not re.search(RE_SAMPLE, os.path.basename(filename_or_filepath)):
         return False
 
     # Long media files are never a sample, no matter what its name suggests
-    if filepath:
-        if duration := get_media_duration(filepath):
-            logging.debug("Media duration of %s is %s seconds", filepath, duration)
+    if os.path.isfile(filename_or_filepath):
+        if duration := get_media_duration(filename_or_filepath):
+            logging.debug("Media duration of %s is %s seconds", filename_or_filepath, duration)
             return duration <= SAMPLE_MAX_DURATION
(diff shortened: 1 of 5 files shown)


jcfp:
Any reason for not using guessit for the pre-selection part? Its sample detection has a much lower false positive rate than the current regular expression, correctly determining something like `The.Moment.of.Proof.S01E01` isn't a sample.

Safihre:
Didn't we try that and it didn't work well enough?

jcfp:
We did indeed (in #1964) but I completely forgot about that - never mind.

sabnzbd/misc.py
```diff
@@ -1073,9 +1085,38 @@ def get_all_passwords(nzo) -> list[str]:
...
+    "proof" (e.g. "The.Moment.of.Proof.S01E01")."""
+    if not re.search(RE_SAMPLE, filename):
+        return False
+
+    # Long media files are never a sample, no matter what its name suggests
+    if filepath:
+        if duration := get_media_duration(filepath):
+            logging.debug("Media duration of %s is %s seconds", filepath, duration)
+            return duration <= SAMPLE_MAX_DURATION
+    return True
```
mnightingale: I think False is a safer default.

Currently failure to determine the duration is treated as evidence that the file is a sample.
`get_media_duration` returning `None` or `0.0` falls through to True, what if it has broken metadata or `hachoir` can’t parse it for some reason? 

I think it should be explicit and only return `True` if the duration could be determined i.e. at least handle the `None` case. Alternatively, `get_media_duration` could return `float(‘inf’)` and keep files if it could not be determined - not sure how that would play with files that are not video files in the first place.

However, I think you wanted the current behaviour and fallback to filename, in which case could it at least debug log if duration is None and explain it could not determine duration so is using filename.

I think I would still prefer if it couldn’t determine duration, return False and keep the file.

---

Separately the signature, could it take it as `filename_or_filepath` and `os.path.basename(...) and os.path.isfile(...) - get_media_duration handles that` but I'm not sure if I really prefer it. The only places that only the filename is provides is `parring` and several tests.

Safihre: Indeed, I want the fallback to filename.

PR #3476, opened 2026-06-30
Improve Config search accessibility

coopa11y:
This improves the Config search dropdown for keyboard and screen reader users.

What changed:
- Gives the icon-only search trigger and search field translated accessible names using the existing Search text.
- Keeps the dropdown expanded state in sync with the Bootstrap open and close events.
- Lets Arrow keys move from the search field into the generated results, and lets Escape close the dropdown from the search field or results.
- Builds search results with DOM helpers instead of inserting raw HTML strings.

Checked:
- git diff --check
- node --check interfaces/Config/templates/staticcfg/js/script.js
- Local Playwright harness for Config search keyboard behavior
- Local mobile and tablet viewport smoke checks for Config search

### DIFF
--- a/interfaces/Config/templates/_inc_header_uc.tmpl
+++ b/interfaces/Config/templates/_inc_header_uc.tmpl
@@ -170,14 +170,14 @@
                     </a>
                 </li>
                 <li class="dropdown" id="search-menu">
-                    <a data-toggle="dropdown" href="#">
+                    <a data-toggle="dropdown" href="#" aria-label="$T('cmenu-search')" aria-controls="search-dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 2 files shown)


thezoggy:
but you can already do this today with tab, arrows and esc?

and dont forget to test out on mobile/tablet

coopa11y:
Good call. I checked current develop against a small local harness.

Tab already reaches the search UI, and once focus is on a result Bootstrap handles arrows and Escape. The gap this is trying to cover is from the search field itself: ArrowDown moves into the generated results, and Escape closes the dropdown from either the field or a result.

I also ran the local harness in mobile and tablet viewports to make sure opening, searching, and selecting still works there. I tightened the PR text so it does not sound like all keyboard support was missing before.

PR #3477, opened 2026-07-01
Improve Glitter queue row accessibility

coopa11y:
This improves keyboard and screen reader access for the Glitter queue row controls.

Changes include:

- Adds accessible labels for row actions like pause/resume, rename, details, delete, and settings.
- Adds row-specific labels for the select-job checkbox in multi-edit mode.
- Adds labels for the multi-edit toolbar controls.
- Keeps the desktop row action buttons reachable by keyboard and reveals them when focused.
- Preserves the existing mobile layout behavior.

Tested with desktop, phone, small phone, and tablet browser harnesses, plus the static accessibility harness and syntax checks.

### DIFF
--- a/interfaces/Glitter/templates/static/javascripts/glitter.queue.js
+++ b/interfaces/Glitter/templates/static/javascripts/glitter.queue.js
@@ -626,6 +626,44 @@ function QueueModel(parent, data) {
         return rewriteTime(self.timeLeft());
     });
 
+    self.labelWithName = function(label) {
+        var name = self.name();
+        return name ? label + ' ' + name : label;
+    };
+    self.selectLabel = ko.pureComputed(function() {
+        return self.labelWithName(glitterTranslate.selectJob);
+    });
+    self.pauseToggleLabel = ko.pureComputed(function() {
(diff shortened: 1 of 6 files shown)


interfaces/Glitter/templates/static/javascripts/glitter.queue.js
```diff
@@ -626,6 +626,44 @@ function QueueModel(parent, data) {
         return rewriteTime(self.timeLeft());
     });
 
+    self.labelWithName = function(label) {
```
Safihre: I don't want to add all these extra pureCompute, maybe it can be solved in another way?

coopa11y: That makes sense. I updated this to avoid adding extra computed label helpers for each queue item. The labels are now handled directly in the template where they can be, with only the pause/resume label staying dynamic since that text actually changes.

I've had similar feedback on a couple of these accessibility PRs now, so I'll carry that forward: keep the labels simple, use the existing translation/template patterns, and avoid adding extra state just to build accessible names.

Safihre:
Much better 👍

thezoggy:
looks like this change broke stuff in 5.0rc1 can no longer click on icons on hover
image

PR #3478, opened 2026-07-01
unrar 7.23

jcfp:
[changes](https://www.rarlab.com/rarnew.htm) include several security fixes:
```
  Version 7.23

  1. Heap overflow vulnerability is fixed in RAR5 recovery volume
     data reconstruction code. It affects WinRAR, RAR and UnRAR.
     UnRAR.dll library doesn't include recovery volume processing,
     so it is not affected.

     We are thankful to Arjun Basnet from Securin Labs for letting us know
     about this security issue.

  2. Symbolic link pointing outside of destination folder could be created
     even without -ola switch, when extracting a specially crafted
     RAR archive by WinRAR, RAR, UnRAR or UnRAR.dll library.
     
     Further check in extraction code prevents placing files to such folder
     even in case of multiple extraction commands, excluding the possibility
     of path traversal attack for WinRAR, RAR or UnRAR based extraction.
     It limits the potential threat to a case where another tool uses this
     symbolic link to store files.

     We are thankful to scofaild23-bnomran for letting us know about this
     security issue.

  3. 7zxa.dll 7z extraction library is updated to version 26.02 to include
     bug and vulnerability fixes by the library developer.
  
  4. Switch -iver prints RAR version even if -idc is specified
     in the command line, configuration file or RARINISWITCHES environment
     variable. Previously -idc blocked -iver action.
     
     Also a new line character is added to -iver output.
```


Safihre:
@mnightingale @thezoggy do we even need 7zip if unrar can unpack also 7zips?

mnightingale:
~7zip is also used for zips~ actually ive read the changes, it just includes 7z libraries so can unrar also handle .zip and everything 7zip does?

I think maybe it's just WinRAR that can use 7zxa to handle 7zip and maybe zip, it doesn't appear unrar can.

thezoggy:
probably could take this to https://github.com/sabnzbd/sabnzbd/discussions/2646 (needs to be unlocked and things validated for how we use stuff today)

iirc not all the unrar variants out there can do stuff (non-free vs free) and then unrar itself doesnt do the zip but basically does it via the dll which is why were stuck having to use each lib. but tar support for example could easily just be done by one of the existing lib vs us trying to use python..

Safihre:
> ~7zip is also used for zips~ actually ive read the changes, it just includes 7z libraries so can unrar also handle .zip and everything 7zip does?
> 
> I think maybe it's just WinRAR that can use 7zxa to handle 7zip and maybe zip, it doesn't appear unrar can.

Indeed, I checked:
```
UNRAR 7.22 x64 freeware      Copyright (c) 1993-2026 Alexander Roshal

tests\data\test_zip\testfile.zip is not RAR archive
No files to extract
--- zip exit code: 10 ---
total 0
drwxr-xr-x 1 FYSICON+jiri.van.bergen 4096 0 Jul  2 10:51 .
drwxr-xr-x 1 FYSICON+jiri.van.bergen 4096 0 Jul  2 10:51 ..
```

```
UNRAR 7.22 x64 freeware      Copyright (c) 1993-2026 Alexander Roshal

tests\data\test_7zip\testfile.7z is not RAR archive
No files to extract
```

PR #3482, opened 2026-07-02
Add workflow for automated bundled tool updates

Safihre:
Introduces a GitHub Actions workflow to automatically check for new releases of unrar, 7-Zip, and par2cmdline-turbo. When new versions are found, it creates a pull request with the updated binaries for review.

Fun use of Fable 5 😁

What do you think @jcfp @thezoggy @mnightingale ?

### DIFF
--- a/.github/workflows/update_tools.yml
+++ b/.github/workflows/update_tools.yml
@@ -23,32 +23,31 @@ jobs:
       - name: Determine latest unrar version
         id: version
         run: |
-          curl -fsSL --retry 3 https://www.rarlab.com/download.htm -o "$RUNNER_TEMP/download.htm"
-          RAW=$(grep -oE 'rarmacos-x64-[0-9]+\.tar\.gz' "$RUNNER_TEMP/download.htm" | grep -oE '[0-9]+' | sort -n | tail -1)
-          if [ -z "$RAW" ]; then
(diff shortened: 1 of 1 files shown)


thezoggy:
Looks good, would be nice to validate hashes for release as well.
- https://github.com/ip7z/7zip/issues/231

.github/workflows/update_tools.yml
```diff
@@ -0,0 +1,202 @@
...
+    env:
+      GH_TOKEN: ${{ github.token }}
+      BRANCH: automation/update-unrar
+    steps:
+      - uses: actions/checkout@v6
+      - name: Determine latest unrar version
+        id: version
+        run: |
+          curl -fsSL --retry 3 https://www.rarlab.com/download.htm -o "$RUNNER_TEMP/download.htm"
+          RAW=$(grep -oE 'rarmacos-x64-[0-9]+\.tar\.gz' "$RUNNER_TEMP/download.htm" | grep -oE '[0-9]+' | sort -n | tail -1)
```
jcfp: This can be done more effectively with sed, directly extracting the version string you're looking for:
`curl -fsSL --retry 3 https://www.rarlab.com/download.htm -o- | sed -n 's%.*/rarmacos-x64-\([0-9]\+\)\.tar\.gz.*%\1%p'`

Safihre: 👍

PR #3489, opened 2026-07-07
[Enhancement] Adding Search Bar to Queue

fallenreaper:
Task:
- Add the ability to filter the queue in scenarios where you have queues longer than a few pages.

Purpose:
- The ability to see where something is in the queue.  Ease of finding so you could move it up or down in the overall list.

What Happens:
- There currently is no search or filter mechanism to understand what is where for longer queues.

What Should Happen
- When a user supplies a query param, it will be used to filter the existing queue to confirm it is in the list.  The desired end state exists the same, where you could move it to the top or bottom, or conduct all normal actions.


Safihre:
?? We already have a search bar? It's at the bottom left. Your AI just moved it...

fallenreaper:
Closing.  Found the search bar.   Was not in an intuitive location.   I am used to search bars appearing that the top of lists and not at bottom.

PR #3490, opened 2026-07-08
Verify RAR5 passwords when header encryption is not used

### LINKED ISSUE #3488: if password_file is set, SAB will not bail out on data-only-encrypted RARs for which there is no password
### SABnzbd version

5.0.4

### Operating system

Ubuntu and Windows

### Using Docker image

None

### Description

See https://forums.sabnzbd.org/viewtopic.php?p=134551 ... in Dutch, but your browser will translate

With pause_on_pwrar set, SAB should bail out if RARs are encrypted and no password is provided.

But: if password_file is set, SAB will not bail out an encrypted rar for which there is no password, and will download the whole post.

As soon as you remove the password_file setting, SAB agains bails out at about 150MB. Good

I can reproduce.

According to Claude Code:

Root cause, one sentence: the early password-verification check in assembler.check_encrypted_and_unwanted_files() assumes that trigger_parse() not raising an exception means the password is correct, but that's only true for header-encrypted RARs — for data-only-encrypted RARs (like this test file) it's a no-op that "succeeds" for any password, so it always accepts the first password in the list as correct even when it's wrong.

mnightingale:
It turns out because we only need to generate the psw_check_value it's much simpler and can be achieved in Python, rarfile does similar to verify passwords when RAR5 headers are encrypted but doesn't make the method available so I've pretty much duplicated it.

lru_cache to avoid computing it for each rar and each file within, and I need to store True/False
Also added lru_cache to `rar*_s2k` because we open many rar files with the same password.

I also removed `trigger_parse` by making  `setpassword` always trigger a parse.

Fixes #3488 but the small test nzb usually downloads before being paused, I think the deobfuscation checks take a while.


sabnzbd/misc.py
```diff
@@ -1750,14 +1766,60 @@ def info_callback(self, rar_obj: rarfile.RarInfo):
...
+        # Encrypted headers already verify the password
+        # We ignore that each file could have a different password
+        if self._file_parser and self._file_parser.has_header_encryption():
+            return
+        if isinstance(rar_obj, rarfile.Rar5FileInfo) and rar_obj.is_file():
+            if rar_obj.needs_password():
+                _algo, flags, kdf_count, salt, _iv, checkval = rar_obj.file_encryption
+                if flags & rarfile.RAR5_XENC_CHECKVAL:
+                    if not rarfile_rar5_file_verify_check_value(self._password, salt, kdf_count, checkval):
+                        raise rarfile.RarWrongPassword()
```
Safihre: Raising from the info callback aborts `parse()` midway, but `RarFile._parse` assigns `self._file_parser` *before* calling `parse()`, so a rejected password leaves the handle with a partially populated parser:

```python
with SABRarFile('tests/data/basic_rar5_64/testfile.rar', part_only=True) as zf:
    try:
        zf.setpassword('nope')
    except rarfile.RarWrongPassword:
        pass
    zf.namelist()  # ['My_Test_Download.bin'] — should be 3 files
```

This affects `check_encrypted_and_unwanted_files`: when all passwords fail, the unwanted-extension loop iterates that truncated `namelist()` and can miss unwanted files. Before this PR the full file list stayed readable in that scenario (headers aren't encrypted), so it's a small regression. Retrying with a correct password recovers fine — only the all-passwords-fail path is affected.

Suggestion: don't raise mid-parse. Let `setpassword()` complete the parse, then verify the check value from the infolist (the first encrypted `Rar5FileInfo` is enough) and raise afterwards. Same detection, full namelist preserved, and it also removes the dependence on rarfile assigning `_file_parser` before `parse()` runs.

Safihre (commented):
Ran the test suite against rarfile 4.3 (all green) and verified the check-value math matches rarfile's `_check_password` byte-for-byte. One issue found, posted inline.

sabnzbd/misc.py
Safihre: (Claude did this of course)

mnightingale: That's a nice edge case, added a test.
Annoyingly I made a couple of variations previously that handled it correctly.

sabnzbd/misc.py
```diff
@@ -1754,10 +1756,60 @@ def filelist(self) -> list[str]:
...
+            # Encrypted headers already verify the password, and we assume all files use the same password
+            if self._file_parser.has_header_encryption():
+                return
+            for rar_obj in self.infolist():
+                if rar_obj.is_file() and rar_obj.needs_password():
+                    _algo, flags, kdf_count, salt, _iv, checkval = rar_obj.file_encryption
+                    if flags & rarfile.RAR5_XENC_CHECKVAL:
+                        if not rar5_check_password(self._password, salt, kdf_count, checkval):
+                            raise rarfile.RarWrongPassword()
+                        # RAR5 salts are per-file, so lru_cache won't dedupe across files.
```
Safihre: Small nit on this comment: it turns out RAR reuses one salt across all files of an archive created in the same session (and even across volumes of a set). I checked empirically — the 3 files in `tests/data/basic_rar5_64/testfile.rar` share a single `file_encryption` salt, and all 6 volumes of `tests/data/fully_encrypted_and_obfuscated_rars` use one salt too. So `lru_cache` *would* dedupe across files here.

The early exit is still the right call — checking one file is strictly cheaper than N cache hits — it's just the stated reason that's off. Maybe reword to something like: "All files typically share the same password (and usually even the same salt), so one successful check is enough."

mnightingale: Argh, I shouldn't have trusted Claude; I didn't think they would be different.

Safihre: Hahah my Claude disagreed with yours? 😂😂

PR #3492, opened 2026-07-11
Optimise scanning for orphan jobs and retry all

### LINKED ISSUE #3484: build_history: UI refresh loads whole history into memory
### SABnzbd version

develop

### Operating system

N/A

### Using Docker image

None

### Description

Related to https://github.com/sabnzbd/sabnzbd/issues/3473#issuecomment-4867866215 with large history.

https://github.com/sabnzbd/sabnzbd/blob/e929a0c6b66cb6336d60e08ed052b2c0a3ac6ef8/sabnzbd/nzbqueue.py#L151-L160

I'm not familiar with what this is doing but I think we should see if there is a way to limit scope to avoid loading all jobs.

Also this is an advantage of the generator I used in `get_feed_jobs`, it avoids allocating a list of the whole history.

mnightingale:
Fixes #3484 - n.b. I might make separate improvements to `build_history` but this PR address the issue that `scan_jobs` and `api_retry_all` load the whole database into memory and prepare all the columns for use.

`sabnzbd.api.build_history()[0]` grabs all history regardless of status with a limit of 1,000,000 records so effectively all.

It filters the history in Python with `if item["retry"] or item["loaded"] or item["status"] == Status.QUEUED`

- **retry:** only True when status = Status.Failed and path exists or report == "future"
- **loaded:** always False for DB rows, set to True when postproc starts
- **status = Status.QUEUED:** waiting to download or postproc

We can reduce how many records and columns we need to select for scan_jobs.
Registered / known paths that are not eligible for repair:

- All active items in the nzbqueue (unchanged)
- `get_retryable_jobs()` not archived and (status = 'Failed' OR report = 'future'), future or with a path that exists
- All items in postproc queue - note duplicates in pp and history do not matter

`_api_retry_all` retries anything in `get_retryable_jobs()`

The test ran without this PR also passes, so I don't think the behaviour has changed but `scan_jobs` is tested more than `_api_retry_all`

---

A few things of note:

`FakeHistoryDB` adds to history database with impossible statuses, history only stores completed and failed but it also adds records with statuses that only exist in nzbqueue or postproc. I haven't changed it because some tests appear to actually rely on this.

`build_history` start and limit are now required, removed the 1,000,000 default. It is now only used from `_api_history_default` which always provides start/limit even if 0. It might be useful to document that limit=0 means postproc only. As mentioned previously I may look at this some other time, perhaps taking the prepending of postproc out of build_history and/or simpler handling of duplicates in postproc and history.

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1930,8 +1930,8 @@ def build_header(webdir: str = "", for_template: bool = True, trans_functions: b
 
 
 def build_history(
-    start: int = 0,
-    limit: int = 1000000,
+    start: int,
+    limit: int,
     archive: bool = False,
     search: Optional[str] = None,
     categories: Optional[list[str]] = None,
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -868,11 +868,7 @@ def _api_rss_now(name: str, kwargs: ApiParams) -> bytes:
 
 def _api_retry_all(name: str, kwargs: ApiParams) -> bytes:
     """API: Retry all failed items in History"""
-    items = sabnzbd.api.build_history()[0]
-    nzo_ids = []
-    for item in items:
-        if item["retry"]:
-            nzo_ids.append(retry_job(item["nzo_id"]))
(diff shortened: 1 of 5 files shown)


Safihre:
Much better.

PR #3497, opened 2026-07-13
Add SAB_FILES environment variable for postproc scripts

### LINKED ISSUE #3118: postprocessing script: does/can SABnzbd tell which files it generated?
### SABnzbd version

4.5.2

### Operating system

Linux

### Using Docker image

None

### Description

Based on https://sabnzbd.org/wiki/configuration/4.5/scripts/post-processing-scripts : AFAIK SAB does not tell to the post-processing script which files it 'generated' in that download. Correct?

That causes this problem for me: if the download is from a series, the file lands in a common/shared directory (for that series) with already existing files, and my post-processing script does not know which file(s) it has to process. 

If the above is indeed as it is: can there be an env var like `SAB_RESULTING_FILENAMES` that tells that? And the real resuling filenames, so after SAB's renaming.

My workaround now: the post-processing script tries to find files which were generated in last 30 seconds, based on  `os.path.getctime()` ... tricky, for example with small downloads or very big downloads.

mnightingale:
Fixes #3118

Adds `SAB_FILES` containing a JSON encoded array of files relative the to complete directory.
Primary reason for use is it avoids problems with path separators, a simple comma separated list of paths could have issues.

Will need a wiki change if you're happy with this.

### DIFF
--- a/sabnzbd/newsunpack.py
+++ b/sabnzbd/newsunpack.py
@@ -21,6 +21,7 @@ sabnzbd.newsunpack
 
 import os
 import sys
+import json
 import re
 import subprocess
 import logging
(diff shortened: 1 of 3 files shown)


Safihre:
Let's add a test that the receiving script nicely decodes to the same data as was the input.
Including some escape chars correctly passed.
If that's possible.

mnightingale:
Had a little help but I'm happy with what it does, receives the env vars and echos it back then check what comes back matches.

`pytest.param` was new to me and is quite useful for adding an ID to test cases, I'd always done a separate ids=["a", "b"]

Safihre:
Could you also create a PR to update the documentation?

PR #3498, opened 2026-07-14
Add rss age rule type

### LINKED ISSUE #2952: Age Filtering in RSS

mnightingale:
Fixes #2952

image

Rules are:

- <10d: younger than 10 days
- \>15d: older than 15 days
- 12d: same as <12d

Supported units: `y / mo / w / d / h / m / s`

Will need wiki changes.


thezoggy:
Might want to add year as well?

what about and = comparison, so age `=1d` as example

sabnzbd/skintext.py
```diff
@@ -625,6 +625,7 @@
     "rss-atmost": TT("At most"),  #: Config->RSS filter-type selection menu
     "rss-from": TT("From SxxEyy"),  #: Config->RSS filter-type selection menu "From Season/Episode"
     "rss-from-show": TT("From Show SxxEyy"),  #: Config->RSS filter-type selection menu "From Show Season/Episode"
+    "rss-age": TT("Age"),  #: Config->RSS filter-type selection menu (e.g. ">3d", "<12h")
```
Safihre, line 628: Need to make clearer what Age, Post Age? RSS Post Age?

mnightingale:
> Might want to add year as well?

That's a good idea, added years (y) and months (mo)

> what about and = comparison, so age `=1d` as example

I thought about this for ranges but it makes it complicated, so like the size rules (at least and at most) you can just handle it with multiple rules and there is a test for it.

I suppose I could do an `=` comparator, but I'm not sure it's worth the complication when it can already be accomplish with multiple rules.

For example for between 1 and 2 days:

Rule | Type | Value
-|-|-
0|Age|>1d
1|Age|<2d
2|Accept|*

---

Will look at making which age it is clearer, will have to check what newznab spec calls for but afaik it can depend on what the indexer populates

thezoggy:
The lack of equal with comparison like <= or >= means you have to +1 or use more grainular measurement (ex: use hours instead of day) which isn't horrible.. but when you want to do something like give me stuff for this month then that falls a little flat. But anyways still way better than before.

Also what about adding w for weeks?

sabnzbd/skintext.py
mnightingale: The couple of indexers I checked populate age as 'posted to usenet', does changing it to 'Post Age' make it any clearer to users, do they know what a post refers to and does it matter?

The table showing rss readout, just says 'Age'.

mnightingale:
> The lack of equal with comparison like <= or >= means you have to +1 or use more grainular measurement (ex: use hours instead of day) which isn't horrible.. but when you want to do something like give me stuff for this month then that falls a little flat. But anyways still way better than before.

I'll add `<=` and `>=`, but I think they'll just be aliases because the `=` point is an exact second.

> Also what about adding w for weeks?

I've already added weeks.

```
"y": "years",
"mo": "months",
"w": "weeks",
"d": "days",
"h": "hours",
"m": "minutes",
"s": "seconds",
```

Overall it's a pretty niche feature request, the original request was really about ignoring new entries for a while. But I think there is a useful use case for age filters, for example if you don't run SAB often and only want it to grab very recent entries.

mnightingale:
I've added `>=`, `=>`, `<=`, and `=<` but they're just aliases to `<` and `>`.

I'm not 100% sure about them, they prevent punishment of a plausible typo/habit but are slightly misleading. The issue is age is measured against a live clock, advancing every second so `<=` compared to `<` if implemented would be allowing an extra second exact boundary instant so no real benefit.

Safihre:
Indeed I was just about to write that 😊 
Still good to add the aliases.

And agreed, Age is fine since it's indeed exactly displayed like that in the tables.

sabnzbd/rss.py
```diff
@@ -312,25 +357,31 @@ def matches(
             None  -> rule does not apply
         """
         # Category rule
-        if category and self.type == "C":
+        if category and self.type == FeedRuleType.CATEGORY:
```
Safihre, line 362: Nice change.

sabnzbd/rss.py
```diff
@@ -341,18 +392,52 @@ def matches(
...
+        recent entries pass. Unparseable expressions are ignored (match).
+        """
+        m = _RE_AGE.match(expr or "")
+        if not m:
+            logging.debug("Ignoring unparseable age filter %r", expr)
+            return True
+
+        comparator = m.group(1) or "<"
+        amount = int(m.group(2))
+        unit = m.group(3).lower() or "d"
```
Safihre: This crashes with an `AttributeError` when the unit is omitted (e.g. `>30` or `12`): the regex makes the unit group optional, so `m.group(3)` is `None`.

```suggestion
        unit = (m.group(3) or "d").lower()
```

The docstring promises "optional unit suffix … default days", so this is a supported input. Since `run_feed` only catches `RuntimeError` around `_process_entry`, the exception aborts the whole feed scan. Worth adding a regression test for the unitless case alongside the fix.

sabnzbd/rss.py
```diff
@@ -70,6 +85,29 @@ class RSSState(str, Enum):
...
+    REJECT = "R"  # Reject on title regex match
+    CATEGORY = "C"  # Reject unless category regex matches
+    AT_MOST = "<"  # Reject if size is larger than value
+    AT_LEAST = ">"  # Reject if size is smaller than value
+    FROM = "F"  # Reject episodes before the given SxxEyy
+    FROM_SHOW = "S"  # Accept given show from the given SxxEyy onwards
+    AGE = "G"  # Reject if entry age is outside the given bound
+
+    @classmethod
+    def _missing_(cls, value):
```
Safihre: Returning `None` from `_missing_` is already the default `Enum` behavior (it produces the normal `ValueError`), so this override is dead code and can be removed.

sabnzbd/rss.py
```diff
@@ -70,6 +85,29 @@ class RSSState(str, Enum):
...
+        return None
+
+    @classmethod
+    def non_regex_types(cls) -> frozenset["FeedRuleType"]:
```
Safihre: Minor: this rebuilds the frozenset on every call, and it's called from `__post_init__` for every rule. A module-level constant (or `functools.cache`) would be cleaner.

interfaces/Config/templates/config_rss.tmpl
```diff
@@ -157,6 +157,7 @@
                                     <option value="<"> $T('rss-atmost')</option>
                                     <option value="F"> $T('rss-from')</option>
                                     <option value="F"> $T('rss-from-show') ($T('rss-accept'))</option>
```
Safihre: Pre-existing issue (not from this PR), but since you're touching this dropdown: this option has `value="F"`, duplicating the "From SxxEyy" option above it — it should presumably be `"S"`, as in the other two filter blocks (lines 233 and 310).

Safihre (commented):
Nice feature and thoughtful `Optional[age]` cleanup — the comparator matrix tests and the NOT NULL persistence fallback are well covered. One real bug (crash on a unitless value) plus a few minor notes inline.

interfaces/Config/templates/config_rss.tmpl
mnightingale: I'm not sure if this causes any problems currently, it's in the "default" row. Fixed it anyway.

Safihre:
@claude review

PR #3503, opened 2026-07-18
Restore keyboard access to queue row actions

coopa11y:
This follows up on #3477 and the queue display changes made afterward.

The row action buttons work with a mouse again, but they are completely hidden until the row is hovered. This prevents keyboard and screen reader users from reaching the move, rename, and details actions.

This change keeps the existing mouse behavior while making those actions reachable again. The buttons appear when the row is hovered or when a keyboard user moves focus to them, and the mobile layout remains unchanged.

I confirmed that the actions have clear screen reader labels, appear when focused, remain clickable, and do not change the height of the queue row.

### DIFF
--- a/interfaces/Glitter/templates/static/stylesheets/glitter.css
+++ b/interfaces/Glitter/templates/static/stylesheets/glitter.css
@@ -685,8 +685,15 @@ td.name .row-wrap-text {
     height: 1.5em;
 }
 
-.queue-table td.name .name-options,
-.queue-table td.name:hover .queue-item-password {
+.queue-table td.name .name-options {
+    display: block;
+    opacity: 0;
+    pointer-events: none;
+    position: absolute;
+}
+
+.queue-table td.name:hover .queue-item-password,
+.queue-table td.name:focus-within .queue-item-password {
     display: none;
 }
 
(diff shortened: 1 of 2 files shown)


thezoggy:
Please double check that this doesnt break normal mouse usage on desktops again.

coopa11y:
I double-checked this against both the earlier accessibility change and the current desktop behavior.

This PR is different because the hidden controls are absolutely positioned, transparent, and excluded from pointer interaction. That keeps them available in the keyboard tab order without occupying space or intercepting mouse clicks. When the name cell is hovered, the PR explicitly restores full opacity, normal static positioning, and `pointer-events: auto`. The same visible state is used for keyboard focus. This avoids the earlier regression where `pointer-events: none` was left active after the hover behavior was changed.

I tested these cases using the branch's actual Glitter styles and queue-row markup:

- With no hover or focus, the controls were transparent, ignored pointer input, and did not affect the row's 22 px height.
- In the exposed desktop state, the controls returned to static positioning with full opacity and pointer events enabled, while the row remained 22 px tall.
- Move to top, move to bottom, rename, and details were each visible, enabled, and their click handlers fired.
- Moving keyboard focus into the controls exposed the same layout without requiring hover.
- With the mobile stylesheet applied, the controls remained visible and clickable through the existing mobile override.

Based on those checks, this keeps the current desktop mouse behavior while adding the keyboard path.

Safihre:
I verified on desktop and mobile that it works.

PR #3505, opened 2026-07-18
Improve Glitter search and pagination accessibility

coopa11y:
The Queue and History search fields and pagination controls are difficult to use from a keyboard or screen reader.

This gives both search fields accessible names and makes their search and clear controls normal keyboard-focusable links. The accessible name changes directly from Search to Clear search when text is entered, and clearing the search returns focus to the field.

Pagination page numbers are now keyboard-focusable links, the current page is exposed to assistive technology, and separator dots are skipped.

### DIFF
--- a/interfaces/Glitter/templates/include_history.tmpl
+++ b/interfaces/Glitter/templates/include_history.tmpl
@@ -135,15 +135,20 @@
     <div class="history-footer">
 
     <div class="input-group search-box">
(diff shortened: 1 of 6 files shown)


thezoggy:
in case you were not aware we do have some keyboard shortcuts already in sab
image

benhaney:
There's something darkly funny about posting a UI screenshot as a response to a blind dev in an accessibility PR

coopa11y:
Thanks for pointing this out. The pagination markup change moved page activation from the list item to its accessible link, so the existing Shift+Arrow shortcuts also needed to target that link. I've updated all four shortcuts and tested previous, next, first, and last page navigation in both Queue and History, along with direct link activation.

Issue #3506, opened 2026-07-18
Tray icon fails on Ubuntu 26.04 despite XApp loading successfully


Amajiro:
### SABnzbd version

5.0.4

### Operating system

Ubuntu Linux 26.04

### Description

**Description**

SABnzbd 5.0.4 does not create a system tray icon on Ubuntu 26.04.

The application is installed from the official SABnzbd PPA and launched within the logged-in graphical session, rather than as a system service.

SABnzbd successfully detects and loads XApp, but GTK immediately emits the following error:

2026-07-18 16:30:54,821::DEBUG::[sabtraylinux:33] XApp found: <OverridesProxyModule <IntrospectionModule 'XApp' from '/usr/lib/x86_64-linux-gnu/girepository-1.0/XApp-1.0.typelib'>>

(sabnzbdplus:170200): Gtk-CRITICAL **: 16:30:54.871: gtk_widget_get_scale_factor: assertion 'GTK_IS_WIDGET (widget)' failed

(sabnzbdplus:170200): Gtk-CRITICAL **: 16:30:55.828: gtk_widget_get_scale_factor: assertion 'GTK_IS_WIDGET (widget)' failed

SABnzbd otherwise starts and operates normally.

**Environment**

SABnzbd: 5.0.4
Commit: 128e0d03d7cc61af7e73b18376b880219fbc3596
Installation source: Official SABnzbd PPA
OS: Ubuntu 26.04 LTS
Desktop: GNOME / Ubuntu desktop
Session: Wayland
Python: 3.14.4
Architecture: x86_64

Relevant packages and services:

gir1.2-xapp-1.0 3.2.2-1 installed
ubuntu-appindicators@ubuntu.com enabled
sabnzbdplus system service inactive

SABnzbd is launched directly in the user session:

/usr/bin/python3 /usr/bin/sabnzbdplus --browser 1

**Steps to reproduce**

Install SABnzbd 5.0.4 from the official PPA on Ubuntu 26.04.
Enable the tray icon in SABnzbd.
Ensure the Ubuntu AppIndicators GNOME extension is enabled.
Launch SABnzbd from the application menu or terminal:
sabnzbdplus --browser 0
Observe that no tray icon appears.
Enable debug logging and restart SABnzbd.
Observe that XApp is found, followed immediately by the GTK assertion shown above.

**Expected behaviour**

SABnzbd should display its tray icon through XApp and the Ubuntu AppIndicators extension.

**Actual behaviour**

[...]

PR #3508, opened 2026-07-19
Add Ayatana AppIndicator tray support on GNOME

Amajiro:
## Summary

Adds an Ayatana AppIndicator backend for the Linux tray icon.

On current GNOME/Wayland environments, both `Gtk.StatusIcon` and
`XApp.StatusIcon` may fail to produce a visible tray icon. This adds
`AyatanaAppIndicator3` as the preferred backend when available, while
retaining the existing XApp and Gtk fallbacks.

## Tested on

- Ubuntu 26.04 LTS
- GNOME / Wayland
- SABnzbd develop, 5.1.0RC1
- Python 3.14
- `gir1.2-ayatanaappindicator3-0.1`
- Ubuntu AppIndicators extension

## Result

The SABnzbd tray icon appears correctly and exposes the existing menu.
The icon also updates for idle, downloading, and paused states.

## Notes

The Ayatana typelib is optional. Systems without it continue to use
the existing XApp or Gtk implementation.

Proposed fix for #3506.

### DIFF
--- a/sabnzbd/sabtraylinux.py
+++ b/sabnzbd/sabtraylinux.py
@@ -23,6 +23,9 @@ import gi
 from gi.repository import Gtk, GLib
 import logging
 
+HAVE_APPINDICATOR = False
+HAVE_XAPP = False
+
 try:
     gi.require_version("AyatanaAppIndicator3", "0.1")
     from gi.repository import AyatanaAppIndicator3
@@ -30,20 +33,19 @@ try:
     HAVE_APPINDICATOR = True
     logging.debug("AyatanaAppIndicator3 found: %s", AyatanaAppIndicator3)
 except Exception:
-    HAVE_APPINDICATOR = False
     logging.debug("AyatanaAppIndicator3 not available")
 
-try:
-    gi.require_version("XApp", "1.0")
-    from gi.repository import XApp
+if not HAVE_APPINDICATOR:
+    try:
+        gi.require_version("XApp", "1.0")
+        from gi.repository import XApp
 
-    if not hasattr(XApp, "StatusIcon"):
-        raise ImportError
-    HAVE_XAPP = True
-    logging.debug("XApp found: %s" % XApp)
-except Exception:
-    HAVE_XAPP = False
-    logging.debug("XApp not available, falling back to Gtk.StatusIcon")
+        if not hasattr(XApp, "StatusIcon"):
+            raise ImportError
+        HAVE_XAPP = True
+        logging.debug("XApp found: %s", XApp)
+    except Exception:
+        logging.debug("XApp not available, falling back to Gtk.StatusIcon")
 from time import sleep
 import subprocess
 from threading import Thread
--- a/sabnzbd/sabtraylinux.py
+++ b/sabnzbd/sabtraylinux.py
@@ -24,6 +24,16 @@ from gi.repository import Gtk, GLib
 import logging
 
 try:
+    gi.require_version("AyatanaAppIndicator3", "0.1")
+    from gi.repository import AyatanaAppIndicator3
+
+    HAVE_APPINDICATOR = True
+    logging.debug("AyatanaAppIndicator3 found: %s", AyatanaAppIndicator3)
+except Exception:
+    HAVE_APPINDICATOR = False
+    logging.debug("AyatanaAppIndicator3 not available")
+
+try:
     gi.require_version("XApp", "1.0")
     from gi.repository import XApp
 
@@ -66,17 +76,31 @@ class StatusIcon(Thread):
             logging.debug("language file not loaded, waiting")
 
         self.sabpaused = False
-        if HAVE_XAPP:
+        if HAVE_APPINDICATOR:
+            self.statusicon = AyatanaAppIndicator3.Indicator.new(
+                "sabnzbd",
+                self.sabicons["default"],
+                AyatanaAppIndicator3.IndicatorCategory.APPLICATION_STATUS,
+            )
+            self.statusicon.set_status(AyatanaAppIndicator3.IndicatorStatus.ACTIVE)
+        elif HAVE_XAPP:
             self.statusicon = XApp.StatusIcon()
         else:
             self.statusicon = Gtk.StatusIcon()
-        self.statusicon.set_name("SABnzbd")
-        self.statusicon.set_visible(True)
+
+        if not HAVE_APPINDICATOR:
+            self.statusicon.set_name("SABnzbd")
+            self.statusicon.set_visible(True)
+
         self.icon = self.sabicons["default"]
         self.refresh_icon()
         self.tooltip = "SABnzbd %s" % sabnzbd.__version__
         self.refresh_tooltip()
-        if HAVE_XAPP:
+
+        if HAVE_APPINDICATOR:
+            self.menu = self.create_menu()
+            self.statusicon.set_menu(self.menu)
+        elif HAVE_XAPP:
             self.statusicon.connect("activate", self.right_click_event)
         else:
             self.statusicon.connect("popup-menu", self.right_click_event)
@@ -85,14 +109,19 @@ class StatusIcon(Thread):
         Gtk.main()
 
     def refresh_icon(self):
-        if HAVE_XAPP:
+        if HAVE_APPINDICATOR:
+            self.statusicon.set_icon_full(self.icon, "SABnzbd")
+        elif HAVE_XAPP:
             # icon path must be absolute in XApp
             self.statusicon.set_icon_name(self.icon)
         else:
             self.statusicon.set_from_file(self.icon)
 
     def refresh_tooltip(self):
-        self.statusicon.set_tooltip_text(self.tooltip)
+        if HAVE_APPINDICATOR:
+            self.statusicon.set_title(self.tooltip)
+        else:
+            self.statusicon.set_tooltip_text(self.tooltip)
 
     # run this every updatefreq ms
     def run(self):
@@ -115,8 +144,7 @@ class StatusIcon(Thread):
         self.refresh_tooltip()
         return 1
 
-    def right_click_event(self, icon, button, time):
-        """menu"""
+    def create_menu(self):
         menu = Gtk.Menu()
 
         maddnzb = Gtk.MenuItem(label=T("Add NZB"))
@@ -148,6 +176,11 @@ class StatusIcon(Thread):
         menu.append(mshutdown)
 
         menu.show_all()
+        return menu
+
+    def right_click_event(self, icon, button, time):
+        """menu"""
+        menu = self.create_menu()
         menu.popup(None, None, None, self.statusicon, button, time)
 
     def addnzb(self, icon):


jcfp:
Seems to work in Debian testing with Xfce, menu shows up and is functional. I haven't tried with Gnome and/or Wayland because I didn't have a VM with any of that readily available.

You should change the import bits though, so that fallbacks are only tried if an earlier option failed. Currently xapp will be imported even if the appindicator bits were succesfully loaded, and you could end up with both `HAVE_APPINDICATOR` and `HAVE_XAPP` set to `True` if both are present.

And then there's this:
```
[sabtraylinux:31] AyatanaAppIndicator3 found: <IntrospectionModule 'AyatanaAppIndicator3' from '/usr/lib/x86_64-linux-gnu/girepository-1.0/AyatanaAppIndicator3-0.1.typelib'>
libayatana-appindicator-WARNING [...] libayatana-appindicator is deprecated. Please use libayatana-appindicator-glib in newly written code.
```
Yet more (future?) trayicon ""fun""... with appindicator-glib expecting a `Gio.Menu` rather than a `Gtk.Menu`.

Amajiro:
Thanks, good catch. I’ve updated the import logic so XApp is only attempted when AyatanaAppIndicator3 is unavailable, which avoids both backends being marked as present.

I’ve retested on Ubuntu 26.04 with GNOME/Wayland and the tray icon and menu still work.

I’ve left the appindicator-glib/Gio.Menu migration out of this patch, since that looks like a broader rewrite rather than a small fallback backend change and is probably beyond me.

mnightingale:
Are there likely to be even more icon systems to support in the future?

GTK 3.14 has been deprecated since 2014 perhaps time to let that go, but I’m not familiar with what might still need it.

From a scan through it, I think I’d favour a bit of a refactor with a probably abstract StatusIconBase rather than if statements and use of different APIs everywhere.

I've not taken a lot of time considering and would probably make changes:

<details>
  <summary>Claudes plan</summary>
  
  # Refactor sabtraylinux.py into pluggable status-icon backends

## Context

Linux has an ever-growing set of tray-icon systems. The `linux-appindicator-tray` branch added Ayatana AppIndicator support alongside XApp and the legacy `Gtk.StatusIcon`, and did it with module-level `HAVE_*` flags plus `if/elif/else` chains scattered through `dowork()`, `refresh_icon()`, `refresh_tooltip()` and the signal-connection code. Adding a fourth backend means touching four more branches, and the caller in `SABnzbd.py` has to know about `gi`/Gtk version pinning and swallow exceptions itself.

Goal: one abstract base holding all the shared logic, one small subclass per backend implementing only what genuinely differs, and a single factory so `SABnzbd.py` needs no knowledge of the specifics.

Out of scope (decided): notifications. Linux tray backends have no notification API — AppIndicator, XApp and `Gtk.StatusIcon` are icon+menu only. Linux desktop notifications already work via `send_notify_osd()` in `sabnzbd/notifier.py:219` (notify2/libnotify, gated on `ntfosd_enable`). `notifier.py` is left untouched.

## Design

All in `sabnzbd/sabtraylinux.py`.

### `StatusIconBase` (abstract, subclasses `Thread`)

Owns everything backend-independent:

- Class attrs `SABICONS` (state -> icon path) and `UPDATE_FREQ = 1000`, matching the `SABICONS` module constant style already used in `sabnzbd/sabtraywin.py:50`.
- `run()`: wait for `sabnzbd.WEBUI_READY`, call `self.setup()`, `GLib.timeout_add(self.UPDATE_FREQ, self.update)`, `Gtk.main()`.
- `update()`: the existing `run()` body — `api.fast_queue()`, `to_units()`, pick state, then call `self.set_icon(path)` / `self.set_tooltip(text)`. Returns `True` to stay scheduled.
- `stop()`: `GLib.idle_add(Gtk.main_quit)`.
- `create_menu()` and every menu handler (`addnzb`, `browse`, `opencomplete`, `pauseresume`, `restart`, `shutdown`, `rss`, `pause`, `resume`) — unchanged behaviour, moved verbatim.
- `right_click_event()` — used by the two backends that pop their own menu.

Abstract surface each backend implements (three methods, one classmethod):

| member | purpose |
|---|---|
| `is_available()` (classmethod) | attempt the guarded `gi.require_version` + import, return bool. Keeps the "only load fallback backends when needed" behaviour from `ab0920b7e`. |
| `setup()` | construct the native icon object, set name/visibility, attach the menu or the right-click signal |
| `set_icon(path)` | `set_icon_full` / `set_icon_name` / `set_from_file` |
| `set_tooltip(text)` | `set_title` / `set_tooltip_text` |

### Backends

`AyatanaAppIndicatorIcon`, `XAppStatusIcon`, `GtkStatusIcon` — each ~15 lines, each doing its own import inside `is_available()` and stashing the module on the class. Keeping all three, ordered:

```python
BACKENDS = (AyatanaAppIndicatorIcon, XAppStatusIcon, GtkStatusIcon)
```

Note the real behavioural differences to preserve: AppIndicator sets a persistent menu via `set_menu()` and uses `set_title()` for the tooltip; XApp/Gtk set a name, `set_visible(True)`, and build the menu on demand from a signal (`activate` vs `popup-menu`).

### Factory — the only public entry point

```python
def start_tray() -> Optional[StatusIconBase]:
    """Start the first available tray backend, or None if none work"""
```

Iterates `BACKENDS`, returns the first whose `is_available()` passes (instantiated and started), logging which one was chosen and `logging.info("No supported tray icon backend found")` otherwise. Wrapped so any backend failure falls through to the next rather than killing startup.

The `import gi` / `gi.require_version("Gtk", "3.0")` / `from gi.repository import Gtk, GLib` block moves to the module top under `try/except`, setting `Gtk = GLib = None` on failure; `start_tray()` returns `None` early in that case. Classes only touch `Gtk`/`GLib` at call time, so this is safe.

### Caller — `SABnzbd.py:1175`

Collapses to:

```python
elif sabnzbd.LINUX_POWER and os.environ.get("DISPLAY"):
    import sabnzbd.sabtraylinux

    sabnzbd.LINUXTRAY = sabnzbd.sabtraylinux.start_tray()
```

No `gi` import, no `require_version`, no `try/except` in the caller.

### Lifecycle

- Add `LINUXTRAY = None` next to `WINTRAY = None` in `sabnzbd/__init__.py:171`.
- In `halt()` (`sabnzbd/__init__.py:349`), extend the existing tray-stop block to also stop `LINUXTRAY`.

## Incidental fixes rolled in

- `StatusIcon` currently subclasses `Thread` but never uses it — `__init__` builds a second `self.mythread` and `dowork()` is the body. Becomes a proper `Thread` subclass with `run()`.
- `restart()` and `shutdown()` assign `self.hover_text`, an attribute that does not exist on this class (copied from the Windows tray). Dead; remove.
- `SABICONS` uses `abspath("icons/...")`, which silently depends on the `os.chdir(sabnzbd.DIR_PROG)` at `SABnzbd.py:1162`. Build from `sabnzbd.DIR_PROG` instead, as `send_notify_osd` does at `notifier.py:226`.

## Files

- `sabnzbd/sabtraylinux.py` — the refactor
- `SABnzbd.py` — simplify the tray-start block
- `sabnzbd/__init__.py` — `LINUXTRAY` global + `halt()` stop

## Verification

No test suite covers the tray (it needs a live X/Wayland session), so verify by running:

1. `python3 -m py_compile sabnzbd/sabtraylinux.py SABnzbd.py sabnzbd/__init__.py` and `black --check` per repo style.
2. On a GNOME box with `gir1.2-ayatanaappindicator3-0.1` installed: start SABnzbd with `tray_icon` enabled, confirm the debug log names the AppIndicator backend, the icon appears, and it cycles default -> green while downloading -> gray when paused. Exercise every menu entry.
3. Force the fallbacks to prove the chain works — temporarily reorder `BACKENDS` (or test on Cinnamon/Mint for XApp) and repeat the icon/tooltip/menu checks for `XAppStatusIcon` and `GtkStatusIcon`.
4. Confirm graceful degradation: run with `gi` unavailable (or no tray libs) and check SABnzbd still starts, logging that no backend was found.
5. Confirm shutdown: trigger a restart and a shutdown from the tray menu and verify the process exits cleanly with no lingering GTK thread.
  
</details>

Safihre:
Guess the Windows and macOS trays are too specific to merge in.

mnightingale:
Maybe, macOS is quite unique but there is shared functionality that could be extracted so they all behave the same.

There is currently different behaviour and actions between them all, which is probably worth standardising.

> State derivation from api.fast_queue() + to_units(). These have already drifted: sabtraywin.py:187 shows "Paused - Remaining: X" when paused with bytes left, sabtraylinux.py:134 just shows "Paused". One shared helper fixes that silently.

<details>
  <summary>Refactor the tray icons: pluggable Linux backends + a shared action core</summary>
  
# Refactor the tray icons: pluggable Linux backends + a shared action core

## Context

Linux has an ever-growing set of tray-icon systems. The current branch added Ayatana AppIndicator alongside XApp and the legacy `Gtk.StatusIcon`, using module-level `HAVE_*` flags and `if/elif/else` chains scattered through `dowork()`, `refresh_icon()`, `refresh_tooltip()` and the signal-connection code. Adding a fourth backend means touching four more branches, and the caller in `SABnzbd.py` has to know about `gi`/Gtk version pinning and swallow exceptions itself.

Separately, the three platform trays (`sabtraylinux.py`, `sabtraywin.py`, `macosmenu.py`) each reimplement the same queue-state derivation and the same one-line action handlers. They have already drifted: when paused with bytes left, Windows shows `"Paused - Remaining: X"` ([sabtraywin.py:188](sabnzbd/sabtraywin.py:188)) but Linux shows only `"Paused"` ([sabtraylinux.py:135](sabnzbd/sabtraylinux.py:135)). Opening a folder is written three different ways, one of which (`os.system('open "%s"' % path)` at [macosmenu.py:509](sabnzbd/macosmenu.py:509)) breaks on paths containing a double quote.

Two goals:
1. Linux: one abstract base plus a small subclass per backend, and a factory so `SABnzbd.py` needs no knowledge of the specifics.
2. All platforms: share the state derivation and action handlers that are currently triplicated.

### Deliberately not unified

The **icon/menu layer stays per-platform.** A single cross-platform `BaseStatusIcon` was considered and rejected — the three differ in run loop (`Gtk.main()` in our thread / manual `PumpWaitingMessages` / `NSRunLoop` with the object built by ObjC), menu model (rebuilt per right-click / static with integer command IDs / mutated every 3s with live submenus), icon format (`.svg` paths / preloaded `HICON` handles / `.tiff` template images plus a text title), and menu content (7 flat items / 3 submenus / warnings badge, live queue and history submenus, purge, speed-limit checkmarks, disk space in titles). macOS is decisive: `SABnzbdDelegate` is an `NSObject` whose methods must be ObjC selectors, instantiated by the NIB, and it carries app-lifecycle callbacks that aren't tray concerns. A shared base would be hooks every platform overrides.

**Notifications stay as they are.** Linux tray backends have no notification API — AppIndicator, XApp and `Gtk.StatusIcon` are icon+menu only. Linux desktop notifications already work via `send_notify_osd()` ([notifier.py:219](sabnzbd/notifier.py:219), notify2/libnotify, gated on `ntfosd_enable`). `notifier.py` is untouched.

## Part 1 — New `sabnzbd/trayactions.py`

Platform-neutral, imports only `sabnzbd` and stdlib. No GUI toolkit.

### `tray_status() -> tuple[bool, str, str]`

Returns `(paused, icon_state, tooltip)` from `api.fast_queue()` + `to_units()`. Single definition of the drifted logic — adopt the Windows variant (it shows remaining bytes while paused, which is strictly more useful):

- paused + bytes left -> `"Paused - Remaining: X"`, state `"pause"`
- paused -> `"Paused"`, state `"pause"`
- bytes left -> `"XB/s - Remaining: XB (time)"`, state `"green"`
- else -> `"Idle"`, state `"default"`

Translated strings are cached at module level as `sabtraywin.py:70` already does, since this is called on a tick.

State keys are `default`/`green`/`pause`, matching the existing Linux and Windows `SABICONS` dicts. macOS uses different keys (`idle`/`pause`/`clicked`) and only distinguishes paused; it maps the returned state itself.

### `open_folder(path: str)`

One implementation replacing all three. `subprocess.Popen` with an argument list per platform (`explorer`/`open`/`xdg-open`), no shell, wrapped in try/except. Fixes the quote-injection in the macOS version.

### `class TrayActions`

Plain mixin, no base class, so Windows and Linux tray classes can inherit it. Methods: `browse()`, `pause()`, `resume()`, `pauseresume()`, `pausefor(minutes)`, `rss()`, `restart_sab()`, `nologin()`, `defhost()`, `shutdown()`, `open_complete_folder()`.

Bodies are lifted from `sabtraywin.py:316-363` (the most complete set). `pauseresume()` reads `self.sabpaused`, which both tray classes already maintain. `shutdown()` keeps the Windows behaviour of running `sabnzbd.shutdown_program` on a separate thread — the tray stopping itself from inside its own event loop is a hazard on Linux too.

## Part 2 — `sabnzbd/sabtraylinux.py` restructure

### `StatusIconBase(TrayActions, Thread)` — abstract

Owns everything backend-independent:

- Class attrs `SABICONS` (state -> path) and `UPDATE_FREQ = 1000`, matching the module-constant style in `sabtraywin.py:50`.
- `run()`: wait for `sabnzbd.WEBUI_READY`, `self.setup()`, `GLib.timeout_add(self.UPDATE_FREQ, self.update)`, `Gtk.main()`.
- `update()`: calls `trayactions.tray_status()`, stores `self.sabpaused`, then `self.set_icon()` / `self.set_tooltip()`. Returns `True` to stay scheduled.
- `stop()`: `GLib.idle_add(Gtk.main_quit)`.
- `create_menu()`, `right_click_event()`, and the Gtk-specific `addnzb()` file chooser. Every other handler comes from `TrayActions`.

Abstract surface per backend — three methods and one classmethod:

| member | purpose |
|---|---|
| `is_available()` (classmethod) | guarded `gi.require_version` + import, return bool. Preserves the "only load fallback backends when needed" behaviour from `ab0920b7e`. |
| `setup()` | construct the native icon, set name/visibility, attach menu or right-click signal |
| `set_icon(path)` | `set_icon_full` / `set_icon_name` / `set_from_file` |
| `set_tooltip(text)` | `set_title` / `set_tooltip_text` |

### Backends

`AyatanaAppIndicatorIcon`, `XAppStatusIcon`, `GtkStatusIcon` — each roughly 15 lines, importing inside `is_available()` and stashing the module on the class.

```python
BACKENDS = (AyatanaAppIndicatorIcon, XAppStatusIcon, GtkStatusIcon)
```

Behavioural differences to preserve: AppIndicator sets a persistent menu via `set_menu()` and uses `set_title()` for the tooltip; XApp and Gtk set a name, `set_visible(True)`, and build the menu on demand from a signal (`activate` vs `popup-menu`).

### `start_tray() -> Optional[StatusIconBase]` — the only public entry point

Iterates `BACKENDS`, returns the first whose `is_available()` passes (instantiated and started), logging the choice, or `None` with `logging.info` if none work. Each attempt is wrapped so a failing backend falls through to the next rather than killing startup.

The `import gi` / `require_version("Gtk", "3.0")` / `from gi.repository import Gtk, GLib` block moves to module top under `try/except`, setting `Gtk = GLib = None` on failure; `start_tray()` returns `None` early in that case. Classes only touch `Gtk`/`GLib` at call time, so this is safe.

## Part 3 — Callers

`SABnzbd.py:1175` collapses to:

```python
elif sabnzbd.LINUX_POWER and os.environ.get("DISPLAY"):
    import sabnzbd.sabtraylinux

    sabnzbd.LINUXTRAY = sabnzbd.sabtraylinux.start_tray()
```

- `sabnzbd/__init__.py:171`: add `LINUXTRAY = None` beside `WINTRAY = None`.
- `sabnzbd/__init__.py:349`: extend the existing tray-stop block in `halt()` to also stop `LINUXTRAY`.

`sabtraywin.py`: `SABTrayThread(TrayActions, Thread)`; delete the handler bodies now in the mixin; `do_updates()` calls `tray_status()`. Its `menu_options` tuple, win32 plumbing and `send_notification()` are unchanged.

`macosmenu.py`: cannot inherit the mixin (PyObjC metaclass, and its methods must be `name_(sender)` selectors), so each selector becomes a one-line delegation, e.g. `rssAction_` -> `trayactions.rss()`, `openFolderAction_` -> `trayactions.open_folder(sender.representedObject())`. Module-level functions are used rather than the mixin. `stateUpdate()` keeps its bespoke menu-bar title formatting — that is genuinely macOS-only — but sources `paused`/`bytes_left` from the shared helper.

## Incidental fixes rolled in

- `StatusIcon` subclasses `Thread` but never uses it: `__init__` builds a second `self.mythread` and `dowork()` is the real body. Becomes a proper `Thread` subclass with `run()`.
- `restart()` and `shutdown()` in `sabtraylinux.py` assign `self.hover_text`, an attribute this class doesn't have (copied from the Windows tray). Dead; removed.
- `SABICONS` uses `abspath("icons/...")`, silently depending on the `os.chdir(sabnzbd.DIR_PROG)` at `SABnzbd.py:1162`. Build from `sabnzbd.DIR_PROG` instead, as `send_notify_osd` does at `notifier.py:226`.

## Files

- `sabnzbd/trayactions.py` — new, shared
- `sabnzbd/sabtraylinux.py` — backend restructure + mixin
- `sabnzbd/sabtraywin.py` — adopt mixin and `tray_status()`
- `sabnzbd/macosmenu.py` — selectors delegate to the shared functions
- `SABnzbd.py`, `sabnzbd/__init__.py` — factory call, `LINUXTRAY` global, `halt()`

## Verification

No test suite covers the trays (they need a live desktop session), so this is manual. Order matters: shared core first, then per platform.

1. `python3 -m py_compile` the five files; `black --check` per repo style.
2. **Linux/GNOME** with `gir1.2-ayatanaappindicator3-0.1` installed: start with `tray_icon` enabled, confirm the debug log names the AppIndicator backend, the icon cycles default -> green while downloading -> gray when paused, and the tooltip now shows remaining bytes while paused. Exercise every menu entry, including Add NZB and Open complete folder.
3. **Fallback chain**: temporarily reorder `BACKENDS` (or test on Cinnamon/Mint for XApp) and repeat the icon/tooltip/menu checks for `XAppStatusIcon` and `GtkStatusIcon`.
4. **Graceful degradation**: run with `gi` unavailable or no tray libs present; SABnzbd must still start, logging that no backend was found.
5. **Windows**: confirm the tray icon, tooltip text, both submenus, left-click pause/resume, double-click browse, and that balloon notifications still fire via `notifier.send_windows`.
6. **macOS**: confirm the menu-bar title formatting is unchanged, queue/history submenus still populate, speed-limit checkmarks track, and Complete/Incomplete Folder open correctly — including from a path containing a double quote, which is the bug being fixed.
7. **Shutdown** on each platform: trigger restart and shutdown from the tray menu, verify a clean exit with no lingering GTK/message-pump thread.
  
</details>

Safihre:
@Amajiro will you address the requests?
@mnightingale could also just merge and we do it ourselves afterwards.

Amajiro:
Honestly, I think a full refactor is a bit beyond me. I'd only vibe code it and I think your wonderful app deserves a little more TLC than that. 

My fix addresses a small specific problem with wayland - I'll leave any larger work to greater minds than mine, if that's okay.

Safihre:
Sounds good, appreciate the honesty and willingness to help us add features we might have overlooked 😊

mnightingale:
I setup a fresh Mint VM and tested AyatanaAppIndicator3, XApp, and Gtk.

Also verified the reported issue #3508 on a fresh Ubuntu 26.04 install, Gtk-CRITICAL every second.

However, switching to this branch, I still get the same critical log.
```
2026-07-27 11:02:47,678::DEBUG::[sabtraylinux:36] AyatanaAppIndicator3 not available
2026-07-27 11:02:47,678::DEBUG::[sabtraylinux:48] XApp not available, falling back to Gtk.StatusIcon
```
I’m not familiar with Linux desktop distros, do I need to install something extra to get it to work on a fresh Ubuntu install?
I think we should test if the Gtk method is working, rather than the spam of Gtk-CRITCAL

The other side of that is do we need to update the install from source instructions at all or changes to how we package it (snaps, etc.)

Once resolved I think merge this for now, and I’ll add it to the list of things to cleanup sometime.

Steps I think I used along with adding the optional dependencies in requirements.txt

```
sudo add-apt-repository ppa:jcfp/nobetas
sudo apt-get update
sudo apt-get install par2-turbo unrar python3.14-venv build-essential cmake libcairo2-dev pkg-config python3-dev libgirepository-2.0-dev libdbus-1-dev

cd sabnzbd
python3 -m venv .venv
source .venv/bin/activate
pip install -r requirements.txt
python SABnzbd.py
```

jcfp:
@mnightingale you also need to install the `gir1.2-ayatanaappindicator3-0.1` package to test the changes in this PR.

mnightingale:
Thanks that has sorted it, I think we can merge this and sort a few problems out later.

Since I had to install extra things to get it working from source, do we need to do anything with the ways we distribute it (from repo, snap, flatpak) to make the tray icon work?

jcfp:
Release notes and documentation about running from source on the sab end should be updated to mention the newly added optional requirement. Package maintainers for distros should add the necessary package dependencies. For sandboxes such as Flatpak and Snap, tray icons have been difficult/impossible to get to work AFAIK.

Safihre:
@mnightingale also want to submit a PR for docs?
Still.km vacation so can't help at the moment 😅

mnightingale:
I'm not sure I'm best suited to provide such information, I just grabbed an Ubuntu based distro and it supported them all.

Maybe it should have some extra information on https://sabnzbd.org/wiki/installation/install-unix about tray requirements. I think tray support should be available for installation from repositories or Flatpak.

https://sabnzbd.org/wiki/installation/install-off-modules could also do with some updates

jcfp:
I can go over the docs, no earlier than the weekend though

jcfp:
Also found a way to possibly get the tray icon to work with the flatpak, will try once this code is part of a final release.

Safihre:
Could it work with Snap?

jcfp:
> Could it work with Snap?

Not sure, there are snapcraft.yaml out there with a `libappindicator3-1` (plus a bunch of gtk/glib stuff) in stage-packages, suggesting there is some level of support.

Safihre:
Will you give it a try? 😇

jcfp:
> Also found a way to possibly get the tray icon to work with the flatpak, will try once this code is part of a final release.

Got this to work with 5.1.0rc2, still need to tidy things up but I expect appindicator support to make it into the flatpak once there's a final release that supports it. Will take a look at the snap too, don't expect quick results though as real life keeps me plenty busy.

PR #3511, opened 2026-07-20
Improve Add NZB form accessibility

coopa11y:
The visible labels in the Add NZB dialog were not associated with their form controls, so screen readers could announce some fields without a useful name.

This associates the existing translated labels with the name, password, category, priority, processing, and script fields. It does not change the dialog's layout, translations, or existing mouse and keyboard behavior.

Testing:
- Verified each label points to the intended form control.
- Verified the new control IDs are unique.
- Confirmed the existing field names and bindings remain unchanged.

### DIFF
--- a/interfaces/Glitter/templates/include_overlays.tmpl
+++ b/interfaces/Glitter/templates/include_overlays.tmpl
@@ -524,29 +524,29 @@
                 <div class="clearfix"></div>
                 <hr />
                 <div class="form-group">
-                    <label class="col-sm-4 control-label">$T('name')</label>
+                    <label class="col-sm-4 control-label" for="nzbname">$T('name')</label>
                     <div class="col-sm-6">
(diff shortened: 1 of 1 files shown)


thezoggy:
can see the naming scheme being used by coopa11y is this-that while we have some things using thisThat notation. do we want to favor one/make the style guideline decision for going forward?

coopa11y:
I followed the hyphenated naming already used by the Add NZB modal and other Glitter modal elements. This template currently has a mix of hyphenated, camelCase, and underscore-separated IDs. I'm happy to update these four IDs if there is a preferred convention going forward.

coopa11y:
@thezoggy After looking through the surrounding template, my preference would be to use hyphenated names for HTML IDs. That keeps these four IDs consistent with nearby names such as `modal-add-nzb` and `add-local-nzb`, while camelCase remains useful for JavaScript properties. I would leave the existing mixed IDs alone in this PR and use hyphenated names for new HTML IDs unless the maintainers prefer a different convention.

PR #3512, opened 2026-07-23
feat: add 'Sort by Remaining Size' queue sort option

QuixThe2nd:
## Summary

Adds a new `remaining_bytes` sort field that sorts the queue by **absolute remaining bytes** (`bytes - bytes_tried`).

This complements the existing options:
- `size` — sorts by total download size
- `remaining` — sorts by percentage downloaded

The new option lets users sort by how much data is actually left to download, which is useful when the queue contains a mix of large partially downloaded jobs and small fresh ones.

The UI and automatic-sort setting expose only ascending order (smallest to largest). Descending automatic sorting is intentionally omitted because periodic re-sorts would make actively downloading jobs flap between positions.

## Changes

| File | Change |
|------|--------|
| `sabnzbd/nzbqueue.py` | Add the `remaining_bytes` field, route `remaining_bytes asc` auto-sort without changing existing percentage-sort behavior, and log non-empty periodic sorts at debug level |
| `sabnzbd/skintext.py` | Add the ascending Remaining Size label and update the auto-sort explanation |
| `interfaces/Glitter/.../include_queue.tmpl` | Add the ascending Remaining Size dropdown entry |
| `interfaces/Glitter/.../glitter.queue.js` | Add the `sortRemainingBytesAsc` handler |
| `interfaces/Config/.../config_switches.tmpl` | Add the ascending `remaining_bytes` auto-sort option |
| `tests/test_functional_api.py` | Test ascending `remaining_bytes` sorting against the `sizeleft` slot field |

## Auto-sort behavior

`update_sort_order()` handles `remaining_bytes asc` explicitly and calls `sort_queue("remaining_bytes")` without passing a direction. Existing percentage-based `remaining` behavior remains unchanged and likewise ignores direction, preventing periodic sort flapping.

## Test Plan

- [x] Functional queue-sort test (`7 passed`)
- [x] Queue unit suite (`6 passed, 1 skipped`)
- [x] Black, Ruff, and `git diff --check`
- [ ] Manual: sort the Glitter queue by Remaining Size and confirm smallest-to-largest order
- [ ] Manual: set auto-sort to `remaining_bytes asc` and confirm the periodic resort

### DIFF
--- a/interfaces/Config/templates/config_switches.tmpl
+++ b/interfaces/Config/templates/config_switches.tmpl
@@ -152,6 +152,8 @@
                         <select name="auto_sort" id="auto_sort">
                             <option value="">$T('default')</option>
                             <option value="remaining asc" <!--#if $auto_sort == "remaining asc" then 'selected="selected"' else ""#--> >$T('Glitter-sortRemaining')</option>
+                            <option value="remaining_bytes asc" <!--#if $auto_sort == "remaining_bytes asc" then 'selected="selected"' else ""#--> >$T('Glitter-sortRemainingBytesAsc')</option>
+                            <option value="remaining_bytes desc" <!--#if $auto_sort == "remaining_bytes desc" then 'selected="selected"' else ""#--> >$T('Glitter-sortRemainingBytesDesc')</option>
                             <option value="avg_age desc" <!--#if $auto_sort == "avg_age desc" then 'selected="selected"' else ""#--> >$T('Glitter-sortAgeAsc')</option>
                             <option value="avg_age asc" <!--#if $auto_sort == "avg_age asc" then 'selected="selected"' else ""#--> >$T('Glitter-sortAgeDesc')</option>
                             <option value="name asc" <!--#if $auto_sort == "name asc" then 'selected="selected"' else ""#--> >$T('Glitter-sortNameAsc')</option>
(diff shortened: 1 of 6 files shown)


mnightingale:
On the surface I don’t think sort by remaining bytes is a bad idea; however, I think there is probably a reason the functionality doesn’t already exist because it seems so obvious.

My main concern would be the queue order flapping; if sorted by remaining bytes largest to smallest:

With an initial queue sort:
JobA: remaining =20,000 
JobB: remaining =10,000

But once JobA has downloaded more than 10,000, JobB jumps to the front, then every time the periodic sort triggers, they may switch places. The result is every job converging on equal `remaining` and completing together.

I don’t think this problem affects ASC sort (smallest to largest), which would probably be the most beneficial anyway.

For similar reasons `remaining asc` for percentage sort is not a handled option (direction is ignored), and `size desc` is available to sort by total size which won’t have the flapping problem.

`field, direction = auto_sort.split()`could raise but needs removing anyway; it's intentional that the direction of `remaining asc` is ignored, it would have the same flapping problem
` logging.info` should be debug like the other paths - edit: also `if self.__nzo_list:` to avoid log spam when empty

I would suggest just adding a `remaining_bytes` (asc / smallest to largest) and remove the other breaking changes.

QuixThe2nd:
Thanks, updated as suggested:

- kept only `remaining_bytes asc` (smallest to largest);
- removed the descending option and the generic `auto_sort.split()` change;
- preserved the existing percentage-based `remaining` periodic-sort behavior, including intentionally ignoring direction;
- moved the remaining-size sort log to `debug` and suppressed it when the queue is empty.

The focused queue/API tests pass, as do Black and Ruff.

PR #3518, opened 2026-07-24
Improve Glitter message accessibility

coopa11y:
The warnings and messages table did not have an accessible name, and its icon-only clear actions were unlabeled.

This names the table and message column using the existing Warnings translation, labels both clear actions using the existing Clear translation, and hides their decorative icons from assistive technology. It does not change the existing layout or behavior.

Testing:
- Verified the warnings table, message heading, and both clear actions have the intended accessible names.


Safihre:
I'm currently on vacation so I can't check the changes. Maybe if @mnightingale had a chance to check if warnings display is still fine? Just double checking because some html was changed.

mnightingale (approved):
No issues with warnings, the extra html for `sr-only`uses a class from bootstrap and is not visible as intended. The remove/dismiss button still works.

Issue #3507, opened 2026-07-19
DirectUnpack hangs indefinitely (zombie unrar process, orphaned _UNPACK_ folder) — still reproducible on 5.0.4


jamjop:
Related to #2533 (closed by stale-bot, never reproduced/fixed) and #2948 (different trigger — disk full — but same symptom class).

**SABnzbd version:** 5.0.4, Docker (linuxserver/sabnzbd image)
**OS:** Ubuntu (host), container runtime gVisor (runsc) for sandboxing

**Description:**
DirectUnpack occasionally leaves a job stuck indefinitely: the `unrar -vp` process for a job keeps running but stops making progress (output directory stops growing), and the job never completes or fails — it just sits in the queue forever. No disk-full condition, and nothing obviously wrong in the UI or logs at the time it happens. Matches the 'zombie DirectUnpacker' behavior described in #2533, not the disk-full trigger in #2948.

**Workaround in place:** a script that kills any `unrar -vp` process running >2 hours with a non-growing output directory for >30 minutes. This fires a few times a month, so the underlying hang is still a live issue on current stable, not just the 4.0 betas discussed in #2533. The killed job then leaves an orphaned `_UNPACK_*` folder behind that SABnzbd never revisits or cleans up (also worth fixing on its own — currently cleaned up by a separate script on our end).

Happy to enable `--log-all` and capture a full repro if useful.

Safihre:
Indeed, if you could get Debug logging (don't need log all I think, it's very verbose). 
Enable Debug logging in the Status window and then after it happens again click Show Logging there, since it will anonymise the logs and include your settings.
First kill unrar, so the unrar logs are included.

mnightingale:
Difficult to be sure without logs but I have found a few problems that can cause the direct unpacker to get stuck, will send a PR.

Is it reproducible with the same NZB? i.e. adding the same NZB causes direct unpacker to get stuck again?

mnightingale:
Hey @jamjop two things would be useful for diagnosing this.

First, as @Safihre suggested please enable Debug logging on the "Status and interface options" screen.
Then the next time it happens (at least a few minutes after) "Show Logging" from that same screen will create a log with sensitive information removed, it might help us find what's happening.

Secondly if able it may be useful to dump the threads so we can see what it's up to:

The host host, as root

Find the process id, in this case 645565

```bash
ps aux | grep SABnzbd

[...]

PR #3521, opened 2026-07-25
DirectUnpack: reading output no longer blocks assembly, dropped aborts, slow failure on missing volumes

mnightingale:
Possibly related to #3507 but I don't think it explains their symptoms (need dump or at least logs) most of the changes are in tests or comments explaining what's going on.

Claude incoming.

Investigation started from a report of DirectUnpack occasionally leaving a job stuck: the `unrar -vp` process stays alive but stops making progress, and the job sits in the queue without ever completing or failing.

**That report is not explained by anything here.** The state it describes — unrar alive, output directory not growing — is what `unrar -vp` looks like while parked at its `[C]ontinue, [Q]uit ` prompt, and reaching that prompt means unrar produced output, which `run()` read and acted on outside the lock. So the unpacker holds nothing while it waits, and the job staying in the *queue* rather than moving to history points at the download not finishing, with the parked unrar as a symptom rather than the cause. Diagnosing it properly needs a thread dump from a running instance (`py-spy dump`) and confirmation of whether other jobs kept downloading at the same time.

What the investigation did turn up is the set of real defects below, found by reading the code around the reported behaviour. (1) can produce exactly this symptom, but only in a narrower situation than the report describes, and is worth fixing on its own merits either way.

---

## 1. Reading unrar's output blocked the Assembler thread

**When it happens.** `run()` held the module-wide `START_STOP_LOCK` across a blocking `stdout.read(1)`. `add()` is `@synchronized` on that same lock and is called from the single Assembler thread, so no job in the queue could assemble for as long as unrar produced no output. With `-idp` unrar is silent for the whole of each volume it decompresses, so this is a continuous throughput drag proportional to the number of active unpackers, not a rare event.

It becomes a permanent freeze only if unrar stays alive and never produces output again — wedged in uninterruptible I/O on an unresponsive disk, rather than merely idle. Then assembly never resumes for any job, no job reaches post-processing, and neither of the two things that can release a DirectUnpacker (`rar_unpack()` or `purge_data()` → `abort_direct_unpacker()`) is reached. `abort()` and `create_unrar_instance()` need the same lock, so the 2-minute watchdog in `rar_unpack()` cannot help either. That case is the whole queue stalling at once, not one job.

**Why this fixes it.** The read never blocks, so no caller ever waits behind it. `build_and_run_command` already sets `bufsize=0`, so `stdout` is a raw `FileIO` and both platforms can read exactly what is already available, leaving the stream in blocking mode:

```python
if sabnzbd.WINDOWS:
    handle = msvcrt.get_osfhandle(self.active_instance.stdout.fileno())
    if available := win32pipe.PeekNamedPipe(handle, 0)[1]:
        return self.active_instance.stdout.read(available)
    time.sleep(WINDOWS_POLL_INTERVAL)
elif select.select([self.active_instance.stdout], [], [], OUTPUT_TIMEOUT)[0]:
    return self.active_instance.stdout.read(OUTPUT_CHUNK_SIZE)
```

`select` cannot take a pipe on Windows and `os.set_blocking()` only supports Windows pipes from 3.12, so Windows peeks the pipe through pywin32, which is already a dependency there. Both branches also re-check `killed`, so an instance that was killed without its pipe reaching end-of-file no longer leaves `run()` waiting.

That also allows the char-by-char loop to go: `run()` appends each chunk to the buffer, takes whole lines off it with `split(b"\n", 1)`, and checks whatever is left for the prompts, which is where the newline-less new-disk message lives.

## 2. Slow DirectUnpack work blocked unrelated jobs

**When it happens.** `START_STOP_LOCK` was module-level and shared by every `DirectUnpacker`, while everything it guards is per-instance state. `abort()` holds it across `kill()`, two 0.2s sleeps and `remove_all(extraction_path, recursive=True)` — seconds on a slow or network disk. `reset_active()` holds it across `wait(timeout=2)`. With `direct_unpack_threads` defaulting to 3, all of that serialised against the Assembler thread.

**Why this fixes it.** The lock becomes `self.lock`, using the existing no-argument `@synchronized()` convention (`NzbObject`/`NzbFile` already do this). `ACTIVE_UNPACKERS` keeps a small dedicated `ACTIVE_UNPACKERS_LOCK`, held across the `direct_unpack_threads` check *and* `create_unrar_instance()` so two callers can never claim the same slot — a plain lock around `append`/`remove` would add nothing, since those are already atomic. Lock order is instance lock → `ACTIVE_UNPACKERS_LOCK`, never the reverse; `abort_all()` therefore iterates a copy taken under the lock, which also fixes it silently skipping entries when the list mutates mid-iteration.

## 3. `abort()` was ignored between sets

**When it happens.** `abort()` required `self.cur_setname`, but `run()` clears it in `reset_active()` before starting the next set, and `reset_active()` can sit in `active_instance.wait(timeout=2)`. An abort arriving in that window returned having done nothing — `killed` stayed `False`, unrar was not killed, files were not removed — and `run()` then went on to unpack the next set of a job being purged or removed from the queue.

**Why this fixes it.** `abort()` now keys only on `self.killed`, so it is effective in that window, and logs the job name when there is no set name yet.

## 4. `abort()` could raise `AttributeError` at the caller

**When it happens.** In the `one_folder` cleanup the `except` handler itself dereferenced `rarfile_nzf.filename`, so with no rarfile the error escaped `abort()` into callers such as `NzbQueue.remove()` → `purge_data()`, where nothing catches it. Reachable today: `run()`'s next-set path sets `cur_setname` and then waits in `wait_for_next_volume()` before `create_unrar_instance()` assigns `rarfile_nzf`.

**Why this fixes it.** Both the call and the handler are guarded. The files are left in place rather than falling through to `remove_all()` — with `one_folder` the extraction path is shared with the user's other files.

## 5. A missing volume failed slowly and reported a volume that was never unpacked

**When it happens.** With volume 2 missing but 3 present, the unpack waited until post-processing, then `run()` incremented `cur_volume` speculatively before checking anything. That made `have_next_volume()` match volume *3*, disarming the `not have_next_volume()` half of the repeating-prompt guard, so it took 11 identical prompts to give up and the UI showed `02/NN` for a volume that never existed.

**Why this fixes it.** Post-processing now tells the unpacker explicitly, via `set_no_more_files()`, immediately after `parring()` — the point where repair is done and nothing new can arrive. `run()` acts on it directly after the wait:

```python
if self.no_more_files and not self.have_next_volume():
    logging.info("DirectUnpack failed due to missing files %s", self.cur_setname)
    self.abort()
```

`cur_volume` is still correct at that point, so the check looks for the volume actually being waited on. `wait_for_next_volume()` now waits on `no_more_files` instead of `nzo.pp_active`, which was never a safe signal: `pp_active` is set at the start of post-processing, *before* `parring()`, so acting on it would feed unrar an unrepaired volume. `rar_unpack()` calls `set_no_more_files()` too, as an idempotent fallback that states the same thing.

Trade-off: `have_next_volume()` is now the authority once repair finishes, so a volume par2 rebuilt that has no matching assembled `NzbFile` aborts rather than continuing. That was already the behaviour whenever no later volume existed; it only survived by accident when a later volume made the check pass after the phantom increment. Renames are covered, since `verify_all_filenames_and_resort()` re-runs `set_volumes_for_nzo()`.

## 6. Minor: post-processing noticed completion up to 2s late

`rar_unpack()` polled with `time.sleep(2)`. `DirectUnpacker` is a `Thread`, so it now uses `join(timeout=2)` and returns the moment the unpack is done. The `wait_count > 60` stuck-detection and the periodic bump of the file-lock are both unchanged — iterations where the thread is alive still take the full 2s. `set_no_more_files()` is sent once before the loop, since it is a state change rather than a nudge.

---

## Deliberately not changed

- **`add()` still returns without notifying when the download is damaged.** This looks like a lost wakeup but is load-bearing: it parks the unpacker on the volume it needs so `rar_unpack()` can resume it after par2 repaired the damage. Waking it would feed unrar a damaged volume, and the resulting `abort()` would `remove_all()` everything unpacked so far. Comments added at both ends to record this.
- **No timeout on `wait_for_next_volume()`.** For the same reason — a periodic re-check would release the wait before repair. Every exit is covered: `add()`'s notify, `set_no_more_files()`, and `purge_data()` → `abort_direct_unpacker()`, which runs on every terminal outcome of `process_job()`.

## Still open

`rar_unpack()`'s watchdog aborts after two minutes without progress but then keeps looping on `while nzo.direct_unpacker.is_alive():`, calling `abort()` every two seconds — and `abort()` is a no-op once `killed` is set. An unpacker that will not stop therefore holds the single post-processing thread forever. Left alone here because stopping the wait means deciding what to do next, and extracting into a directory a surviving unrar may still be writing to is worse than waiting.

## Testing

New `tests/test_directunpacker.py`, 13 tests, using a real `os.pipe()` for unrar's output so the polling and `select` paths are actually exercised. Each fix has a test verified to fail against the pre-change implementation:

- reading output does not block `add()`
- all output is picked up whether it arrives a byte at a time or in one chunk, including the prompt that has no newline after it
- claiming a `direct_unpack_threads` slot is atomic, and unpackers do not share a lock
- abort works between sets, and without a rarfile leaves the shared folder alone
- a missing volume aborts as soon as repair is done, with and without a later volume present
- `add()` sends no notification while the download is damaged, and does send one otherwise

### DIFF
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -40,9 +40,10 @@ from sabnzbd.postproc import prepare_extraction_path
 from sabnzbd.misc import SABRarFile
 from sabnzbd.utils.diskspeed import diskspeedmeasure
 
-# Only ever appended to from add(), which runs on the single Assembler thread, and only
-# removed by the unpacker itself. So the direct_unpack_threads limit needs no extra lock,
-# at worst a slot is freed just after we decided we had to wait for one.
+# Unpackers that have a live unrar instance. Held while checking the direct_unpack_threads
+# limit and claiming a slot, so that two callers of add() can never take the same one.
+# Always taken after the lock of an individual unpacker, never before it.
+ACTIVE_UNPACKERS_LOCK = threading.RLock()
 ACTIVE_UNPACKERS = []
 
 RAR_NR = re.compile(r"(.*?)(\.part(\d*).rar|\.r(\d*))$", re.IGNORECASE)
@@ -154,13 +155,18 @@ class DirectUnpacker(threading.Thread):
             logging.debug("DirectUnpack queued %s for %s", nzf.filename, self.cur_setname)
             # Is this the first one of the first set?
             if not self.active_instance and not self.is_alive() and self.have_next_volume():
-                # Too many runners already?
-                if len(ACTIVE_UNPACKERS) >= cfg.direct_unpack_threads():
-                    logging.info("Too many DirectUnpackers currently to start %s", self.cur_setname)
-                    return
+                # Claim a slot and create the instance in one go, so we can never hand the
+                # same slot to another caller of add()
+                with ACTIVE_UNPACKERS_LOCK:
+                    # Too many runners already?
+                    if len(ACTIVE_UNPACKERS) >= cfg.direct_unpack_threads():
+                        logging.info("Too many DirectUnpackers currently to start %s", self.cur_setname)
+                        return
 
-                # Start the unrar command and the loop
-                self.create_unrar_instance()
+                    # Start the unrar command
+                    self.create_unrar_instance()
+
+                # Start the loop
                 self.start()
         elif not any(test_nzf.setname == nzf.setname for test_nzf in self.next_sets):
             # Need to store this for the future, only once per set!
@@ -272,7 +278,10 @@ class DirectUnpacker(threading.Thread):
                     # Did we reach the end?
                     # Stop timer and finish
                     self.unpack_time += time.time() - start_time
-                    ACTIVE_UNPACKERS.remove(self)
+
+                    # Free the slot, there is no instance running until we start the next set
+                    with ACTIVE_UNPACKERS_LOCK:
+                        ACTIVE_UNPACKERS.remove(self)
 
                     # Take note of the correct password
                     if self.nzo.password and not self.nzo.correct_password:
@@ -397,8 +406,9 @@ class DirectUnpacker(threading.Thread):
             self.killed = True
             # Make more space
             self.reset_active()
-            if self in ACTIVE_UNPACKERS:
-                ACTIVE_UNPACKERS.remove(self)
+            with ACTIVE_UNPACKERS_LOCK:
+                if self in ACTIVE_UNPACKERS:
+                    ACTIVE_UNPACKERS.remove(self)
             logging.debug("Closing DirectUnpack for %s", self.nzo.final_name)
 
     def have_next_volume(self):
@@ -502,7 +512,8 @@ class DirectUnpacker(threading.Thread):
         ).start()
 
         # Add to runners
-        ACTIVE_UNPACKERS.append(self)
+        with ACTIVE_UNPACKERS_LOCK:
+            ACTIVE_UNPACKERS.append(self)
 
         # Doing the first
         logging.info("DirectUnpacked volume %s for %s", self.cur_volume, self.cur_setname)
@@ -603,7 +614,10 @@ def analyze_rar_filename(filename):
 def abort_all():
     """Abort all running DirectUnpackers"""
     logging.info("Aborting all DirectUnpackers")
-    for direct_unpacker in ACTIVE_UNPACKERS:
+    # Abort outside the lock, since abort() needs the lock of the unpacker itself
+    with ACTIVE_UNPACKERS_LOCK:
+        active_unpackers = ACTIVE_UNPACKERS.copy()
+    for direct_unpacker in active_unpackers:
         direct_unpacker.abort()
 
 
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -42,7 +42,6 @@ from sabnzbd.utils.diskspeed import diskspeedmeasure
 
 if sabnzbd.WINDOWS:
     import msvcrt
-    import pywintypes
     import win32pipe
 
 # Unpackers that have a live unrar instance. Held while checking the direct_unpack_threads
--- a/sabnzbd/directunpacker.py
+++ b/sabnzbd/directunpacker.py
@@ -20,6 +20,7 @@ sabnzbd.directunpacker
 """
 
 import os
+import queue
 import re
 import subprocess
 import time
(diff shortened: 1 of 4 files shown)


sabnzbd/directunpacker.py
```diff
@@ -165,6 +178,47 @@ def add(self, nzf: NzbFile):
         with self.next_file_lock:
             self.next_file_lock.notify()
 
+    def read_output(self, instance: subprocess.Popen, output_queue: queue.Queue[Optional[bytes]]):
```
Safihre: This really is the only way? Because read on the output has no timeout?

mnightingale: It's Windows fault.

Unix we could use `selectors/select.select`.
From Python 3.12 we can use `os.set_blocking(fd, False)` - Windows support for pipes added in 3.12
A maybe complicated split to use `msvcrt.get_osfhandle` + `win32pipe.PeekNamedPipe` via `pywin32` and `selectors/select.select` for other platforms.
Reader thread and queue seems the most straightforward way to support all platforms.

Safihre: Our windows releases use Python 3.14 so we can use this then?
This is extra complexity is just too much for a bug that was only reported now while we have direct unpack for years..

mnightingale: `select.select` and `win32pipe.PeekNamedPipe` isn't too bad.
Draft until I test it out on Windows a little.

mnightingale:
I don’t think the main problem found explains the users’ symptoms, or we at least need more information.
I worry that the symptoms that unrar is running but not processing are the outcome not the cause.

However, it and the other issues identified are worth fixing.

1. Needing module lock blocks other unpackers from progressing. Luckily unrar is quite chatty so this is probably mostly unnoticed unless extracting multiple jobs, one with large volumes and the other small
2. Reading 1 character at a time is wasteful – extra syscalls, reading whatever is available and splitting on `\n` is much better and still lets us read the trailing prompt
3. Changes to allow multiple assembler threads in the future - better SSD queuing
4. `abort()` between sets is a race, although it’s rare to have a job with multiple sets
5. `one_folder` `abort()`could raise AttributeError when logging a failure, rare
6. Missing volumes, noted in comment but never handled properly, ` cur_volume ` is incremented for the wrong volume then abort is after asking for the same missing volume 10 times 
```
# If lines did not change and we don't have the next volume, this download is missing files!
# In rare occasions we can get stuck forever with repeating lines
```
7. Postproc waiting for DirectUnpacker called `time.sleep(2)`in between polling for if it has finished, slower finished jobs vs `join(timeout=2)`

There is one are (6) that I think it could get stuck in and probably should be solved for the case where it’s stuck and we don’t know why. That postproc check every 2 seconds aborts after 2 minutes, but it doesn’t break out of the loop so we are trusting that `abort` can’t fail.

There is very little testing of DirectUnpacker, so the tests can at least protect against regressions.

Maybe leave it for after 5.1.0

mnightingale:
These are issues but I'm going to split them up and better explain why

PR #3522, opened 2026-07-25
Fix flaky test_api_queue_sort test and to_unit at boundaries

### LINKED ISSUE #3520: Flaky tests in TestQueueApi::test_api_queue_sort due to size sorting
### SABnzbd version

develop

### Operating system

macOS

### Using Docker image

None

### Description

A note for myself to fix this flaky test sometime.

I've looked at it before and it is an issue with `_create_random_queue` and file sizes close to unit barriers.
I suspect it decided `1024 KB` is equal to `1.0 MB` so the order depends on the queue order or something.
But the main issue is probably how it decides they are not both `1.0 MB`.

```
        # Verify the result
>       assert new_order == original_order
E       AssertionError: assert ['798 KB', '8...'1.2 MB', ...] == ['798 KB', '8...'1.2 MB', ...]
E         
E         At index 3 diff: '1024 KB' != '1.0 MB'
E         
E         Full diff:
E           [
E               '798 KB',
E               '877 KB',
E               '1000 KB',
E         +     '1024 KB',
E               '1.0 MB',
E         -     '1024 KB',
E               '1.2 MB',
E               '1.3 MB',
E               '1.7 MB',
E           ]
```

```
FAILED tests/test_functional_api.py::TestQueueApi::test_api_queue_sort[size-size-asc] - AssertionError: assert ['798 KB', '8...'1.2 MB', ...] == ['798 KB', '8...'1.2 MB', ...]
  
  At index 3 diff: '1024 KB' != '1.0 MB'
  
  Full diff:
    [
        '798 KB',
        '877 KB',
        '1000 KB',
  +     '1024 KB',
        '1.0 MB',
  -     '1024 KB',
        '1.2 MB',
        '1.3 MB',
        '1.7 MB',
    ]
```

mnightingale:
Fixes #3520

Verified my hypothesis, found it's two issues.

- The api sorts by `nzo.bytes` exact value, the test maps lower resolution units back to bytes. Fix is to compare at the same resolution
- `to_units` outputting `1024 KB` instead of `1.0 MB`, calculation is `trunc(log2(val)/10)` but value is then printed at reduced precision. Fixed by applying the same formatting as the output.

bytes | before | after
-- | -- | --
1048575 | 1024 K | 1.0 M
1073741823 | 1024.0 M | 1.0 G
1099511627775 | 1024.0 G | 1.0 T

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -653,25 +653,26 @@ def to_units(val: int | float, postfix="") -> str:
         # Limit it to 5 as the maximum defined index.
         n = min(5, math.trunc(math.log2(val) / 10))
 
-    # Now we scale our value to the appropriate power of 1024
-    # It is written as 2^10n for symmetry with the
-    # selection above.
-    val = val / 2 ** (10 * n)
-
     # Showing the single decimal per doc string
     if n > 1:
         decimals = 1
     else:
         decimals = 0
 
-    # Rounding to the displayed precision can carry the value up to the next unit,
-    # for example 1048575 would be shown as "1024 K" instead of "1.0 M". Move it to
-    # the next unit instead, unless we are already at the maximum defined index.
-    if n < 5 and float(f"{val:.{decimals}f}") >= 1024:
+    # Now we scale our value to the appropriate power of 1024
+    # It is written as 2^10n for symmetry with the
+    # selection above. Round it to the precision we are going to display, so
+    # what we check below is what ends up in the output.
+    val = round(val / 2 ** (10 * n), decimals)
+
+    # That rounding can carry the value up into the next unit, for example 1048575
+    # would be shown as "1024 K" instead of "1.0 M". Move it up a unit instead,
+    # unless we are already at the maximum defined index.
+    if n < 5 and val >= 1024:
         n += 1
-        val = val / 1024
         if n > 1:
             decimals = 1
+        val = round(val / 1024, decimals)
 
     # We might not have anything at all to append
     if n == 0 and postfix == "":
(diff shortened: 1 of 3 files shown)


sabnzbd/misc.py
Safihre, line 654: This performance related log change probably introduced this. Now I wonder with the extra check if the old boring if-statemenys way isn't actually faster again?

mnightingale: Yes https://github.com/sabnzbd/sabnzbd/pull/3076 caused the regression, but it also improved accuracy without the looping.

bytes | pre-#3076 | after #3076 | with my fix
-- | -- | -- | --
1047961 | 1.0 M | 1023 K | 1023 K
1048505 | 1.0 M | 1024 K | 1.0 M
1048575 | 1.0 M | 1024 K | 1.0 M
1073741823 | 1.0 G | 1024.0 M | 1.0 G
1099511627775 | 1.0 T | 1024.0 G | 1.0 T
1024**6 | 1024.0 P | 1024.0 P | 1024.0 P

> Reverting just the index computation back to the loop would fix the flakiness, but the old loop was over-eager in the other direction: it promoted on > 1023 in the scaled value, so 1047961 bytes printed as 1.0 M even though it rounds to 1023 K. Over a 400k-value random sample, pre-[#3076](https://github.com/mnightingale/sabnzbd/issues/3076) disagrees with the current fix 372 times; #3076-as-shipped disagrees 17 times. Every one of the old code's extra disagreements is it promoting a value that shouldn't be — losing precision that #3076 legitimately gained.
>
> The fix I applied keeps the accuracy #3076 introduced and restores the invariant the loop gave for free: promote only when the value would actually print as 1024 at its display precision. It's a superset of both, not a compromise between them.

Safihre:
@claude review and suggest if fastest approach

PR #3524, opened 2026-07-26
Label SABnzbd login fields and help link for screen readers

coopa11y:
The login form relied on placeholder text to identify the username and password fields, and its icon-only help link did not have an accessible name.

This adds visually hidden labels associated with both login fields and names the help link using the existing Help translation. The decorative help icon is hidden from assistive technology. It does not change the existing layout or behavior.

Testing:
- Verified both labels are associated with their intended fields and the help link has the translated accessible name.

### DIFF
--- a/interfaces/Config/templates/login/main.tmpl
+++ b/interfaces/Config/templates/login/main.tmpl
@@ -37,10 +37,8 @@
                 <div class="alert alert-danger" role="alert">$error</div>
                 <!--#end if#-->
 
-                <label class="sr-only" for="login-username">$T('srv-username')</label>
-                <input type="text" class="form-control" id="login-username" name="username" placeholder="$T('srv-username')" autocomplete="username" required autofocus>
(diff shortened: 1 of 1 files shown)


mnightingale:
Would it be acceptable to use aria-label on the inputs, rather than adding invisible labels?

Reference: https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA14

Or for accessibility reasons is it preferable to have the separate label + input tags?

coopa11y:
@mnightingale Yes, `aria-label` should be fine here, I originally used visually hidden labels because they provide a native association, but these fields have simple, stable names and `aria-label` gives screen readers what they need without changing the design.

I?ve updated the username and password inputs and removed the extra hidden labels and IDs. The placeholders can stay as visual cues, while `aria-label` keeps the fields properly named after someone starts typing, thanks for the suggestion and the link.

PR #3525, opened 2026-07-26
Label schedule time, enable, and delete controls for screen readers

coopa11y:
The Scheduling configuration page did not distinguish the hour and minute selections for screen readers. Existing schedules also used unnamed enabled checkboxes and icon-only delete buttons, without identifying which schedule each control affected.

This gives the hour and minute selections translated accessible names. Each enabled checkbox and delete button is associated with its schedule's visible time, frequency, and action. The delete icon is hidden from assistive technology, and an unused duplicate ID is removed.

It does not change the existing layout or behavior and uses existing translated text.

Testing:
- Verified the control labels, unique IDs, and schedule descriptions.
- Compiled the updated Cheetah template.


Safihre:
Could you use aria-label here also?

coopa11y:
@Safihre Sure, I?ve switched the enabled checkbox to use `aria-label` and removed the hidden label and ID. The existing `aria-describedby` is still there so the checkbox keeps the context for the specific schedule.

PR #3527, opened 2026-07-27
Reduce BPS allocations and fix scale

mnightingale:
Noticed this while tracing allocations and I love a good deque.

`self.bps_list[-BPS_LIST_MAX:]`allocates a new 275 element list every second
`extend([0] * nr_diffs)` creates a temporary list to extend

Deque can do the 275 limited append in C without allocating.

---

The other changes are to initially populate the bps_list with 275 zeros, this fixes the x-axis scale where before SAB has been running for 275 seconds the speed is stretched over the whole sparkline.

Before:

https://github.com/user-attachments/assets/9387d835-c1c6-422b-9226-e229c02dc78b

After:

https://github.com/user-attachments/assets/7c2b77d4-1427-4dec-bda8-bc35b877b2a7

---

Set `self.speed_log_time = t` in `add_empty_time` fixes an issue that the graph moves left too quick if the page is refreshed lots while idle; download a job, hold F5 to refresh a bunch of times and it quickly jumps.

Before:

https://github.com/user-attachments/assets/ebdb39d5-1c99-4565-9763-fa376393d9ca

After:

https://github.com/user-attachments/assets/807d2b79-5b92-43b9-a601-ae5d17209fe3

### DIFF
--- a/sabnzbd/bpsmeter.py
+++ b/sabnzbd/bpsmeter.py
@@ -22,6 +22,8 @@ sabnzbd.bpsmeter - bpsmeter
 import time
 import logging
 import re
+from collections import deque
+from itertools import islice, repeat
 from typing import Optional
 
 import sabnzbd
@@ -132,7 +134,7 @@ class BPSMeter:
         self.speed_log_time = t
         self.last_update = t
         self.bps = 0.0
-        self.bps_list: list[int] = []
+        self.bps_list: deque[int] = deque(maxlen=BPS_LIST_MAX)
 
         self.server_bps: dict[str, float] = {}
         self.cached_amount: dict[str, int] = {}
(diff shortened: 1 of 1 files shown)


Safihre:
Did you benchmark it? 😊

sabnzbd/bpsmeter.py
```diff
@@ -432,7 +433,7 @@ def get_bps_list(self):
         refresh_rate = int(cfg.refresh_rate()) if cfg.refresh_rate() else 1
         self.add_empty_time()
         # We record every second, but display at the user's refresh-rate
-        return self.bps_list[::refresh_rate]
+        return list(islice(self.bps_list, 0, None, refresh_rate))
```
Safihre, line 436: Could also just do this in the front end? Do we need to list() it?

mnightingale: It's only used on a page load, and can't return the islice or bps_list due to `deque mutated during iteration`

mnightingale:
> Did you benchmark it? 😊

We are talking nanoseconds speed difference, but allocations are much lower. It's only every second anyway.
It's still worth doing - take care of the pennies and the pounds will take care of themselves

https://gist.github.com/mnightingale/049a1dee64c84693c8c9d3a095e43095

per-second tick | time | peak alloc
-- | -- | --
legacy list | 670 ns | 2776 B
deque | 229 ns | 144 B

> 2.9× faster, 19× less memory churn per second. That's the change paying off.

page-load read | time | peak alloc
-- | -- | --
legacy list[::r] | 702 ns | 2432 B
list(islice(deque)) | 2343 ns | 2856 B
list(deque)[::r] | 1583 ns | 4688 B

> The read path got ~3× slower — islice walks the deque one element at a time through the iterator protocol, where list slicing is a single C copy. I left it as islice because this runs once per Glitter page load, so 1.6 µs of absolute cost is noise, and it allocates only the output list. list(deque)[::r] is the faster option but needs both the intermediate and the result alive at once. Both are in the file so the trade-off is documented rather than re-litigated later.

PR #3528, opened 2026-07-28
cgroup-aware memory limits for the article cache

mnightingale:
On a clean install `misc.get_cache_limit()` was `min(physical/4, 1G)`
The ceiling in `ArticleCache.__cache_upper_limit` clamps whatever the user has configured and varied by platform and 64bit.

## Summary table
 
| Scenario | Default `cache_limit` before | after | Ceiling before | after |
|---|---|---|---|---|
| 64-bit host, 64G RAM | 1G | 1G | 4G | 4G |
| 64-bit host, 16G RAM | 1G | 1G | 4G | 4G |
| 64-bit host, 4G RAM | 1G | 1G | 4G | **2G** |
| 64-bit host, 2G RAM | 512M | 512M | 4G | **1G** |
| Docker `--memory=8g` (64G host) | 1G | 1G | 4G | **2G** |
| Docker `--memory=2g` (64G host) | 1G | **512M** | 4G | **512M** |
| Docker `--memory=512m` (64G host) | 1G | **128M** | 4G | **128M** |
| k8s `memory.high=1g` (64G host) | 1G | **256M** | 4G | **256M** |
| NAS/LXC 2g limit, 8G host | 1G | **512M** | 4G | **512M** |
| 32-bit Linux (RPi OS), 8G | 1G | 1G | 1G | 1G |
| 32-bit Windows, 4G | 1G | 1G | 4G | **1G** |
| 32-bit Windows in 2g container | 1G | **512M** | 4G | **512M** |
| memory undetectable | - | - | 4G | 4G |

Bold = changed

tl;dr: **Nothing increases** - every difference is a tightening

Ordinary installs are untouched. 64-bit bare metal with ≥8G sees no change at all to either value. The changes land on containers and small machines, which is the intent. I'm saving probably drastically reducing limits for a future change.

The `--memory=512m` row is the bug being fixed: 1G default cache in a 512M container - twice the entire budget. That's the OOM kill.

## What drives each change

**Containers now see their own budget.** `get_memory()` previously read `/proc/meminfo` `MemTotal`, which is not namespaced, so a container reported the host's RAM. A 512M container therefore computed its default from 64G, sailed past the `1G` cap check, and settled on a 1G cache - twice its entire budget. This is the OOM-kill case the work exists to fix.

**The ceiling is now memory-derived, not just platform-derived.** Previously the ceiling ignored memory entirely, so a user on a 2G machine could set `cache_limit=4G` and it would be honoured. It is now `min(address_space_cap, effective_memory // fraction)`.

**Containers get a smaller fraction (1/4) than bare metal (1/2).** Under cgroup v2 the page cache generated by our own writes to `download_dir` is charged against the same limit, so a given fraction leaves less real headroom inside a container than on a host where the kernel reclaims globally.

**32-bit Windows ceiling corrected.** The old condition was `MACOS or WINDOWS or 64-bit`, so `WINDOWS` short-circuited the bitness test and a 32-bit Windows build got a 4G ceiling inside a ~2G address space. Bitness now decides the address-space cap on its own. I'm not sure it's even usable on 32-bit Windows anymore so this probably doesn't matter.

**Undetectable memory keeps a working cache.** When neither physical memory nor a cgroup limit can be determined, the memory-derived cap is skipped entirely rather than evaluating to zero. A zero ceiling would clamp `__cache_limit` to 0, making `save_article()`'s `if self.__cache_limit` guard falsy and pushing *every* article down the spill-to-disk path.

## Unrelated bug spotted nearby

`get_cache_limit()` did (WINDOWS **and** MACOS) so never true.

### DIFF
--- a/sabnzbd/articlecache.py
+++ b/sabnzbd/articlecache.py
@@ -21,7 +21,7 @@ sabnzbd.articlecache - Article cache handling
 
 import logging
 import threading
-import struct
+import sys
 import time
 from typing import Collection, Optional
 
@@ -34,7 +34,7 @@ from sabnzbd.constants import (
     ARTICLE_CACHE_NON_CONTIGUOUS_FLUSH_PERCENTAGE,
 )
 from sabnzbd.nzb import Article, NzbFile
-from sabnzbd.misc import to_units
+from sabnzbd.misc import to_units, get_memory
 
 # Operations on the article table are handled via try/except.
 # The counters need to be made atomic to ensure consistency.
@@ -56,11 +56,18 @@ class ArticleCache(threading.Thread):
         self.__last_flush: float = 0
         self.__non_contiguous_trigger: int = 0  # Force flush trigger
 
-        # On 32 bit we only allow the user to set 1GB
-        # For 64 bit we allow up to 4GB, in case somebody wants that
-        self.__cache_upper_limit = GIGI
-        if sabnzbd.MACOS or sabnzbd.WINDOWS or (struct.calcsize("P") * 8) == 64:
-            self.__cache_upper_limit = 4 * GIGI
+        # A 32 bit process runs out of address space long before it runs out of memory,
+        # so it never gets to use more than 1GB no matter how much the machine has
+        self.__cache_upper_limit: int = int(4 * GIGI) if sys.maxsize > 2**32 else int(GIGI)
+
+        # Beyond that, never claim more than a share of the memory we are allowed to use.
+        # Skipped when it could not be determined, we would end up without any cache at all
+        memory = get_memory()
+        if memory.effective:
+            # A cgroup limit also has to cover the page cache our own writes generate,
+            # so leave more headroom there than on a host that can reclaim globally
+            memory_fraction = 4 if memory.cgroup_limit else 2
+            self.__cache_upper_limit = min(self.__cache_upper_limit, memory.effective // memory_fraction)
 
     def change_direct_write(self, direct_write: bool) -> None:
         self.__direct_write = direct_write
@@ -124,7 +131,7 @@ class ArticleCache(threading.Thread):
             self.__cache_limit = self.__cache_upper_limit
         else:
             self.__cache_limit = min(limit, self.__cache_upper_limit)
-        self.__non_contiguous_trigger = self.__cache_limit * ARTICLE_CACHE_NON_CONTIGUOUS_FLUSH_PERCENTAGE
+        self.__non_contiguous_trigger = int(self.__cache_limit * ARTICLE_CACHE_NON_CONTIGUOUS_FLUSH_PERCENTAGE)
         if self.__cache_limit:
             logging.debug("Article cache trigger:%s", to_units(self.__non_contiguous_trigger))
 
--- a/sabnzbd/articlecache.py
+++ b/sabnzbd/articlecache.py
@@ -32,9 +32,10 @@ from sabnzbd.constants import (
     GIGI,
     ANFO,
     ARTICLE_CACHE_NON_CONTIGUOUS_FLUSH_PERCENTAGE,
+    ARTICLE_CACHE_RESERVED_MEMORY,
 )
 from sabnzbd.nzb import Article, NzbFile
-from sabnzbd.misc import to_units, get_memory, cgroup_memory_limit
+from sabnzbd.misc import to_units, get_memory
 
 # Operations on the article table are handled via try/except.
 # The counters need to be made atomic to ensure consistency.
@@ -60,13 +61,12 @@ class ArticleCache(threading.Thread):
         # so it never gets to use more than 1GB no matter how much the machine has
         self.__cache_upper_limit: int = int(4 * GIGI) if sys.maxsize > 2**32 else int(GIGI)
 
-        # Beyond that, never claim more than a share of the memory we are allowed to use.
-        # Skipped when it could not be determined, we would end up without any cache at all
+        # Whatever is configured also has to fit in the memory we are allowed to use,
+        # leaving room for the rest of SABnzbd and anything else sharing the limit.
+        # Skipped when memory could not be determined, we would end up without any cache
         if memory := get_memory():
-            # A cgroup limit also has to cover the page cache our own writes generate,
-            # so leave more headroom there than on a host that can reclaim globally
-            memory_fraction = 4 if cgroup_memory_limit() else 2
-            self.__cache_upper_limit = min(self.__cache_upper_limit, memory // memory_fraction)
+            available = max(0, memory - ARTICLE_CACHE_RESERVED_MEMORY)
+            self.__cache_upper_limit = min(self.__cache_upper_limit, available)
 
     def change_direct_write(self, direct_write: bool) -> None:
         self.__direct_write = direct_write
--- a/sabnzbd/articlecache.py
+++ b/sabnzbd/articlecache.py
@@ -34,7 +34,7 @@ from sabnzbd.constants import (
     ARTICLE_CACHE_NON_CONTIGUOUS_FLUSH_PERCENTAGE,
 )
 from sabnzbd.nzb import Article, NzbFile
-from sabnzbd.misc import to_units, get_memory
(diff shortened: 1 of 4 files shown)


thezoggy:
Are the values in your table is if the user did not specify a specific value? what if they used `-1` ?

Using `-1` would allow user to just let them use the artificial limit of 2G on 32bit os, 4G on 64bit os).
Which as you noted could actually result in a value larger than desired when doing min(artificial limit, physical ram)
So updating it to min(artificial limit, effective ram) makes sense.

But I wanted to confirm you are not changing the 'artificial limit' part, so for example:
> host os with 64G ram but container with 2G ram set (--memory=2g) and in sab article cache set to `-1`.

I would expect cache to be able to use the full 2G in sab.. is this correct?

mnightingale:
> Are the values in your table is if the user did not specify a specific value? what if they used `-1` ?

In the table "Default `cache_limit` before" is what would be applied on a clean install; previously it was 1/4 of physical memory capped at 1G or 512M in `get_cache_limit()`

"Ceiling before" refers to the maximum cache_limit the article cache will actually allow to be used, it's now derived from how much memory is available to the container and not just the type of platform. For example, a bare-metal host with 2G of RAM actually allowed and used an article cache of 4G which makes no sense and missed some of our attempts to avoid usage running away.

The changes are primarily to make it more friendly when running in a container environment, previously containers only saw the host physical ram and not the limits applied to them.

In your example a container with a 2G ram limit `--memory=2g` would be killed with OOM if it exceeds 2G, so the change caps the article cache at 512M regardless of the user-controlled setting.

| Scenario | Default `cache_limit` before | after | Ceiling before | after |
|---|---|---|---|---|
| Docker `--memory=8g` (64G host) | 1G | 1G | 4G | **2G** |
| Docker `--memory=2g` (64G host) | 1G | **512M** | 4G | **512M** |
| Docker `--memory=512m` (64G host) | 1G | **128M** | 4G | **128M** |

These are probably the ones worth considering, possibly the 1/4 rule for containers could be more related to available memory - will do a separate PR after this one.

So take the top line, previously an 8G limit would use 1G on a fresh install and allow 4G as a maximum user setting, it's now 2G.
The 2G limit would default to 1G, now 512M and a maximum value of 512M instead of 4G (more than the container is allowed).
The 512M limit would default to 1G(!), now 128M and a ceiling of 128M instead of 4G(!) -  I doubt a 512M limit even works; maybe needs no cache or much smaller 32-64M.

Keep in mind that 2GB needs to allow for SAB and any unrar or par2 operations, therefore the ceiling calculation is a little stricter on containers. Bare-metal can use 1/2 of physical memory and containers 1/4.

Obviously if you're running in a container without a cgroup limit it can see and use as much as if running directly on the host.

> Using `-1` would allow user to just let them use the artificial limit of 2G on 32bit os, 4G on 64bit os). Which as you noted could actually result in a value larger than desired when doing min(artificial limit, physical ram) So updating it to min(artificial limit, effective ram) makes sense.

Setting `cache_limit` to `-1` is shorthand for use the maximum (ceiling) and other values apply `min(want, ceiling)`, this is unchanged, the only change is that ceiling is restricted to a value within the available memory.

I left the `get_memory()` returning physical memory for now, but did consider making it look at what's is available which is probably more sensible. For example if you launch SAB on a system that has 16G of memory, but 12G is being used for a game or something then allowing a 4G cache is not ideal - but I'll leave any such change for a possible future PR

> But I wanted to confirm you are not changing the 'artificial limit' part, so for example:
> 
> > host os with 64G ram but container with 2G ram set (--memory=2g) and in sab article cache set to `-1`.
> 
> I would expect cache to be able to use the full 2G in sab.. is this correct?

[Docker: Limit a container's access to memory](https://docs.docker.com/engine/containers/resource_constraints/#limit-a-containers-access-to-memory)

`-m` or `--memory=`: The maximum amount of memory the container can use. ...
`--oom-kill-disable`: By default, if an out-of-memory (OOM) error occurs, the kernel kills processes in a container. ...

For this reason, a container ran with `--memory=2g` will restrict the maximum article cache to 512M - previously it would allow 4G and trigger OOM if total container usage exceeds 2G.

---

Possibly we should change the config behaviour, this PR doesn't change it and currently it's:

`-1` = ceiling
`(empty)` = 0 (no article cache)
`0` = also no article cache
`(else)` = bytes or units - **PR change** to respect the calculated ceiling

I think empty should apply the clean install logic.

---

The article cache allowed values (4G) are likely way too large anyway - probably by a factor of 10 - but that's part of a larger problem to be addressed in the coming months as I implement changes to rebalance the download, assemble, and post-process pipeline. This was just the first step, low risk change that has some real benefit for container and low RAM users.

I can't find it right now but I have seen some reports of containers being killed for OOM.

Safihre:
Seems a bit convoluted with the dataclass? Just have it return the effective memory from get_memory? Since that's all we care about anyway?

mnightingale:
> Seems a bit convoluted with the dataclass? Just have it return the effective memory from get_memory? Since that's all we care about anyway?

Originally, I was going to do that, but I ended up needing ArticleCache to know why the effective value is what it is. 

`memory_fraction = 4 if memory.cgroup_limit else 2`

If it has a cgroup_limit then reduce the fraction to 1/4 rather than 1/2 – because it’s more critical that we don’t reach the ceiling and there are other costs like page caching and not just the article cache size.

The alternative is letting the caller call `cgroup_memory_limit()` a second time so they can determine what the result of `get_memory()` means, but it just makes it more complicated for callers.

It’s like what I already did with `class Diskspace` and probably won’t be the last.

---

I am going to rework `get_cache_limit()` a little but a separate PR.

Safihre:
I theoretically get that want to be more strict, but don't agree we need seperate case for it. Either physical or effective memory, limit should be the same.

mnightingale:
I’ve removed `MemoryInfo` so `get_memory()` returns effective memory, the minimum of physical and cgroup limit.

Then `cgroup_memory_limit()` can be used to see if there is a hard memory limit we need to stay below - it's a limit not a target.

The overall change is protecting users from themselves, in several cases the defaults even on a clean install want to use more memory than the container allows which can cause out-of-memory issues.  Additionally, the user is allowed to manually configure a value higher than the system or cgroup allows.

Then ceiling is calculated as half physical memory or a quarter of cgroup_limit, previously it was 1G or 4G on macOS, Windows, or 64-bit (note bug 32-bit Windows would allow 4G)

Bare-metal or no limits are unaffected, 8G or more memory can continue to use 4G article cache (even though it shouldn’t), but a system with 4G or 2G can use a maximum of 2G or 1G.

thezoggy:
the value the user want needs to land within the effective memory.  we all agree on that. 

so fundamentally, we should do: min(artificial_limit, effective memory).  

but you are afraid that if that the delta between artificial_limit and  effective memory is too close artificial_limit should be reduced to avoid an oom. which I can understand and you want to guard against that.

I dont think we need to do anything crazy but if you wanted to just carve off some value from the effective memory like -500mb, so min(artificial_limit, effective memory - 500mb) just protects more scenarios that is probably good enough.

I don't think we really need to try and handle all scenarios if user is very ram starved.. they prob need to adjust their stuff/figure out if they really should be limiting their setup like they are.

mnightingale:
That seems quite reasonable to me and a better fit than a percentage.

Scenario | Default before → after | Ceiling before → after
-- | -- | --
64-bit host, 8G+ | 1G → 1G | 4G → 4G
64-bit host, 2G | 512M → 512M | 4G → 1.5G
Docker --memory=2g (64G host) | 1G → 512M | 4G → 1.5G
Docker --memory=512m | 1G → 128M | 4G → 0, cache off
Docker --memory=256m | 1G → 64M | 4G → 0, cache off
32-bit Windows, 4G | 1G → 1G | 4G → 1G
memory undetectable | unchanged | 4G → 4G

thezoggy:
the docker memory=512 one where it results in 12M id question if that actually provides value over just not using cache..   but id personally be shocked if someone was able to run sab effectively with only 512mb ram in the first place. overall everything else looks good

mnightingale:
Changed the constant so 512 limit gets 0 ceiling, forcing no cache.
I left it initially because I wouldn't really expect it to work below 1GB anyway.

PR #3530, opened 2026-07-29
Fix os.umask race and unnecessary

mnightingale:
We don't need this, it's not doing anything except opening a window where umask is 0

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -544,8 +544,6 @@ class Assembler(Thread):
         :returns (file_descriptor, current_offset, can_direct_write)
         """
         with nzf.file_lock:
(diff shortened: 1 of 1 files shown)


Safihre:
Why was is there initially?

thezoggy:
looks like was added for direct write?

found this article that talks about umask(0) thing:
https://orkus.wordpress.com/2011/04/17/python-getting-umask-without-change/

mnightingale:
AI added it, but it's effectively a no-op.

```
old = os.umask(0)  # change umask to 0 and return old value
# Window here to create files with zero umask
os.umask(old) # immediately restore the old value
```

In this case is it does nothing, except getting the current umask and restoring it, umask is not threadsafe so any files created between the get and set have a umask of 0.

Safihre:
AI 🫣😅

PR #3532, opened 2026-07-30
feat: add PBKDF2-HMAC-SHA256 web password hashing

mission-deny-the-mission:
Replace the plaintext `password` storage for the web interface login with PBKDF2-HMAC-SHA256 hashing (600,000 iterations, 16-byte random salt, constant-time verification).

This means the web password stored in `sabnzbd.ini` is no longer recoverable plaintext — if someone gains read access to the config file, they cannot obtain the password.

### Changes
- New `OptionPassword` config type in `cfg.py` that stores and validates the hash in `$algorithm$iterations$salt$hash` format
- New `password_hash` config option under `[misc]`, stored as `protect=True` (not exposed via API)
- Automatic migration on first run: reads the old `password`, writes `password_hash`, clears the old field
- Constant-time HMAC-SHA256 verification in `check_login()` and `check_basic_auth()`
- All existing auth paths (cookie login, basic auth, API key) continue to work unchanged
- Backward compatible: configs with only the old `password` continue to work (hashed on first access)

### Checks
- All 7,322 unit tests pass
- Black and Ruff clean

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -32,6 +32,7 @@ from sabnzbd.config import (
     OptionBool,
     OptionNumber,
     OptionPassword,
+    OptionPasswordHash,
     OptionDir,
     OptionStr,
     OptionList,
@@ -372,7 +373,10 @@ web_host = OptionStr("misc", "host", DEF_HOST, validation=validate_host)
 web_port = OptionStr("misc", "port", DEF_PORT)
 https_port = OptionStr("misc", "https_port")
 username = OptionStr("misc", "username")
+# password is retained only as a read-once migration source for existing web credentials.
+# New web passwords are persisted as one-way PBKDF2 hashes in password_hash.
 password = OptionPassword("misc", "password")
+password_hash = OptionPasswordHash("misc", "password_hash")
 bandwidth_max = OptionStr("misc", "bandwidth_max")
 cache_limit = OptionStr("misc", "cache_limit")
 web_dir = OptionStr("misc", "web_dir", DEF_STD_WEB_DIR)
--- a/sabnzbd/config.py
+++ b/sabnzbd/config.py
@@ -19,10 +19,14 @@
 sabnzbd.config - Configuration Support
 """
 
+import base64
+import hashlib
+import hmac
 import logging
 import os
 import re
 import shutil
+import tempfile
 import threading
 import time
 import uuid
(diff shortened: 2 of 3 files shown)


Safihre:
But if they have access to the sabnzbd.ini, they also have access to the API key, so that's basically the same attack vector.

mnightingale:
I don't know if passwords would want PBKDF2-HMAC-SHA256 https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#password-hashing-algorithms suggests Argon2id, scrypt, or bcrypt (in that order) - probably need extra dependencies for them.

I think it could be kept in a single config item though just check for the prefix - also better for version downgrades, the current changes remove the legacy password so a downgrade leaves it open.

It's unclear what some of the `set_config` changes are doing.

As mentioned other things would need to be encrypted instead so they are just unreadable at rest; probably server passwords, api keys, maybe rss urls. With the encryption key stored in macOS keychain, Windows Credential Manager, file?, I'm not sure how linux/docker would do it.

But I think this is an area we should be looking at.

Safihre:
First ask us if we want something, before spending tokens.

mission-deny-the-mission:
> But if they have access to the sabnzbd.ini, they also have access to the API key, so that's basically the same attack vector.

Well yeah that's fairly obvious and also not the point. The reason to do this is to prevent passwords getting leaked that might be used for other services. Not to prevent SAB getting compromised. We assume that's already happened if they can see the ini file.

mission-deny-the-mission:
> I don't know if passwords would want PBKDF2-HMAC-SHA256 https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#password-hashing-algorithms suggests Argon2id, scrypt, or bcrypt (in that order) - probably need extra dependencies for them.
> 
> 
> 
> I think it could be kept in a single config item though just check for the prefix - also better for version downgrades, the current changes remove the legacy password so a downgrade leaves it open.
> 
> 
> 
> It's unclear what some of the `set_config` changes are doing.
> 
> 
> 
> As mentioned other things would need to be encrypted instead so they are just unreadable at rest; probably server passwords, api keys, maybe rss urls. With the encryption key stored in macOS keychain, Windows Credential Manager, file?, I'm not sure how linux/docker would do it.
> 
> 
> 
> But I think this is an area we should be looking at.

Yeah I agree. I wasn't sure what to do about encrypting the other credentials. Just thought this would be a start. I think it would be quite difficult to handle on Linux servers potentially as they wouldn't have anything like Gnome or KDE keyrings if they are headless. Best I can think to do is tie them into infisical or something like that. I already do this for some stuff so adding support for that would actually be helpful to me personally but I would think it's a fairly niche audience. For Linux desktop it wouldn't be too hard I imagine to use the desktop keyring. Although you would have to account for different desktop environments and stuff. Alternatively you could build in a keyring that gets unlocked the first time someone logs into the admin console/interface.

I could submit another PR with bcrypt instead or another hashing algorithm. However that hinges on convincing the maintainer(s) that hashing passwords is necessary.

mission-deny-the-mission:
> First ask us if we want something, before spending tokens.

I did this because it's security best practice to hash passwords wherever possible and that's something I personally care about. I am not sure who "we" is you are referring to. If you don't care that's fine. I would bet that at least one other person cares though. In fact there already is one in this thread who has presented ideas about how this could be implemented better.

If you also think there is a better way of doing this I am all ears. Otherwise I don't really know what to say to you.

Safihre:
Open an issue first, before dumping code. 
We have this every week now and it's exhausting.

mission-deny-the-mission:
> Open an issue first, before dumping code. 
> 
> We have this every week now and it's exhausting.

I am happy to do this going forward.

Would it not be wise to add some documentation explaining what the rules for contributing are? If you want issues for all changes there should be something spelling that out. Unless I missed some somewhere.

I considered adding an issue first anyway but found that not all commits were tied to one so assumed it wasn't necessary and would be a waste of time.

mnightingale:
Agreed these kinds of big changes need discussion first, we must consider the implications and all the other moving parts, restoring backups, downgrading versions, etc. Not least because we are committing to long-term support of a new feature.

For example, say we did encrypt the usenet server passwords, what should a scheduled backup look like? Also encrypted so they can only be restored on a system with the same secret? Decrypted? Require a password prompt on import? I'm not certain the best security practice.

I'm not intending to prompt discussion on it here - feel free to open an issue or maybe a discussion would be more appropriate, there are a bunch of things to consider before implementation.

I do care about this; credential stealing is a big concern these days. But the solution needs more thought on its implementation and all the areas it needs to touch.

If no one else does, I’ll probably pick it up eventually but either way it’s not as simple as rushing in and changing just the password hashing.

PR #3533, opened 2026-07-30
feat: add Authentik proxy-header SSO authentication

mission-deny-the-mission:
Add support for Authentik (and compatible) forward-auth proxies that set the `X-Authentik-Username` header.

The feature is configured by adding trusted proxy IP(s) to the new `authentik_proxy_trusted_proxies` config option (a protected, API-hidden list under `[misc]`).

### How it works (hybrid mode)
- Requests from a **trusted proxy** with a valid `X-Authentik-Username` header → authenticated via SSO
- Requests from a **trusted proxy without the header** → rejected (cannot bypass the proxy)
- Requests **not from the proxy** → fall through to normal password/cookie/basic auth

This means existing local/direct access continues to work even when SSO is configured — useful for mixed-access environments.

### Design decisions
- The TCP peer address (`cherrypy.request.remote.ip`) is used for trust, not `X-Forwarded-For` — prevents header spoofing from untrusted clients
- When a request is authenticated via SSO, the logout button is hidden (SSO logout is handled at the proxy level, not within SAB)
- The login page returns 401 for requests that arrived through the proxy (they must have a valid Authentik session)
- Config is `protect=True` and `public=False` — not exposed in the web UI or API

### Testing
- New parametrized test covering all three-state return values (True/False/None)
- Integration tests for `check_login()` and `set_auth()` in hybrid mode
- All 7,326 unit tests pass (including 8 new SSO-specific tests)
- Black and Ruff clean

### DIFF
--- a/sabnzbd/cfg.py
+++ b/sabnzbd/cfg.py
@@ -160,6 +160,18 @@ def cleanup_list_validator(value: str | list) -> tuple[None, str | list]:
     return None, validate_entry(value)
 
 
+def validate_trusted_proxy_addresses(value: list[str]) -> tuple[None, list[str]]:
+    """Accept only exact IP addresses for trusted Authentik proxy peers."""
+    normalized = []
+    for entry in value:
+        try:
+            address = ipaddress.ip_address(entry.strip())
+        except ValueError:
+            return "Invalid trusted proxy address: %s" % entry, None
+        normalized.append(str(address))
+    return None, normalized
+
+
 def validate_single_tag(value: list[str]) -> tuple[None, list[str]]:
     """Don't split single indexer tags like "TV > HD"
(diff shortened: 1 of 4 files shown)


mnightingale:
Needs the password hashing commit removing.

I wonder if we could add more generic SSO support like OIDC, rather than specific Authentik support?
OIDC I think could avoid the extra xff config option and maybe simpler for the future uvicorn change.

https://integrations.goauthentik.io/applications/

> Most third-party services that support authentication protocols such as SAML, OAuth, and OpenID Connect can be integrated with authentik, allowing users to log in to these services using their authentik credentials.

Safihre:
First ask us if we want something, before spending tokens.

PR #3535, opened 2026-07-31
Relax disk space requirements when not unpacking and on the same device

### LINKED ISSUE #3531: Incorrect calculation of required free disk space pauses downloads too soon when Unpack is turned off
### SABnzbd version

5.0.4

### Operating system

Windows

### Using Docker image

None

### Description

The calculation to check how much free disk space is required for a download does not factor in whether +Repair or +Unpack is enabled/disabled on the current download's Category. So when +Repair and +Unpack are both off (so the Category is set to just Download), the download might be paused even though there is more than enough free disk space for the download to complete. Manually resuming the download and it just pause again at the completion of the next file.

Checking the code, I identified in sabnzbd/assembler.py, in the diskspace_check() function is this section of code (lines 334 to 347 in master branch) :

```
        if not full_dir:
            complete_free = cfg.complete_free.get_float()
            required_space = 0
            if cfg.direct_unpack():
                # We unpack while we download, so we should check every time
                # if the unpack maybe already filled up the drive
                required_space = complete_free / GIGI
            elif nzo.bytes_tried > (nzo.bytes - nzo.bytes_par2) * 0.90:
                # Since only at 100% unpack is started, continue
                # downloading until 95% complete before checking
                required_space = (complete_free + nzo.bytes) / GIGI

            if required_space and complete_dir.free < required_space:
                full_dir = complete_dir.path
```

I have Unpack turned off for all Categories - My default Category has Processing=Download. All custom Categories have Processing=Default. I also have Direct Unpack turned off.

I ran the app from source and added some extra logging to the above block of code to trace what it was doing.

It always fell through the first if statement (so it correctly sees Direct Unpack is turned off) and continues into the "elif nzo.bytes_tried > (nzo.bytes - nzo.bytes_par2) * 0.90:" block. This checks if at least 90% of the download has completed (ignoring the PAR files).

If that elif evaluates to True (>90% of non-PAR files are complete), the code calculates that I need at least "(complete_free + nzo.bytes) / GIGI" free disk space, which is equal to the threshold we set in the config + full download size of the nzb. This is the amount it wants free, even though there is only 10% of the download remaining (+ what is needed for the PAR files) and even if all Unpacking is turned off.

As a result, the download can be incorrectly paused when when there is still more than enough space available. Manually unpausing the download and it just re-pauses again at the end of the next file within the nzb collection because the same free disk space is expected again.

To avoid this, the elif block needs to check if Repair/Unpack is enabled or disabled for the current download Category and adjust the required free disk space accordingly. At the moment it just assumes Unpack at download completion is always turned on .

Example, I started a 62479MB (61GB) download with 118GB free disk space. All Unpack options are off. Completed Folder free space threshold is set to 5GB in my config.
The download paused at the ~90% completion point, with only 14GB remaining to download and 65GB free disk space.

mnightingale:
Fixes #3531

Past 90%, `diskspace_check()` required `complete_free + nzo.bytes` in the complete folder for every job. In the users 61GB Download-only job paused with 60GB still free, and re-paused on every following file.

Post-processing moves the data to the complete folder, so that space is genuinely needed across devices. On a *single* device it is not: the downloaded bytes are already deducted from the free space the check reads, and the move is a rename, so the job size was counted twice.

The job size is now only added when the data really is duplicated, either because the job unpacks or because the folders are on different devices. `same_device()` compares `st_dev`, guarded by a `splitdrive` check for Windows drives and shares, and assumes separate devices when it cannot tell.

Also fixed `cfg.direct_unpack()` took effect for jobs that never direct unpack, skipping the size check entirely. Tests were written against the old code first, so the ones that change are the ones the fix intends to change.

---

There is another existing issue with `+Unpack` since it keeps the archives, when they are different devices it should reserve 2x the space. I can add that fix to this PR if you'd like?

### DIFF
--- a/sabnzbd/assembler.py
+++ b/sabnzbd/assembler.py
@@ -41,6 +41,7 @@ from sabnzbd.filesystem import (
     get_filename,
     has_unwanted_extension,
     get_basename,
+    same_device,
 )
 from sabnzbd.constants import (
     Status,
@@ -335,14 +336,19 @@ class Assembler(Thread):
         if not full_dir:
             complete_free = cfg.complete_free.get_float()
             required_space = 0
-            if cfg.direct_unpack():
+            if cfg.direct_unpack() and nzo.unpack:
                 # We unpack while we download, so we should check every time
                 # if the unpack maybe already filled up the drive
                 required_space = complete_free / GIGI
             elif nzo.bytes_tried > (nzo.bytes - nzo.bytes_par2) * 0.90:
                 # Since only at 100% unpack is started, continue
                 # downloading until 90% complete before checking
-                required_space = (complete_free + nzo.bytes) / GIGI
+                required_space = complete_free / GIGI
+                if nzo.unpack or not same_device(download_dir.path, complete_dir.path):
+                    # Unpacking writes a second copy of the data before the archives are removed
+                    # and moving to another device copies it, so both need room for the whole job.
+                    # Without unpacking on the same device the move is only a rename.
+                    required_space += nzo.bytes / GIGI
 
             if required_space and complete_dir.free < required_space:
                 full_dir = complete_dir.path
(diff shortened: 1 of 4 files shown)


Safihre:
@claude review for edge cases, just to be sure

PR #3537, opened 2026-08-01
Early exit shutdown_program when it is called concurrently

### LINKED ISSUE #2148: Shutdown stalls indefinitely

anchpop:
Sorry, my claude went crazy and made a PR when I didn't want it to (#3534). I was embarrassed and closed the PR, but I actually think its change is correct having reviewed it. I can't reopen the PR because I force-pushed to the branch, so I'm making a new one. This probably closes #2148.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -176,6 +176,7 @@ RESTORE_DATA = None
 
 # Condition used to handle the main loop in SABnzbd.py
 SABSTOP_CONDITION = Condition(Lock())
+SHUTDOWN_LOCK = Lock()
 
 # General threadpool
 THREAD_POOL = concurrent.futures.ThreadPoolExecutor(max_workers=2)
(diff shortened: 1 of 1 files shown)


mnightingale:
`SHUTDOWN_LOCK` is not released if it fails to shutdown, I would suggest try/finally to release it

```py
def shutdown_program():
    """Stop program after halting and saving"""
    # Non-blocking acquire so a concurrent caller returns instead of running the teardown twice.
    # Releasing it again keeps a failed shutdown retryable, SABSTOP blocks the successful case
    if sabnzbd.SABSTOP or not SHUTDOWN_LOCK.acquire(blocking=False):
        logging.debug("[%s] Shutdown already in progress", misc.caller_name())
        return

    try:
        logging.info("[%s] Performing SABnzbd shutdown", misc.caller_name())
        sabnzbd.halt()
        cherrypy.engine.exit()
        sabnzbd.SABSTOP = True
        notify_shutdown_loop()
    finally:
        SHUTDOWN_LOCK.release()
```

I would move `SHUTDOWN_LOCK = Lock()` to after `SABSTOP_CONDITION`

There are some other changes I will make with how we handle the signal, but this is a good improvement to prevent the deadlock thanks!

PR #3538, opened 2026-08-02
Fix assembler ready bytes accounting

mnightingale:
`ArticleCache.save_article` > `_flush_article_to_disk` > `Assembler.assemble_article` > `Assembler.write` subtracts ready bytes before they are added via `NzbQueue.register_article`, results in incorrect tracking and higher memory usage - assembler thinks it has little to do meanwhile it keeps hitting the out-of-order write path.

Fix is to record the bytes in the article cache before we decide how they will be written, then it doesn't matter if all assembler paths subtract.


Safihre:
@claude review, most effecient?

PR #3541, opened 2026-08-02
fix active pagination for light theme

thezoggy:
before:
image

after:
image


thezoggy:
this does make the hover color and the active color pretty close. 
looking up bootstrap spec looks like technically we should be using "pagination-active-bg" which would be something more like `--clr-toggle-active` 

ex:
image

but imho that is a bit much

PR #3546, opened 2026-08-03
Speed up tests with pytest-xdist

mnightingale:
Removed `-s` because not supported [pytest: Output (stdout and stderr) from workers](https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#output-stdout-and-stderr-from-workers) you still get captured logging around failures.

At some point we may be able to use `--dist loadscope` to better spread the load, I was still finding too many failures. It's not a huge difference on GitHub runners anyway, locally 25s vs 45s. But overall this should be around 1-2 minutes faster overall, any much faster locally.

Most of the changes are removing global module state, mostly in tests with a little bit in `sabnzbd/config.py` but it also improves testing with more use of monkeypatch to ensure clean state between tetsts.

Fixes a number of tests that were dependant on the order of execution.

We could also maybe add the default args to pyproject.toml.

I've ran this loads over the last few weeks, I kept forgetting to change selftest_host and getting blocked 😆 however, I wouldn't be surprised if we get occasional failures due to order of execution.

### DIFF
--- a/.github/workflows/integration_testing.yml
+++ b/.github/workflows/integration_testing.yml
@@ -68,4 +68,4 @@ jobs:
         pip install -r requirements.txt --no-dependencies
         pip install --upgrade -r tests/requirements.txt
     - name: Test SABnzbd
-      run: pytest -s -n auto --dist loadscope
+      run: pytest -s -n auto
--- a/.github/workflows/integration_testing.yml
+++ b/.github/workflows/integration_testing.yml
@@ -68,4 +68,4 @@ jobs:
         pip install -r requirements.txt --no-dependencies
         pip install --upgrade -r tests/requirements.txt
     - name: Test SABnzbd
-      run: pytest -s
+      run: pytest -s -n auto --dist loadscope
--- a/.github/workflows/integration_testing.yml
+++ b/.github/workflows/integration_testing.yml
@@ -68,4 +68,4 @@ jobs:
         pip install -r requirements.txt --no-dependencies
         pip install --upgrade -r tests/requirements.txt
     - name: Test SABnzbd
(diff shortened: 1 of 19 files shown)


mnightingale:
TestInternetSpeed is a little flakey, I think Cloudflare must block the runners occasionally - not a new failure

Safihre:
@claude review if any tests functionally changed, and if changes are the smartest way to implement this

mnightingale:
Claude got the descriptions a little wrong I think https://pytest-xdist.readthedocs.io/en/stable/distribution.html
I was working towards `pytest -n auto --dist worksteal` reliability, it's the fastest because we have a load of tests in one file that all get pinned to a single worker otherwise.

Default is `load` (46s), `loadscope` (31s) groups by class and I feel is a bit safer than `worksteal` (28s).

If you're prepared to have occasional test failures I'd go with loadscope, worksteal could give obscure test failures if I haven't fixed all the issues and the speed difference especially in GitHub runners is not significant.

Rule of thumb, tests should not change SABnzbd global state at all and that's what almost all the changes are.

Safihre:
Loadscope seems like logical choice.

mnightingale:
I've switched to loadscope, I expect occasional failures but I don’t think it’s particularly bad if inconsistent tests end up getting fixed, and I’ll probably be me fixing them anyway.

The only other thing we could do is change the default in pyproject.toml

```toml
[tool.pytest]
addopts = [
    "-n",
    "auto",
    "--dist",
    "loadscope"
]
markers = [
    "config",
    "platform",
    "fake_fs"
]
```

Merge this before uvicorn please

PR #3549, opened 2026-08-05
Don't log a warning during clean shutdown (#3548)

### LINKED ISSUE #3548: Clean shutdown posts notification "Warning: Signal 15 caught, saving and exiting..." (Linux/macOS only)
### SABnzbd version

5.0.4 [128e0d0]

### Operating system

Unraid

### Using Docker image

linuxserver

### Description

I configured sabnzbd notifications for Warnings to be posted to Pushover.

I set up a nightly backup job for my server, which cleanly shuts down my Docker containers, which sends sabnzbd `SIGTERM` (signal 15).

This posts a notification every night:

```
Warning: Signal 15 caught, saving and exiting...
```

Since `SIGTERM` is an expected part of clean shutdown, it shouldn't post a warning.

bhamiltoncx:
`SIGTERM` is an expected part of a clean shutdown on POSIX-like OSes (Linux and macOS, not Windows).

Previously, the signal handler posted a log warning for any signal received, which turned into Warning notifications every time the server shut down.

This PR turns down the log level to `info` for handled signals (the log was issued on non-Windows OSes only). This only affects the two signals we handle, `SIGTERM` (aka `kill`) and `SIGINT` (aka Control-C), both of which are user-issued and shouldn't cause warnings.

Tested:

```
% python3 -m venv venv
% source venv/bin/activate
% pip install -r dependencies.txt
% python3 -OO SABnzbd.py
```

In another terminal, ran `kill -TERM $PID`.

Confirmed in logs that `Signal X caught, saving and exiting...` appeared at `INFO` level (previously it appeared with `WARNING` level and pushed a push notification.)

Repeated with Control-C, confirmed it worked and logged at `INFO`.

Fixes #3548.

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -205,7 +205,7 @@ def sig_handler(signum=None, frame=None):
         # Ignore the "logoff" event when running as a Win32 daemon
         return True
     if signum is not None:
-        logging.warning(T("Signal %s caught, saving and exiting..."), signum)
+        logging.info(T("Signal %s caught, saving and exiting..."), signum)
         sabnzbd.shutdown_program()
 
 


Safihre:
Why is any signal a Warning actually?

bhamiltoncx:
> Why is any signal a Warning actually?

Great question. The only other signal which is handled is `SIGINT`, aka Control-C:

https://github.com/sabnzbd/sabnzbd/blob/70f89e98e8a38301fe8acb6e63b768550f2faab9/SABnzbd.py#L1631-L1632

It'd be fine to use info level for that, since it's an intentional user action.

I went ahead and changed this PR to have the log level be info unconditionally.

mnightingale:
> Why is any signal a Warning actually?

I would guess was so users on "Errors/Warnings" logging saw the message in logs.
I think I would just go with logging.debug since shutdown_program already does a logging.info that's more meaningful to the user, normal signal logs almost look like something went wrong.

Safihre:
Thanks for helping out!

PR #3550, opened 2026-08-05
Replace CherryPy with Uvicorn and Starlette

mnightingale:
This builds upon the work done in #3373

The first commit combines all the work from #3373 with a few unrelated changes left out.
Each of the many following commits fix or reimplement some missing feature.
I tried really hard but there are a few new features that are either required or worth adding and are related to the PR I promise!

You won't like one part where I used `data-form` I couldn't decide what approach you would prefer, maybe classes since id couldn't be used without creating unique ones - the same form is on a page multiple times.

I purposely made as few changes as possible to the api route implementations themselves, except preparing them for a future when they may be async but it requires more changes than I felt comfortable with, better to land this first. This means Uvicorn delegates almost everything to it's threadpool to run them async, note Uvicorn by default can use 40 threads - I've not checked how it works but I presume it typically uses far fewer.

The main thing to be aware of is there is one async loop, so if you do something synchronous you block all other requests; for example if you do a DB query, or try to grab an RSS feed you block every other request unless you use async methods. So for now pretty much everything runs in the threadpool.

Given our performance requirements I didn't bother adding uvloop.
For anyone wondering, Starlette is used instead of FastAPI because the existing API is not rest routing so you don't gain any of the auto-documentation or validation benefits.

Missing features/changes:

- The dual-server http and https is gone; single uvicorn.Config that is either HTTPS with a dedicated https_port, HTTPS on web_port, or HTTP on web_port - note https_port could be removed but i didn't want to break existing configs
- Basic authentication is removed
- Uvicorn can't be shutdown from its own thread, so such requests are async: you send a restart request via the api and it starts a BackgroundTask actioned after a response is sent to the browser, I made the webserver stop accepting new connections so you can't immediately connect to the shutting down server
- secure cookies if request was served over https or enable_https, the XFF middleware proxies the request scheme. Note I have not made this quite correct for `SessionMiddleware` yet
- I need to check 'Fix return type of retry job for future types'

New features:

- Fixed existing global RSS state shared between webserver threads without locking - uses a session cookie to flash data
- Pool database connections; previously each webserver thread (10) would open a database connection, Uvicorn could use 40 for its threadpool for handling sync requests. The previous threadlocal approach was a bit messy and you still had other areas creating short-lived connections. Therefore there is now `with sabnzbd.db_pool.connection() as history_db` used throughout, it manages a pool of 5 connections but more can still be requested they're just not returned to the pool. This may result in lower memory usage because idle connections are not sat in webserver threads.
- SQLite PRAGMA tweaks to compliment the pool changes; see `database.py` for explanation tl;dr journal_mode=WAL,synchronous=NORMAL,busy_timeout=5000,cache_size=-4000,temp_store=MEMORY,mmap_size=16777216
- Backup changes to facilitate database WAL changes; note WAL won't play nice over the network, but if someone has their config directory on a network share they've got bigger problems
- $url helper in interfaces, a lot nicer to work with from nested routes and required because Uvicorn/Starlette changes removed the trailing slashes - accounts for ~300 line changes

Future:

There are loads of future changes this opens up, some not strickly unlocked by this PR but I thought of having looked at webserver/interface/api code:

- Pydantic validation of api requests, like FastAPI would provide - I'm imagining passing models to `ApiEntry` and also using it to generate API documentation
- Config changes without restart; all but about 2 usages of the `guard_restart` callback can be replaced with live changes to the webserver, without restarting the whole application
- Database-backed sessions for cookie authentication and removing apikey usage from the frontend - use aiosqlite, can resolve some issues with cookies in containers
- Constant time comparison of login credentials - prevent timing based leaks of credentials
- add_nzbfile by processing UploadFile without a temporary file (except 7zip)
- SQLite incremental vacuum upon DELETE operations
- UI state via SSE or WebSockets (probably SSE), perhaps in-browser log output via SSE
- I left it out of these changes but add trigger_shutdown() and act on it from the main thread, with careful use of a wait_for_shutdown() from threads that require it, for now I kept it as is and spawned threads
- ~~Detect network filesystems and disable WAL~~

Probably some things I've forgotten so I'll probably edit this message.

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -144,7 +144,7 @@ jobs:
       run: |
         python3 --version
         pip3 install --upgrade pip wheel
-        pip3 install -r requirements.txt --no-binary cffi,CT3,PyYAML --no-dependencies
+        pip3 install -r requirements.txt --no-binary cffi,CT3,PyYAML,httptools --no-dependencies
         pip3 install -r builder/requirements.txt --no-dependencies
     - name: Import macOS codesign certificates
       # Taken from https://github.com/Apple-Actions/import-codesign-certs/pull/27 (comments)
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1354,7 +1354,8 @@ def main():
     notifier.send_notification("SABnzbd", T("SABnzbd shutdown finished"), "startup")
     logging.info("Leaving SABnzbd")
     sabnzbd.pid_file()
-    sabnzbd.WEB_SERVER.stop()
+    if sabnzbd.WEB_SERVER:
+        sabnzbd.WEB_SERVER.stop()
 
     try:
         sys.stderr.flush()
--- a/SABnzbd.py
+++ b/SABnzbd.py
@@ -1215,7 +1215,14 @@ def main():
         ssl_ca_certs=https_chain if enable_https else None,
     )
     sabnzbd.WEB_SERVER = sabnzbd.interface.ThreadedServer(config=server_config)
-    sabnzbd.WEB_SERVER.run_in_thread()
+    try:
+        sabnzbd.WEB_SERVER.run_in_thread()
+    except Exception:
+        # The webserver runs in a separate thread; if it fails to start (bad cert,
+        # port taken between the free-check and bind, bad host) surface the error
+        # and abort instead of hanging. Details are also shown on the console.
+        logging.error(T("Failed to start web-interface: "), exc_info=True)
+        abort_and_show_error(browserhost, web_port)
 
     # Set URL for browser
     if enable_https:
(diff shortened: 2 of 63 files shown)


Safihre:
@claude review high level and security validations still correct

Safihre:
Checked the PR in the plane. 
Only comment would be to remove the first argument takes priority. External tools just need to fix their setup. Or do we also have problem internally?

Also noticed a few small functions or functions only called once inside another function. Are those really needed?

mnightingale:
> Only comment would be to remove the first argument takes priority. External tools just need to fix their setup. Or do we also have problem internally?

I'd suggest keeping it, it's only 3 lines.

> Also noticed a few small functions or functions only called once inside another function. Are those really needed?

I’ve implemented the secure cookies and removed or inlined some functions, note some are required or are a common  middleware pattern.

PR #3552, opened 2026-08-07
Refactor Config object

Safihre:
@mnightingale In #3546 I saw changes that made me realize that I wanted to refactor this for a while.
Personally, I like this version much better then the dataclass.
I hope you agree 😁

### DIFF
--- a/sabnzbd/config.py
+++ b/sabnzbd/config.py
@@ -751,6 +751,15 @@ class SABnzbdConfig(configobj.ConfigObj):
     is held in the module-global CONFIG; tests get a clean slate by replacing that instance.
     """
 
+    # INI sections that hold multiple named sub-sections, each backed by a Config* class.
+    # Single source of truth for the "special" sections handled differently from flat options.
+    SPECIAL_SECTIONS: dict[str, type] = {
+        "categories": ConfigCat,
+        "rss": ConfigRSS,
+        "servers": ConfigServer,
+        "sorters": ConfigSorter,
+    }
+
     def __init__(self, *args, **kwargs):
         # SABnzbd always reads and writes the INI as UTF-8
         kwargs.setdefault("encoding", "utf-8")
@@ -831,7 +840,7 @@ class SABnzbdConfig(configobj.ConfigObj):
 
         # Use CFG data to set values for all static options
         for section in self.database:
-            if section not in ("sorters", "servers", "categories", "rss"):
+            if section not in self.SPECIAL_SECTIONS:
                 for option in self.database[section]:
                     config_option = self.database[section][option]
                     try:
@@ -839,19 +848,11 @@ class SABnzbdConfig(configobj.ConfigObj):
                     except KeyError:
                         pass
 
-        # Define the special settings
-        if "categories" in self:
-            for cat in self["categories"]:
-                ConfigCat(cat, self["categories"][cat])
-        if "rss" in self:
-            for rss_feed in self["rss"]:
-                ConfigRSS(rss_feed, self["rss"][rss_feed])
-        if "servers" in self:
-            for server in self["servers"]:
-                ConfigServer(server, self["servers"][server])
-        if "sorters" in self:
-            for sorter in self["sorters"]:
-                ConfigSorter(sorter, self["sorters"][sorter])
+        # Build the special sections, each backed by its own Config* class
+        for special_section, section_class in self.SPECIAL_SECTIONS.items():
+            if special_section in self:
+                for name in self[special_section]:
+                    section_class(name, self[special_section][name])
 
         self.modified = False
         return True, ""
@@ -867,7 +868,7 @@ class SABnzbdConfig(configobj.ConfigObj):
             return False
 
         for section in self.database:
-            if section in ("sorters", "servers", "categories", "rss"):
+            if section in self.SPECIAL_SECTIONS:
                 if section not in self:
                     self[section] = {}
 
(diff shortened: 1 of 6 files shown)


mnightingale:
My changes were very much "what are the fewest changes I can make to fix the test bug", overall I much prefer this approach.

The first thing I noticed is `SPECIAL_SECTIONS: dict[str, type]` it makes little difference but I'd tighten that up.

```py
SpecialConfigTypes: TypeAlias = ConfigCat | ConfigRSS | ConfigServer | ConfigSorter
AllConfigTypes: TypeAlias = Option | SpecialConfigTypes

SPECIAL_SECTIONS: dict[str, type[SpecialConfigTypes]] = {...}
```

While dealing with related code `def get_servers(self) -> dict[str, ConfigServer]:` actually return `dict[str, AllConfigTypes]` I think it would end up with  `cast(dict[str, ConfigServer], self.database["servers"])` (eww)

I think it may be better to not merge them like that at all now that you've encapsulated it in a class anyway.

```py
self.database: dict[str, dict[str, Option]] = {}
self.categories: dict[str, ConfigCat] = {}
self.rss: dict[str, ConfigRSS] = {}
self.servers: dict[str, ConfigServer] = {}
self.sorters: dict[str, ConfigSorter] = {}
self.special_database: dict[str, dict[str, SpecialConfigTypes]] = {...}
```

This lets us resolve a bunch of warnings, `add_to_database` still uses `special_database` accessing the same object instances, but `get_servers()` etc. get the correct types. But someone messing with the object, `CONFIG.servers = {}` would make them out of sync - but that's not to dissimilar to what you can already do bypassing `add_to_database`.

Patch for above: https://gist.github.com/mnightingale/a609dca2161c69356e4e834d4f56d65a

Notifications page has the wrong title `#if $pane == "Email" then $T('cmenu-notif') else ""#`

`reload()` destroys the config on parse failure, it used to only switch them on success. This doesn't matter on normal startup (aborts with error) but `restore_config_backup` ignores the error and calls save_config(), so restoring a broken/corrupt ini writes partial state and saves.

Some other warnings in the file `'Option' object is not callable` and a couple in `OptionPassword`.

---

Some Claude things - 4 seems important

1. SPECIAL_SECTIONS is called the single source of truth, but get_dconfig:1027-1035 still hardcodes "categories" / "sorters" / ("servers", "rss"). Those need per-section behaviour so they can't all collapse into the map, but the tuple at line 1031 could at least be self.SPECIAL_SECTIONS.keys() - {"categories", "sorters"}.
1. CONFIG now exists from import with filename = None, so a save_config() before any read_config hits TypeError on filename + ".bak" instead of the old NameError. That's why test_misc.py:950 had to widen to bare except Exception. An early if not self.filename: return False in save_config would make it explicit and let that test narrow again.
1. read_config sets self.filename = path before parsing, so a failed load leaves the object pointing at the bad file. Combined with the previous point, a later save_config writes there.
1. Re-reading doesn't clear database, so special-section entries from a previous read survive and get written back — pre-existing, but restore-from-backup is now a live path for it.
1. The rest of the refactor is clean: no CFG_OBJ/CFG_DATABASE/CFG_MODIFIED/CONFIG_LOCK references remain anywhere, the @synchronized() no-arg form resolving to self.lock works, RLock keeps the re-entrant calls (read_config recursion, get_categories → save_config) safe, and none of the new attribute/method names collide with ConfigObj/Section internals (checked against the full member list).

Safihre:
I decided to not apply the gist for now, the changes feel forced and maybe better to change the whole config-backend if that's what we want.
I will merge this and then we can merge the xdist PR!

mnightingale:
@claude review, the primary goal was to replace several module-globals with a single instance. Take particular notice of any change of behaviour this introduces.

mnightingale (commented):
The action itself also checks permissions, according to the docs you could use `allowed_non_write_users`

> Action failed with error: Actor does not have write permissions to the repository

I'll run it locally:

1. Appears intentional and I have no problem with the silent errors are just if the config was used before loading but I don't see that as an issue
2. I would do the fresh instance thing, xdist is going to use monkeypatch to ensure that happens in a bunch of places anyway
3. Re-reading replacing special sections instead of merging, this change seems like the behaviour we would want and probably assumed it was already doing
4. Possibly pull the lock out of the instance, we already do this throughout the codebase
5. I have no issue with

The change look OK but just check points 1-3

---

The stated goal (four module globals → one instance, callers untouched) is met, and most of the conversion is genuinely behaviour-preserving: the `@synchronized()` no-arg form resolves to `self.lock` giving the same RLock re-entrancy, `del self[section][keyword]` matches `del CFG_OBJ[...]`, encoding still ends up `utf-8` after `reload()` (the constructor kwargs survive `_initialise`), and the module-level delegates keep every external caller identical. But several behaviour changes ride along, and two of them are load-bearing.

## Verified behaviour changes

### 1. `CONFIG` always exists, so pre-`read_config` misuse went from loud to silent

This is the core semantic shift of the refactor and everything else below descends from it. `CFG_OBJ` was a bare annotation — *unbound* until the first successful read — so any config access before startup finished raised `NameError: name 'CFG_OBJ' is not defined`. Now there is always a real instance with `filename = None` and an empty INI, and those paths quietly return falsy values instead:

| call before `read_config` | develop | this PR |
|---|---|---|
| `config.get_filename()` | `NameError` | `None` |
| `config.save_config()` | `NameError` | `False` |
| `api.py:1552` / `interface.py:717` `clip_path(get_filename())` | `NameError` | `None` reported as `configfn` |
| `api.py:787` `open(get_filename(), "rb")` | `NameError` | `TypeError` |

The PR leans into this deliberately — `tests/test_misc.py` now drops its `except NameError` and relies on `save_config()` returning `False`. That's a reasonable call, but it converts a class of "you used config too early" bugs from immediate crashes into silent no-ops. Since the new guard at [config.py:883](sabnzbd/config.py:883) is the one place that knows this happened, I'd log there:

```python
if not self.filename:
    logging.error("Cannot save settings, no INI file has been read yet")
    return False
```

### 2. `CONFIG.clear()` is not equivalent to the old `CFG_OBJ = ConfigObj()` — the tests now write real files

`tests/test_rss.py:137` swapped `sabnzbd.config.CFG_OBJ = configobj.ConfigObj()` for `sabnzbd.config.CONFIG.clear()`. Those aren't the same thing: rebinding gave a fresh object with `filename = None`, which made writing to disk *impossible*. `Section.clear()` wipes the INI contents but leaves `filename` and the whole option `database` intact — so the next `save_config` reconstructs a full INI from the database and writes it to whatever path was last read.

Measured, with `save_config` instrumented to report its return value:

```
tests/test_rss.py alone                    →  0 saves (filename is None)
tests/test_config.py + tests/test_rss.py   → 30 saves, all wrote=True,
                                              into tests/cache/test_config_backup/sabnzbd.ini
```

`test_config.py:192` leaves `CONFIG.filename` pointing at its own INI, and `test_rss` then rewrites it 30 times. The suite passes either way, so this is latent, but the comment sitting directly above the call — *"Need to create the Default category / Otherwise it will try to save the config"* — is now describing behaviour the code no longer has. It does save the config.

Fix is one line in the test (`sabnzbd.config.CONFIG.filename = None`), or better, do what the class docstring already claims and give the test a fresh instance: `sabnzbd.config.CONFIG = config.SABnzbdConfig()`.

Related artefact from the same mechanism: `clear()` removes `__version__` and `__encoding__` from the INI structure, and `save_config` never re-adds them (only `read_config` does). So a save after a clear produces a **version-less INI** — reread as `KeyError` → silently skipped by the version check. Test-only today, but it's the concrete proof that `clear()` shouldn't be treated as "reset to fresh".

## Behaviour changes worth a changelog line

### 3. Re-reading now *replaces* special sections instead of merging — [config.py:867](sabnzbd/config.py:867)

```python
if special_section in self.database:
    self.database[special_section].clear()
```

On develop, a second `read_config` (in practice: restore-from-backup) left previously registered servers, categories, RSS feeds and sorters in the database, and `save_config` wrote them back — so restoring a backup *merged* into the running config. Now they're dropped and only what's in the file survives. That's the right semantics for a restore and it fixes real stale-entry weirdness, but it is a user-visible change in what "restore backup" does, and it's independent of the globals goal.

Side effect: categories are momentarily empty after a read, until `get_categories()` re-seeds the defaults — which itself calls `save_config(True)`, so a restore can trigger a write mid-restore.

### 4. Per-instance lock, if instance-swapping becomes the test idiom

`CONFIG_LOCK` was a module singleton every path shared; the lock now belongs to the instance. Identical while there's exactly one — but the class docstring proposes "tests get a clean slate by replacing that instance", and that idiom is not safe as written: `Option.set()` resolves the module global `CONFIG` at call time ([config.py:118](sabnzbd/config.py:118)), so `Option` objects registered against the *old* instance would flip the *new* instance's `modified` flag while `@synchronized()` guards a different lock. Either don't advertise swapping, or note that the whole database must be rebuilt with it.

### 5. `Option.set()` now has a hard dependency on module-global `CONFIG`

Previously it flipped a plain bool; now it reaches for an object defined at the *bottom* of the module. Nothing constructs a `Config*` at module scope inside `config.py` today, but since all four call `set_dict()` from `__init__`, doing so would `NameError` on import. Worth a brief comment near `CONFIG = SABnzbdConfig()`.

## Changes unrelated to the stated goal

These are all improvements, but they're independent of "replace the globals" and each carries its own risk. If the maintainer wants this PR to stay reviewable as a pure refactor, they'd be easier to reason about split out:

- **Base `Option.__call__`** ([config.py:138](sabnzbd/config.py:138)) — plain `Option` instances become callable; `save_config`'s `self.database[section][option]()` no longer `TypeError`s on one. A latent-bug fix, currently only reachable from `test_config.py`.
- **`OptionPassword.get_string`** — instance-attribute rebinding replaced by a real override; `get_stars` annotation corrected to `str`. No runtime change (develop already returned `""`).
- **validate-before-`reload()`, the `save_config` filename guard, and the `restore_config_backup` result check** — the three fixes from the previous round.
- **`SPECIAL_SECTIONS.keys() - {"categories", "sorters"}`** ([config.py:1057](sabnzbd/config.py:1057)) — pure style, and arguably a regression in readability.

## Checked and confirmed *not* changed

- Lock re-entrancy for the nested paths (`read_config` recursion, `get_dconfig` → `get_ordered_categories` → `get_categories` → `save_config`) — all still one RLock on one thread.
- `reload()` preserves `lock`, `database` and `modified` (it only re-runs `_initialise` on the INI-structure attributes and `Section._initialise`) — verified directly.
- No attribute or method name on `SABnzbdConfig` collides with the `ConfigObj`/`Section` API.
- `get_ordered_sorters()` / `get_ordered_categories()` still iterate a live dict outside the lock — unchanged pre-existing raciness, not introduced here.
- No `CFG_OBJ` / `CFG_DATABASE` / `CFG_MODIFIED` / `CONFIG_LOCK` references remain anywhere in the tree.

## Bottom line

As a globals→instance conversion it's sound; the delegating wrappers keep the blast radius at zero for callers. The two items I'd want addressed before merge are **#2** (the test comment is now false and the suite is writing files it used to be structurally incapable of writing) and **#1** (decide explicitly whether silent-falsy is the wanted contract, and log if so). **#3** is a genuine user-facing change to restore-from-backup that belongs in the release notes regardless of which way it goes.

Safihre:
Isn't 3 a good thing?

mnightingale:
Yes I think so, I think it's saying previously we would only change those that were present. Now those new specials reset to default. If that's what happens it seems reasonable to me.

I probably should have said just to check 1 and 2, I think the change in 3 is what was intended to happen all along.

Issue #1304, opened 2019-06-21
SABnzbd thinks unraring fails when filename contains ?


puzzledsab:
The unpacking works but the SABnzbd decoder does not understand this output:
```
UNRAR 5.71 x64 freeware      Copyright (c) 1993-2019 Alexander Roshal


Extracting from C:\incomplete\testnzb\Great!.part01.rar

Cannot create C:\complete\_UNPACK_testnzb\Great?.mkv
The filename, directory name, or volume label syntax is incorrect.
WARNING: Attempting to correct the invalid file name
Renaming C:\complete\_UNPACK_testnzb\Great?.mkv to C:\complete\_UNPACK_testnzb\Great_.mkv
```

puzzledsab:
Don't know how to write backslash underscore so that Github doesn't think it's markup.

Safihre:
Use code-blocks :) with `

thezoggy:
>  WARNING: Attempting to correct the invalid file name
but that's not invalid. ! is legal on windows. would need to see debug lines for this to be useful

sanderjo:
No problem on my Linux, not with py2, not with py2.

So it must be Windows things, or a setting.

NZB: https://github.com/sanderjo/NZBs/blob/master/ExclamationMarkInResultingFile.nzb

puzzledsab:
@thezoggy : Yes, that's where it stops. There isn't any more useful stuff in the debug log, it only says " [newsunpack:746] ERROR: write error ..." and the output until the "Cannot create" line. I got the full output by running the same command on the commandline. It ends with
```
Extracting from C:\incomplete\testnzb\rarfile.part10.rar

...         Great?.mkv  OK
Total errors: 1
```

If you can create a file with a char that's invalid in whatever file system you are trying to unpack to it will probably give the same error. I'm not sure what the best way to fix it would be because it's hard to properly understand the output. It says OK after the filename when it has finished extracting but it's hard to know if all the files have been extracted successfully.

thezoggy:
just extract the rarset yourself and see the error as sab would.
you can see `Total errors: 1` so obviously its having a problem extracting.
I've been running sab on windows forever now and never had any problems with `!`.

and of course testing with sander's .nzb works just fine in 2.3.9
```
2019-06-21 18:10:19,960::INFO::[__init__:681] Adding ExclamationMarkInResultingFile.nzb
...
[...]

PR #3553, opened 2026-08-07
Unpack rar sets that par2 moved into a subfolder

mnightingale:
User described issue on Discord, and shared logs 5.1.0RC2

### What happens

Jobs finish "successfully" in seconds having unpacked nothing, leaving subfolders of raw `.rar`/`.rNN` volumes in the completed folder:

```
build_filelists(): rars: []
Unpacked files []
Moving …/Foo.Bar.S05E01…/…r00 => …/_UNPACK_…/Foo.Bar.S05E01…/…r00
```

If Direct Unpack is also running, post-processing moves and purges the job out from under unrar and the extraction aborts part-way.

### How to trigger

Only one thing is needed: **a job whose par2 set stores its files in a subdirectory** — `Sample/sample.mkv`, or a season pack with a folder per episode. No race involved; it fires every time.

SABnzbd sanitises `/` to `_` when assembling, so a file par2 calls `Sub/name.rar` lands on disk as `Sub_name.rar`. The names can therefore never match, `quick_check_set()` always takes its obfuscation branch, and the set is relocated into a newly created subfolder (`newsunpack.py:1707`, added in #1802). `build_filelists()` only scanned the download folder itself, found no archives, and `unpacker()` had nothing to do. Direct Unpack only changes which symptom you see. The two halves have conflicted since the non-recursive scan (8ce6c26f9, 2020) met subdirectory-creating renames (#1802, 2021).

### Changes

- **`NzbObject.relocated_directories()`** reports the subfolders par2-driven renames moved files into, derived from `nzf.filename` and `nzo.renames` — no new persisted state. `build_filelists()` gains an `extra_dirs` argument for them and the `unpacker()`/`parring()` callsites pass it. The download folder itself stays non-recursive, so the #1304 behaviour is unchanged: folders we did not create are still never scanned. A containment check keeps a hostile par2 from pointing the scan outside the job.
- **`wait_for_direct_unpacker()`** extracted from `rar_unpack()`, now also called from `unpacker()` and from `process_job()` before `parring()` — so it no longer depends on archives being found, and the DirectUnpacker is finished before quick-check renames the files it is reading.
- **`have_next_volume()`** confirms the volume is still on disk, so unrar aborts instead of spinning on a phantom volume that post-processing already moved.
- **`rename_deobfuscated_nzf()`** applies the de-obfuscation renames in `quick_check_set()` and `sfv_check()` to everything derived from the old name. `nzf.setname`/`nzf.vol` were left describing the old name, and the sets DirectUnpack had already finished stayed keyed by it — so `rar_unpack()` no longer recognised them as done and extracted them a second time on top of the existing output. `analyze_rar_filename()` now takes the basename as well, matching what `rar_unpack()` derives with `setname_from_path()`, and the cleanup of a completed set uses the volumes found on disk rather than the paths recorded before the rename.

### Tests

`TestUnpackNestedRarSets` covers a rar set one level down — that `build_filelists()` picks it up only when told about the folder, that it then gets unpacked, that the DirectUnpacker is waited on (parametrised against the flat layout as a control), plus the relocated-folder derivation, the setname/volume refresh in both rename paths, and that a set already completed by DirectUnpack is not extracted again.

---

This is from a direct unpack disabled run, results in quite a mess but I think it'll do for now.

<details>
  <summary>file tree - flat_unpack = 0</summary>

  - proof and sample are not in par2 data = root
  - mkv is extracted to the root
  - at the end the directories with nfo and sfv are moved
    
  ```
└── foo.bar.S05.1080p.BluRay.x264-blah
    ├── foo.bar.S05E01.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e01.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e01.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E01.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e01.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e01.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E02.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e02.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e02.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E02.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e02.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e02.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E03.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e03.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e03.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E03.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e03.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e03.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E04.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e04.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e04.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E04.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e04.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e04.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E05.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e05.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e05.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E05.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e05.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e05.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E06.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e06.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e06.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E06.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e06.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e06.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E07.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e07.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e07.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E07.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e07.1080p.bluray.x264-blah.proof.jpg
    ├── foo.bar.s05e07.1080p.bluray.x264-blah.sample.mkv
    ├── foo.bar.S05E08.1080p.BluRay.x264-blah
    │   ├── foo.bar.s05e08.1080p.bluray.x264-blah.nfo
    │   └── foo.bar.s05e08.1080p.bluray.x264-blah.sfv
    ├── foo.bar.S05E08.1080p.BluRay.x264-blah.mkv
    ├── foo.bar.s05e08.1080p.bluray.x264-blah.proof.jpg
    └── foo.bar.s05e08.1080p.bluray.x264-blah.sample.mkv
  ```
  
</details>


Safihre:
@claude this PR fixes the problem as described, great. However, it feels convoluted. Can we not solve this problem (par2 creating folders with rar's inside) in a simpler way?

mnightingale:
Lets have a Claude-off

tl;dr PR A simplify this to fix postproc not waiting for direct unpacker, PR B fix par2 subdirectory names at creation so the rename and tracking does not need fixing after

Or do both in this PR, which simplifies the overall solution

---

<html><head></head><body><p>Let me verify the review's claims before planning around them.</p>
<p>I read the thread — Safihre's "feels convoluted" and the bot's suggestion to track directories at <code>renamer()</code> instead of deriving them. Two corrections to that suggestion before the plan, then your sequencing question, which I think is the more important one.</p>
<h2>The bot's suggestion is incomplete</h2>
<p>It's right that <code>create_local_directories=True</code> has exactly one caller (<code>quick_check_set</code>, <a href="https://claude.ai/epitaxy/sabnzbd/newsunpack.py:1735">newsunpack.py:1735</a>) and that <code>renamer()</code> already validates containment. But it misses that <strong>par2cmdline creates directories itself during a real repair</strong>, restoring targets at their stored paths without going through <code>renamer()</code>. Those only surface via <code>PAR2_IS_MATCH_FOR_RE</code> → <code>nzo.renamed_file()</code> (<a href="https://claude.ai/epitaxy/sabnzbd/newsunpack.py:1506">newsunpack.py:1506</a>) — which is exactly why <code>relocated_directories()</code> reads <code>nzo.renames</code> as a second source. Tracking at <code>renamer()</code> alone would silently miss the repair path.</p>
<p>It also understates the cost: a tracked set is new state on <code>NzbObject</code>, so it needs a <code>NzbObjectSaver</code> entry plus a <code>__setstate__</code> fallback for existing pickles. The derived version needs no migration at all.</p>
<p>The one part I'd take from it now: <strong>the containment check in <code>relocated_directories()</code> is genuinely redundant</strong>. Both sources are already validated — <code>renamer()</code> refuses to escape, and par2cmdline names come from a completed repair. That's a few lines gone.</p>
<h2>Your par2 change: do it first, and it answers Safihre directly</h2>
<p>This is the real answer to "can we do this more simply". The PR feels convoluted because it's compensating for a self-inflicted wound: SABnzbd flattens <code>Sub/name.rar</code> to <code>Sub_name.rar</code> at assembly, then un-flattens it mid-post-processing, then every derived thing has to be repaired afterwards. Stop flattening and most of the machinery has nothing to compensate for.</p>
<p>Concretely, here's what the naming change does to each piece:</p>

Piece | After par2-correct names
-- | --
rename_deobfuscated_nzf() + success_sets re-key | Mostly gone. Names match, so quick_check_set() takes the file == nzf.filename branch and never renames. Survives only as a safety net for jobs where the par2 lands after the file was written.
nzo.renames as a directory source | Gone. par2cmdline finds files already at their stored paths, so no repair-time renames to track.
Containment check in relocated_directories() | Gone. The sanitizer guarantees locality by construction — it splits on / and drops .. before any path exists.
The subfolder-aware scan | Still needed, but becomes nzo.sub_directories() derived purely from nzf.filename. Not "where did things get relocated to", just "where this job's files are".
wait_for_direct_unpacker() | Still needed. PP moving and purging while unrar reads is independent of naming.
have_next_volume() disk check | Still needed. Same.
analyze_rar_filename() basename | Still needed — needed because nzf.filename now carries a path.


<p>I checked the timing assumption it rests on: the small main <code>.par2</code> is fetched among the first articles and <code>handle_par2()</code> → <code>postpone_pars()</code> → <code>verify_all_filenames_and_resort()</code> populates <code>md5of16k</code> before most data files are written. So correct-from-the-start is the normal case, not the lucky one.</p>
<p><strong>So: first, not after, and not instead.</strong> Not "instead" because the scan fix is still required either way — the volumes are in a subfolder regardless of who put them there. Doing it after means shipping compensation machinery you're about to delete, and Safihre reviewing exactly the part you'd remove.</p>
<h2>Proposed plan</h2>
<p>I'd split rather than grow this PR, because the naming change touches the assembler write path and has a security dimension that doesn't belong in an RC bugfix:</p>
<p><strong>PR A — Direct Unpack safety (this PR, reduced).</strong> <code>wait_for_direct_unpacker()</code> extracted and called from <code>unpacker()</code> and <code>process_job()</code>; <code>have_next_volume()</code> disk check. Self-contained, no naming involvement, fixes the abort-and-purge-under-unrar from the first log. Easy review.</p>
<p><strong>PR B — par2 subdirectory names.</strong> <code>sanitize_filename(allow_subdirs=True)</code> splitting on <code>/</code>, sanitizing per component, dropping <code>.</code>/<code>..</code>/leading slashes; <code>get_unique_filepath()</code> creating the folder and returning <code>(filename, path)</code>; <code>analyze_rar_filename()</code> on the basename; the scan fix with the simplified derivation. This is where the headline "nothing gets unpacked" bug actually dies, and it lands as a design fix rather than a workaround.</p>
<p>The cost of splitting: PR A alone doesn't fix the headline symptom, so users stay broken until B. If that's not acceptable for the RC, the alternative is one PR containing A + B, dropping <code>rename_deobfuscated_nzf()</code> and the <code>nzo.renames</code> source — still a smaller diff than what's on the branch now, and a much better story for the review.</p>
<p>I have the PR B work from earlier in this session (sanitizer, <code>get_unique_filepath</code>, the knock-ons, tests across all three platforms) — it was dropped from the branch but I can reconstruct it. Which split do you want?</p></body></html>Let me verify the review's claims before planning around them.

I read the thread — Safihre's "feels convoluted" and the bot's suggestion to track directories at `renamer()` instead of deriving them. Two corrections to that suggestion before the plan, then your sequencing question, which I think is the more important one.

## The bot's suggestion is incomplete

It's right that `create_local_directories=True` has exactly one caller (`quick_check_set`, [[newsunpack.py:1735](https://claude.ai/epitaxy/sabnzbd/newsunpack.py:1735)](sabnzbd/newsunpack.py:1735)) and that `renamer()` already validates containment. But it misses that **par2cmdline creates directories itself during a real repair**, restoring targets at their stored paths without going through `renamer()`. Those only surface via `PAR2_IS_MATCH_FOR_RE` → `nzo.renamed_file()` ([[newsunpack.py:1506](https://claude.ai/epitaxy/sabnzbd/newsunpack.py:1506)](sabnzbd/newsunpack.py:1506)) — which is exactly why `relocated_directories()` reads `nzo.renames` as a second source. Tracking at `renamer()` alone would silently miss the repair path.

It also understates the cost: a tracked set is new state on `NzbObject`, so it needs a `NzbObjectSaver` entry plus a `__setstate__` fallback for existing pickles. The derived version needs no migration at all.

The one part I'd take from it now: **the containment check in `relocated_directories()` is genuinely redundant**. Both sources are already validated — `renamer()` refuses to escape, and par2cmdline names come from a completed repair. That's a few lines gone.

## Your par2 change: do it first, and it answers Safihre directly

This is the real answer to "can we do this more simply". The PR feels convoluted because it's compensating for a self-inflicted wound: SABnzbd flattens `Sub/name.rar` to `Sub_name.rar` at assembly, then un-flattens it mid-post-processing, then every derived thing has to be repaired afterwards. Stop flattening and most of the machinery has nothing to compensate for.

Concretely, here's what the naming change does to each piece:

| Piece | After par2-correct names |
|---|---|
| `rename_deobfuscated_nzf()` + `success_sets` re-key | **Mostly gone.** Names match, so `quick_check_set()` takes the `file == nzf.filename` branch and never renames. Survives only as a safety net for jobs where the par2 lands after the file was written. |
| `nzo.renames` as a directory source | **Gone.** par2cmdline finds files already at their stored paths, so no repair-time renames to track. |
| Containment check in `relocated_directories()` | **Gone.** The sanitizer guarantees locality by construction — it splits on `/` and drops `..` before any path exists. |
| The subfolder-aware scan | **Still needed**, but becomes `nzo.sub_directories()` derived purely from `nzf.filename`. Not "where did things get relocated to", just "where this job's files are". |
| `wait_for_direct_unpacker()` | **Still needed.** PP moving and purging while unrar reads is independent of naming. |
| `have_next_volume()` disk check | **Still needed.** Same. |
| `analyze_rar_filename()` basename | **Still needed** — needed *because* `nzf.filename` now carries a path. |

I checked the timing assumption it rests on: the small main `.par2` is fetched among the first articles and `handle_par2()` → `postpone_pars()` → `verify_all_filenames_and_resort()` populates `md5of16k` before most data files are written. So correct-from-the-start is the normal case, not the lucky one.

**So: first, not after, and not instead.** Not "instead" because the scan fix is still required either way — the volumes are in a subfolder regardless of who put them there. Doing it after means shipping compensation machinery you're about to delete, and Safihre reviewing exactly the part you'd remove.

## Proposed plan

I'd split rather than grow this PR, because the naming change touches the assembler write path and has a security dimension that doesn't belong in an RC bugfix:

**PR A — Direct Unpack safety (this PR, reduced).** `wait_for_direct_unpacker()` extracted and called from `unpacker()` and `process_job()`; `have_next_volume()` disk check. Self-contained, no naming involvement, fixes the abort-and-purge-under-unrar from the first log. Easy review.

**PR B — par2 subdirectory names.** `sanitize_filename(allow_subdirs=True)` splitting on `/`, sanitizing per component, dropping `.`/`..`/leading slashes; `get_unique_filepath()` creating the folder and returning `(filename, path)`; `analyze_rar_filename()` on the basename; the scan fix with the simplified derivation. This is where the headline "nothing gets unpacked" bug actually dies, and it lands as a design fix rather than a workaround.

The cost of splitting: PR A alone doesn't fix the headline symptom, so users stay broken until B. If that's not acceptable for the RC, the alternative is one PR containing A + B, dropping `rename_deobfuscated_nzf()` and the `nzo.renames` source — still a smaller diff than what's on the branch now, and a much better story for the review.

I have the PR B work from earlier in this session (sanitizer, `get_unique_filepath`, the knock-ons, tests across all three platforms) — it was dropped from the branch but I can reconstruct it. Which split do you want?

mnightingale:
Since this is for 5.2 I've fixed the underlying "par2 allows sub-directories" issue, which is the cause.

> **One correction to what I told you earlier**
>
>I said the naming change would let you delete 685a3c5dc and 6c102aabc. Having built it, that's not right — keep both. quick_check_set() still renames when the par2 lands after the file was written (verify_nzf_filename returns early once nzf.filepath is set), and when that happens the setname staleness and the success_sets key mismatch are exactly as before. What changes is that they're now a narrow fallback rather than the main path — which is still a much better answer for Safihre than the current branch, but the diff doesn't shrink as much as I implied.
>
> Two things I did not do, for the record: the containment check in sub_directories() stays, because nzo.renames keys come from parsing par2cmdline output and are never sanitized (the bot's review missed that repair path entirely). And the existing_filename in nzf.filename substring heuristic in check_existing_files() is untouched — retry matching works because verify_nzf_filename() always records the par2 name as a rename, so there's a mapping to match on; a file that reached a subfolder with no rename recorded would still miss, but I couldn't construct a real path that does that.

AFAIK you can't do the last thing it mentioned because it would need the name to come from the nzb/yenc, which I don't think we'd use with directories.

So I didn't remove as much as hoped but most (66%) of the changes are for tests.

What I have not addressed is although we download to the correct paths, we extract to the download_path. However, that should be handled in a different PR.

PR #3559, opened 2026-08-12
Persistent sessions, CSRF tokens, and no apikey in the web interface

### LINKED ISSUE #3063: setting to specify the cookie duration
### Description

14 days for "remember me" is quite short - at least in my opinion.

My idea would be to leave the 14 days as a default, but add a setting to change the time the cookie stays valid.

mnightingale:
Replaces the web interface's authentication with persistent sessions, takes the apikey out of the UI, and adds CSRF protection.

| Before | After |
|---|---|
| Sessions died on restart, bound to an address | Survive restarts, capped at 90 days, revoked on credential change |
| apikey in every page's HTML | Never in a page; refused on page routes |
| CSRF rested on `SameSite` alone; `/api` unprotected | Per-session token on every state change |
| Unlimited password guessing | 5 attempts per address, then a 5-minute cooldown |

## Sessions

- A dict saved to `admin/sessions.sab` with `save_admin`/`load_admin`, like the queue — no database, no new dependency. Persisted so a restart does not log everyone out
- Cookie holds a 256-bit random token, of which only the SHA-256 hash is stored. 14-day idle timeout, sliding on use, with a 90-day cap from creation
- Each session stores a fingerprint of the credentials, so changing the username or password revokes every existing session
- `HttpOnly`, `SameSite=Strict` and `Secure` decided per request, so it works behind a TLS-terminating proxy

## CSRF protection

- Per-session token, `hmac(key, cookie)` — nothing stored, no lookup, and it dies with the session
- Required on every state-changing page POST, and as a header on cookie-authorised `/api` calls so it cannot ride in a URL
- Not gated on whether a login is configured: `SameSite` is not origin-scoped, so another port on the same host is same-site and its forms do carry the cookie
- `/shutdown`, `/logout`, the wizard steps and the log download became POSTs

## The apikey

- No page embeds it; `check_api_key` is gone from ~35 config routes, and an apikey on a page route is refused with a message pointing at `/api`
- New `/log` route serves the log as a guarded POST, so `mode=showlog` needs no CSRF exemption and stays available for automation

## Breaking changes

- **An apikey no longer opens the config `*_save` routes.** Use the matching `mode=set_config` call; the 403 names the replacement.
- A new `sessions.sab` in `admin_dir`, excluded from config backups.

## Future work

- Stop showing the apikey at all — the end state is a stored hash, shown once when generated
- Hash the ini password: `credential_fingerprint` becomes `sha256(username + ":" + hash)`, provided the stored hash is stable
- Persist the CSRF and anonymous keys, so a page left open across a restart still holds a valid token

---

Fixes #3063

### DIFF
--- a/builder/package.py
+++ b/builder/package.py
@@ -246,7 +246,7 @@ def test_sab_binary(binary_path: str):
 
         # Parse API-key so we can do a graceful shutdown
         sab_config = configobj.ConfigObj(os.path.join(config_dir, "sabnzbd.ini"))
-        urllib.request.urlopen(base_url + "shutdown/?apikey=" + sab_config["misc"]["api_key"], timeout=10)
+        urllib.request.urlopen(base_url + "api?mode=shutdown&apikey=" + sab_config["misc"]["api_key"], timeout=10)
         sabnzbd_process.wait()
 
         # Print logs for verification
--- a/interfaces/Config/templates/_inc_header_uc.tmpl
+++ b/interfaces/Config/templates/_inc_header_uc.tmpl
@@ -41,10 +41,10 @@
         var formWasSubmitted = false;
 
         // Information we need
-        var sabSession = '$apikey';
+        var csrfToken = '$csrf_token';
         var rootURL = '$url()'
         var urlBase = '${url_base}'
-        var folderBrowseUrl = "$url('api', mode='browse', output='json', apikey=$apikey)";
+        var folderBrowseUrl = "$url('api', mode='browse', output='json')";
         var folderSeperator = '#if $os.sep == '\\' then '\\\\' else '/'#'
 
         // Translations
--- a/interfaces/Config/templates/config_cat.tmpl
+++ b/interfaces/Config/templates/config_cat.tmpl
@@ -28,7 +28,6 @@
                                 <span class="glyphicon glyphicon-option-vertical"></span>
                             </td>
                             <td>
-                                <input type="hidden" name="apikey" value="$apikey" />
                                 <input type="hidden" name="order" value="$slot.order" />
                                 <input type="hidden" value="$slot.name" name="name" />
                                 <!--#if $slot.name != '*'#-->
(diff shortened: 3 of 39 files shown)


Safihre:
Let's move all the auth/cookie/etc stuff outside interface.py to separate file.

To me, oldschool me, the CSRF protection really seems something an application like would not need. When would this really be a risk? For how many % of users?
You know by now that my rule is that we keep things out of SABnzbd that don't affect less than 1-5% of users.

Maybe we can also ask Claude to be a bit more concise in it's code-comments? But that's not really important 😁

Oh and fix the build of course.

mnightingale:
> To me, oldschool me, the CSRF protection really seems something an application like would not need. When would this really be a risk? For how many % of users?
> You know by now that my rule is that we keep things out of SABnzbd that don't affect less than 1-5% of users.

It is not an addition, it's a replacement. Currently every state changing request requires the apikey (not just the api) and a foreign page can't - that is CSRF protection.

It's protecting some site being able to make requests to a SAB instance. Specifically anything served as the same site.

The simplest case it protects against is if SAB was configured for local access, with no username and password.

```
POST /config/switches/save (no cookie, no token, no key) -> 403 "Missing or invalid session token"
```

The CSRF check is the only thing preventing that request.

Percentage wise, it's at least every install without a login.

Additionally we use SameSite scoped cookies, they are site-scoped not origin-scoped so every app even on different ports counts as the same site. Therefore, requests do send cookies. So if Sonarr was at 127.0.0.1:8181, or SomeRandomCredentialStealingApp on 127.0.0.1:1337 they could potentially issue requests to SAB with cookies.

mnightingale:
Rebased and cleaned up the commits, recent api_route fix is restricted to just /api routes because the interface pages no longer accept apikeys at all. Adapted `test_check_apikey_ignores_mode_off_api_route` and `test_apikey_does_not_open_the_page_routes` tests similar for page routes.

From a security design perspective this split of api vs cookie authentication is a good argument for this direction, it completely separates two different authorisation decisions so such issue is less likely.

I also removed most moves to `security.py`, can follow up after but it made this PR look too complicated and hid the minor changes I did make in `interface.py`

PR #3560, opened 2026-08-12
Less hacky cache busting for Auto.css

mnightingale:
## Why?

Because it's bound to catch us out again in the future

Also the previous fix was only applied to glitter


Safihre:
I don't like this. My Claude also came up with it 😂
If anything, we can add something like $commit that is replaced in the builder, in the patch_version.

mnightingale:
I have a long term fix to modernise it properly but I'll wait until next week to start trying to convince you.

For now at least apply the same horrible fix in `interfaces/Config/templates/staticcfg/css/Auto.css` so that it applies to the config pages also.

PR #3562, opened 2026-08-13
Write articles through sabctools.FileWriter, and optionally straight from the decoder

mnightingale:
## Goal

Every article currently crosses back into Python at each step:

```
Downloader (Py) > Unlocked SSL (C) > Decoder (C) > Article Cache (Py) > Assembler (Py) > Disk write (Py/C)
```

This shortens it to:

```
Downloader (Py) > Unlocked SSL (C) > Decoder (C) > Disk write (C)
```

## 1. All article writes go through FileWriter (unconditional)

- `Assembler.assemble`, `assemble_article` and `open` use `sabctools.FileWriter` instead of `os.pwrite` / `lseek`+`write` / `sabctools.sparse`
- Drops `nzf.file_lock` from the write path; it only existed because `os.pwrite` is Unix-only, so Windows had to seek and write as two steps
- The handle is kept on the `NzbFile` and reused, bounded by an LRU of 32 open writers
- Evicted writers are dropped, not closed - another thread may be mid-write, and the last reference closes it once that returns
- Same bytes, same offsets, same append/positional decisions

## 2. `direct_decode` (opt-in, off by default)

The decoder writes each article's decoded bytes straight to its offset, so it never enters the article cache or reaches the assembler. Used only when `direct_write` is on, the file is yEnc, the filename is resolved, and it is not the first article of a file (`decode_yenc` needs those bytes for `md5of16k`). Otherwise the existing path runs.

- A failed write fails the article, not the connection: it is refetched, and anything already written is overwritten at the same offset
- **Natural backpressure**: the write happens inside the decode call with the GIL released, so a disk that cannot keep up blocks that receive thread rather than filling a cache. Download rate self-limits to write rate, including while par2 or unrar are competing for the same disk

Same 46.6 GB job, both capped at 250 MB/s so they are comparable:

|                   | cache + assembler | direct_decode |
|-------------------|------------------:|--------------:|
| peak RSS          |          867.7 MB |      243.5 MB |
| total CPU         |           147.9 s |       138.8 s |
| assembler CPU     |            16.3 s |         0.1 s |
| max write latency |          169.5 ms |        4.8 ms |

### Why off by default

Articles are written as they arrive, so the writes land all over the file instead of being grouped together. On an SSD that is fine until at least 10 Gbps on most NVMe SSDs. On a drive that has to seek for each one it is slower, and SABnzbd cannot currently tell the two apart.

Testing the disk at startup does not work either, because the drive's own cache hides how fast it really is.

The answer is for SABnzbd to watch how long its writes actually take and adjust while downloading. That comes in a later PR. Until then, turning this on is up to the user.

## Testing

- Byte-identical output across all three write paths: streamed, cached-positional, append
- Missing articles with par2 removed entirely, so nothing can mask a misplaced write: every byte outside the holes matches the source, and the holes are exactly the missing articles. With par2 present it repairs to a byte-identical file
- Articles served only by a backup server land at the same offsets, no repair needed


mnightingale:
"Build macOS binary" only fails because it's building sabctools from source and the `CFLAGS / ARCHFLAGS` breaks the configure step, sabctools has to be compiled per architecture then combined as a normal wheel build does.
Once requirements.txt points at a published version it'll use the wheels and not cause a failure, the alternative is installing sabctools separately with no CFLAGS so it builds a universal2 wheel but not worth it when usually there will be wheels.

PR #3572, opened 2026-08-16
Use root relative URLs

mnightingale:
It could break behind proxies otherwise.

Assets are included as `href="/static/bootstrap/css/bootstrap.min.css"` rather than `href="https://domain/static/..."`

Needed to also change the access URL shown at the end of the wizard, so made it sort so https are first then alphabetically

### DIFF
--- a/sabnzbd/api.py
+++ b/sabnzbd/api.py
@@ -1920,7 +1920,7 @@ def clear_trans_cache():
     sabnzbd.WEBUI_READY = True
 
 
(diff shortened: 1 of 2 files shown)


Safihre:
I was wondering how this was going to work when I saw the uvicorn pr.
What was the reason to not use relative urls in first place?

mnightingale:
I needed the full URL for access URL so I figured why not and used the same helper
Now I know 😆

PR #3573, opened 2026-08-16
Fix pytest-xdist failure due to selenium driver in class global

mnightingale:


Safihre:
Only have to make make Ruff happy

Safihre:
Tests complete so fast!

mnightingale:
It's great locally, 25-30s, uvicorn also helped a bit because it starts and stops quickly
There are still a few slow tests we can improve, but they're mostly where the UI has a timeout before updating

PR #3576, opened 2026-08-17
Fix rss clearing at startup and downloaded items still in the feed

### LINKED ISSUE #3574: RSS items cleared from the RSS database while still present in the feed
### SABnzbd version

5.1.0

### Operating system

n/a

### Using Docker image

None

### Description

Will investigate https://forums.sabnzbd.org/viewtopic.php?p=134649 - I suspect they don't run 24/7 and we purge at startup which was a mistake

> SABnzbd 5.1.0 broke RSS functionality because downloaded RSS items are now automatically cleared from the RSS database after 3 days while still present in the feed

Also requested:
- clear limit value of 3 days a special option so we can set it to a much higher value like 90 or even 365 days globally

mnightingale:
Fixes #3574

~~As requested I also wanted to add a `rss_retention` config option, I've put it on /config/rss instead of specials because  history retention is part of the normal switches config screens and it's on it's own page anyway so doesn't look cluttered.~~

- Increased default retention to 7 days
- ~~Configurable retention on /config/rss~~ - separate PR
- Only purges unconfigured rss feeds at startup, and only expired items so removing and readding with the same name doesn’t purge immediately
- Only purges if the feed was `readout=True`
- Main reported issue is because `seen_at` was not updated for downloaded items

### DIFF
--- a/sabnzbd/__init__.py
+++ b/sabnzbd/__init__.py
@@ -536,8 +536,8 @@ def delayed_startup_actions():
     # aren't on 24/7 and typically don't benefit from the daily scheduled call at midnight
     sabnzbd.database.scheduled_history_purge()
 
-    # Purge links older than 3 days
-    sabnzbd.rss.expired_purge()
+    # Drop leftover records of feeds that are no longer configured
+    sabnzbd.rss.purge_removed_feeds()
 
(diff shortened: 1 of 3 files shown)


mnightingale:
Configurable retention will move to a separate PR

PR #3580, opened 2026-08-18
Add xff_trusted_hosts and refuse requests whose client a proxy hides

mnightingale:
## Background

Behind a reverse proxy, requests arrive from the proxy's address, so the real client is invisible. Proxies add an `X-Forwarded-For` header naming the client. Anyone can send it, so it only counts coming from a proxy you trust.

## The problem

When SABnzbd did not read the header it fell back to the address that opened the connection: the proxy. A proxy is usually on the LAN, so everyone behind it looked local, gaining local-level access and, at `inet_exposure: 5`, no login prompt. `verify_xff_header: off` both skips the header and is what people turn off when a proxy is denied, so the workaround for one problem opened this one. 5.1.x behaved the same way.

## What this changes

Unless the client was resolved from a forwarded header, the request is now refused rather than credited to the proxy. `inet_exposure` still governs what an unknown client may reach, so an instance deliberately published behind a proxy keeps working. A forged header grants nothing: it is read only from a trusted peer, and the rightmost untrusted hop wins.

New setting `xff_trusted_hosts` says who may speak for other clients, and grants no access of its own. `local_ranges` used to decide both; it now only narrows that trust, never widens it.

Two smaller fixes ride along: old-style `local_ranges` entries never matched in the header parser, and log lines naming a client went unredacted in `showlog`, failed logins among them.

## Breaking

Three setups change, each because the old behaviour handed one host's access to others:

- **`verify_xff_header: off` behind a proxy** is now refused. Set `xff_trusted_hosts` and leave verification on.
- **A public range in `local_ranges`** keeps its own access but can no longer name other clients. Being called local should not confer the right to speak for someone else.
- **6to4 and Teredo clients** are no longer local. Both are globally routable, so treating them as a LAN let remote clients in. Tailscale's `100.64.0.0/10` is unchanged, never having counted as local.

Nothing changes without a proxy, or where the forwarded chain is entirely local. Behind Cloudflare, put its published ranges in `xff_trusted_hosts`, not `local_ranges`. Clients then count as external, so `inet_exposure` and a password apply, and it is only sound if your origin is unreachable except through your own Cloudflare account.

Refusals name the address that connected and the setting that would admit it, and a range that loses trust says so at start-up, so nobody has to read this to fix their own setup.

A forwarded header is something a scanner sends as readily as a proxy, so that detail is kept for a peer that could be one of ours: on the local network, or already listed in `xff_trusted_hosts`. Such a peer gets the reason in the response as well as a warning in the log. Anyone else is told only that they were refused, and it is logged at info, so a passing scan neither learns the proxy's address nor fills the log with advice to trust the scan itself. Being listed counts wherever the proxy sits, so a Cloudflare origin whose header names nobody is still reported.

```
403 External internet access denied - https://sabnzbd.org/access-denied - X-Forwarded-For received while verify_xff_header is off, turn it on and list 192.168.0.201 in xff_trusted_hosts

Refused connection from: 192.168.0.201:47512 [curl/8.0] - X-Forwarded-For received while verify_xff_header is off, turn it on and list 192.168.0.201 in xff_trusted_hosts
Refused connection from: 104.16.0.1:47512 (X-Forwarded-For: ,,,) [curl/8.0] - X-Forwarded-For from 104.16.0.1 does not name a client address
Refused connection from: 192.168.0.201:47512 [curl/8.0] - Forwarded is not supported, configure the proxy to send X-Forwarded-For instead
100.64.0.0/10 in local_ranges is not a private network, list it in xff_trusted_hosts to trust it
```

## Summary by CodeRabbit

* **Security**
  * Improved handling of proxy and forwarded-client headers with trusted-network validation.
  * Untrusted, malformed, or unresolved forwarded addresses are rejected before access checks.
  * Added protected settings for configuring trusted forwarded hosts.

* **Bug Fixes**
  * Improved sanitization of logged addresses, including IPv6 addresses, ports, and proxy metadata.
  * Restricted trusted proxy ranges to valid local networks and improved invalid-range handling.

* **Performance**
  * Static files now use long-lived caching for faster repeat access.

* **Tests**
  * Expanded coverage for proxy trust, forwarded headers, address validation, IPv6 handling, and log sanitization.


mnightingale:
I also have some changes in progress for `wiki/extra/howto-reverse-proxy` but you might have remind me nearer release since that doesn't really want merging yet

Safihre:
@jcfp If you have time, what do you think? The TLDR is that we become more restrictive.

jcfp:
Starlette/uvicorn is designed to work this way, that alone makes switching over to explicitly allowed proxies a logical choice. The original xff implementation was intended to avoid letting everyone in just because the proxy's client ip appeared local to sab, nothing more. Most users will be fine, with extra restrictions only really affecting those using some exotic ip ranges (test nets and the likes) on their local network.

One thing though: if there's any chance of the `"X-Forwarded-For received while verify_xff_header is off, turn it on and list %s in xff_trusted_hosts"` msg showing up for requests made by random clients sending a fake xff header, we should add an explicit "**to allow such requests, do this and that...**" in there. I understand the target audience is users that need to adapt their setups to the new design on the sab side (for whom this is valid advice), but the current wording could trigger other users to make these config changes too - even when inappropriate. A significant part of the userbase lacks the level of understanding of networking and protocols required to make that distinction on their own.

mnightingale:
> One thing though: if there's any chance of the "X-Forwarded-For received while verify_xff_header is off, turn it on and list %s in xff_trusted_hosts" msg showing up for requests made by random clients sending a fake xff header

That's a good point, it shouldn't be a `logging.warning` just an error message returned to the client on refused requests. Then the user only sees the error and advice for their own requests.

jcfp:
> Then the user only sees the error and advice for their own requests.

Better, although that comes with its own concern: the error message as-is could leak information about the local network topology to outsiders (by mentioning the internal address of the proxy). On the other hand, listing the address is very convenient for legit users.

Safihre:
@coderabbitai review

mnightingale:
I have tweaked the error messages a little, messages are only returned to the client if `cfg.api_warnings()` is enabled (unchanged) so that's probably where I'd expect users to remaining going to stop getting the messages.

Some are downgraded to info rather than warning, but some will continue with warnings.

PR #3582, opened 2026-08-20
Fix applying changes to RSS filters

mnightingale:
https://forums.sabnzbd.org/viewtopic.php?p=134661&sid=6cffff58ad01eff79bcfe3006d9aee1b#p134661

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -1006,7 +1006,6 @@ class RSSReader:
                     new_links.add(entry.link)
 
                     downloaded = self._process_entry(
-                        repo,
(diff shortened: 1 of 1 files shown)


Safihre:
I don't understand the changes but I trust you 😂

mnightingale:
`gen = repo.get_feed_jobs(feed=feed)` queries the DB, lazily getting entries.
`repo.upsert` and `self.enqueue_download` used the same connection and changed the result set so the filter changes only apply to the first item

With the pool of connections it's no bother to just temporarily grab one for each change, so one queries for the existing items and one performs updates to them.

PR #3583, opened 2026-08-20
Add configurable speedgraph duration

QuixThe2nd:
The header speedgraph in Glitter was hardcoded to 275 samples, which is about 4.5 minutes at the default 1s refresh rate. This makes the graph window configurable from 1 minute up to 24 hours.

**Changes:**
- New `speedgraph_duration` setting (misc section, default 300s = 5 minutes, allowed range 60-86400)
- Dropdown in Glitter's interface options (next to refresh rate): 1m / 5m / 15m / 30m / 1h / 3h / 6h / 12h / 24h
- `BPSMeter` now retains up to 24h of 1-second samples (`BPS_LIST_MAX`), and `get_bps_list()` returns only the configured window, still strided by the refresh rate
- The Glitter client caps its history at `speedgraph_duration / refresh_rate` samples instead of a hardcoded 275, so per-device refresh rates keep working
- Like the refresh rate, the setting is saved to the config when global interface settings are used, and kept per-device otherwise

**Testing:**
Ran 5.2.0-dev with the change and verified via the API and rendered page:
- `get_config&section=misc` exposes `speedgraph_duration` (default 300), `set_config` updates it live
- Main page renders `var sabSpeedgraphDuration = <value>;` and seeds `sabSpeedHistory` with exactly the configured number of samples (3600 after setting 1h, 300 after reverting)
- The new dropdown and label appear in the interface options modal

Note: the new `Glitter-speedgraphDuration` string was added to `skintext.py`; translations for other languages still need to be extracted.


thezoggy:
you can change the update stuff in glitter via: click wrench > web interface
if you were going to add an option for it, it should ideally go on that page.

I'm not sure there is much value in showing graph so far in the past when people wanting to look at it is only the recent stuff ~5min. but I'll defer to others if they think its worth adding

Safihre:
I think this is not worth adding. 
If we wanted real graphing, we should add a graphing library (like nzbget).
This graph is really only for graphic flash-view.

PR #3585, opened 2026-08-21
Add SafeUnpickler to guard against pickle-attacks

Safihre:


sabnzbd/filesystem.py
```diff
@@ -1153,6 +1153,39 @@ def get_new_id(prefix: str, folder: str, check_list: Optional[list] = None) -> s
...
+
+class SafeUnpickler(pickle.Unpickler):
+    """Unpickler that blocks code-execution gadgets: only sabnzbd classes and safe data types"""
+
+    def find_class(self, module, name):
+        if module == "sabnzbd" or module.startswith("sabnzbd."):
+            # Wildcard our own package, but classes defined in it only (never re-exported callables)
+            obj = super().find_class(module, name)
+            if isinstance(obj, type) and getattr(obj, "__module__", "").startswith("sabnzbd"):
+                return obj
```
mnightingale: I think we can either be more restrictive here by not allowing vulnerabilities from anything in `sabnzbd.*`

Either check for `__module__ == "sabnzbd.nzb"` or use the `_SAFE_GLOBALS` and be explicit about `("sabnzbd.nzb", "file")` etc. that at least reduced the attack surface.

Also note `sys.modules["sabnzbd.nzbstuff"]` so either needs a prefix check or be explicit.

For example, `os.kill(pid, SIGKILL)` is reachable via `sabnzbd.utils.kronos.ForkedScheduler.__del__` and you can probably get to anything else we import.

sabnzbd/filesystem.py
```diff
@@ -1153,6 +1153,39 @@ def get_new_id(prefix: str, folder: str, check_list: Optional[list] = None) -> s
...
+    ("collections", "deque"),
+    ("builtins", "set"),
+    ("builtins", "frozenset"),
+    ("builtins", "bytearray"),
+    ("builtins", "complex"),
+    ("copyreg", "_reconstructor"),
+}
+
+
+class SafeUnpickler(pickle.Unpickler):
```
mnightingale: Nitpicking but I wouldn't describe anything using pickler as safe, RestrictedUnpickler maybe

sabnzbd/filesystem.py
```diff
@@ -1153,6 +1153,39 @@ def get_new_id(prefix: str, folder: str, check_list: Optional[list] = None) -> s
     raise IOError
 
 
+# Non-sabnzbd globals our pickles reference: plain data types only, nothing that runs code
+_SAFE_GLOBALS = {
```
mnightingale, line 1159: dirscanner saves `os.stat_result`, but I wouldn't add it to the allow list, instead make it store the properties it needs, losing previous dirscanner state does not matter

mnightingale (commented):
I've made my thoughts on pickler known, but I think this closes most of the risk but I'm not confident in that.

A bunch of them don't particularly need pickler at all, especially in jobs folders so they could just be JSON and reduce the attack surface eventually 

- bpsmeter
- postproc queue (nzo_id and paths)
- postproc verified
- postproc on-disk
- queue (list of nzo_ids)
- NzbObject renames dict[str, str]
- NzbObject attribs
- NzbFile raw_article_db (article_id and size)

Safihre:
Claude is sure we have every object covered now.

mnightingale (approved):

PR #3586, opened 2026-08-21
Enable direct_decode by default, paced against what the disk can do

mnightingale:
The downloader thread becomes throttled by; the share of article cache free, and the rate the disk is measured to write at.

This allows `direct_decode` to default to on. A monitor watches what the download directory does with the writes it is given and withdraws the permission when it stops keeping up, so the setting grants permission rather than forcing the behaviour. `/sys/block/*/queue/rotational` only picks the opening mode; the measurement overrules it either way.

This also massively improves performance for append mode where the disk is slower than network speed.

Measured on a USB hard drive purposely limited to ~41 MB/s (connected to USB2.0) on the 10 GB test job, append mode:

| | before | after |
|---|---|---|
| articles spilled to `__ADMIN__` | 7,988 (5.3 GB) | 0 |
| speed | 33–45 MB/s, thrashing | steady 40.0 MB/s |

Those spill writes were consuming the whole device budget, starving the assembler that would have freed the cache. It's actually 5.3GB*2 excess writes being performed, because the temporary files have to read and then write to the actual file.

Needs a sabctools release, and I'll update the wiki entry for `direct_decode` and remove `assembler_max_queue_size`


sabnzbd/downloader.py
```diff
@@ -834,33 +834,22 @@ def process_nw_read(self, nw: NewsWrapper, generation: int) -> None:
...
-                    "Delayed - %d seconds - Assembler queue: %s",
-                    logged_counter,
-                    to_units(sabnzbd.Assembler.total_ready_bytes()),
-                )
-                next_log += 1.0
+        sabnzbd.BPSMeter.update()
+
+        if time.monotonic() >= self.next_delay_log:
+            self.next_delay_log = time.monotonic() + 1.0
+            sabnzbd.BPSMeter.delayed_assembler += 1
```
mnightingale: Instead of incrementing `delayed_assembler` once per second, I could make it only do so if the delay is larger than a set amount?

The delay isn't really saying there is a problem any more, it's saying we are adapting to the disk performance so theoretically download speed "may" be impacted by the slow disk. So if the delays are small we are only intervening minimally and probably not worth telling the user.

Safihre: Agreed.

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+                    return flag.read().strip() == "1"
+            except OSError:
+                continue
+    except OSError:
+        pass
+
+    return None
+
+
+class WriteMonitor:
```
Safihre: Let's move this to a separate file.
Then we can also move all those constants there.
Those are probably way to specific to belong in the general constants.py.

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+        self.slow_samples = 0
+        self.retry_after = WRITE_MONITOR_RETRY_AFTER
+        self.retry_at = 0.0
+        self.sampled_at = 0.0
+        # Counters as of the previous sample
+        self.seen = (0, 0, 0)
+        self.reset()
+
+    @synchronized()
+    def reset(self, path: Optional[str] = None):
```
Safihre: Why the path parameter if it's never used?

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+        """Start again for a directory, reading the hint but measuring nothing"""
+        self.path = path if path is not None else sabnzbd.cfg.download_dir.get_path()
+        self.hint = is_rotational(self.path) if self.path else None
```
Safihre: Line above says it can never be none?

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+        self.allow_direct_decode = self.hint is not True
```
Safihre: Can it also be 0?

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+        self.rebaseline()
+
+    def measured(self) -> str:
+        """The smoothed cost as MB/s"""
+        if self.cost is None:
+            return "not measured yet"
+        return "%.0f MB/s inside write" % (1e9 / self.cost / (1024 * 1024))
+
+    def log_profile(self):
+        """Log the mode the download directory is in"""
```
Safihre: Is this really needed? Just log it when initializing

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+        logging.info("Storage profile for %s: %s", self.path, self)
+
+    @synchronized()
+    def __str__(self) -> str:
```
Safihre: Where is this used?
Otherwise `measured()` can also go.

Sorry for being nitpicky, it's just the AI's tendency to function-everything is annoying me as you know 😂

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+
+    def rebaseline(self):
+        """Start counting writes from this moment, keeping what they have cost so far"""
+        self.slow_samples = 0
+        self.seen = self.read_counters()
+        # So the next sample measures from here, not from whenever the process started
+        self.sampled_at = time.monotonic()
+
+    @synchronized()
+    def allows_direct_decode(self) -> bool:
```
Safihre: Does this really need a synced method?

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+    @synchronized()
+    def write_rate(self) -> Optional[float]:
+        """Bytes per second the destination is draining, None if unmeasured"""
+        return self.throughput
+
+    @staticmethod
+    def read_counters() -> tuple[int, int, int]:
+        """Writes, bytes and nanoseconds every file has cost since sabctools loaded"""
+        stats = sabctools.write_stats()
+        return stats["count"], stats["bytes"], stats["nanos"]
```
Safihre: Shouldn't there just be a trigger on sabctools to trigger a reset? 
Why do the bookkeeping here if this is the only consumer?

sabnzbd/filesystem.py
```diff
@@ -1390,6 +1400,181 @@ def check_sparse_and_disable(test_dir: str) -> bool:
...
+        self.allow_direct_decode = False
+        self.hint: Optional[bool] = None
+        # Nanoseconds per byte spent inside write(), smoothed
+        self.cost: Optional[float] = None
+        # Bytes per second the destination actually drained, over WRITE_RATE_WINDOW
+        self.throughput: Optional[float] = None
+        self.window: deque[tuple[int, float]] = deque()
+        # Consecutive samples in which the device did not keep up
+        self.slow_samples = 0
+        self.retry_after = WRITE_MONITOR_RETRY_AFTER
```
Safihre: Why do we even try again?

sabnzbd/filesystem.py
mnightingale: I'd rather not lock in that there can only be one consumer and it keeps the sabctools side simple

sabnzbd/filesystem.py
mnightingale: Could be just temporarily slow, par2, unrar, direct unpack, another process.

Safihre:
I should have named it sabctools 9.8.0 because with 9.7.0 this would break, it needs 9.7.1 for the exception.

PR #3595, opened 2026-08-25
Keep the client address of incoming connections when a proxy is set

### LINKED ISSUE #3594: Setting a SOCKS5 proxy locks you out of the web UI ("External internet access denied")
On current `develop`, setting `socks5_proxy_url` makes SABnzbd refuse every request with "External internet access denied", including from `127.0.0.1` and including the web UI itself. Only way back in is editing `sabnzbd.ini` by hand.

**To reproduce**

1. Config > General, click Advanced, set SOCKS5 Proxy to `socks5://socks.example` (doesn't need to be reachable). Leave External internet access at "No access", the default.
2. Save, restart, hit any URL, get a 403

Log shows why: `Refused connection from: :0`, the client address is just gone.

**Why**

PySocks proxies the process by replacing `socket.socket`, and `socket.accept()` builds accepted sockets from that same global, so our own web server's accepted connections become `socksocket` instances too. Its `getpeername()` only knows about outbound proxy connections and returns `None` for anything else instead of raising, so Uvicorn's `get_remote_addr()` bails out with `None` right there instead of falling back to the real address it already has. `request.client` ends up `None`, and `check_access()` correctly fails closed.

Regression from #3550 (CherryPy to Uvicorn), Cheroot never called `getpeername()` so it never hit this. Hasn't shipped, 5.1.1 was cut before #3550 landed.

FrankGasparovic:
Fixes #3594

Setting `socks5_proxy_url` made SABnzbd refuse every request with "External internet access denied", including from `127.0.0.1` and including the web UI itself, leaving no way back in other than editing the ini by hand.

PySocks proxies the process by replacing `socket.socket`, and `socket.accept()` builds accepted sockets from that very global, so the connections our own web server accepts are PySocks sockets too. Its `getpeername()` returns `proxy_peername`, which is only ever set by an outgoing `connect()` through the proxy, and it returns `None` rather than raising. Uvicorn asks the socket before falling back to the address asyncio kept from `accept()`, so it never got that far and the request arrived without a client address, which `check_access()` correctly refuses.

Cheroot used the address returned by `accept()` and never called `getpeername()`, so this only started with #3550.

## Test plan

- [x] `pytest tests/test_misc.py -k Socks5` covers the accepted-socket case, confirmed it fails without the fix
- [x] Full suite passes locally
- [x] Verified manually: with `socks5_proxy_url` set and `inet_exposure` at its default of 0, `/`, `/config/general/`, and `/api` all returned 403 before this change and 200 after, with the log line changing from `Request GET /api from :0` to the real client address

### DIFF
--- a/sabnzbd/misc.py
+++ b/sabnzbd/misc.py
@@ -69,6 +69,9 @@ from sabnzbd.decorators import conditional_cache, synchronized
 from sabnzbd.encoding import ubtou, platform_btou
 from sabnzbd.filesystem import userxbit, make_script_path, remove_file, strip_extensions, safe_fnmatch
 
+# Keep the real socket around, setting a proxy replaces socket.socket
+_ORIGINAL_SOCKET = socket.socket
+
 if sabnzbd.WINDOWS:
     try:
         import winreg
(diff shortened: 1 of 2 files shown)


mnightingale:
Overriding getpeername() may now raise OSError on an un-connected socket.

Try this, I've not verified it with a socks proxy myself https://gist.github.com/mnightingale/c49399a60c7ba84fe34e3da7d32f6bad

FrankGasparovic:
Good catch, confirmed it. Pushed your fix plus a test for it.

mnightingale:
I do wonder if rather than a workaround, is it worth reporting as a bug in uvicorn?

Claude thinks `get_remote_addr()` and `get_local_addr()` has bugs, but I haven't verified them nor tested what it proposes with a proxy https://gist.github.com/mnightingale/e7a169d1682fbf340f1ee808996e2b96

PR #3596, opened 2026-08-27
Replace anonymous session cookie with stable CSRF identity

Safihre:
Previously, UI access for login-bypassed clients relied on issuing a stateless "anonymous session" cookie to enable CSRF protection. This commit removes that cookie mechanism.

Instead, when the login is bypassed (e.g., for local clients without credentials), the CSRF token is now bound to a stable, internal "anonymous" identity. This maintains robust CSRF protection for browser interactions without requiring an explicit session cookie, simplifying overall session management.

### DIFF
--- a/sabnzbd/security.py
+++ b/sabnzbd/security.py
@@ -42,7 +42,7 @@ _MSG_APIKEY_NOT_ON_PAGES = (
 _MSG_SESSION_EXPIRED = "Session expired, reload the page"
 
 
-# Holds a database-backed login token, or the anonymous tag when the login is bypassed
+# Holds the login token for an authenticated session; absent where the login is bypassed
 SESSION_COOKIE_USER = "sabnzbd_user"
 # The SessionMiddleware cookie, used for RSS flash messages only. Not authentication.
 SESSION_COOKIE_FLASH = "sabnzbd_flash"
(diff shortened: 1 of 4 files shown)


sabnzbd/security.py
```diff
@@ -243,10 +225,10 @@ def presented_csrf_token(request: Request, header_only: bool = False) -> str:
 
 
 def csrf_token_matches(request: Request, header_only: bool = False) -> bool:
-    """Whether the request echoes the CSRF token belonging to the cookie it sent"""
+    """Whether the request echoes the CSRF token belonging to its identity"""
     return constant_time_equals(
```
mnightingale:
```suggestion
    identity = csrf_identity(request)
    return bool(identity) and  constant_time_equals(
```

Not reachable currently, but tighten requirements for anyone calling this directly

sabnzbd/security.py
```diff
@@ -243,10 +225,10 @@ def presented_csrf_token(request: Request, header_only: bool = False) -> str:
...
         presented_csrf_token(request, header_only=header_only),
-        csrf_token_for(request.cookies.get(SESSION_COOKIE_USER, "")),
+        csrf_token_for(csrf_identity(request)),
```
mnightingale:
```suggestion
        csrf_token_for(identity),
```

mnightingale (approved):
I’m satisfied by this change, defence in depth reduces (deliberately) because a forged post needed the token and a cookie, now it only needs the token but that’s not exploitable without leaking the page source.

PR #3597, opened 2026-08-28
Remove the downloader sleep that throttled the read loop

mnightingale:
The downloader tracked the largest single read per pass and, when that peak fell below a third of a running high-water mark, slept 1ms before returning to `select()`. The intention was reasonable: if we read much less than the connections were recently capable of, there is probably little to do, so yield the CPU rather than spin.

The signal does not mean that. A small read means the receive buffers were near-empty when we drained them, which on a healthy connection is what keeping up looks like. Sleeping then delays the next drain, lets buffers fill while we sit idle, and shrinks the advertised TCP window, so it fires during steady throughput and acts as a speed ceiling. The high-water mark only ratcheted upward for the life of the session, so one burst read permanently raised the bar and made ordinary reads trip the sleep continuously.

It saved no CPU either. The same bytes still have to be read, decoded and written, just later and in burstier, larger passes. Real busy-loop protection is already present and untouched: `select(timeout=1.0)` blocks in the kernel, the no-sockets branch waits on `DOWNLOADER_CV`, and `check_assembler_levels()` applies backpressure from a measurement when the disk is genuinely behind.

It also put calculations in the hot path to reach that wrong decision: a float division by three plus comparisons on every pass of the loop, and a compare-and-store against `last_max_chunk_size` on every socket read, the latter while holding `DOWNLOADER_LOCK`. None of it is expensive in isolation, but it is work done at the highest frequency in the downloader purely to feed a heuristic that should not have been acting.

The `downloader_sleep_time` special setting is removed so will also need updating on the wiki.


Safihre:
Oke I believe what you write, however, why did it reduce the CPU usage percentage while keeping the speed the same?

Safihre:
I tested this branch vs develop without a speed-limit on this 100MBit connection (so clearly hot-looping) using the 10GB test file.
On develop my CPU usage is in the 1-2% range, on this branch 2-3%.
I tested a few times.

mnightingale:
I have never liked the approach, it always seemed an XY problem, and I think any CPU reduction was the result of batching rather than a saving. The sleep does not remove any work, it still eventually gets read, decoded, and written.
The 1ms sleep allows the buffers to build, so each recv returns more and fewer loops are needed. Every loop of the downloader has a cost, it is doing a bunch of checks on `for server in self.servers`, a select, queue put per socket, locks, bpsmeter. Reduce the number of loops and CPU% falls because the fixed cost per byte downloaded falls, not because work was skipped.

So, I would rather focus on reducing the fixed cost than accept time.sleep as the answer, and I concede did mostly achieve it’s intended outcome. I have two changes ready that I was going to follow up with, but kept separate to avoid conflating things:

https://github.com/mnightingale/sabnzbd/tree/perf/gate-downloader-server-scan

- Gate the server walk behind a flag whenever any connection goes idle, plus the the existing 0.5s periodic check, instead of running every pass
- Put selector events back in the queue with put_multiple

I can add those commits to this PR if you’d like, because that’s what it should be judged on. ` DOWNLOADER_LOCK` on bpsmeter Is something I will probably look at next since the lock is taken on every read.

On speed, “speed the same” is what would be expected, the sleep only fires when the largest read in a pass drops below 1/3 of the largest seen recently. So, if the bandwidth or server is the bottleneck you never see it and it’s only when the loop itself is the limit with high connection counts on a fast link. `max_chunk_size` only grows until the queue is empty, one large read raises the bar and reads for the rest of the queue continuously sleep.

What it did allow is yielding time to other threads, especially the assembler. But the WriteMonitor now better handles that in `Assembler.delay()` because it holds the download back depending on how full the cache is and if the disk is keeping up. It also find a balance and settles at what speed is sustainable unlike the chunk-size sleep which fired regardless of if anything was struggling to cope. Also Python already takes care of general thread fairness, and we allow controlling it with `cfg.switchinterval`.

Safihre:
This PR is right now making it worse, using more CPU than is needed.
Even though you might theoretically be right.
If the CPU usage stays low with those extra commits, then we can merge it.

mnightingale:
I wasn’t going to combine several changes but if the requirement is to not increase CPU usage, then this is probably the smallest change I can make.

This is testing the 1GB test file from a local NNTP server over SSL, 60 connections, three runs per configuration. CPU is whole-process user+sys sampled across the download window only, so startup and post-processing are excluded.

**Whole process, CPU seconds per 100MB, this branch against its base:**

| | base | branch | |
|---|---|---|---|
| 100Mbps speed limit | 0.591 | 0.541 | -8.5% |
| no limit (~110 MB/s) | 0.370 | 0.315 | -15% |

Throughput was identical in both: 12.1 MB/s limited, ~110 MB/s unlimited.

**Where it moves, per thread, unlimited:**

| | base | branch |
|---|---|---|
| Downloader (main loop) | 0.030 | 0.086 |
| receive threads (3, summed) | 0.307 | 0.207 |
| Assembler | 0.032 | 0.028 |

So you were right that removing the sleep costs CPU. It nearly triples the main loop, which is the extra passes. What pays for it is `put_multiple`: the per-event `put()` calls `notify()` once per event and wakes a worker each time, so batching the whole select result into one call cuts the receive threads by a third. That is twice the size of the regression, hence the net gain and removing the speed ceiling from faster connections.

The token bucket contributes nothing to the unlimited figure and is the difference in the limited one. It also holds the requested rate more accurately, since the old loop waits for a five second EMA to decay past the limit and undershoots: at a 50M limit it delivered 48.9 and 49.4 MB/s where the bucket delivered 49.6 and 49.5.

sabnzbd/downloader.py
```diff
@@ -823,17 +807,28 @@ def process_nw_read(self, nw: NewsWrapper, generation: int) -> None:
...
+            time.sleep(delay)
+
+    def bandwidth_delay(self, bytes_received: int) -> float:
+        """How long this connection waits so the download stays within the speed limit"""
+        with BANDWIDTH_LOCK:
+            if not (limit := self.bandwidth_limit):
+                return 0.0
+
+            now = time.time()
+            self.bandwidth_tokens = min(
```
Safihre: The term "tokens" is confusing me a bit, since it seems also to be bytes?
So can we have a better term for this?

sabnzbd/downloader.py
```diff
@@ -692,27 +684,15 @@ def run(self):
...
-                        now = time.time()
-                    self.last_max_chunk_size = 0
-
                 # Use select to find sockets ready for reading/writing
                 if self.selector.get_map():
                     if events := self.selector.select(timeout=1.0):
-                        for key, ev in events:
-                            nw = key.data
-                            process_nw_queue.put((nw, ev, nw.generation))
+                        process_nw_queue.put_multiple([(key.data, ev, key.data.generation) for key, ev in events])
```
Safihre, line 681: Can't we just pass the events on without creating a new list?
Claude also said we need: `self.not_empty.notify(len(multiple_items))` in `put_multiple` so we wake up all processors.

mnightingale: Creating the new list was one of the reasons I didn’t use it previously, the notify issue is why it appeared to reduce CPU usage – it serialised and only used one receive thread.

At this point, I’ll drop `put_multiple` altogether (separate PR but it’s unused) it only ever looked like a CPU utilisation win because it only used one receive threads. Once you add `notify(len(…))` it’s the same as just adding them individually and saves creating a list.

I’m going to rethink this, keeping sleep_time but ignore at high bps.

PR #3600, opened 2026-08-30
Store the Windows API connection info per config file

mnightingale:
On Windows the URL of a running instance is stored in a single registry value, `HKCU\Software\SABnzbd\api\url`. Every instance under one account writes to that one slot, so a starting instance reads whichever one happened to write last.

When it finds that URL alive, `check_for_sabnzbd` hands off and calls `exit_sab(0)`. That happens before the logging handlers are installed, so the instance disappears without a single line of output.

Two instances on separate config files are not duplicates of each other, so each config file now gets its own subkey, named after a hash of the config path (a subkey name cannot contain a path separator). `get`/`set`/`del_connection_info` take the config path, and `halt()` passes `config.CONFIG.filename`.

Restarting an instance running alongside another one no longer silently exits into the other one's web interface. It also fixes a flaky `test_queue_repair` on Windows CI, where four xdist workers each run their own SABnzbd and fight over that shared value.

No fallback to the old value: a starting instance that misses the registry still finds a running one through the port check.

### DIFF
--- a/sabnzbd/utils/apireg.py
+++ b/sabnzbd/utils/apireg.py
@@ -19,6 +19,9 @@
 util.apireg - Registration of API connection info
 """
 
+import hashlib
+import os
+import sys
 import winreg
 from typing import Optional, Tuple
 
@@ -47,8 +50,8 @@ LANGUAGE_MAP = {
 }
 
 
-def reg_info(user: bool) -> Tuple[int, str]:
-    """Return the registry hive and key path for the API info
+def reg_info(user: bool, inifile: str) -> Tuple[int, str]:
+    """Return the registry hive and key path for the API info of one instance
 
     The URL of a running instance is stored so that a second start of SABnzbd
     (for example by double-clicking an NZB) can hand off to the running instance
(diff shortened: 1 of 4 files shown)


Safihre:
This is only for testing. It's not relevant for real users.
Certainly also not for Windows Services, there can only be 1 by definition.

mnightingale:
It's a probably rarely used but supported configuration though - well not on Windows

We give `--config-file`, `--pid`, and `--new`.

> ... If you really want to run multiple instances, you will need to use the --new parameter. 

Users could add multiple services with different config files, not sure why they would.

It causes test failures on Windows otherwise, because the queue repair restart isn't with `--new` and it may see the reg key for a different instance.

I could maybe add special behaviour when `CI` is defined, or make the restart the test triggers use `--new`.
Or just skip `test_queue_repair` on Windows, it's the only one which calls `restart_repair`

sabnzbd/utils/apireg.py
```diff
@@ -47,28 +50,33 @@
...
     instead of starting a duplicate. A desktop run stores it per-user in HKCU,
     but a Windows Service runs under a service account whose HKCU is invisible
     to the desktop user, so it uses the machine-wide service key in HKLM instead.
     Readers check HKCU first and fall back to HKLM to find either kind of instance.
+
+    Each config file gets its own subkey, so instances running side by side on
+    separate configs do not share an entry.
     """
+    # Subkey names cannot contain a path separator
+    instance = hashlib.sha256(os.path.normcase(os.path.abspath(inifile)).encode("utf-8")).hexdigest()[:16]
```
Safihre: Why the limit to 16?

mnightingale: No reason it can't store the full 64 characters, there is no need for them to stay easily readable
But the chance of a collision is infinitesimally small, so it doesn't make a lot of difference either way

PR #3601, opened 2026-08-30
Serve static assets with Cache-Control headers

mnightingale:
Serves assets within `/static`, `/staticcfg`, and `/wizard/static` with `Cache-Control` headers, after the first download the browser wont even request them again.

For this reason every reference to such asset now includes the cache_buster.

Repeated page loads drop from 175KB to 21KB

Also fixes config restart button with hardcoded `../../api` url.

> Inclusion is faster than external scripts

The original commit that added this 0864fd9a6 said "SABNZB doesn't do caching!", it was only slower because Cache-Control headers were not sent and the browser checked Last-Modified for several assets before it could load the UI. The cache_buster on all asset urls means we can set Cache-Control headers now and any penalty is only on the first load.

In 2015 it was ~10 revalidations vs 1 large response, now it is 0 requests vs ~450KB read from disk.


Safihre:
I remember making that, I was still a young man 😂

sabnzbd/interface.py
```diff
@@ -2537,17 +2537,32 @@ async def not_found_redirect(request: Request, exc):
...
 
 
+class CachedStaticFiles(StaticFiles):
+    """Static files that may be held indefinitely.
+
+    Every reference to these is written by $url() with the version and commit appended, so
+    an upgrade asks for a different URL rather than a fresh copy of the same one.
+    """
+
+    def file_response(self, *args, **kwargs) -> Response:
```
Safihre, line 2543: starlette doesn't have this build in regularly for static files? Or some switch? Really?

mnightingale: Nope, extending StaticFiles is the best approach.
I can trim comments here if you'd like.

Safihre: Yes

PR #3603, opened 2026-08-31
Load one copy of jQuery, jQuery UI and Bootstrap instead of two

mnightingale:
Glitter and Config each ship their own jQuery and jQuery UI (identical), and Bootstrap at two different versions. Glitter now loads all three from staticcfg/, which Glitter already uses for favicons. Removes 170 KB of duplicated JavaScript from the repo and the release.

Glitter's Bootstrap JS moves 3.3.5 to 3.3.7. The CSS and glyphicon fonts are deliberately left alone: the two releases ship different font files, so unifying those would change every icon in Glitter.


Safihre:
Should we update our jQuery and jQuery UI versions to latest?

mnightingale:
Added, I wasn't planning on doing so but the risk of breaking anything is low.

Kept to jQuery 3.7.1 so we don't have to deal with whatever has changed in 4.0.0

My long-term intention is to remove jQuery, jQuery UI, momentjs, datejs, and possibly Bootstrap but there are loads of things before that i.e. I wouldn't remove it all at once

PR #3604, opened 2026-08-31
Downloader hot path improvements

mnightingale:
1. bpsmeter could miss updates because check_assembler_levels doens't hold DOWNLOADER_LOCK, give it its own lock
2. update bps without taking DOWNLOADER_LOCK
3. remove unused MultiAddQueue, put_multiple didn't notify all threads which was the only reason it looks like it used less CPU
4. skip the downloader sleep on fast connections
5. update now after a sleep

### DIFF
--- a/sabnzbd/bpsmeter.py
+++ b/sabnzbd/bpsmeter.py
@@ -21,6 +21,7 @@ sabnzbd.bpsmeter - bpsmeter
 
 import time
 import logging
+from threading import Lock
 import re
 from collections import deque
 from itertools import islice, repeat
@@ -38,6 +39,8 @@ BPS_LIST_MAX = 275
 
 RE_DAY = re.compile(r"^\s*(\d+)[^:]*")
 RE_HHMM = re.compile(r"(\d+):(\d+)\s*$")
+
+BPSMETER_LOCK = Lock()
 
 
 def tomorrow(t: float) -> float:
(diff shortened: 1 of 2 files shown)


Safihre:
@mnightingale I cherry picked the "easy" commits from this, but can you explain maybe a bit more of the gains of the remaining ones?

mnightingale:
# Guard the speed counters with their own lock

This is a bug fix rather than performance related and one that you took on it's own made it worse.

`BPSMeter.update()` reads cached sums and zeros them, it needs a lock somewhere because we have multiple recieve threads.

Until "Check the speed limit on its own lock", those calls shared `DOWNLOADER_LOCK` so couldn't overlap. Basically there are non-downloading actions we might need to do and separating BPSMETER updates from this is necessary and puts the responsibility in the correct place.

# Update the speed counters without taking the Downloader lock

Continues the above, could be merged with it I suppose. Now bpsmeter has it's own lock a caller doesn't need DOWNLOADER_LOCK.

I've added a little more to it, eliminating the lock entirely because each receieve thread competes for it. It makes `last_max_chunk_size` not 100% accurate, but it doesn't matter. At most it causes a 1ms sleep on the next loop.

# Skip the downloader sleep once the loop is the limit

This is my compromise for not removing sleep_time, skip it when sleeping may affect the download speed which I set at 150M but would happily do 300M. Basically at higher speeds all the sleeping places a ceiling on the maximum download speed and achieving more requires allowing smaller more frequent request processing.

# Remove unused variable

Added this, "Refresh the time after waiting" makes this redundant

Safihre:
Does it still sleep at such speeds at all? The calculations should prevent that right?

mnightingale:
Not if the current speed is above `_SLEEP_MAX_BPS`, but that value can raise to at least 300-500M
Below that the calculation is unchanged, just using multiplication instead of division.

I don't think it's a real CPU saving. I measured CPU seconds rather than utilisation percentage at ~100MB/s and it was 0.338s per 100MB with the sleep vs 0.336s without. But I haven't tested at faster or slower speeds.

We already sleep for bandwidth limiting against the measured rate, so when a limit is set it is already pacing the downloader and this sleep adds nothing on top.

But I'll compromise, the sleeps make it harder for any machine to reach their max but especially slower ones or on faster connections and I want to avoid the need for such users to go looking for special switches.

thezoggy:
rather than focus on network/disk io would it make more sense to leverage if there was just more articles/connections to be more aggressive on updating? or use an iterator rather than a loop?

mnightingale:
As a brief explanation of how it currently works.

`max_chunk_size` the largest read since the downloader last went idle, it resets to the default when there are no active sockets
`last_max_chunk_size` the largest read from a single socket in the last loop

```py
if self.last_max_chunk_size > self.max_chunk_size:
    self.max_chunk_size = self.last_max_chunk_size
elif self.last_max_chunk_size < self.max_chunk_size / 3:
    time.sleep(self.sleep_time)
self.last_max_chunk_size = 0
```

If `last_max...` is greater than ever seen before update `max_chunk_size`
If the max we saw last loop is less than 1/3 of the max we have ever seen, then sleep for 1ms

I have problems with `max_chunk_size` because one large read raises the bar for all future reads causing sleeps, and that's probably more likely at high speeds (>10 Gbps)

The other side of it is that 1ms sleep. Worst case it bounds the loop to 1000 passes a second, the receive window fills and the sender backs off ceiling the maximum speed.

Personally, I favour removing the complexity, but it doesn't really hurt at slow speeds, so I'd like to just ignore it if bps is above some fast speed.

sabnzbd/bpsmeter.py
```diff
@@ -317,31 +320,33 @@ def update(self, server: Optional[str] = None, amount: int = 0):
         dt_total = max(t - start_time, 1e-6)
         dt_last = max(last_update - start_time, 1e-6)
 
+        with BPSMETER_LOCK:
```
Safihre, line 289: Wondering why not just have the stuff below wrapped in the lock instead of this pre-drain?

mnightingale: We do need to, else it races anyway.

This change looks more substantial, but it's just the original surrounded by a `with BPSMETER_LOCK` block.
It is needed because multiple threads can call update, don't use `@synchronized` it'll lock into other calls and deadlock. Which is why `check_quota()` is outside the lock.

A note on `cached_amount[server] += amount` a problem/race with that is rare with the GIL, but this prepares it for free-threaded use in the future.

Also it used to lock the whole call anyway `with DOWNLOADER_LOCK:` but that competes with everything else that lock is required for.

Safihre: What happens without a lock? Pratically?
Do we mis a few bytes in some counters or are we wildly wrong?

mnightingale: I couldn’t get it to be wrong with many threads, so it’s rare due to the GIL.

I'd keep the locks anyway. Eventually we won't have the GIL, and the cost is negligible. The risk is mis-accounting the per-server quota.

PR #3608, opened 2026-08-31
Let the browser pick the dark theme instead of an @import

mnightingale:
Gets rid of the Auto `Night.css?lastmajorupdate=510` by calculating the filename and media values in `build_header` allowing the browser to decide to load it or not.

This is the last immediate change I have for now; gets into much larger changes and features that I'll discuss first


Safihre:
Or we just drop Light/Dark and only keep auto.

thezoggy:
> Or we just drop Light/Dark and only keep auto.

please god no.  many of us run our os in one mode (light) but specifically want sab to be in (dark). auto would break things. if anything id say remove auto and only give me light/dark like we used to if your going to go that route.

mnightingale:
My ideal would have been to have it purely as a client-side preference stored in local storage. That’d be an extra icon in the navigation, and it still needs to populate by dynamic values possibly with icons.

But for now I just wanted to fix the cache problem.

PR #3611, opened 2026-09-01
Fix macOS tray crashing when the notification center is unavailable

mnightingale:
Running form source, there is no bundle id so it fails to init - there are hacky workarounds but I would not suggest using them.

This change lets the tray icon still function, but without notifications.

It may be worth switching to UNUserNotificationCenter, but it still won't let notifications work.

```
  File "/Users/mnightingale/personal/sabnzbd/SABnzbd.py", line 1568, in <module>
    sabnzbd.MACOSTRAY.awakeFromNib()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/mnightingale/personal/sabnzbd/sabnzbd/macosmenu.py", line 93, in awakeFromNib
    DefaultUserNotificationCenter.setDelegate_(self)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'setDelegate_'
```

### DIFF
--- a/sabnzbd/macosmenu.py
+++ b/sabnzbd/macosmenu.py
@@ -90,7 +90,8 @@ class SABnzbdDelegate(NSObject):
             time.sleep(0.5)
 
         # Set this thread as default handler for notification actions
-        DefaultUserNotificationCenter.setDelegate_(self)
+        if DefaultUserNotificationCenter:
+            DefaultUserNotificationCenter.setDelegate_(self)
 
         # Do we want the menu
         if sabnzbd.cfg.tray_icon():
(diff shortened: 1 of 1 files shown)


mnightingale:
Not tested but I would guess it's because I use https://github.com/pyenv/pyenv so there is no Info.plist next to the python executable, but I don't want to use system or homebrew Python.

As it is the failure stops the tray initialising, even without notifications working.

Safihre:
Haven't booted up my ancient Macbook Pro in a while, but running from source using a regular venv from a Python.org installer both tray and notifications used to work.

PR #3612, opened 2026-09-03
Fix "database is locked" when re-evaluating stored RSS jobs (readout=False)

btli:
## Problem

Since RSS state moved into `history1.db` (#3253), every RSS filter add / update / delete from the web interface or the API (`mode=set_config section=rss filter_action=add|update|delete`) calls `process_feed(feed, readout=False)`. That path iterates the stored jobs lazily through one `HistoryDB` connection (`repo.get_feed_jobs()`), while `_process_entry()` writes each re-evaluated job through a **second** connection to the same file.

In the default rollback-journal mode the still-open SELECT cursor holds a SHARED lock, so every upsert waits out the busy timeout (5 attempts × 5 s + 4 × 0.5 s ≈ 27 s) and then fails:

```
ERROR::[database:188] SQL Command Failed, see log
INFO::[database:189] SQL: INSERT INTO rss ( ... ) ON CONFLICT(feed, url) DO UPDATE SET ...
sqlite3.OperationalError: database is locked
```

Observed on 5.1.2 (Docker, local ZFS storage, single process): a feed with ~100 GOOD/BAD jobs turns one filter change into 45–80 minutes of these errors (one every 27 s). The filter change is not applied to the stored jobs, and because `RSS_LOCK` is held the whole time, scheduled read-outs are blocked as well. Every "Writing settings to INI file" line in the last week was followed 27 s later by such an error storm.

`develop` no longer shows this because #3550 moved the database to WAL mode with a connection pool, but the `5.1.x` line is still affected.

## Fix

Materialize the stored jobs before evaluating them (`list(repo.get_feed_jobs(...))`) so no read cursor is open while writing. Minimal and safe to backport.

## Test

`test_process_feed_without_readout_updates_all_stored_jobs` stores five BAD jobs, adds an accept-all filter, replays with `readout=False`, and asserts every job became GOOD.

- On `5.1.x` without the fix: fails after ~3.5 min (`database is locked` on every job but the last).
- With the fix: `37 passed in 6.14s` for `tests/test_rss.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -989,9 +989,7 @@ class RSSReader:
             if readout:
                 gen = self.fetch_rss(feed, uris)
             else:
-                # Read all stored jobs before evaluating them: each evaluation writes
-                # through its own connection, which would otherwise be blocked by the
-                # still-open read cursor ("database is locked")
+                # Materialize before evaluating: each write needs the read cursor closed
                 gen = list(repo.get_feed_jobs(feed=feed))
 
             # Evaluate rules and apply side effects
--- a/sabnzbd/rss.py
+++ b/sabnzbd/rss.py
@@ -989,7 +989,10 @@ class RSSReader:
             if readout:
                 gen = self.fetch_rss(feed, uris)
             else:
-                gen = repo.get_feed_jobs(feed=feed)
+                # Read all stored jobs before evaluating them: each evaluation writes
+                # through its own connection, which would otherwise be blocked by the
+                # still-open read cursor ("database is locked")
+                gen = list(repo.get_feed_jobs(feed=feed))
 
             # Evaluate rules and apply side effects
             try:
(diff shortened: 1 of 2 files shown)


btli:
A note on why this is a separate fix even though `develop` already runs in WAL mode (#3550):

- **#3550 is a large rewrite that will ship with the next major release**, so the `5.1.x` line stays affected. Every RSS filter add / update / delete on 5.1.0–5.1.2 currently produces this error storm, blocks scheduled read-outs for the duration, and silently fails to apply the filter change to stored jobs. This PR is intentionally minimal (one line plus a test) so it can go into a 5.1.3.
- **WAL masks the pattern rather than fixing it.** `process_feed(readout=False)` still keeps a read cursor open on one connection while `_process_entry()` writes through another connection to the same file. WAL happens to tolerate that; a rollback journal does not. Materializing the rows first makes the code correct in either journal mode, and the regression test catches a reintroduction regardless of the pragmas in effect.

Happy to also open the same test against `develop` if you want it guarding the pooled/WAL path there.

mnightingale (approved):
@Safihre won't like the three lines of comments to one line of code so probably remove that, it's unnecessary

I'd vote yes to adding the test to develop, but not the fix because some users can have thousands of entries in their RSS history and don't particularly want to load them all in to memory

PR #3614, opened 2026-09-03
Optimise NzbQueue add, remove and reorder

mnightingale:
The download queue is a single list kept in priority order. Adding a job and changing a job's priority both scanned it for the right slot, with the same branching written out in `add()` and `set_priority()`. Both now call one helper that uses `bisect` on the priority key. It uses `bisect_right` rather than `insort` because `set_priority()` needs the resulting position for its return value, which `insort` would cost a second scan to recover.

That gives every job one rule, join the back of your own priority group, which the two paths used to disagree about. `add()` shared a branch with FORCE and inserted at the very top, so a repair re-add went in newest first and kept sinking as later jobs completed. `set_priority()` put REPAIR above FORCE and oldest first, but sent FORCE to the very top. Repair now outranks force in both paths and a re-add no longer starves.

Forced jobs lose their front-of-group insert, a leftover from before REPAIR existed (396e77951, 2011, which moved fetching extra par2 files to priority 3). The visible effect is that a multi-select no longer reverses: selecting a, b, c and pressing Force gave c, b, a and now gives a, b, c.

Every add and remove also walked the queue to rebuild the nzo_id admin file before writing it, so removing twenty jobs meant twenty walks and twenty writes. `remove()` takes a `save` flag so `remove_multiple()` writes once per batch, and `save()` accepts a list so adding an archive of NZBs does the same. Those are the only signature changes.

The last commit is two micro-optimisations: `attrgetter` for the attribute sort keys, and `itertools.chain` instead of concatenating the queue and post-processing lists in the two duplicate checks, which run on every added NZB and copied the whole queue each time.

`tests/test_nzbqueue.py` had nothing covering add, remove, reorder, priority or sorting, so this adds 25 tests.

### DIFF
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -24,6 +24,8 @@ import os
 import logging
 import time
 import uuid
+from itertools import chain
+from operator import attrgetter
 from typing import Optional
 
 from starlette.datastructures import UploadFile
@@ -569,11 +571,11 @@ class NzbQueue:
             sort_function = lambda nzo: nzo.final_name.lower()
         elif field == "size" or field == "bytes":
             logging.info("Sorting by size (reversed: %s)", reverse)
-            sort_function = lambda nzo: nzo.bytes
+            sort_function = attrgetter("bytes")
         elif field == "avg_age":
             reverse = not reverse
             logging.info("Sorting by average date... (reversed: %s)", reverse)
-            sort_function = lambda nzo: nzo.avg_date
+            sort_function = attrgetter("avg_date")
         elif field == "remaining":
             if self.__nzo_list:
                 logging.debug("Sorting by percentage downloaded...")
@@ -581,14 +583,14 @@ class NzbQueue:
         elif field == "remaining_bytes":
             if self.__nzo_list:
                 logging.debug("Sorting by remaining size...")
-            sort_function = lambda nzo: nzo.remaining
+            sort_function = attrgetter("remaining")
         else:
             logging.debug("Sort: %s not recognized", field)
             return
 
         # Apply sort by requested order, then restore priority ordering
         self.__nzo_list.sort(key=sort_function, reverse=reverse)
-        self.__nzo_list.sort(key=lambda nzo: nzo.priority, reverse=True)
+        self.__nzo_list.sort(key=attrgetter("priority"), reverse=True)
 
     def update_sort_order(self):
         """Resorts the queue if it is useful for the selected sort method"""
@@ -947,7 +949,7 @@ class NzbQueue:
         """Check whether this name or md5sum is already
         in the queue or the post-processing queue"""
         lname = name.lower()
-        for nzo in self.__nzo_list + sabnzbd.PostProcessor.get_queue():
+        for nzo in chain(self.__nzo_list, sabnzbd.PostProcessor.get_queue()):
             # Skip any jobs already marked as duplicate, to prevent double-triggers
             # URL's do not have an MD5!
             if not nzo.duplicate and (
@@ -960,7 +962,7 @@ class NzbQueue:
     def have_duplicate_key(self, duplicate_key: str) -> bool:
         """Check whether this duplicate key is already
         in the queue or the post-processing queue"""
-        for nzo in self.__nzo_list + sabnzbd.PostProcessor.get_queue():
+        for nzo in chain(self.__nzo_list, sabnzbd.PostProcessor.get_queue()):
             # Skip any jobs already marked as duplicate, to prevent double-triggers
             if not nzo.duplicate and nzo.duplicate_key == duplicate_key:
                 return True
--- a/sabnzbd/nzbqueue.py
+++ b/sabnzbd/nzbqueue.py
@@ -235,16 +235,23 @@ class NzbQueue:
             self.__nzo_table[nzo_ids[0]].reuse = None
 
     @NzbQueueLocker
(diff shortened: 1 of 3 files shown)


Safihre:
I can't say I'm a huge fan to be honest... It feels very convoluted with the buckets for a performance gain that only applies to a super-minimal set of users that actually have such queues.

Safihre:
Unless there's a strong argument, I would like to keep the batched-save from this PR and a REPAIR-always-goes-first fix.
Claude also suggested we could use `bisect.insort` to have smarter insertion point.

sabnzbd/nzbqueue.py
```diff
@@ -581,26 +571,26 @@ def sort_queue(self, field: str, direction: Optional[str] = None):
             sort_function = lambda nzo: nzo.final_name.lower()
         elif field == "size" or field == "bytes":
             logging.info("Sorting by size (reversed: %s)", reverse)
-            sort_function = lambda nzo: nzo.bytes
+            sort_function = attrgetter("bytes")
```
Safihre, line 574: This is nice

sabnzbd/nzbqueue.py
```diff
@@ -340,32 +348,8 @@ def add(self, nzo: NzbObject, save: bool = True, quiet: bool = False) -> str:
...
-                        break
-                    pos += 1
-                if not added:
-                    # if there are no other items classed as a lower priority
-                    # then it will be added to the bottom of the queue
-                    self.__nzo_list.append(nzo)
-            else:
-                # if the queue is empty then simple append the item to the bottom
-                self.__nzo_list.append(nzo)
+        self.__insert_in_priority_order(nzo)
```
Safihre, line 351: Yes, very nice.

sabnzbd/nzbqueue.py
```diff
@@ -698,6 +657,16 @@ def set_priority(self, nzo_ids: list[str], priority: int) -> int:
         except Exception:
             return -1
 
+    def __insert_in_priority_order(self, nzo: NzbObject) -> int:
+        """Insert a job into the priority-ordered queue and return its position.
+        Repair jobs outrank forced ones, forced jobs go to the front of their own
```
Safihre: Why do Forced jobs go to the front of their own priority group?
I see it used to be like that, but why?

sabnzbd/nzbqueue.py
```diff
@@ -981,7 +950,7 @@ def have_name_or_md5sum(self, name: str, md5sum: str) -> bool:
         """Check whether this name or md5sum is already
         in the queue or the post-processing queue"""
         lname = name.lower()
-        for nzo in self.__nzo_list + sabnzbd.PostProcessor.get_queue():
+        for nzo in chain(self.__nzo_list, sabnzbd.PostProcessor.get_queue()):
```
Safihre, line 952: Never knew about that function.
Just for fun I had Claude run a test with lists of 50 items:
for x in a + b: pass	550 ns
for x in chain(a, b): pass	857 ns

Surprisingly, `chain` is slower. Only faster for early-exit loops, which we don't have much.
Still, good to know and use.

sabnzbd/nzbqueue.py
mnightingale: Best I can work out from the comments "usually a completed download fetching pars" it's leftover from when the repair priority didn't exist so it wanted jobs returning to the queue to go to the top.

It is the existing behaviour, but it's a little weird.

select a,b,c => set force => c,b,a (odd)
select a,b,c => set high => a,b,c

Where it might be move expected for the user is moving one item, that could be interpreted as the priority for the user so going to the top could be intended.

Consider making the behaviour consistent, but it changes what it currently does.

sabnzbd/nzbqueue.py
mnightingale: The speed different is tiny, it's really used to keep memory usage flat and early-exit - which this case is although it probably never does so both loops are practically the same cost.

A loop with just `pass` flatters a+b, but with real processing that difference will disappear.

`a + b` allocates a new list and 8 bytes per item, chain will be flat at a few bytes of memory.

sabnzbd/nzbqueue.py
Safihre: Let's change it to do the same for all priority types.

thezoggy: > Why do Forced jobs go to the front of their own priority group?
> I see it used to be like that, but why?

We used force for the test nzb

mnightingale: > Let's change it to do the same for all priority types.

I have changed this.

> We used force for the test nzb

If we want the test ones to go to the top still, then we can send an api move request after adding.
But this was the kind of behaviour that I pointed out users might expect for force - "I want this now"

mnightingale:
> Keep send_back() priority ordered.
>
> If the replacement has a lower priority than old_nzo, these lines place it at the old index ahead of higher-priority jobs. process_single_nzb() already adds the replacement through add(), which inserts it in priority order. Do not move it back to old_position.

The bug coderabbitai noticed, is not introduced by this change and a smaller issue than noted.

`end_job()` => `save_to_disk()` => `save_attribs()` => `send_back()`

Restoring calls `load_attribs()` which restored the priority, however it does `if attribs.get(attrib):` so not if falsy (NORMAL_PRIORITY = =0) so it stays DEFAULT_PRIORITY and takes the category default, so a pretty small gap that might restore jobs with the wrong priority.

PR #3619, opened 2026-09-10
Add active web-UI session management

### LINKED ISSUE #3581: Session tracking and invalidation
### Description

The just announced security vulnerability highlights a  need for a administration page where active sessions/cookies can be tracked, and when necessary invalidated. It would also be useful information to have the ip address, and country/state of origin of that ip address.

I did examine the codebase and see nothing like this exists. It also looks like a complete refactor of the authentication would be required. Currently it seems like cookies are just blindly validated, and not tracked. This seems like a huge room for improvement. 

Currently it seems like the closest fix is to restart the server. It would be better to give the administrator more options and better visibility.

1. Track cookies
2. Allow them to be invalidated
3. Provide a page for them to be tracked and invalidated

Alternatively without having to dig into the logs, a page that lists all recent validated sessions with the same ip address information mentioned above.

https://sabnzbd.org/wiki/extra/authentication-vulnerability.html

Safihre:
Allow users to view and revoke active login sessions. This enhances security and control by providing insights into authenticated access and the ability to sign out remote devices.

Closes #3581

## Summary by CodeRabbit

- **New Features**
  - Added an **Active Sessions** view accessible from the main menu, showing signed-in devices, IP addresses, sign-in times, last activity, and expiration.
  - Added session list loading, refresh, individual session revocation, and an option to sign out all sessions with confirmation.
  - Added localized labels and messages for session management.

- **Bug Fixes**
  - Improved session activity and client information tracking.
  - Sessions associated with previous login credentials are no longer displayed after credentials change.


Safihre:
@mnightingale What do you think?

mnightingale:
I've not had a chance to look through it since you made further changes.
I saw coderabbit picked up on my intentional design that it was only going to be used from the event loop so didn't need locks, I haven't looked at how you have addressed that yet
The note about save_admin() on the event loop I can live with because it's such a rare request and would be a pain to change

Will try and look more closely tomorrow

mnightingale (changes requested):
A few things I'd suggest changing.

- `public_list` shows sessions that are invalid due to a password change; don't just make changing the username/password delete sessions because the config might have been edited directly, check cred_fingerprint
- `session["last_seen"] = now` handle that in session.py rather than security.py modifing sessions internals
- `last_seen` stored in memory and lost on restart, persist it on shutdown
- Like `last_seen` I'd consider not having IP/User-Agent changes persist to disk from the event loop on every change
- Might be nice to highlight your own/current session
- I should have documented that sessions stuff should be async, so it runs on the single event loop and doesn’t require locking
- maybe simplify `$have_logout` and `$have_sessions`
- Revoke all button when none are available, disable or hide?
- `text: user_agent || '$T('notAvailable')'` should go via a binding like the others else breaks if the translatable ever contains `'`

sabnzbd/__init__.py
```diff
@@ -480,6 +480,7 @@ def save_state():
     sabnzbd.BPSMeter.save()
     sabnzbd.DirScanner.save()
     sabnzbd.PostProcessor.save()
+    sabnzbd.SessionStore.flush()
```
mnightingale, line 483: `save_state()` isn't shutdown-only: `postproc.handle_empty_queue()` calls it from the **PostProcessor thread** every time the post-processing queue drains (`postproc.py:1139`), while the web server is live. So `flush()` mutates and pickles the store from a second thread, breaking the invariant this same commit documents in `api.py`:

> SessionStore assumes it is only ever touched from the single event loop and has no lock

Reproduced by running `flush()` in one thread against `add()` in another:

```
File "sabnzbd/sessionstore.py", line 158, in flush
    self._sessions = self._unexpired(self._sessions, int(time.time()))
File "sabnzbd/sessionstore.py", line 76, in _unexpired
    return {token: s for token, s in sessions.items() if s["expires"] > now}
```

Neither `handle_empty_queue()` nor `PostProcessor.run()` wraps this, so it propagates out of `run()` and the post-processing thread dies for the rest of the run - every later completed download then sits unprocessed, with only a bare traceback on stderr. On the shutdown path `halt()` does catch it and logs "Fatal error at saving state", leaving the sessions unflushed.

Either give `SessionStore` a lock (and drop the "needs no locking" claim from its docstring and from the `api.py` comment), or marshal the flush onto the event loop.

sabnzbd/sessionstore.py
```diff
@@ -67,20 +100,79 @@ def get(self, token_hash: str) -> Optional[Session]:
...
+        self._save()
+
+    def flush(self):
+        """Force a persist of state the throttled per-request save may not have
+        written yet (mainly last_seen/ip/user_agent), and prune anything expired
+        since. Called at shutdown so neither is lost until the next start; a no-op
+        if the store was never loaded this run. Unlike _prune_and_persist, this
+        always saves - that is the point of a forced flush."""
+        if self._sessions is not None:
+            self._sessions = self._unexpired(self._sessions, int(time.time()))
```
mnightingale, line 158: Same root cause as the `save_state()` call site, but a second and silent failure mode: read -> rebuild -> reassign is not atomic, so any session `add()` installs between the read and the assignment is discarded.

A user logging in exactly when post-processing drains the queue gets a cookie back for a session that no longer exists - the login looks like it worked, then the next request bounces them to the login page. Under contention I measured **869 of 20000** concurrently-added sessions lost. The same window lets `flush()` resurrect sessions that `delete()`/`delete_all()` just removed, so "Revoke all sessions" can silently fail to revoke.

Minor, while you're here: `flush()` saves unconditionally, so it rewrites `sessions.sab` on every queue drain even when nothing changed.

sabnzbd/sessionstore.py
```diff
@@ -67,20 +100,79 @@ def get(self, token_hash: str) -> Optional[Session]:
...
+            {
+                "id": public_session_id(token_hash),
+                "created": s["created"],
+                "last_seen": s["last_seen"],
+                "expires": s["expires"],
+                "ip": s["ip"],
+                "user_agent": s["user_agent"],
+            }
+            for token_hash, s in self._sessions.items()
+            if s["cred_fingerprint"] == fingerprint
```
mnightingale, line 176: Nice fix for the listing, but the records themselves still stay on disk: `_unexpired()` - which backs both `_prune_and_persist()` and `flush()` - filters on `expires` alone.

`test_public_list_excludes_a_stale_credential_fingerprint` asserts exactly that (`session_store.get("e" * 64) is not None`), and `_validate_session` only cleans one up if that client comes back. A client that never returns leaves its recorded IP and user-agent in `sessions.sab` for the full 14-day `SESSION_DURATION` after the password change that was meant to kill it.

Dropping mismatched fingerprints in `_unexpired()` (or in `flush()`) would clear them at the next write.

tests/test_sessionstore.py
```diff
@@ -21,42 +21,181 @@
...
+
+        session_store.flush()
+
+        reopened = sessionstore.SessionStore()
+        session = reopened.get("hash1")
+        assert session["ip"] == "5.6.7.8"
+        assert session["user_agent"] == "other"
+        assert session["last_seen"] == now + 100
+
+    def test_flush_is_a_noop_when_never_loaded(self, session_store):
```
mnightingale, line 87: No assertion here, so this only checks that the call doesn't raise - remove the `if self._sessions is not None` guard from `flush()` and the test still passes.

That guard is what stops a never-used store from loading the file purely to write it straight back at shutdown. `patch.object(store, "_save")` and asserting it wasn't called (or that no `sessions.sab` appears) would cover what the name claims.

mnightingale (commented):
A few things from re-reading the latest push. The credential-fingerprint filter, the memory-only `mark_seen` and the `glitterTranslate` binding all look right. The new `flush()` wiring is the problem: `save_state()` is not shutdown-only, so it reaches `SessionStore` from other threads.

sabnzbd/sessionstore.py
```diff
@@ -19,68 +19,181 @@
...
-                    now = int(time.time())
-                    self._sessions = {token: s for token, s in sessions.items() if s["expires"] > now}
+                    self._prune_and_persist(sessions, int(time.time()))
         except Exception:
             logging.info("Failed to load sessions", exc_info=True)
 
     def _save(self):
-        save_admin((SESSIONS_VERSION, self.sessions), SESSIONS_FILE_NAME)
+        """Caller holds self.lock."""
+        save_admin((SESSIONS_VERSION, self._sessions), SESSIONS_FILE_NAME)
```
mnightingale, line 111: The lock is held across the disk write, so a save started on *any* thread now blocks every session lookup - including the ones the event loop makes on each request.

`handle_empty_queue()` -> `save_state()` -> `flush()` -> `_save()` runs on the PostProcessor thread; meanwhile `_validate_session()` calls `SessionStore.get()` on the event loop and blocks on the same lock, so all HTTP handling stalls rather than just one request. With a 0.5s `save_admin` I measured a `get()` from another thread waiting **0.446s**.

Not purely a slow-disk hypothetical: `save_data()` retries three times on a transient failure with two `time.sleep(0.1)` calls in between, so a single failing write holds the lock for 0.2s plus I/O - and plenty of installs keep the admin dir on a NAS or an SD card.

Getting the write out of the locked region means trading `@synchronized()` on the public methods for explicit `with self.lock:` around just the dict work, then saving a snapshot after the block:

```python
def flush(self):
    with self.lock:
        if self._sessions is None:
            return
        self._sessions = self._live(self._sessions, int(time.time()))
        payload = (SESSIONS_VERSION, {token: dict(s) for token, s in self._sessions.items()})
    save_admin(payload, SESSIONS_FILE_NAME)
```

The per-session `dict(s)` matters: a shallow `dict(self._sessions)` would still share the `Session` values that `mark_seen()` mutates. That copy also happens to answer the CodeRabbit comment above about handing out live `Session` objects.

sabnzbd/api.py
```diff
@@ -840,6 +840,29 @@ def _api_showlog(name: str, kwargs: QueryParams) -> StreamingResponse:
     return build_log_response()
 
 
+# All three kept async def rather than sync, for consistency with the other
```
mnightingale: Two problems with the replacement text.

`security.py` defines no routes - it has helpers (`create_session`, `validate_session`, ...) called from middleware - so "the other session-related route handlers in security.py" points at nothing a reader can go and find.

More importantly the reasoning is inverted. The old comment stated a real invariant: `async` kept these off `run_in_threadpool`, which the lock-free store depended on. Now that `SessionStore` is locked, thread-safety is exactly what *removes* that constraint - sync handlers would be dispatched via `run_in_threadpool`, and the pickle write inside `public_list()` / `delete_by_id()` / `delete_all()` would stop running on the event loop. That is the fix for the `_save()` comment above, i.e. the opposite of what this comment recommends.

So either drop it, or have it say why `async` is still wanted on its own merits.

mnightingale (commented):
Re-checked the thread-safety work - the lock holds up. 20000 concurrent `add()` calls against a continuous `flush()` loop now produce no `RuntimeError` and no lost sessions (previously 869 lost), and the fingerprint pruning in `_live()` clears revoked sessions from disk. Two smaller things left.

sabnzbd/__init__.py
mnightingale: Addressed in f78a692 - `SessionStore` now takes an `RLock` via `@synchronized()`, and the class docstring records that `flush()` reaches it from the PostProcessor thread. Verified: 20000 concurrent `add()` calls against a continuous `flush()` loop produce no `RuntimeError` and no lost sessions, against 869 lost before.

For the record on the alternative I floated in this thread: marshalling onto the event loop would not have been a good fit. `ThreadedServer` keeps no reference to the loop (`uvicorn.Server.run()` creates it inside the `WebServer` thread), you would have to block on `.result()` for the shutdown flush to be worth anything, and `halt()` is reachable on paths where the server thread has already gone. The lock was the right call. Feel free to resolve - I only have triage access here.

mnightingale (approved):

PR #3623, opened 2026-09-14
Implement Indexer Search

Safihre:
image

## Summary by CodeRabbit

* **New Features**
  * Added NZB search across configured Newznab indexers.
  * Added indexer management, including adding, editing, enabling, deleting, and connection testing.
  * Added category filters, result counts, password indicators, favicon support, and direct result links.
  * Search results can be sent directly to the NZB add queue.
  * Added search access from both the main interface and configuration menu.
* **Bug Fixes**
  * Improved favicon fallback handling and search result loading.
* **Tests**
  * Added coverage for indexer searching, XML parsing, categories, filtering, and error handling.


interfaces/Config/templates/config_nzbsearch.tmpl
```diff
@@ -0,0 +1,113 @@
...
+                        <th>$T('srch-host')</th>
+                        <th>$T('srch-apikey')</th>
+                        <th>&nbsp;</th>
+                    </tr>
+                    <!--#end if#-->
+                    <tr class="data-row">
+                        <td class="nzbsearch-favicon-col"><div class="favicon" style="background-image: url(//$indexer['baselink']/favicon.ico);" data-domain="$indexer['baselink']"></div></td>
+                        <td><input type="checkbox" name="enable" value="1" <!--#if $indexer['enable'] then 'checked="checked"' else ""#--> /></td>
+                        <td><input type="text" name="host" value="$indexer['host']" size="40" /></td>
+                        <td><input type="password" name="api_key" value="$indexer["api_key"]" size="32" autocomplete="new-password" /></td>
```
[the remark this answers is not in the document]
Safihre: Not true

mnightingale (commented):
First few items of feedback, not looked at the code at all yet:

It works OK for a first attempt, but there are a few things I think need addressing;
- I’d take some inspiration from Prowlarr, checkbox on the left for adding multiple, column order, indexer name - when multiple are configured you can’t tell which it’s coming from and that can be useful if you have some that are typically more reliable.
- Indexer name, like RSS would probably help Prowlarr syncing indexers to SAB, assuming that’s something they’d add support for.
- Adding a second indexer doesn’t work, doesn’t give any error or anything just clears the form and makes no config change. - wasn't able to test it but I'd probably to be able to pick which indexer to search like Prowlarr
- Using Prowlarr as the indexer, category number translations are missing.
- At least on “Always use full screen width” it wastes a lot of screen space, copping off long titles.
- Light mode (not checked dark) the results table seems to be missing some borders, the first column has a left border in the header row.
- Table row height feels too spaced out and not consistent without elsewhere with similar layout, queue is 38px, search is 51px and on compact it’s 24.5px vs 34.55px

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+        """GET this indexer's API."""
+        url = self._build_url(apikey=self.api_key, **params)
+        request = urllib.request.Request(
+            url, headers={"User-Agent": f"SABnzbd/{sabnzbd.__version__}", "Accept-Encoding": "gzip"}
+        )
+        try:
+            with urllib.request.urlopen(request, timeout=DEFAULT_TIMEOUT) as response:
+                body = response.read()
+                if "gzip" in (response.headers.get("Content-Encoding") or ""):
+                    body = gzip.decompress(body)
```
mnightingale: Why not use `requests` and not deal with this manually? It's already in requirements

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+    async def _search_one(
+        self,
+        indexer: Indexer,
+        text: str,
+        category_ids: tuple,
+    ) -> tuple[list[SearchResult], int | None]:
+        """Search one indexer off the event loop; never raises - failures return no results."""
+        try:
+            logging.debug("Searching indexer %s", indexer.name)
+            results, reported_total = await asyncio.to_thread(indexer.search, text, category_ids)
```
mnightingale: Should use `starlette.concurrency.run_in_threadpool`

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+    values = conf.get_dict()
+    return Indexer(values["name"], values["host"], values["api_path"], values["api_key"])
+
+
+class CategoryCache:
+    """Build the category tree once, until indexers are added or removed."""
+
+    def __init__(self):
+        self.lock = Lock()
+        self.categories: list[dict] | None = None
```
mnightingale, line 210: There are a lot of UP045 `X | None`, maybe time to give up on this fight or do what I did and add something to CLAUDE.md telling it your preference

sabnzbd/api.py
```diff
@@ -1173,6 +1223,9 @@ def _api_gc_stats(name: str, kwargs: QueryParams) -> Response:
     ("resume_pp", ""): ApiEntry(_api_resume_pp, 2),
     ("pause_pp", ""): ApiEntry(_api_pause_pp, 2),
     ("rss_now", ""): ApiEntry(_api_rss_now, 2),
+    # mode=nzbsearch
+    ("nzbsearch", ""): ApiEntry(_api_nzbsearch, 2),
+    ("nzbsearch", "caps"): ApiEntry(_api_nzbsearch_caps, 2),
```
mnightingale, line 1235: Should this be level 2? or level 3 like get_config?

sabnzbd/misc.py
```diff
@@ -1477,10 +1477,12 @@ def ip_extract() -> list[str]:
...
 
 def get_base_url(url: str) -> str:
-    """Return only the true root domain for the favicon, so api.oznzb.com -> oznzb.com
-    But also api.althub.co.za -> althub.co.za
+    """Return the true root domain for a URL or bare hostname, e.g.
+    api.oznzb.com -> oznzb.com and api.althub.co.za -> althub.co.za.
     """
     url_host = urllib.parse.urlparse(url).hostname
+    if not url_host:
+        url_host = urllib.parse.urlparse("//" + url).hostname
```
mnightingale: Two problems with this fallback, and it regresses callers outside this feature.

**It can now raise.** `get_base_url` previously returned `""` for anything unparseable and never raised. `urlparse("//" + url)` goes through `_check_bracketed_netloc`, so any input containing brackets raises `ValueError`:

```
>>> get_base_url("[]")
ValueError: '' does not appear to be an IPv4 or IPv6 address
>>> get_base_url("[bad]")
ValueError: 'bad' does not appear to be an IPv4 or IPv6 address
>>> get_base_url("a[b]c")
ValueError: Invalid IPv6 URL
```

The pre-existing callers feed this third-party data: rss.py:1114 (`get_base_url(uri)`) and interface.py:2095 (`get_base_url(entry.link)` on an RSS entry link). A feed link containing brackets now 500s the RSS log page. It also takes out `/config/nzbsearch` via interface.py:2023 and the add form via interface.py:2037.

**It invents hostnames.** For scheme-only or path-only input it reads a hostname out of what is actually the scheme or the first path segment:

```
>>> get_base_url("http://")
'http'
>>> get_base_url("...")
'..'
```

So pasting `http://` into the add form creates an indexer whose config section is named `http`, with a favicon pointing at `//http/favicon.ico`.

Worth noting these need two separate fixes: wrapping the fallback in try/except stops the crash but still returns `'http'`, so the result needs validating as well.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+        try:
+            with urllib.request.urlopen(request, timeout=DEFAULT_TIMEOUT) as response:
+                body = response.read()
+                if "gzip" in (response.headers.get("Content-Encoding") or ""):
+                    body = gzip.decompress(body)
+                return parse_xml_response(body)
+        except urllib.error.HTTPError as error:
+            raise IndexerError(f"Request failed (HTTP {error.code})")
+        except (urllib.error.URLError, TimeoutError, OSError) as error:
```
mnightingale: This misses two exception types that a plain typo in the host produces, and neither is an `OSError`:

```python
>>> Indexer("x", "exa mple.com", "/api", "K").test()
InvalidURL: URL can't contain control characters. 'exa mple.com' (found at least ' ')
>>> Indexer("x", "api.example.com:notaport", "/api", "K").test()
InvalidURL: nonnumeric port: 'notaport'
>>> Indexer("x", "[::1", "/api", "K").test()
ValueError: Invalid IPv6 URL
```

`http.client.InvalidURL` subclasses `HTTPException`, not `OSError`, so it escapes `fetch` untouched. `_api_config_test_indexer` (api.py:1137) only catches `IndexerError`, so a single stray space in the host field gives the user an HTTP 500 instead of "Cannot reach indexer". In the search path the same escape is swallowed by `_search_one`'s broad `except Exception` and the indexer silently vanishes from the results.

This is another argument for my `requests` comment above: `requests` normalises all of these into `RequestException` subclasses, so switching would fix it as a side effect rather than needing `InvalidURL` and `ValueError` bolted onto this tuple.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+
+class IndexerError(Exception):
+    """A single indexer request failed."""
+
+
+def text_from_element(element: ET.Element, tag: str) -> str:
+    return (element.findtext(tag) or "").strip()
+
+
+def valid_details_url(url: str) -> str:
```
mnightingale, line 60: `urlparse` raises on a malformed IPv6 URL, and this is called inside the per-item loop, so one bad field from an indexer throws away every result from that indexer rather than just the one item:

```python
ix.fetch = lambda **_: ET.fromstring(
    '<rss><channel>'
    '<item><link>https://e.com/a.nzb</link><comments>http://[::1</comments></item>'
    '<item><link>https://e.com/b.nzb</link></item>'
    '</channel></rss>'
)
ix.search("q")
# ValueError: Invalid IPv6 URL
```

The exception propagates out of `Indexer.search`, `_search_one` catches it, and all 100 results are gone. A `try/except ValueError` here (or validating in the loop and skipping the item) keeps the rest of the page.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+                    for subcategory in caps_category.findall("subcat"):
+                        subcategory_id = int_conv(subcategory.get("id"))
+                        name = subcategory.get("name")
+                        if subcategory_id and name and subcategory_id not in existing_ids:
+                            category["subcats"].append({"id": subcategory_id, "name": name})
+                            existing_ids.add(subcategory_id)
+            for category in categories:
+                category["subcats"].sort(key=lambda subcategory: subcategory["id"])
+            logging.info("Loaded %s subcategories", sum(len(category["subcats"]) for category in categories))
+            self.categories = categories
```
mnightingale: This assigns unconditionally, so a run where every indexer's caps request failed gets cached as the bare 8-entry tree and is never retried.

Open the search modal once while the indexer is briefly unreachable (or during startup before DNS is up), line 233 swallows the error with `continue`, and this line caches the empty result. Every later open returns a category dropdown with no subcategories for the rest of the process lifetime, and the only way out is editing an indexer to trigger `invalidate_categories()`.

Only caching when at least one indexer actually answered would avoid it.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+                {"id": 2000, "name": "Movies", "subcats": []},
+                {"id": 3000, "name": "Audio", "subcats": []},
+                {"id": 4000, "name": "PC", "subcats": []},
+                {"id": 5000, "name": "TV", "subcats": []},
+                {"id": 6000, "name": "XXX", "subcats": []},
+                {"id": 7000, "name": "Books", "subcats": []},
+                {"id": 8000, "name": "Other", "subcats": []},
+            ]
+            categories_by_id = {category["id"]: category for category in categories}
+            for indexer in indexers:
```
mnightingale: This loop issues blocking 15s caps requests one indexer at a time, while holding `self.lock` for the whole thing, even though `search()` on the same objects fans out concurrently with `asyncio.gather`.

With four indexers and one slow one, the first open of the search modal costs up to 4 x `DEFAULT_TIMEOUT` = 60s. `glitter.nzbsearch.js:55` gives `callAPI` a 30s timeout, so the caps call times out, `capabilitiesRequested` resets to `false`, and the user gets an empty category dropdown. Meanwhile any concurrent `invalidate_categories()` (saving or toggling an indexer on the config page) blocks on the lock for that entire window.

Fanning these out the same way `search()` does, and not holding the lock across the network I/O, fixes both.

sabnzbd/interface.py
```diff
@@ -2011,6 +2011,72 @@ def config_sorting_toggle_sorter(request: Request):
...
+        search_list=conf,
+    )
+
+
+@secured_expose(route="/config/nzbsearch/add_indexer", check_configlock=True, methods=["POST"])
+def config_nzbsearch_add_indexer(request: Request):
+    params = request_params(request)
+    host = Strip(params.get("host"))
+    # The indexer's root domain doubles as its identifier and its favicon domain
+    name = config.clean_section_name(get_base_url(host)) if host else ""
```
mnightingale: This is the root cause of "adding a second indexer doesn't work, doesn't give any error or anything just clears the form and makes no config change" from my earlier review.

The config section name is derived from the host's root domain, so any two indexers sharing one root domain collide. `get_config("indexers", name)` is truthy for the second, the `if` on the next line fails, and the handler falls straight through to `base_redirect_response` having done nothing. Same silent no-op when `get_base_url(host)` returns `""`.

It bites in realistic setups: two local indexers on `http://192.168.1.5:8080` and `http://192.168.1.5:9090`, or accounts on both `api.nzbgeek.info` and `nzbgeek.info`. Ties in with the "indexer name, like RSS" point from that review, a separate user-supplied name would remove the collision and the naming problem at once.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+    Knows how to build its own request URLs and fetch and parse search results.
+    """
+
+    def __init__(self, name: str, base_url: str, api_path: str, api_key: str):
+        self.name = name
+        self.base_url = base_url
+        self.api_path = api_path
+        self.api_key = api_key
+
+    def _build_url(self, **params) -> str:
```
mnightingale: Everything this computes is invariant for the life of the `Indexer`, but it's redone on every request, and `get_base_url(self.base_url)` at :189 is redone for every single result item (up to `SEARCH_LIMIT` = 100 full `urlparse` calls per search).

Normalising once in `__init__` into `self.api_url` and `self.baselink` collapses this to one `urlencode` plus an f-string, at which point it folds into `fetch` and stops being a separate method. It also drops the `path or 'api'` fallback, which duplicates the `/api` default already on `ConfigIndexer.api_path`, and gives one obvious place to validate the URL, which the `InvalidURL` comment above needs anyway.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+from threading import Lock
+
+import sabnzbd
+import sabnzbd.config as config
+from sabnzbd.decorators import synchronized
+from sabnzbd.misc import get_base_url, int_conv, to_units
+
+SEARCH_LIMIT = 100  # latest results requested per indexer
+DEFAULT_TIMEOUT = 15  # seconds per indexer request; not user-configurable
+UNKNOWN_AGE = datetime(1970, 1, 1, tzinfo=timezone.utc)
```
mnightingale, line 49: An item with no parseable date gets 1970-01-01, which `calc_age` in `_api_nzbsearch` renders as roughly `20713d` in the Age column.

The test at tests/test_nzbsearch.py:126 is named `test_missing_item_date_defaults_to_zero_days` but asserts the value is 1970-01-01, so the name documents behaviour the code doesn't have and nothing covers how it renders. Either make it `Optional[datetime]` and emit `""`, or have the API skip `calc_age` when age is `UNKNOWN_AGE`.

mnightingale (commented):
Another pass, Python only this time. `ruff check` and `black` are clean and `tests/test_nzbsearch.py` passes 24/24, so everything below is behaviour rather than style.

I'm not repeating the four things I already commented on (`run_in_threadpool` at nzbsearch.py:307, the UP045 unions at :213, the access level at api.py:1228, `requests` at :125), but two of them gained some detail worth adding:

- On `run_in_threadpool` vs `asyncio.to_thread`: the thread itself can't go away. There's no async HTTP client anywhere in requirements.txt (only `requests`, and that's pulled in as an apprise dependency), so making `Indexer.search` natively async would need a new dependency. The `asyncio.gather` fan-out is already the right shape for waiting on several indexers and combining the results, it's only the offload call that should change. Same for `category_tree` at :319.
- On `requests`: it would also fix a real bug on its own, see the comment on the `except` clause below.

The comment on interface.py:2037 is the root cause of "adding a second indexer doesn't work, doesn't give any error" from my first review.

Three of these I confirmed by actually running them rather than just reading, marked below. The worst is the `get_base_url` change, which regresses callers that have nothing to do with this feature.

For what it's worth, CodeRabbit's claim that saving resubmits the masked API key is indeed wrong as you said: `OptionPassword.set` (config.py:431) already ignores star-only values.

sabnzbd/nzbsearch.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+class IndexerError(Exception):
+    """A single indexer request failed."""
+
+
+def text_from_element(element: ET.Element, tag: str) -> str:
```
mnightingale, line 56: Minor, but this module adds six single-line functions: `text_from_element` (:56), `valid_details_url` (:60), `strip_namespace` (:64), `Indexer.test` (:132), `invalidate_categories` (:260) and `category_tree` (:317), plus `config.get_indexers` at config.py:1327.

`strip_namespace` in particular only exists to do namespace-agnostic matching that ElementTree already does with the `{*}` wildcard. `item.findall("{*}attr")` at :154 and `channel.find("{*}response")` remove both the helper and the manual re-scan of `channel`'s children at :196-198, which currently walks the same children a second time after `findall("item")`.

sabnzbd/interface.py
```diff
@@ -2011,6 +2011,72 @@ def config_sorting_toggle_sorter(request: Request):
...
+    if name and host and not config.get_config("indexers", name):
+        kwargs = dict(params)
+        kwargs["host"] = host
+        # An unchecked checkbox is simply absent from the POST body
+        kwargs.setdefault("enable", 0)
+        config.ConfigIndexer(name, kwargs)
+        sabnzbd.nzbsearch.invalidate_categories()
```
mnightingale, line 2045: interface.py never imports `sabnzbd.nzbsearch`, this only resolves because `sabnzbd/__init__.py:97` happens to import it first.

Every other submodule used here is imported explicitly, and api.py adds its own `import sabnzbd.nzbsearch` for exactly this reason. As it stands, reordering or dropping that line in `__init__.py` turns into an `AttributeError` at request time on three handlers (2044, 2060, 2069) instead of an `ImportError` at startup.

sabnzbd/interface.py
```diff
@@ -2548,11 +2614,19 @@ async def not_found_redirect(request: Request, exc):
 
 
 class CachedStaticFiles(StaticFiles):
-    """Static files the browser may hold indefinitely, as $url() versions every reference"""
+    """Static files the browser may hold indefinitely, as $url() versions every reference.
+
+    Development and pre-release builds revalidate instead: their `$url()` version is
+    the git commit, which does not change when the working tree is edited."""
+
+    _immutable = not bool(re.search(r"(alpha|beta|rc|dev)", sabnzbd.__version__, re.IGNORECASE))
```
mnightingale: This has nothing to do with indexer search and changes the `Cache-Control` header on every static asset for every skin. Worth splitting into its own PR so it can be reviewed (and reverted) on its own.

The regex is also looser than intended: `rc` matches anywhere in the string, so any future version suffix containing those two letters in sequence (`arch`, `src`) would silently drop immutable caching on release builds. Anchoring it, or matching on a separator, would be safer. `not bool(...)` is redundant too, `not` already returns a bool.

mnightingale: Human: I get why you'd want this, eventually Vite + manifest loaded at startup would sort it properly rather than the cache buster by commit

Safihre: No clue how I missed this. Should not be in.

Safihre:
Will fix the issues.
However, this was never meant to replace Prowlarr or anything like that.
So that's why there is no multi-add, no sorting, etc etc. 
In case of multiple indexers the favicon of the indexer is shown in front, so that will indicate what is the source.

PR #3628, opened 2026-09-17
Fix unwanted extensions not running on deobsufacted files correctly

### LINKED ISSUE #3627: Unwanted extensions can fail to run for obfuscated file extensions
### SABnzbd version

5.1.3

### Operating system

Linux

### Using Docker image

binhex

### Description

A download with an obfuscated, extensionless file passed the unwanted-extension check, even though `exe` is in my unwanted extensions list and the action is set to "Fail job (move to History)".

The file had no extension while downloading, so the download-time check and the post-unpack check both saw nothing to block. Later, "Deobfuscate final filenames" inspected the contents, recognized a Windows executable, and renamed it to `.exe`. No unwanted-extension check runs after that point, so the job completed successfully with an executable in the completed folder.

Tooltip on the job showed: 

```
Path: /media/Downloads/{video_name}.exe
Deobfuscate: Deobfuscate corrected the extension of 1 file(s)
```

My unwanted extensions was set to blacklist: `exe, lnk, zipx, bat` with the action to take as "Fail job (move to History)" but the `.exe` never got deleted and the job was marked as successfully downloaded

lukezbihlyj:
Closes #3627 - also ensures that if the unwanted file action is set to fail to job that it gets failed correctly which was not previously occuring as intended

## Summary by CodeRabbit

* **Bug Fixes**
  * Unwanted-file checks now run after files are renamed or deobfuscated, improving detection accuracy.
  * Detected unwanted files are removed according to the configured action.
  * Jobs are marked as failed when the configured policy requires it or unwanted files cannot be removed.
  * Conditional end-job email notifications now include unwanted-file detection results.
  * Added a message when files with unwanted extensions cannot be removed.


lukezbihlyj:
Both of the remaining 2 minor cautions from coderabbit are for pre-existing functions or behavior that already existed, so for the sake of keeping the PR low-scope I'll leave that stuff as-is

PR #3629, opened 2026-09-19
Mark jobs with missing articles as failed when there is no par2

mnightingale:
Fixes the issue described on https://forums.sabnzbd.org/viewtopic.php?p=134711

## Problem

With `fail_hopeless_jobs` disabled, a job that downloads with missing articles is reported as Completed and moved to the Complete folder, even when thousands of articles are missing and the files are unusable.

`fail_hopeless_jobs` is the only thing that ever consults article completeness: `check_availability_ratio()` is called from `remove_article()` only when the option is on. With it off, `nzo.fail_msg` is never set and post-processing starts with `all_ok = True`.

Post-processing then derives the verdict purely from verification and unpack results, and never looks at `nzo.bytes_missing`. If the NZB has no par2 (or the par2 were themselves unavailable), there is no SFV, and no identifiable RARs, nothing sets `par_error` or `unpack_error`, so `all_ok` survives and the job is marked Completed.

This also contradicts the documented behaviour of `safe_postproc`, whose help text states that turning it *off* is what causes incomplete jobs to be marked Completed.

## Fix

Fail the job when articles are missing and no par2 is available to repair them:

```python
# Without par2 there is nothing to repair missing articles with
if all_ok and nzo.bytes_missing and not nzo.extrapars and cfg.safe_postproc():
```

`bytes_missing` already tracks only non-par2 files. `not nzo.extrapars` means no par2 set exists; when one does exist, a failed repair already sets `par_error`.

Gated on `safe_postproc` rather than `fail_hopeless_jobs`, since this is a post-processing verdict and not the mid-download abort decision.

## Notes

- Applies to Download-only jobs too, matching what `fail_hopeless_jobs` already does to the same job during download.
- Unchanged: a job whose only par2 sets are samples under `ignore_samples`, and a Download-only job whose par2 downloaded successfully, both still skip the check.

## Tests

Four tests in `TestProcessJobMissingArticles` covering the bug plus guards against over-failing (par2 present, no missing articles, `safe_postproc` off). The first fails on current develop with `assert 'Completed' == 'Failed'`; the guards pass on both sides.

## Summary by CodeRabbit

* **Bug Fixes**
  * Downloads with missing articles are now correctly marked as failed when safe post-processing is enabled and no additional PAR2 files are available.
  * Downloads with complete data, available PAR2 files, or disabled safe post-processing continue to complete as expected.


thezoggy:
technically a nzb posted with just a rarset and no parset could* have some recovery record info in it and restore using that..
https://www.win-rar.com/recovery-record.html?&L=0

but that is niche and most posters do not create rarsets with it (or disallowed by rules).

sabnzbd/postproc.py
```diff
@@ -475,6 +475,15 @@ def process_job(nzo: NzbObject) -> bool:
         if flag_repair and cfg.safe_postproc():
             all_ok = all_ok and not par_error
 
+        # Without par2 there is nothing to repair missing articles with
```
Safihre: Why only here and not earlier?
And why not use `nzo.check_availability_ratio()`?

mnightingale: > Why only here and not earlier?

The gap is when there's no par2 and no RAR files: RAR jobs are left to unrar, which can repair them from REV files and otherwise fails on the CRC error.

> And why not use `nzo.check_availability_ratio()`?

It allows up to 5 bad articles for jobs without par2, so post-processing only fails them if something like a RAR fails to unpack. Damage to a loose mkv or an nfo goes undetected and is marked Completed, even with fail_hopeless_jobs enabled.

The trade-off is that a no-par2 job without RARs whose only damage is in an nfo or other unimportant file is now marked as failed, while damage to a loose extra in a RAR release is left to unpack.

mnightingale:
> technically a nzb posted with just a rarset and no parset could* have some recovery record info in it and restore using that..

Latest commit addresses that, @coderabbitai also noticed so it now tries to unrar which might recover them

PR #3631, opened 2026-09-21
Fix macOS menu bar item disappearing while downloading

mnightingale:
The title was padded with blank lines then pulled back into view by a baseline offset, for example `\n\n%s\n%sB/s\n` for the downloading state. Which is 5 lines in a 22pt menu bar.

On macOS 27 a status item whose title lays out taller than the menu bar reports a nonsense intrinsic width, and the system drops it.

| title | button width | button height |
| --- | --- | --- |
| `""` (idle) | 32.0 | 22.0 |
| `"\n\n3m 20s\n1.5 MB/s\n"` (downloading, before) | 20041.0 | 33.0 |
| `"3m 20s\n1.5 MB/s"` (after) | 73.0 | 22.0 |

The titles now contain only the lines that are actually shown, two for downloading and paused, one for "Stopping...".

Two follow-on adjustments were needed once the padding was gone:

`setParagraphSpacing_(-3.0)` is removed. Paragraph spacing applies after each paragraph, and each `\n` separated line is its own paragraph.

The baseline offset is now derived from the line count. The title block is bottom anchored in the button, so each extra line raises its centre by half a line height, and a single constant cannot centre both the two line and the one line titles against the icon.

## Deprecated AppKit constants

`NSAlternateKeyMask`, `NSOnState`, `NSOffState` and `NSCenterTextAlignment` are replaced by their current names. All four are pure renames with identical values, and they are pyobjc module level constants, so nothing here depends on the deployment target.

## Import NSColor, NSFont and NSImage from AppKit

These are AppKit classes that were being imported from `Foundation`. That works today only because `SABnzbd.py` imports `AppKit` before it imports `sabnzbd.macosmenu`, which loads the framework and lets pyobjc's `Foundation` resolve them through its ObjC runtime lookup. Importing `sabnzbd.macosmenu` on its own raises `ImportError` on pyobjc 12.2.2. No user visible change, it just removes the dependency on import order.

## Testing

Checked on macOS 26 and macOS 27. The offsets below come from driving `setMenuTitle_` against a live `NSStatusItem` on macOS 27 and measuring the rendered status button. Offset is between the centre of the text block and the centre of the icon:

| title | offset from icon centre | clearance above / below |
| --- | --- | --- |
| `0:01:19 / 9.7 MB/s` | 0.0pt | 3.5 / 1.5 |
| `1h 12m 30s / 10.1MB/s` | -0.25pt | 3.0 / 1.5 |
| `Paused / 29:57` | -1.0pt | 3.0 / 3.0 |
| `Stopping...` | +0.25pt | 8.0 / 5.5 |

The sub-point residual is ink variation between glyph sets rather than layout. `Paused / 29:57` has no descenders, so its ink centre reads slightly high.

The bundle targets `LSMinimumSystemVersion` 10.15, so releases older than macOS 26 are untested, although two line titles and vertical alignment are stock AppKit behaviour and the status bar has been 22pt throughout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated macOS menu behavior for compatibility with current system APIs.
  * Improved status bar title alignment and spacing across paused, speed, and restart states.


Safihre:
Thanks!

